PR #1, opened 2017-01-26
Implement default `navigationOptions`

grabbou:
For context: this is old #222 pull request that already got `lgtm`, just with the difference that Object.assign is removed. 

Todo: 
- <s>docs</s>
- <s>support in TabBar</s>
- support `style` on header (next PR)

This basically makes it possible to define `navigationOptions` inside `NavigationStackRouterConfig`, so that you don't have to repeat them all over the place.

Proper docs have to land tomorrow that explain how to use it and how to mimic `merge` (explain 2nd parameter to the function because I guess it's nowhere documented)

### DIFF
--- a/docs/guides/Screen-Nav-Options.md
+++ b/docs/guides/Screen-Nav-Options.md
@@ -32,7 +32,6 @@ class ProfileScreen extends React.Component {
   ...
 ```
 
-
 #### Generic Navigation Options
 
 The `title` navigation option is generic between every navigator. It is used to set the title string for a given screen.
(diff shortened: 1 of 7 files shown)


src/routers/StackRouter.js
```diff
@@ -288,7 +288,7 @@ export default (
       };
     },
 
-    getScreenConfig: createConfigGetter(routeConfigs),
+    getScreenConfig: createConfigGetter(routeConfigs, stackConfig.navigationOptions),
```
satya164: You also need to do this for `TabRouter`

satya164:
We need to update the docs with the third way of specifying `navigationOptions`.

src/routers/StackRouter.js
grabbou: Done

knowbody:
good work @grabbou, thank you!

docs/guides/Screen-Nav-Options.md
```diff
@@ -47,6 +46,54 @@ class MyScreen extends React.Component {
 
 Unlike the other nav options which are only utilized by the navigator view, the title option can be used by the environment to update the title in the browser window or app switcher.
 
+#### Default Navigation Options
+
+You can configure the default `navigationOptions` for all the screens within particular navigator to save you needing
+to specify same properties all over the place.
```
mkonicek: Long sentences are hard to read, please break it up. Also it's unclear what "needing to specify same properties all over the place." - when I ran into the issue I didn't even know I had to specify the same props all over the place, I was stuck.

It might make sense to introduce the concept of Default Navigation Options to the reader, ideally by using an example (as if explaining what Default Navigation Options are to a person  sitting next to you). You could use the following text, for example:

It's very common to define `navigationOptions` on a screen but sometimes it is useful to define `navigationOptions` on a Navigator too. Imagine the following scenario: Your `TabNavigator` represents one of the screens in the app, and is nested within a top-level `StackNavigator`:

```
StackNavigator:
  - route1: A screen
  - route2: A TabNavigator
```

Now, when `route2` is active, you would like to hide the header. It's easy to hide the header for `route1`, and it should also be easy to do it for `route2`. This is what Default Navigation Options are for - they are simply Navigation Options set on a navigator:

```js
StackNavigator({
  profile: ProfileScreen,
  ...
}, {
  navigationOptions: {
     header: {
       visible: false,
     },
   },
 });
```

Note that you can still decide to **also** specify the `navigationOptions` on the screens at the leaf level - e.g.  the `ProfileScreen` above. The `navigationOptions` from the screen will be merged key-by-key with the default options coming from the navigator, and whenever both define the same option (e.g. `header`), the screen wins. Therefore, you could make the header visible when `ProfileScreen` is active.

**Extending defaults:** In order to extend default config with screen-specific properties instead of overriding it, you could dynamically configure an option:
 
 ```js
 class ProfileScreen extends React.Component {
   static navigationOptions = {
     header: (navigation, defaultHeader) => ({
       ...defaultHeader,
       visible: true,
       color: 'blue',
     }),
   }
   ...
 }
 ```
 
 The 2nd argument passed to the function is the default value as found on the navigator.

grabbou: CC: @knowbody, mind helping out with a PR?

PR #2, opened 2017-01-26
fix Page not found blog bug

knowbody:


ericvicenti:
This seems to address the issue, and this was undocumented behavior. We may want this down the line, but for now this is the way to go.

PR #3, opened 2017-01-26
remove a not needed password

knowbody:


README.md
```diff
@@ -2,7 +2,7 @@
 
 *Learn once, navigate anywhere.*
 
-Browse the docs on [reactnavigation.org](https://navigate:navigate@reactnavigation.org/). (The username/pass is `navigate:navigate`)
+Browse the docs on [reactnavigation.org](https://navigate:navigate@reactnavigation.org/).
```
vramana: Shouldn't it be just `https://reactnavigation.org/` instead of `https://navigate:navigate@reactnavigation.org/`

knowbody: thanks!

PR #10, opened 2017-01-27
Don’t handle click with modifier keys.

### LINKED ISSUE #9: Website: command-click should not handle inside page
Cmd-click on the website is broken

Andreyco:
Ignore clicks with modifier keys such as `alt`, `cmd`, `ctrl` and `shift`.

Closes #9

### DIFF
--- a/website/src/Link.js
+++ b/website/src/Link.js
@@ -1,5 +1,8 @@
 import React, { PropTypes, Component } from 'react';
 
+const isModifiedEvent = (event) =>
+  !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
+
 const Linkable = (Inner) => {
   class LinkableWrapped extends Component {
     render() {
@@ -29,7 +32,7 @@ const Linkable = (Inner) => {
     }
     onClick = e => {
       const action = this.getAction();
-      if (action) {
+      if (!isModifiedEvent(e) && action) {
         this.context.dispatch(action);
         e.preventDefault();
       }


website/src/Link.js
```diff
@@ -1,5 +1,8 @@
 import React, { PropTypes, Component } from 'react';
 
+const isModifiedEvent = (event) =>
```
satya164: Perhaps modified event is not a good name, since the event is not modified

satya164: probably you mean `isModifierKeyEvent`

website/src/Link.js
```diff
@@ -1,5 +1,8 @@
...
+  !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
```
ericvicenti: nit: in this codebase we use semicolons

website/src/Link.js
Andreyco: Actually I think click event is modified by pressing modifier key. My opinion, could be wrong :)

website/src/Link.js
```diff
@@ -29,7 +32,7 @@ const Linkable = (Inner) => {
     }
     onClick = e => {
       const action = this.getAction();
-      if (action) {
+      if (!isModifiedEvent && action) {
```
jcarbo: I think you need to call the function here with the event.

Andreyco:
Rebased... Sorry, should be sleeping already.

website/src/Link.js
satya164: @Andreyco better to name it `hasModifier` then

satya164: As I understand by modified event, it would mean an event got mutated/changed by some code

PR #23, opened 2017-01-27
Allow custom style

### LINKED ISSUE #18: Make it possible to pass `style` to header
As Martin suggested in old repo, it would be great to have an ability to define a custom style for a header on a Navigator level.

This is already getting implemented in #1. As a follow-up, we will expose style.

grabbou:
Allows setting custom `style` property on a header to change e.g. backgroundColor. Fixes #18. Also fixes `headerComponent` by fixing an old property reference

**Test plan (required)**

Go to any example in playground and define `header.backgroundColor`. Reload.

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -30,6 +30,7 @@ import type {
 
 import type {
   HeaderMode,
+  HeaderProps,
 } from './Header';
 
 import type { TransitionConfig } from './TransitionConfigs';
@@ -41,6 +42,7 @@ const NativeAnimatedModule = NativeModules && NativeModules.NativeAnimatedModule
 type Props = {
   screenProps?: {};
   headerMode: HeaderMode,
+  headerComponent?: ReactClass<HeaderProps>,
   mode: 'card' | 'modal',
   navigation: NavigationScreenProp<NavigationState, NavigationAction>,
   router: NavigationRouter,
@@ -61,6 +63,7 @@ type Props = {
 type DefaultProps = {
   mode: 'card' | 'modal',
   gesturesEnabled: boolean,
+  headerComponent: ReactClass<HeaderProps>,
 };
 
 class CardStack extends React.Component<DefaultProps, Props, void> {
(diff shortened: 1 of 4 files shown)


src/views/CardStack.js
```diff
@@ -184,16 +184,17 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
 
   _renderHeader(props: NavigationTransitionProps, headerMode: HeaderMode): ?React.Element<*> {
     const navigation = this._getChildNavigation(props.scene);
-    const header = this.props.router.getScreenConfig(navigation, 'header');
+    const header = this.props.router.getScreenConfig(navigation, 'header') || {};
 
     let HeaderComponent = Header;
-    if (header && header.bar) {
+    if (header.bar) {
```
satya164: Is the option to change the header component called `header.bar`? It doesn't match what we've in the type definitions https://github.com/react-community/react-navigation/blob/master/src/TypeDefinition.js#L244

grabbou: Yeah, will send a sep. PR

grabbou: Actually good note. Good we fix bugs at the same time :)

src/views/CardStack.js
```diff
@@ -184,16 +185,14 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
...
   _renderHeader(props: NavigationTransitionProps, headerMode: HeaderMode): ?React.Element<*> {
     const navigation = this._getChildNavigation(props.scene);
-    const header = this.props.router.getScreenConfig(navigation, 'header');
+    const header = this.props.router.getScreenConfig(navigation, 'header') || {};
 
-    let HeaderComponent = Header;
-    if (header && header.bar) {
-      HeaderComponent = header.bar;
-    }
+    const HeaderComponent = this.props.headerComponent || Header;
```
satya164: We should use `deafultProps` for this

grabbou:
Since this PR combines two issues, I'll rebase them to two commits that can be rebase and merged.

PR #24, opened 2017-01-27
Correct Redux example with addNavigationHelpers

hilkeheremans:
Correction for docs, addNavigationHelpers was missing, which if applied copy/paste style would lead to crashes


satya164:
Thank you for the PR

PR #32, opened 2017-01-27
withCachedChildNavigation: set proper component name

hilkeheremans:
When a component is wrapped with the withCachedChildNavigation, the component name was not always displayed. This fixes that which facilitates debugging.

When using debugger tools such as `react-native-debugger`, attempting to inspect the component tree shows various components as `withCachedChildNavigation(undefined)`. Closed inspection showed that this HOC relies only in `.displayName` of the wrapped component, which is not always set.

This change resolves that issue. Tested locally.


satya164:
Thank you for the PR.

PR #39, opened 2017-01-27
Update Routers.md

philraj:
Removing `const` (line 43):

Can't declare a property on the MyApp object as const (probably a typo).


satya164:
thank you

PR #41, opened 2017-01-27
Minor fix in documentation in the "Screen Navigation Options" sample (state undefined)

machadogj:
The current code as it was, depended on the "state" param, but the title was being defined as a string template literal.


satya164:
Thank you for the PR

PR #44, opened 2017-01-27
Use es6 default values for initial state

sagivo:
motivation - using es6 default values.
no functionality change, just cosmetics.


satya164:
Thanks for the PR

PR #45, opened 2017-01-27
Change tabBar.icon and drawer.icon to be react elements

### LINKED ISSUE #15: Change tabBar.icon and header.icon to be react elements rather than thunks
It's currently like this,

```js
static navigationOptions = {
  drawer: {
    icon: ({ route, focused, tintColor }) => React.Element<*>
  }
}
```

Change it to

```js
static navigationOptions = {
  header: ({ state, focused, tintColor }) => ({
    icon: React.Element<*>
  })
}
```

satya164:
This makes the API consistent with how header uses `left` and `right` elements.

Fixes #15

### DIFF
--- a/docs/guides/Screen-Nav-Options.md
+++ b/docs/guides/Screen-Nav-Options.md
@@ -19,13 +19,14 @@ class MyScreen extends React.Component {
 
 Or, each option can be a function that takes the following arguments, and returns the value of the option.
 
-- `navigation` - the [navigation prop](/docs/intro/navigation-prop) for the screen, with the screen's route at `navigation.state`
+- `navigation` - The [navigation prop](/docs/intro/navigation-prop) for the screen, with the screen's route at `navigation.state`
+- `viewOptions` - Options specific to the view, e.g. - header, tabBar or drawer
 - `childRouter` - The child router, if the screen is a navigator
 
 ```js
 class ProfileScreen extends React.Component {
   static navigationOptions = {
-    title: (navigation, childRouter) => {
+    title: (navigation, viewOptions, childRouter) => {
       return navigation.state.params.name + "'s Profile!";
     },
   };
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -91,10 +91,12 @@ export type NavigationRouter = {
   ) => ?any, // todo, fix this any type to become a key of NavigationScreenConfig
 };
 
-export type NavigationScreenOption<T> =
-  | (navigation: NavigationScreenProp<NavigationRoute, NavigationAction>,
-    config: NavigationScreenOptionConfig,
-    router?: NavigationRouter) => T
+export type NavigationScreenOption<T, ViewOptions> =
+  | (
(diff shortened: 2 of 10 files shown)


docs/guides/Screen-Nav-Options.md
```diff
@@ -19,7 +19,8 @@ class MyScreen extends React.Component {
 
 Or, each option can be a function that takes the following arguments, and returns the value of the option.
 
-- `navigation` - the [navigation prop](/docs/intro/navigation-prop) for the screen, with the screen's route at `navigation.state`
+- `navigation` - The [navigation prop](/docs/intro/navigation-prop) for the screen, with the screen's route at `navigation.state`
+- `viewOptions` - Options specific to the view, e.g. - header, tabBar or drawer
```
ericvicenti: Can we describe the full config here, or link to it?

src/TypeDefinition.js
```diff
@@ -165,19 +167,19 @@ export type NavigationScreenOptions = {
...
   /**
    * Options passed to the navigation bar for this screen.
    */
-  header?: NavigationScreenOption<HeaderConfig>;
+  header?: NavigationScreenOption<HeaderConfig, void>;
   /**
    * Options passed to the tab bar for this screen.
    */
-  tabBar?: NavigationScreenOption<TabBarConfig>;
+  tabBar?: NavigationScreenOption<TabBarConfig, { tintColor: string, focused: boolean }>;
```
ericvicenti: re-use this type?

src/views/Drawer/DrawerSidebar.js
```diff
@@ -59,26 +68,51 @@ class DrawerSidebar extends PureComponent<void, Props, void> {
...
+    const activeTintColor =
+      contentOptions && contentOptions.activeTintColor ?
+      contentOptions.activeTintColor :
+      defaultDrawerProps.activeTintColor;
+    const inactiveTintColor =
+      contentOptions && contentOptions.inactiveTintColor ?
+      contentOptions.inactiveTintColor :
+      defaultDrawerProps.inactiveTintColor;
+
+    const configs = navigation.state.routes.reduce((acc: DrawerConfigs, route: *) => {
```
ericvicenti: If you pass the router down then you could handle this stuff in the content component. This seems appropriate because many content components may be custom and wont need all this stuff

src/TypeDefinition.js
satya164: @ericvicenti the type may vary for both over time, so I kept it separate

src/views/Drawer/DrawerSidebar.js
satya164: @ericvicenti I updated pass the router down, but I'll prefer to keep the `renderIcon` and `renderLabel` here so that it's easier to reuse if wanted without having to write the logic again.

docs/guides/Screen-Nav-Options.md
satya164: @ericvicenti we have to finish write documentation about screen nav options for drawer and tab bar before I add that there. I'll create an issue for it.

PR #57, opened 2017-01-28
Update react-native-tab-view

### LINKED ISSUE #50: How to disable the left-to-right transition animation during initial app launch?
I use **TabNavigator** and here's the full code:

```js
class HomeScreen extends React.Component {
    render() {
        return <View style={{flexDirection:'row', backgroundColor: 'red', height: 300,
            justifyContent:'flex-end', paddingTop: 30}}>
            <Text>Hello, Navigation!</Text>
        </View>
    }
}

class SecondScreen extends React.Component {
    render() {
        return <View style={{flexDirection:'row', justifyContent:'flex-end', paddingTop: 30}}>
            <Text>Hello, Second!</Text>
        </View>
    }
}

const AppContainer = TabNavigator({
        HomePage: {screen: HomeScreen},
        SecondPage: {screen: SecondScreen}
    }
)
```

As you can see, when app initially launches, the text "flies" in ... How can I disable this effect?

jan-27-2017 14-00-02

Thanks in advance!

satya164:
Fixed scrollview's style so it doesn't need layout https://github.com/react-native-community/react-native-tab-view/commit/6ef48ed52983272264a0efcd358c346abcf1cadc

Fixes #50

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,12 +65,13 @@
     "clamp": "^1.0.1",
     "fbjs": "^0.8.5",
     "path-to-regexp": "^1.7.0",
-    "react-addons-shallow-compare": "15.4.1",
     "react-native-drawer-layout": "^1.1.0",
-    "react-native-tab-view": "^0.0.52"
+    "react-native-tab-view": "^0.0.53"
   },
   "jest": {
-    "testPathDirs": ["src/"],
+    "testPathDirs": [
+      "src/"
+    ],
     "notify": true
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,12 +65,13 @@
     "clamp": "^1.0.1",
     "fbjs": "^0.8.5",
     "path-to-regexp": "^1.7.0",
-    "react-addons-shallow-compare": "15.4.1",
```
satya164: This was a peer dependency in react-native-tab-view sometime ago. It's not needed anymore for a few releases.

PR #69, opened 2017-01-28
Added nodes.js cross platform path separator in scripts/compile-docs.js

Steviey:
1. _Added nodes.js cross platform path separator in scripts/compile-docs.js_
2. _cross platform path separator_
3. _npm i_
4. _Only tested on Windows, but should work on Mac/Unix too._


Steviey:
May should be changed to:
`var sep = require('path').sep;`

PR #96, opened 2017-01-30
Set up Jest to be able to test RN UI components

ferrannp:
With this we can test UI like `TabNavigator` or `DrawerNavigator`. We can also use [Jest snapshot](https://facebook.github.io/jest/docs/tutorial-react-native.html#snapshot-test).

It required some Jest extra configuration.


package.json
```diff
@@ -69,9 +70,14 @@
     "react-native-tab-view": "^0.0.54"
   },
   "jest": {
-    "testPathDirs": [
```
ferrannp: This config had issues with RN, see https://github.com/facebook/jest/issues/2670.

.gitignore
```diff
@@ -17,3 +17,6 @@ yarn-error.log
 
 # Exponent
 .exponent
+
+# Jest
+coverage
```
ferrannp: Jest will generate a coverage folder if you run `npm test -- --coverage`.

jest-setup.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+
+// See https://github.com/facebook/jest/issues/2208
+jest.mock('Linking', () => {
```
satya164: Can you add the mock for ScrollView too?

package.json
```diff
@@ -26,7 +26,7 @@
     "start": "node ./node_modules/react-native/local-cli/cli.js start --config ./rn-cli.config.js",
     "run-playground-ios": "cd examples/NavigationPlayground && react-native run-ios",
     "run-playground-android": "cd examples/NavigationPlayground && react-native run-android",
-    "test": "BABEL_ENV=development jest",
```
ferrannp: This was not necessary. Actually, with this line, the `test` env in `.babelrc` was totally ignored.

jest-setup.js
ferrannp: You got it ;)

src/navigators/__tests__/__snapshots__/DrawerNavigator-test.js.snap
```diff
@@ -0,0 +1,167 @@
+exports[`DrawerNavigator renders a DrawerNavigator with correct options 1`] = `
```
satya164: Remove this, we can snapshot views later.

src/navigators/__tests__/DrawerNavigator-test.js
```diff
@@ -0,0 +1,40 @@
+/* @flow */
+
+import 'react-native';
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import DrawerNavigator from '../DrawerNavigator';
+
+const _getScreen = (label: string) => {
```
torabian: What is this `label: string` sounds like TypeScript, but in JavaScript. Is it flow?

PR #97, opened 2017-01-30
[WIP] iOS header refinements

grabbou:
Ideas welcome.

This pull request applies various visual tweaks to the header on iOS, most notably it adds support for `back button` label. 

By default, we will display a previous route title. I consider the logic done. The only part left is the styling itself, with a couple of todos as listed below.

Todo:
- [ ] Better visual fixes
- [ ] Fix header glitch

Will request reviews when ready. In the meantime feel free to chime in.

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -224,22 +224,21 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
           if (header && header.left) {
             return header.left;
           }
-          if (props.scene.index === 0 || !props.onNavigateBack) {
-            return null;
+          const { renderLeftComponent } = this.props.headerComponent.defaultProps || {};
+          if (typeof renderLeftComponent === 'function') {
+            return renderLeftComponent(props);
           }
-          const prevSceneTitle = this._getHeaderTitleForScene(props.scenes[props.index - 1]);
-          return (
-            <HeaderBackButton
-              onPress={props.onNavigateBack}
-              title={typeof prevSceneTitle === 'string' ? prevSceneTitle : undefined}
-            />
-          );
+          return null;
         }}
         renderRightComponent={({ scene }) => {
           const navigation = this._getChildNavigation(scene);
           const header = this.props.router.getScreenConfig(navigation, 'header');
           if (header && header.right) {
             return header.right;
+          }
+          const { renderRightComponent } = this.props.headerComponent.defaultProps || {};
+          if (typeof renderRightComponent === 'function') {
+            return renderRightComponent(props);
           }
           return null;
         }}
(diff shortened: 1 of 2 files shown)


src/views/CardStack.js
```diff
@@ -207,11 +224,16 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
...
-          const { renderLeftComponent } = this.props.headerComponent.defaultProps || {};
-          if (typeof renderLeftComponent === 'function') {
-            return renderLeftComponent(props);
+          if (props.scene.index === 0 || !props.onNavigateBack) {
+            return null;
           }
-          return null;
+          const prevSceneTitle = this._getHeaderTitleForScene(props.scenes[props.index - 1]);
+          return (
+            <HeaderBackButton
```
satya164: Why not `this.props.headerComponent.defaultProps. renderLeftComponent `?

satya164: Maybe add `scene.title`

grabbou: Decided to fall back to original approach as it was just easier to define custom back button that way (specify custom title). Otherwise I would've had to change either the signature of `renderLeftComponent` inside Header or make Header itself aware of a previous scene which I wanted to avoid.

satya164: @grabbou but shouldn't any custom header component be able to render the back button with title? otherwise if I want to use custom header button I can't show the title, and if I want to show the title I can't use a custom back button. Probably better to just add `title` to scene.

grabbou: > Maybe add scene.title

That would have to be `previousSceneTitle` kind of thing, which changes the API, right? An alternative if you really want to make the `Header` aware is to pass it `previousScene` as a prop and replace all checks for `props.scene.index === 0` with `!!previousScene`.

satya164: How about a prop to header called `getBackTitleForScene`?

grabbou: > but shouldn't any custom header component be able to render the back button with title?

If you define a custom header, the `this.props.renderLeftComponent(props)` will return you a standard back button anyway as long as your route doesn't override it. 

> if I want to use custom header button I can't show the title

If you want to show a custom header button (`left`), it's already up to you to access the title yourself I believe.

> if I want to show the title I can't use a custom back button

Not sure I understand? The default back button with a title will not show only if either of these is true:
- you have a custom `left` component
- you have a non-string `title` 

> How about a prop to header called getBackTitleForScene

How that improves over the existing interface? All you are saying is the ability to still have an access for a backButtonTitle when `header.left` is specified?

satya164: > If you define a custom header, the this.props.renderLeftComponent(props) will return you a standard back button anyway as long as your route doesn't override it.

yeah, but it's the standard back button for old header. if I have my default left component defined on my header, it won't work.

> If you want to show a custom header button (left), it's already up to you to access the title yourself I believe.

cardstack reads the config and passes the left component.

> Not sure I understand?

The header doesn't have a way to get the back button title for previous scene, so it won't able to show a title in a custom back button.

grabbou: Let's follow up on that one via Slack. Will post updates tomorrow when we get more details.

grabbou:
Will re-open when ready.

PR #99, opened 2017-01-30
CardStack: Don't overwrite custom transitionConfig with default

lintonye:
- The custom `transitionConfig` in `CardStack` is currently always overridden by `defaultConfig`. This causes `transitionConfig` to be ignored.
- The propType of `CardStack` requires `transitionConfig` to be a function. So, `this.props.transitionConfig` => `this.props.transitionConfig()`

**Test plan (required)**

- Write a custom navigation view which renders to a `CardStack` and test if the transition animation behave as expected:

```jsx
class CrossFadeTransitioner extends Component {
    render() {
        const transitionConfig = {
            screenInterpolator: (sceneProps: NavigationSceneRendererProps) => {
                const { position, scene, progress } = sceneProps;
                const { index } = scene;
                const inputRange = [index - 1, index, index + 1];
                    inputRange,
                    outputRange: [0, 1, 0],
                });

                return { opacity };
            }
        }
        return (
            <CardStack mode="card"
                router={this.props.router}
                transitionConfig={() => transitionConfig}
                />
        )
    }
}

```


satya164:
cc @ericvicenti

ericvicenti:
What bug does this fix? I think I've seen a related issue

But the change looks good to me! Is there any documentation that needs to be updated?

grabbou:
Up: @lintonye

nihgwu:
why this is still pending?

lintonye:
@ericvicenti  This PR makes `transitionConfig` in `CardStack` actually work.

I don't think there're any documentation changes needed since `CardStack` itself isn't documented.

brennanerbz:
Let's get this in!

kuka:
Anybody with write access, what’s the hold up?

brennanerbz:
@lintonye once this gets merged, can you include a fully fleshed out example?

lintonye:
@brennanerbz Sure, check out this: https://github.com/lintonye/react-native-diary/blob/master/transitions/app/CrossFadeTransitioner.js

Also this post: http://www.reactnativediary.com/2016/12/22/navigation-experimental-custom-transition-2.html

PR #105, opened 2017-01-31
Add support for custom app bar and status bar heights

### LINKED ISSUE #283: Change the header height
Hi,
First thanks you for the good job 😄 

Is there any way to change the header height?
I tried:

```javascript
static navigationOptions = {
    title: 'Welcome',
    header: navigation => ({
      style: {
        height: 80,
        backgroundColor: 'yellow',
        shadowColor: 'black',
        shadowRadius: 5,
        shadowOpacity: 0.1,
        shadowOffset: {
          height: 3,
          width: 0,
        },
      },
      titleStyle: { color: 'red' },
      left: <Text style={{color: 'red'}} onPress={() => navigation.goBack()}>LEFT</Text>,
      right: <Text style={{color: 'red'}} onPress={() => navigation.goBack()}>RIGHT</Text>,
    })
  }
```

And only get this weird result: https://transfer.sh/IbAum/header-height.png
I guess this is the problem: https://github.com/react-community/react-navigation/blob/master/src/views/Header.js#L283

juhasuni:
Current [Header view](https://github.com/react-community/react-navigation/blob/master/src/views/Header.js) implementation doesn't have support for fullscreen iOS apps (that is apps without status bar). It also lacks support for custom (app) bar height. While it would be ideal to be able to customise the Header using the `style` prop, it is not possible due to the layout mechanism used in CardStack. For that reason my proposal is to introduce two props: `barHeight` and `statusBarHeight`. The same prop names were used with [ex-navigation](https://github.com/exponent/ex-navigation).

Fixes #283

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -55,11 +55,13 @@ const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
 
 class Header extends React.Component<DefaultProps, HeaderProps, *> {
 
-  static HEIGHT = APPBAR_HEIGHT + STATUSBAR_HEIGHT;
   static Title = HeaderTitle;
   static BackButton = HeaderBackButton;
 
   static defaultProps = {
+    barHeight: APPBAR_HEIGHT,
+    statusBarHeight: STATUSBAR_HEIGHT,
+
     renderTitleComponent: (props: SubViewProps) => {
       const title = String(props.scene.route.title || '');
       return <HeaderTitle>{title}</HeaderTitle>;
@@ -86,6 +88,8 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
     renderLeftComponent: PropTypes.func,
     renderRightComponent: PropTypes.func,
     renderTitleComponent: PropTypes.func,
+    barHeight: PropTypes.number,
+    statusBarHeight: PropTypes.number,
     style: PropTypes.any,
   };
 
@@ -189,7 +193,7 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
 
   render(): React.Element<*> {
     // eslint-disable-next-line no-unused-vars
-    const { scenes, scene, style, position, progress, ...rest } = this.props;
+    const { scenes, scene, style, position, progress, barHeight, statusBarHeight, ...rest } = this.props;
 
     let leftComponents = null;
     let titleComponents = null;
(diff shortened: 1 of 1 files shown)


satya164:
I'll not merge this since we're doing some refactoring for header.

juhasuni:
Okay, I'll wait for your solution.

grabbou:
There will be small conflicts (I guess), but refactor is done and waiting for merge. Thanks @satya164 for keeping an eye on that.

grabbou:
@satya164 I'll review that one tomorrow and merge after my back button label lands. We should have then the header refactoring done 100%.

ericvicenti:
This looks like a decent step, but it would be really cool to be able to configure the status bar and header height for each screen in the `navigationOptions`

juhasuni:
> ...it would be really cool to be able to configure the status bar and header height for each screen...

I guess you're right. We're using `statusBarHeight = 0` when the screen is in fullscreen mode (that is without status bar). Now, it is possible that one screen wants to be presented without status bar (fullscreen) and another with status bar. In such case it would be useful to be able to configure this setting on per-screen basis.  However, transition from one screen to another in such case would probably look silly without animations. For that reason I'd implement this feature as I've proposed and leave per-screen options to backlog.

As a side note, we're also leveraging `react-navigation` with menus and when the navigator is used in menu context, we generally want to set the `statusBarHeight` to `0`.

grabbou:
@ericvicenti 

> This looks like a decent step, but it would be really cool to be able to configure the status bar and header height for each screen in the navigationOptions

I believe we can add it in next PR to `CardStack`. 

@juhasuni 

We already apply `style` to headerComponent and that `style` is on per-screen basis. Same limitations apply to the statusBarHeight and barHeight.

Also, do we need `barHeight` prop? Isn't it able to unpack the style we received and check for `height`? Would be nice if `header` respected the prop it was given.

juhasuni:
> Also, do we need barHeight prop? Isn't it able to unpack the style we receiver and check for height? 

I personally prefer using `style` prop over layout props such as height when ever possible. But picking layout props from the style prop is something I've tried to avoid. If we can get rid of the `appBar` child component, we wouldn't need to unpack the style.

juhasuni:
Please see my new PR #341 which is based on Header customisation using only `style` prop. I think that this PR should be closed and we should focus on the alternative method.

grabbou:
Sounds good, thanks for all the work @juhasuni on that :)

Issue #8, opened 2017-01-26
Allow screen option of `screenStyle` to for StackNavigator


ericvicenti:
I think this would will be useful for changing the background color of a screen

brunolemos:
Not sure if it's the same thing, but also a way to change the backgroundColor of the iOS rootView on runtime (and the android equivalent).

ericvicenti:
Ok, closing this because `cardStyle` was implemented in #106.

@brunolemos, you can currently change the root view background by wrapping your app navigator in a view that has a background color set on it.

brunolemos:
@ericvicenti if I understood you correctly, that would change the first view in the react hierarchy, but not the `rootView` from iOS, the one I set with this Obj-C code:
`rootView.backgroundColor = [UIColor colorWithRed:0.29 green:0.83 blue:0.71 alpha:1.0];`

If they have different background colors, which is my case, that would be very noticeable when you rotate the screen, for example:

[![rn-rootview.gif](https://s30.postimg.org/bnky8matd/rn_rootview.gif)](https://postimg.org/image/wx8kjgr3x/)

ericvicenti:
Oh, hmm. You may need to jump into the ObjC to handle this case

brunolemos:
I used to use `react-native-navigation` from wix, they have a `screenBackgroundColor` property that does this job, so I got used to it. (https://github.com/wix/react-native-navigation/pull/437)

But yes, I'll probably have to create a bridge to be able to change it dynamically independently of the navigation library.

ericvicenti:
Ah. It may make sense to add that functionality to core as well. I can see that it may be important to adjust that

sibelius:
I think it should be useful to modify this backgroundColor https://github.com/react-community/react-navigation/blob/e2f5e7fd494a9242edf1951e2398a1d1adf5b1ed/src/views/Card.js#L51

koenpunt:
> I think it should be useful to modify this backgroundColor

This is possible using the `cardStyle` option:

https://github.com/react-community/react-navigation/commit/a64eb46c9ced9c84c11a53e1d25cd1da08a36804

kelset:
Closing this since as @koenpunt said it's possible via `cardStyle`.

PR #106, opened 2017-01-31
Add support for cardStyle prop

juhasuni:
Currently it is not possible to set the style for an individual card in the CardStack when using the default StackNavigator. My proposal is to add `cardStyle` option to `stackConfig`.

### DIFF
--- a/src/navigators/StackNavigator.js
+++ b/src/navigators/StackNavigator.js
@@ -27,6 +27,7 @@ export default (routeConfigMap: NavigationRouteConfigMap, stackConfig: StackNavi
     headerComponent,
     headerMode,
     mode,
+    cardStyle,
     navigationOptions,
   } = stackConfig;
   const stackRouterConfig = {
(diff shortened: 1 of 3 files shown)


src/navigators/StackNavigator.js
```diff
@@ -42,6 +43,7 @@ export default (routeConfigMap: NavigationRouteConfigMap, stackConfig: StackNavi
       headerComponent={headerComponent}
       headerMode={headerMode}
       mode={mode}
+      cardStyle={cardStyle}
```
satya164: How does it get passed to the card?

satya164:
Is it the same as https://github.com/react-community/react-navigation/issues/8 ?

src/navigators/StackNavigator.js
juhasuni: CardStack seems to pass it. I didn't have to do anything for it.

juhasuni:
Yes. This resolves #8

juhasuni:
The proposed `screenStyle` is more universal, so if you need to support something like this in other navigators, it might be better to use `screenStyle` instead of `cardStyle`.

satya164:
I don't think we want this in other navigators. Both tab and drawer navigators don't set a background color, and if user is able to override their styles, they'll break.

ericvicenti:
Ok, lets ship this!

PR #109, opened 2017-01-31
[Header] tintColor, backButton, refactor

### LINKED ISSUE #36: Add support for tintColor in header
This should style a text and icon colors. See #30 for reference

### LINKED ISSUE #123: Support for hiding the default back button
I want to hide back button for some screens in StackNavigator.
So far I've tried to do the following without success. Is it not supported yet or am I doing anything wrong.
The right button does appear though.

(In a individual screen)
```
static navigationOptions = {
    title: "First Filter",
    header: ({ state, setParams, goBack, navigate }) => ({
      left:  <Button title="left"/>
      right: <Button title="right"/>
    })
  }
```
I've also tried to set header.left to null, no luck there also.

Thanks for this great library.

grabbou:
#This pull request serves two purposes:
1) It aims to implement:
    - [x] back button (Fixes #123)
    - [x] tint color (Fixes #36)
2) It refactors header and fixes few design issues that I have discovered after discussing them over with @satya164 

**TODO**
We need to handle case when back button title is too long. To merge it (if that's the case), I propose we turn off the `back button title` for now and just ship with a better back button icon.

**Test plan**

Run example, go to 2nd profile, you will see a different header colour and the back button has a label.
screen shot 2017-01-31 at 13 54 14
screen shot 2017-02-01 at 13 08 07

**Adnotation to point 2**

`tl;dr`

Now, `renderXComponent` that are defined in `CardStack` have only single purpose - check whether there's a custom React element supplied for a left/right/title and use it. Otherwise, return `undefined` so that `Header` can pick its default implementation (render back button, title) etc.


grabbou:
**Note** There will be 3 commits here to merge, back button, tint Color and refactor. Will rebase when `LGTM`

src/views/Header.js
```diff
@@ -99,11 +74,52 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
     );
   }
 
+  getHeaderTitle(props: SubViewProps): ?string {
```
satya164: Why not `_` prefix?

src/views/HeaderBackButton.js
```diff
@@ -27,6 +29,11 @@ const HeaderBackButton = ({ onPress, tintColor }: Props) => (
       source={require('./assets/back-icon.png')}
       tintColor={tintColor}
     />
+    {Platform.OS === 'ios' && title && (
```
satya164: Doesn't `TouchableOpacity` require a single child?

src/views/Header.js
grabbou: True, I don't use `_` at all personally and sometimes I happen to forget.

src/views/HeaderBackButton.js
```diff
@@ -27,6 +29,11 @@ const HeaderBackButton = ({ onPress, tintColor }: Props) => (
...
+      <Text style={{ color: tintColor }}>
```
satya164: Probably this needs additional styles to match the native look? Also the default tint color should be the same as the back button.

grabbou: Yeah, it's not yet done, just made a simple structure, but then decided to move to the refactor

src/views/HeaderBackButton.js
```diff
@@ -39,14 +53,27 @@ const styles = StyleSheet.create({
...
-  button: {
-    height: 24,
-    width: 24,
-    margin: Platform.OS === 'ios' ? 10 : 16,
-    resizeMode: 'contain',
-    transform: [{ scaleX: I18nManager.isRTL ? -1 : 1 }],
+  backButton: {
+    fontSize: 17,
   },
+  button: Platform.select({
```
satya164: This will break on windows, not sure if we support it yet, but better to keep in mind.

src/views/HeaderBackButton.js
```diff
@@ -39,14 +53,27 @@ const styles = StyleSheet.create({
...
     justifyContent: 'center',
+    flexDirection: 'row',
   },
-  button: {
-    height: 24,
-    width: 24,
-    margin: Platform.OS === 'ios' ? 10 : 16,
-    resizeMode: 'contain',
-    transform: [{ scaleX: I18nManager.isRTL ? -1 : 1 }],
+  backButton: {
```
satya164: name it as `title` maybe, since it's not a button

src/views/HeaderBackButton.js
```diff
@@ -12,21 +14,33 @@ import TouchableItem from './TouchableItem';
...
+  title?: string,
   tintColor?: string;
 };
 
-const HeaderBackButton = ({ onPress, tintColor }: Props) => (
+const defaultTintColor = Platform.select({
+  ios: '#037aff',
+});
+
+const HeaderBackButton = ({ onPress, title, tintColor = defaultTintColor }: Props) => (
```
satya164: Use `defaultProps` instead of default arguments

src/views/Header.js
```diff
@@ -227,8 +272,8 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
     return (
       <Animated.View {...rest} style={[styles.container, style]}>
         <View style={styles.appBar}>
-          {leftComponents}
           {titleComponents}
+          {leftComponents}
```
satya164: why the order change?

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+
+  _getHeaderTintColor(navigation: Navigation): ?string {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.tintColor) {
+      return header.tintColor;
+    }
+    return null;
+  }
+
+  renderTitleComponent = (props: SubViewProps) => {
```
satya164: add `_`

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    const color = this._getHeaderTintColor(props.navigation);
+    const title = this._getHeaderTitle(props.navigation);
+    return <HeaderTitle style={{ color }}>{title}</HeaderTitle>;
+  };
+
+  renderLeftComponent = (props: SubViewProps) => {
```
satya164: add `_`

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    return (
+      <HeaderBackButton
+        onPress={props.onNavigateBack}
+        tintColor={tintColor}
+        title={backButtonTitle}
+      />
+    );
+  };
+
+  renderRightComponent = () => {
```
satya164: add `_`

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    }
+    return null;
+  }
+
+  _getHeaderTintColor(navigation: Navigation): ?string {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.tintColor) {
+      return header.tintColor;
+    }
+    return null;
```
satya164: return `undefined` if `tintColor` isn't given. Otherwise default params or default props won't work

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    let title = null;
+    if (header && header.title) {
+      title = header.title;
+    } else {
+      title = this.props.router.getScreenConfig(navigation, 'title');
+    }
+    if (typeof title === 'string') {
+      return title;
+    }
+    return null;
```
satya164: Return `undefined` if not defined

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
     );
   }
 
+  _getHeaderTitle(navigation: Navigation): ?string {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    let title = null;
```
satya164: unnecessary initialization, it gets overriden below

src/views/Header.js
```diff
@@ -53,39 +53,20 @@ type SubViewName = 'left' | 'title' | 'right';
 const APPBAR_HEIGHT = Platform.OS === 'ios' ? 44 : 56;
 const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
 
-class Header extends React.Component<DefaultProps, HeaderProps, *> {
+class Header extends React.Component<*, HeaderProps, *> {
```
satya164: Use `void` instead of `*`

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
 
+  _getHeaderTitle(navigation: Navigation): ?string {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    let title = null;
+    if (header && header.title) {
+      title = header.title;
+    } else {
+      title = this.props.router.getScreenConfig(navigation, 'title');
+    }
+    if (typeof title === 'string') {
```
satya164: remove this check, only back button needs the check

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.tintColor) {
+      return header.tintColor;
+    }
+    return null;
+  }
+
+  renderTitleComponent = (props: SubViewProps) => {
+    const color = this._getHeaderTintColor(props.navigation);
+    const title = this._getHeaderTitle(props.navigation);
```
satya164: it should support both strings and react elements

src/views/Header.js
grabbou: Because title component has left/right 40 absolute position and covers the back button title. Changed the order to make it appear.

src/views/Header.js
grabbou: React Elements won't get into here, the CardStack parent checks for elements. This method is only executed when user doesn't supply a custom element to be rendered.

src/views/Header.js
grabbou: This also needs a check. Technically speaking this will never be reached since `CardStack` provides methods that check for `React.Element`

src/views/Header.js
grabbou: True, copy&paste from master.

satya164:
Also can you provide screenshot for Android too? And a full screenshot for iOS.

src/views/HeaderBackButton.js
grabbou: Out of curiosity, why?

satya164: @grabbou because you're specifying styles only for android and iOS

src/views/Header.js
```diff
@@ -53,39 +53,20 @@ type SubViewName = 'left' | 'title' | 'right';
 const APPBAR_HEIGHT = Platform.OS === 'ios' ? 44 : 56;
 const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
 
-class Header extends React.Component<DefaultProps, HeaderProps, *> {
+class Header extends React.Component<void, HeaderProps, *> {
```
satya164: Change the other `*` also

grabbou: Cool

src/views/Header.js
```diff
@@ -99,11 +80,63 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    if (header && header.tintColor) {
+      return header.tintColor;
+    }
+    return undefined;
+  }
+
+  _renderTitleComponent = (props: SubViewProps) => {
+    const color = this._getHeaderTintColor(props.navigation);
+    const title = this._getHeaderTitle(props.navigation);
+    return <HeaderTitle style={{ color }}>{title}</HeaderTitle>;
```
satya164: `style={color && {{ color }}` is better since otherwise it'll override the default color in the component even if no color is specified

src/views/HeaderBackButton.js
```diff
@@ -35,18 +45,37 @@ HeaderBackButton.propTypes = {
...
-  button: {
-    height: 24,
-    width: 24,
-    margin: Platform.OS === 'ios' ? 10 : 16,
-    resizeMode: 'contain',
-    transform: [{ scaleX: I18nManager.isRTL ? -1 : 1 }],
+  title: {
+    fontSize: 17,
   },
+  button: Platform.select({
```
satya164: Change to `Platform.OS` check to avoid breaking windows

src/views/Header.js
grabbou: From testing, it doesn't override the default value, however I do see the point of keeping it explicit.

src/views/HeaderBackButton.js
```diff
@@ -35,18 +45,36 @@ HeaderBackButton.propTypes = {
...
-    height: 24,
-    width: 24,
-    margin: Platform.OS === 'ios' ? 10 : 16,
-    resizeMode: 'contain',
-    transform: [{ scaleX: I18nManager.isRTL ? -1 : 1 }],
+  title: {
+    fontSize: 17,
   },
+  button: Platform.OS === 'ios'
+    ? ({
```
satya164: unnecessary parens

satya164:
LGTM

@ericvicenti ?

src/views/Header.js
grabbou: This is a temporary solution probably until we figure out how to change the layouts here. But since it changes a few things, I'd say let's ship it and I can keep working on back button layout in the meantime.

src/TypeDefinition.js
```diff
@@ -381,6 +381,7 @@ export type NavigationTransitionProps = {
   // is the index of the scene
   scene: NavigationScene,
   index: number,
+  navigation: NavigationScreenProp<NavigationRoute, NavigationAction>,
```
ericvicenti: Is flow passing? Is this actually provided everywhere the transition prop is used?

grabbou: @ericvicenti hard to tell as there are quite a few issues with `flow`. I didn't notice anything relevant. Anyways, regarding the navigation being provided -> we attach it in `_render()` so it's not included between `render()` -> `Transitioner` -> `_render()`.

grabbou: Actually no! I'll fix it

migueloller:
Hi guys, I'll chime in real quick on what the usual result is when the title is too long. iOS seems to change the text from the title to "Back". An easy way to check for this is to go to Settings > General > Storage & iCloud Usage > Manage Storage. Since "Storage & iCloud Usage" is too long then it will simply show "Back".

Hope this helps! 😄

src/TypeDefinition.js
grabbou: @ericvicenti you are right. In order to make flow pass, we would have to define `navigation` directly inside `Transitioner` `buildTransitionProps`. Do you think it makes sense? Not sure how that corresponds to the `_getChildNavigation()` that seems to have some memoization. Anyway, the only place where we don't have `navigation` is all calls prior to _render(). Any ideas?

ericvicenti: Based on the intention of what the transitioner props are, I think it makes sense for the screen navigation prop to be a part of them.

grabbou: @ericvicenti how do you see the `_getChildNavigation` going forward? Shall we stick to it and pass it around, or remove it entirely and just stick to `addNavigationHelpers`?

ericvicenti: My thought is that hopefully we can use the transitioners' memoization of the transitioner props, instead of `_getChildNavigation`. Do you think that could work?

grabbou: Didn't know about it. I think it's fine as you described. Added in latest commit

grabbou: Here for ref: https://github.com/react-community/react-navigation/pull/109/commits/1b5fb64cd376723ba0e3c044322595741e2009f8

grabbou: Also @ericvicenti sounds like with `navigation` being passed, we can just remove `navigationState`, right?

grabbou:
Thank @migueloller for the update! For now I think it's good to ship it `as is` and just send another PR tomorrow so there are no PRs being blocked.

src/TypeDefinition.js
grabbou: I haven't used NavigationExperimental a lot so I might be missing a bigger picture in few cases, esp. when compatibility is concerned. However, thanks for the reminder, I'll check how's it implemented a bit more in deep to keep that in mind for the future.

ericvicenti: I think we cannot remove `navigationState` because it refers to the parent navigation state, with routes and index, rather than `navigation.state` which is the route/screen's state.

Maybe we can change the API, and one day provide wrappers to help people adapt. Nevermind for now!

grabbou: All righty, sounds good! Thanks for clarifications, some of the things are still mythical to me!

src/views/CardStack.js
```diff
@@ -191,49 +192,42 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
...
-          }
-          return null;
+          return undefined;
         }}
-        renderRightComponent={({ scene }) => {
-          const navigation = this._getChildNavigation(scene);
-          const header = this.props.router.getScreenConfig(navigation, 'header');
+        renderRightComponent={(props: NavigationTransitionProps) => {
+          const header = this.props.router.getScreenConfig(props.navigation, 'header');
           if (header && header.right) {
```
satya164: Check for `(header && typeof header.right !== 'undefined')`

src/views/CardStack.js
```diff
@@ -191,49 +192,42 @@ class CardStack extends React.Component<DefaultProps, Props, void> {
...
-          if (header && header.title) {
-            title = header.title;
-          } else {
-            title = this.props.router.getScreenConfig(navigation, 'title');
-          }
-          if (typeof title === 'string') {
-            return <Header.Title>{title}</Header.Title>;
+        renderTitleComponent={(props: NavigationTransitionProps) => {
+          const header = this.props.router.getScreenConfig(props.navigation, 'header');
+          if (header && header.title && React.isValidElement(header.title)) {
```
satya164: This also needs to support `null`

Check for `(header && typeof header.title !== 'undefined' && (React.isValidElement(header.title) || header.title === null))`

src/views/Header.js
```diff
@@ -99,11 +80,57 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
     );
   }
 
+  _getHeaderTitle(navigation: Navigation): ?string {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.title) {
```
satya164: This needs to be `(header && typeof header.title === 'string')` to address empty string use case, and 0

src/views/Header.js
```diff
@@ -99,11 +80,57 @@ class Header extends React.Component<DefaultProps, HeaderProps, *> {
...
+    const color = this._getHeaderTintColor(props.navigation);
+    const title = this._getHeaderTitle(props.navigation);
+    return <HeaderTitle style={color && ({ color })}>{title}</HeaderTitle>;
+  };
+
+  _renderLeftComponent = (props: SubViewProps) => {
+    if (props.scene.index === 0 || !props.onNavigateBack) {
+      return null;
+    }
+    const tintColor = this._getHeaderTintColor(props.navigation);
```
satya164: Since the `tintColor` is passed by header, while the card stack seems to get `navigationOptions.header.left` and `navigationOptions.header.right` how would we pass `tintColor` to them?

grabbou: `tintColor` is just read from `navigationOptions`, for custom header.left. If you have specified a custom `tintColor` in e.g. defaults, you can:
```
navigationOptions: {
   header: (_, defaults) => {
      left: <Text style={{ color: defaults.tintColor }}>ABC</Text>,
   },
}
```
right?

satya164: Following the discussion in slack, I think we decided not to change the current behaviour and not pass options through the defaults

grabbou: True, but `tintColor` is not passed by the header -> it's read from navigationOptions:
```js
 const header = this.props.router.getScreenConfig(navigation, 'header');
    if (header && header.tintColor) {
       return tintColor;
    }
  return undefined;
```
And so, it means the value can be fully read by the user as well.

satya164: @grabbou this doesn't allow me to do `left: ({ tintColor }) => <Icon tintColor={tintColor} />`

grabbou:
screen shot 2017-02-01 at 13 08 07

ericvicenti:
The colors are now inconsistent in the example :-(

Also on Apple's header, the back text fades into the title when moving back, so this change seems to make it almost less "native".

satya164:
> The colors are now inconsistent in the example :-(

Ouch! I fixed it in my local, but forgot to commit :(

> Also on Apple's header, the back text fades into the title when moving back, so this change seems to make it almost less "native".

@grabbou ^

PR #114, opened 2017-01-31
Update Guide-Headers.md

pronevich:
this.props must be in render


grabbou:
Looks good! Thanks :)

Issue #52, opened 2017-01-28
Help: Navigate from nested navigation


ElliottJRo:
Hi all, I am trying to setup navigation in my app but having some trouble navigating from nested views. Im pretty sure im doing this wrong since I cant go from a 2nd level view to another 2nd level view in another Navigator. 

Basically I want to go from LoginView to MainView somehow but keeping WelcomeNavigator separate from HomeNavigator (mostly for organization purposes but also to clear the stack when transitioning from !isLoggedIn screen to isLoggedIn screen)

Can someone point me in the right direction? Thanks! 

```
const WelcomeNavigator = StackNavigator({
  Login: { screen: LoginView},
  Register: { screen: RegisterView}
}, {
  headerMode: 'none'
});

const HomeNavigator = TabNavigator({
  Main: {screen: MainView },
  Settings: { screen: SettingsView}
});

const MasterNavigator = StackNavigator({
  Welcome: { screen: WelcomeNavigator },
  Home: { screen: HomeNavigator }
},
{
  headerMode: 'none'
});
```

satya164:
I don't think this will work. What do you expect it to do?

satya164:
> I expect it to go from one nested view to another nested view ideally hahah

What'll animation do you expect?

> Can I use a custom router to go from one path to another using absolute keys instead of having local keys? eg: /welcome/login to /home/main or something

Use `navigation.dispatch` to dispatch a nested 'Reset' action which does 'Navigate' after the reset

ericvicenti:
We have an advanced feature that may help here: you can create a nested navigate action. In your case, it would look like this:

```js
this.props.dispatch({
  type: 'Navigate',
  routeName: 'Home',
  action: {
    type: 'Navigate',
    routeName: 'Main',
  },
});
```

ElliottJRo:
@ericvicenti ahaha it works!! So when I dispatch the navigate action its caught by the parent `Navigator` or does it look first in sibling views and then passes to the parent?

One thing I did notice is: I don't see a second `Navigation Dispatch` in console log, is this expected?

ericvicenti:
It is caught by the parent navigator because the action's routeName matches the route config name for that navigator. It is expected that there is only one action logged, because only one action happens at the top level.

I'm closing this out for now, feel free to file more issues as you come across them. If you have a chance, consider improving the docs to help make this feature more apparent!

temitope:
@ericvicenti that was extremely helpful, but i had to use
[...]

PR #118, opened 2017-01-31
Update Guide-Nested.md

ElliottJRo:
Updated nested navigation guide to include navigating between multi level nested screens

reference:  #52


ericvicenti:
Can we document the nested action behavior elsewhere? This guide is pretty focused, and the additional example might confuse people.

Now that I think about it, the two places where this should be documented is:

A. The [navigate function on the navigation prop](https://reactnavigation.org/docs/navigators/navigation-prop#Navigate), which already has a small mention of it
B. Add a new section at the end of the [deep linking advanced guide](https://reactnavigation.org/docs/guides/linking), explaining how URIs can be converted into deep actions

ElliottJRo:
Personally I found it hard to read how to do nested navigating and had trouble understanding what the `action` param did just from the nested navigators guide.

In the [navigation prop md](https://reactnavigation.org/docs/navigators/navigation-prop) it looks like its all API documentation. Also the term 'action' is being overloaded ( 'dispatch an action' and 'action object in dispatch object') which makes things more confusing -> not sure where to put the nested navigation behaviour.

As for putting it in the deep linking guide, i'm having trouble understanding how passing a dispatch object with an action to be handled by the child router has anything to do with deep linking / external URIs.

ericvicenti:
Ok, I see what you mean and I'm thinking it makes sense to clarify this inside the existing nested navigator guide. Do you have an idea on how we can work it in to the existing guide, rather than re-defining the example at the end? It looks like we may need a modified example. I can re-record the gifs, if we need to change them

I do think it deserves a mention in the deep linking guide- it can be highly relevant because a URI may get converted into one of these deep actions.

If you have ideas for improving the navigation prop API docs, that would also be appreciated!

ElliottJRo:
I think we should separate the api docs into its own section, away from the docs explaining ideas. I realized while developing, I cant find the docs I need and need to sift through all the docs until I reach the right API doc. I can make another thread regarding documentation overhaul and some thoughts on how to do it.

As for the deep linking guide, could you explain how a URI would be converted into a deep action? Does `MyApp://stacked/screen` get converted into a `type:Navigate` dispatch with `action`  being set? (ps, the docs regarding [nav prop](https://reactnavigation.org/docs/navigators/navigation-prop) changed so anything under `dispatch` doesn't make sense ei: doesn't actually use the `dispatch` method)

Now back to this current issue, how about we close it for now and chunk in this documentation with a documentation re-organization?

docs/api/navigators/DrawerNavigator.md
```diff
@@ -79,7 +79,12 @@ this.props.navigation.navigate('DrawerClose'); // close drawer
 DrawerNavigator(RouteConfigs, DrawerNavigatorConfig)
 ```
 
-### Drawer Navigator Options
+### RouteConfigs
+
+The route configs object is a mapping from route name to a route config, which tells the navigator what to present for that route, see [example](https://github.com/coodoo/react-navigation/blob/master/docs/api/navigators/StackNavigator.md#routeconfigs) from `StackNavigator`.
```
ericvicenti: This link should point to the main repo.

Also, isn't there someplace else where route configs are documented? We should link there

ElliottJRo:
woah woops, i pulled from master and pushed to my repo so this PR is now all wack. Closing this PR.

PR #120, opened 2017-01-31
Use redux constants and action creator functions

### LINKED ISSUE #40: Move addNavigationHelpers functions to an action creators file
When using with Redux we have to manually create the objects to dispatch:

```js
dispatch({ type: 'Navigate', routeName: 'Home' });
```

This make us dependent on the `type` string value key.

I took a look the source code and found that [`addNavigationHelpers.js`](https://github.com/react-community/react-navigation/blob/master/src/addNavigationHelpers.js) have the so called _Action creators_.

It might me helpful to move those functions to a separate file and include them on `addNavigationHelpers.js` but also expose them to be imported on custom reducers.

```js
export const navigate = (routeName, params, action) => ({
  type: 'Navigate',
  routeName,
  params,
  action,
});

// ...
```

Posible usage on custom reducers:
```js
import * as actions from 'react-navigation/actions';

function reducer(state = initialState, action) {
  switch (action.type) {
    case LOGIN_SUCCESS:
      return AppNavigator.router.getStateForAction(actions.navigate('Home'), state); 
   default:
      return AppNavigator.router.getStateForAction(action, state);
  }
}
```

jcarbo:
First I just to want to say thanks for publishing this library! It feels really well thought-out and it's been easy to use.

----

> Explain the **motivation** for making this change. What existing problem does the pull request solve?

The way redux actions are currently implemented in the codebase is brittle and error prone and doesn't follow the "best practices" espoused by the redux community and other redux-related packages. Namely:
- redux action names are very general (e.g. 'Reset', 'Init') and likely to collide with other libraries or general redux usage.
- referring to redux action names as strings throughout the codebase instead of setting them as constants (http://redux.js.org/docs/basics/Actions.html#actions)
- manually creating the action object (`{ type, payload }`) instead of using action creators (http://redux.js.org/docs/basics/Actions.html#action-creators)

This PR addresses all of those issues. This library now exports the action type constants and action creators for each type and uses a `'navigation/'` namespace for the types to avoid collisions.

_NOTE: I just noticed the "common navigation" spec - cool idea! I still think these actions should be namespaced though and I think the `navigation` is pretty general._

**Test plan (required)**

> Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.
> The code must pass tests and shouldn't add more Flow errors.

All the existing tests are still passing including new ones for the actions and action creators.

**Code formatting**

> Look around. Match the style of the rest of the codebase.

I followed the coding patterns within the codebase.


ericvicenti:
In general I am in support of this change, but it is very breaky. This will break for people who use the existing action names, but I do agree that this redux-style action creators are preferable for those who aren't using flow.

So I'm curious if anybody else feels strongly about this.

jcarbo:
Glad you support the change! I'm certainly aware this is breaky. I figure that given this has only been published for ~5 days and it's technically a beta release (`1.0.0.beta.1`) it'd be more acceptable - better now than when out of beta. That being said, I defer to the contributors here.

One thought: This library only consumes the actions in `getStateForAction` in `StackRouter` and `TabRouter`. We could add a deprecation warning (console.warn) that maps old action types to new ones with a note that says once out of beta the old action types are going to be officially deprecated and to use the action creators instead.  This assumes there will be at least one more `beta` release and I think it's a reasonable compromise.

jcarbo:
@ericvicenti - Just pushed a commit to prevent most of the breakage. So now:
- internally this library will consume/dispatch the new action types.
- when an action with an old type is dispatched from userland we'll map it to the corresponding new type internally (e.g. `Init` becomes `navigation/INIT`) along with a warning, e.g.:
> The action type 'Init' has been renamed to 'navigation/INIT'. 'Init' will continue to work while in beta but will be removed in the first major release. Moving forward, you should use the action constants and action creators exported by this library in the 'actions' object. See https://github.com/react-community/react-navigation/pull/120 for more details.

The only thing this doesn't handle is a case where a user has a custom router with a custom `getStateForAction` that is listening for the old actions. This will break since we're no longer dispatching the old actions.

ericvicenti:
Thanks for adding the informative warning, thats an excellent approach!

Have you run flow on these changes? I know there are a bunch of existing flow issues, but I want to make sure this doesn't introduce more issues.

jcarbo:
There were 110 flow errors on master, 107 on this branch. I also confirmed the playground builds both worked fine but couldn't get the web running locally. The contributing docs refer to a 'prod' npm task that doesn't seem to exist.

Steviey:
Once react-navigation is Redux-best practice confirmed, I will start with app coding.
Otherwise there is to much refactoring involved for me later.
**Please merge as soon as possible.** Thank you for all till now.

Image of Yaktocat

docs/api/routers/Routers.md
```diff
@@ -78,7 +78,7 @@ const MyAppRouter = {
   getStateForAction(action, state) {
     if (
       state &&
-      action.type === 'Back' &&
+      action.type === actions.BACK &&
```
satya164: This is purely visual. But I think we should format the names as `NavigationActions.Back` etc. All lowercase prefix and uppercase action name looks very weird.

docs/api/routers/Routers.md
```diff
@@ -97,6 +97,8 @@ Perhaps your app has a unique URI which the built-in routers cannot handle. You
 
 ```js
 
+import { actions } from 'react-navigation'
```
satya164: Maybe name it as `Actions`

docs/guides/Common-Navigation-Spec.md
```diff
@@ -56,8 +56,8 @@ A navigation-aware component that hosts other navigation-aware components. Most
 The navigation prop should be provided to components who need access to navigation. If provided, it must follow this interface:
 
 ```javascript
-type BackAction = {type: 'Back'};
-type URIAction = {type: 'URI', uri: string};
+type BackAction = {type: 'navigation/BACK'};
```
satya164: Let's make it `Navigation/Back` ? Again, all lowercase prefix and all uppercase name looks weird

ericvicenti:
Ok, I think this is pretty much ready to land.

@satya164, I agree about the capitalization. Want to clean that up and then ship this thing?

satya164:
@ericvicenti Sounds good. If @jacrbo doesn't manage to do it sometime, I can take it up :)

jcarbo:
@ericvicenti @satya164 - Currently the constants are all caps (`BACK`, `SET_PARAMS`) and the action creators are camelCased (`back`, `setParams`). I'm afraid changing the constants to pascal case (`Back`, `SetParams`) may make them look too similar to the action creators:

```js
NavigationActions.Back === 'Navigation/Back'
NavigationActions.back() === { type: 'Navigation/Back' }

NavigationActions.SetParams === 'Navigation/SetParams'
NavigationActions.setParams() === { type: 'Navigation/SetParams' }
```

Two points in favor of ALL_CAPS_SNAKE_CASE:
- that's how action types are typically formatted in the redux docs and almost every redux library I've used.
- it's a very general programming pattern in many different languages to format constant values in that way. Pascal case is typically used for classes or singletons.

I'd be in favor of:
```js
NavigationActions.BACK === 'Navigation/BACK'
NavigationActions.back() === { type: 'Navigation/BACK' }

NavigationActions.SET_PARAMS === 'Navigation/SET_PARAMS'
NavigationActions.setParams() === { type: 'Navigation/SET_PARAMS' }
```

... but ultimately defer to you guys. Let me know what you think and I'll update the PR real quick.

satya164:
I'm fine with that too 👍

examples/HelloHybrid/HybridContainer.js
```diff
@@ -7,6 +7,7 @@ import {
   NativeModules,
 } from 'react-native';
 import {
+  actions,
```
satya164: typo?

jcarbo:
I pushed up a commit but missed a few imports. Just force-pushed so I think what' up there should be good to go 👍

satya164:
LGTM. @ericvicenti should I merge?

hilkeheremans:
Great job!

Suggest we close #40, #172 and #153.

PR #122, opened 2017-01-31
Docs: add info about calling nav from same level component

sagivo:
giving more info for typical use case when you want to invoke navigation from top-level component. 
More info and request to add documentation can be found [here](https://github.com/react-community/react-navigation/issues/71)


ericvicenti:
Can you add a note to mention that this should only be used on the top level navigator?

sagivo:
the title should say it but added another line to make sure.

PR #128, opened 2017-02-01
Support label as component in TabNavigator

Ehesp:
>Explain the **motivation** for making this change. What existing problem does the pull request solve?

The tabBar navigationOptions only allow a label to be a string. When an app requires more complex layout, the label needs to match, for example having multiline text. This change solves that by allowing a component to be passed through as the `label` component.

```
function A() {
  return <Text>Page</Text>;
}

function WithComponent() {
  return <Text>Page</Text>;
}

WithComponent.navigationOptions = {
  tabBar: {
    label: (
      <View>
        <Text>SAT</Text>
        <Text>28</Text>
      </View>
    )
  },
}

const Test = TabNavigator({
  A: { screen: WithComponent },
  B: { screen: A },
  C: { screen: A },
  D: { screen: A },
}, {
  tabBarOptions: {
    scrollEnabled: true,
    tabStyle: {
      width: 50,
      height: 50,
      padding: 0,
    }
  },
})
```

Output (example has more screens):

image

Any issues/changes let me know.

Cheers

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -79,11 +79,7 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
         </Animated.Text>
       );
     }
-    return (
-      <Animated.View style={labelStyle}>
-        {label}
-      </Animated.View>
-    );
+    return label;
   };
 
   _renderIcon = (scene: TabScene) => {
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -71,10 +71,18 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
       inputRange,
       outputRange,
     });
+    const label = this.props.getLabelText(scene);
+    if (typeof label === 'string') {
+      return (
+        <Animated.Text style={[styles.label, { color }]}>
+          {label}
+        </Animated.Text>
+      );
+    }
     return (
-      <Animated.Text style={[styles.label, { color }]}>
-        {this.props.getLabelText(scene)}
-      </Animated.Text>
+      <Animated.View style={labelStyle}>
+        {label}
+      </Animated.View>
     );
   };
 
--- a/src/views/TabView/TabBarTop.js
+++ b/src/views/TabView/TabBarTop.js
@@ -78,10 +78,17 @@ export default class TabBarTop extends PureComponent<DefaultProps, Props, void> 
       outputRange,
     });
     const label = this.props.getLabelText(scene);
+    if (typeof label === 'string') {
+      return (
+        <Animated.Text style={[styles.label, labelStyle, { color }]}>
+          {label}
+        </Animated.Text>
+      );
+    }
     return (
-      <Animated.Text style={[styles.label, labelStyle, { color }]}>
-        {upperCaseLabel && label ? label.toUpperCase() : label}
-      </Animated.Text>
+      <Animated.View style={labelStyle}>
+        {label}
+      </Animated.View>
     );
   };
 
--- a/src/views/TabView/TabBarTop.js
+++ b/src/views/TabView/TabBarTop.js
@@ -85,11 +85,8 @@ export default class TabBarTop extends PureComponent<DefaultProps, Props, void> 
         </Animated.Text>
       );
     }
-    return (
-      <Animated.View style={labelStyle}>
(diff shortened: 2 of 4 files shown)


src/views/TabView/TabView.js
```diff
@@ -86,7 +86,7 @@ class TabView extends PureComponent<void, Props, void> {
 
   _getLabelText = ({ route }: TabScene) => {
     const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[route.key], 'tabBar');
-    if (tabBar && typeof tabBar.label === 'string') {
+    if (tabBar && tabBar.label) {
```
satya164: change to `(tabBar && typeof tabBar.label === 'undefined')`

satya164:
Can you also do the change in `TabBarBottom`?

src/views/TabView/TabView.js
Ehesp: `(tabBar && typeof tabBar.label !== 'undefined')` you mean?

Ehesp:
@satya164 done 👍

src/views/TabView/TabBarTop.js
```diff
@@ -78,10 +78,17 @@ export default class TabBarTop extends PureComponent<DefaultProps, Props, void>
...
     });
     const label = this.props.getLabelText(scene);
+    if (typeof label === 'string') {
+      return (
+        <Animated.Text style={[styles.label, labelStyle, { color }]}>
+          {label}
+        </Animated.Text>
+      );
+    }
     return (
```
satya164: I think this can just be `return label`. `labelStyle` is supposed to be text style

src/views/TabView/TabBarBottom.js
```diff
@@ -71,10 +71,18 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
...
     });
+    const label = this.props.getLabelText(scene);
+    if (typeof label === 'string') {
+      return (
+        <Animated.Text style={[styles.label, { color }]}>
+          {label}
+        </Animated.Text>
+      );
+    }
     return (
```
satya164: I think this can just be `return label`. `labelStyle` is supposed to be text style

satya164:
Thanks. Just two more small changes and we'll merge.

src/views/TabView/TabBarBottom.js
Ehesp: @satya164 Do you mean like:

```
if (typeof label === 'string') {
  return label;
``` 

Won't this be missing the default styles though?

satya164: @Ehesp No string is fine. If a custom component is specified, we should just return it without wrapping in a view.

Also `labelStyle` is for a text label, so you can't apply them to a view.

Ehesp: Isn't the default value (the `routeName`) passed down a string though, meaning by default it'll lose all of the styling? My thought was:

```
label: null // Render the routeName
label: 'Custom Label' // Render a custom string but with default styling
label: <Text>Custom Component</Text> // Render it inside Animated.View
```

With your proposed changes, the first two will lose all default styling. Unless I'm being an idiot :D

satya164: Not sure what you mean, nothing changes when it's string,

```js
    if (typeof label === 'string') {
      return (
        <Animated.Text style={[styles.label, { color }]}>
          {label}
        </Animated.Text>
      );
    }
    return label;
```

When you're specifying a component, we don't need to wrap it in an `Animated.View`, it doesn't serve any purpose. Also `labelStyle` is only valid on `Text` elements and you'll get warnings when you pass to `Animated.View`. And `labelStyle` is the styles passed by the user, not the default styles.

Ehesp: Sorry man, I thought you meant something different. Pushed.

jhilden:
Thanks for this PR.

One question about this:  When providing a custom label component, is there a way to access the current `tintColor` for styling the text as active etc.?  Like it is possible for providing your own icon component (as seen in the [docs](https://reactnavigation.org/docs/navigators/tab)).
Or alternatively, is there a way to determine if the current view is active somehow?

satya164:
I'll do some changes to allow this soon

hose314:
is this thread is active?
i made PR https://github.com/react-community/react-navigation/pull/524, probably related to @jhilden question

Issue #80, opened 2017-01-29
Params with DrawerNavigator


TomiHiltunen:
DrawerNavigator doesn't seem to accept params as the second argument for `props.navigation.navigate`. I'm using custom drawer menu component. Problem is, that it does navigate to the view but the params are not there.

Here's how I'm using the `navigation` prop provided to the custom drawer menu component by `DrawerNavigator`:
```
const DrawerMenuView = (props) => {
    const {
        myLittleItems,
        navigation
    } = props;
    return (
        <View>
            {myLittleItems.map((myLittleItem) =>
                <MyCustomItemComponent
                    key={myLittleItem.uuid}
                    myLittleItem={myLittleItem}
                    onPress={() => {
                        navigation.navigate("MyItem", { "displayName": myLittleItem.displayName });
                    }} />
            )}
        </View>
    );
};
```

Is this by design or a bug?

ps. This is one of my attempts at getting dynamically created navigation items in the drawer menu.

baderahmed:
I have the same problem !

satya164:
AFAIK `TabRouter` (which `DrawerNavigation` uses) doesn't support params currently.

cc @ericvicenti

jeffreymendez1993:
This happens not just for the DrawerStack it happens when you try to pass data from into a new stack of any type. Current issue is that the params object is not being passed between stacks and we should easily be able to implement this into React-Navigation. @grabbou

uxman-sherwani:
Any updates on this issue? Still facing it with "react-navigation": "^1.0.0-beta.5"

No params when using DrawerNavigator
![screenshot from 2017-02-24 16-14-22](https://cloud.githubusercontent.com/assets/3153380/23308786/989f0c54-faac-11e6-8bd7-0c8ae3cd6dbf.png)

grabbou:
Working on a fix over the weekend. We will try to land it to next release

hasLandon:
I just upgraded to beta 6 and as far as I can tell, the issue is still there

coodoo:
Can confirm this is still happening in branch https://github.com/react-community/react-navigation/commit/af787e56fe05183ae1e1e0c26899b8ae76e318d7

renanpupin:
Have save problem :(

Kerumen:
I have the same problem with a `TabNavigator`.

Did you had time to work on your fix @grabbou ?

JulianKingman:
Can I help somehow?

[...]

PR #134, opened 2017-02-01
apply params state change during navigate in TabRouter

danj3:
Related to Issue #80

`params` is not merged into state during the Navigation dispatch, so is unavailable to the Screen. I make no claim of understanding this code base, however, running through the nav transitions in the debugger led me to conclude that this is simply not handled. If it is handled then there is a documentation hole explaining how it is supposed to work, though I don't know what that would be.

**Test**

Object transition during action with params, attached  png

navigate_params

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -129,6 +129,10 @@ export default (
       }
       if (activeTabIndex !== state.index) {
         // console.log(`${order.join('-')}: Normal navigation`, {lastIndex: state.index, newIndex: activeTabIndex});
+        if ( action.params ) {
+          state = {...state, routes: [...state.routes]};
+          state.routes[activeTabIndex] = {...state.routes[activeTabIndex], params: action.params};
+        }
         return {
           ...state,
           index: activeTabIndex,


src/routers/TabRouter.js
```diff
@@ -129,6 +129,10 @@ export default (
       }
       if (activeTabIndex !== state.index) {
         // console.log(`${order.join('-')}: Normal navigation`, {lastIndex: state.index, newIndex: activeTabIndex});
+        if ( action.params ) {
```
ericvicenti: Nit: no spaces around the inside of if statements

src/routers/TabRouter.js
```diff
@@ -129,6 +129,10 @@ export default (
...
+          state = {...state, routes: [...state.routes]};
+          state.routes[activeTabIndex] = {...state.routes[activeTabIndex], params: action.params};
```
ericvicenti: I think we shouldn't change the route unless the params have actually changed. Also, I think we may want to merge the params rather than overwrite them

satya164:
Any updates on this @danj3 ?

danj3:
I hope to attend to the changes shortly.

Radu-Filip:
any progress on this?

ericvicenti:
This is a pretty simple and high-pri fix. @danj3, any updates? If anybody else wants to address the feedback and submit a PR for this, we can land it sooner rather than later

mgtitimoli:
I've just found this one :expressionless: ...

I've also resolved #80 on [this one](https://github.com/react-community/react-navigation/pull/969). Hope it lands to master soon.

PR #137, opened 2017-02-01
Add withNavigation HOC

### LINKED ISSUE #61: add withNavigation HOC, and export navigation context from SceneView
Can we get access to the global dispatch/nagivate/etc functions, so that we don't have to pass them through our component trees?

jakubsta:
This PR resolve #61 

- [x] Expose `navigation` from `SceneView`
- [x] Create HOC `withNavigation`
- [x] Provide example
- [x] Add docs

### DIFF
--- a/docs/api/withNavigation.md
+++ b/docs/api/withNavigation.md
@@ -1,0 +1,17 @@
+
+# withNavigation
+
+[`withNavigation`](/src/views/withNavigation.js) is a Higher Order Component which passes the `navigation` prop into wrapped Component. It's useful when you cannnot pass the `navigation` prop into into the component directly, or don't want to pass it in case of a deeply nested child.
+
+## Example
+
+```js
+import { Button } 'react-native';
+import { withNavigation } from 'react-navigation';
+
+const MyComponent = ({ to, navigation }) => (
+    <Button title={`navigate to ${to}`} onPress={() => navigation.navigate(to)} />
+);
+
+const MyComponentWithNavigation = withNavigation(MyComponent);
+```
--- a/examples/NavigationPlayground/js/SimpleTabs.js
+++ b/examples/NavigationPlayground/js/SimpleTabs.js
@@ -11,21 +11,20 @@ import {
 } from 'react-native';
 import {
   TabNavigator,
+  withNavigation,
 } from 'react-navigation';
 import Ionicons from 'react-native-vector-icons/Ionicons';
 import SampleText from './SampleText';
 
+const ButtonWithNavigation = withNavigation(({ navigation, to, ...rest}) => {
+  return <Button {...rest} onPress={() => navigation.navigate(to)} />;
+});
+
 const MyNavScreen = ({ navigation, banner }) => (
   <ScrollView style={styles.container}>
     <SampleText>{banner}</SampleText>
-    <Button
-      onPress={() => navigation.navigate('Home')}
-      title="Go to home tab"
-    />
-    <Button
-      onPress={() => navigation.navigate('Settings')}
-      title="Go to settings tab"
-    />
+    <ButtonWithNavigation to="Home" title="Go to home tab" />
+    <ButtonWithNavigation to="Settings" title="Go to settings tab" />
     <Button
       onPress={() => navigation.goBack(null)}
       title="Go back"
--- a/examples/NavigationPlayground/js/SimpleTabs.js
+++ b/examples/NavigationPlayground/js/SimpleTabs.js
@@ -11,20 +11,21 @@ import {
 } from 'react-native';
 import {
   TabNavigator,
-  withNavigation,
 } from 'react-navigation';
 import Ionicons from 'react-native-vector-icons/Ionicons';
 import SampleText from './SampleText';
 
-const ButtonWithNavigation = withNavigation(({ navigation, to, ...rest}) => {
-  return <Button {...rest} onPress={() => navigation.navigate(to)} />;
-});
-
 const MyNavScreen = ({ navigation, banner }) => (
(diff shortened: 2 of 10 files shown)


src/views/withNavigation.js
```diff
@@ -0,0 +1,17 @@
+/* @flow */
+
+import React from 'react';
+
+import { ContextWithNavigation } from '../TypeDefinition';
+
+export default function (Component: ReactClass<T>) {
```
satya164: Can you also set displayName ? https://github.com/react-community/react-navigation/blob/master/src/withCachedChildNavigation.js#L25

satya164: nit: unenecessary space after function

src/views/withNavigation.js
```diff
@@ -0,0 +1,17 @@
...
+  const withNavigation = (props: T, { navigation }: ContextWithNavigation) => (
```
satya164: Can you also set displayName? https://github.com/react-community/react-navigation/blob/master/src/withCachedChildNavigation.js#L25

And the function name should be something else, coz `withNavigation` is the name of the HOC

src/views/SceneView.js
```diff
@@ -17,6 +18,16 @@ type Props = {
 export default class SceneView extends PureComponent<void, Props, void> {
   props: Props;
```
satya164: nit: move props below static properties

examples/NavigationPlayground/js/SimpleTabs.js
```diff
@@ -11,21 +11,20 @@ import {
 } from 'react-native';
 import {
   TabNavigator,
+  withNavigation,
 } from 'react-navigation';
 import Ionicons from 'react-native-vector-icons/Ionicons';
 import SampleText from './SampleText';
 
+const ButtonWithNavigation = withNavigation(({ navigation, to, ...rest}) => {
```
satya164: Not sure we want this in simple tabs example cc @mkonicek

ericvicenti:
Exciting, thanks for the contribution! Happy to see this merged once the minor feedback is addressed.

examples/NavigationPlayground/js/SimpleTabs.js
ericvicenti: I think we should keep this out of the simple example, but add it to multiple of the other examples, and/or add new examples. Lets do our best to not pollute the simple examples, and when it makes sense, make new examples rather than making the existing ones endlessly more complicated.

jakubsta:
@satya164 I've just pushed fixes for your comments.

examples/NavigationPlayground/js/SimpleTabs.js
jakubsta: @satya164 @ericvicenti You are right. I reverted changes in this example and created a new one.

satya164:
Cool! Let's add docs and shipit

docs/api/withNavigation.md
```diff
@@ -0,0 +1,14 @@
+# withNavigation
+[`withNavigation`](/src/views/withNavigation.js) is a High Order Component which injects a `navigation` into wrapped Component. It's useful when you can't/don't want pass `navigation` explicitly into nested component.
```
grabbou: nit: `higher`

grabbou: nit: `to pass`

noung1234:
+# withNavigation

docs/api/withNavigation.md
jakubsta: @grabbou eagle eye, fixed

docs/api/withNavigation.md
```diff
@@ -0,0 +1,14 @@
+# withNavigation
+[`withNavigation`](/src/views/withNavigation.js) is a Higher Order Component which injects a `navigation` into wrapped Component. It's useful when you can't/don't want to pass `navigation` explicitly into nested component.
+## Example
```
satya164: Put a new line before the title

docs/api/withNavigation.md
```diff
@@ -0,0 +1,14 @@
+# withNavigation
```
satya164: Put a new line after the title

docs/api/withNavigation.md
```diff
@@ -0,0 +1,14 @@
...
+[`withNavigation`](/src/views/withNavigation.js) is a Higher Order Component which injects a `navigation` into wrapped Component. It's useful when you can't/don't want to pass `navigation` explicitly into nested component.
```
satya164: perhaps

```
which passes the `navigation` prop
```

is better. also,

> It's useful when you cannot pass the `navigation` prop into the component directly, or don't want to pass it in case of a deeply nested child.

src/views/withNavigation.js
```diff
@@ -0,0 +1,19 @@
...
+    <Component {...props} navigation={navigation} />
+  );
+
+  componentWithNavigation.displayName = `withNavigation(${Component.displayName || Component.name})`;
+
+  componentWithNavigation.contextTypes = {
+    navigation: React.PropTypes.object.isRequired,
+  };
+
+  return componentWithNavigation;
```
sibelius: missing hoistStatics like this: https://github.com/exponent/ex-navigation/blob/79a7a5b87007bffe35b0947ac23996b20be9d6e1/src/ExNavigationComponents.js#L199

satya164: @jakubsta can you add that?

jakubsta: Sure, added.

satya164:
@jakubsta you need to update the `package.json` and `yarn.lock` too

PR #144, opened 2017-02-01
Fix back button tinting

lnikkila:
Tinting was just added in #109. In a related issue [it was decided][1] that that the `tintColor` option should control back button tinting as well.

However `<Image>` takes `tintColor` [as a style][0] and not as a prop, so currently the option has no effect on the button’s colour.

Please let me know if any changes are required. Thanks!

[0]: https://facebook.github.io/react-native/releases/0.41/docs/image.html#style
[1]: https://github.com/react-community/react-navigation/issues/36#issuecomment-276294877

### DIFF
--- a/src/views/HeaderBackButton.js
+++ b/src/views/HeaderBackButton.js
@@ -27,9 +27,8 @@ const HeaderBackButton = ({ onPress, title, tintColor }: Props) => (
   >
     <View style={styles.container}>
       <Image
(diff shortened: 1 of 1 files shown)


guigrpa:
Do you know if there is a way to tint the back button and label, but not the header title (center element)? In other words, replace the standard blue by another color, but leaving the title unchanged.

PR #150, opened 2017-02-02
Fix issue where initialRouteParams were not set

juhasuni:
Currently initialRouteParams option has no effect when using StackRouter.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -50,7 +50,7 @@ export default (
   });
 
   const {
-    initialRouteParams = {},
+    initialRouteParams,
   } = stackConfig;
 
   const initialRouteName = stackConfig.initialRouteName || routeNames[0];
@@ -113,9 +113,10 @@ export default (
             params: initialRouteParams,
           }));
         }
-        const params = (route.params || action.params) && {
+        const params = (route.params || action.params || initialRouteParams) && {
           ...(route.params || {}),
           ...(action.params || {}),
+          ...(initialRouteParams || {}),
         };
         route = {
           ...route,
--- a/src/routers/__tests__/StackRouter-test.js
+++ b/src/routers/__tests__/StackRouter-test.js
@@ -384,6 +384,26 @@ describe('StackRouter', () => {
     });
   });
 
+  test('Initial route params appear in nav state', () => {
+    const FooScreen = () => <div />;
+    const router = StackRouter({
+      Foo: {
+        screen: FooScreen,
+      },
+    }, { initialRouteName: 'Bar', initialRouteParams: { foo: 'bar' } });
+    const state = router.getStateForAction({ type: NavigationActions.INIT });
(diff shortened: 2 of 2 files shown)


src/routers/StackRouter.js
```diff
@@ -113,6 +113,8 @@ export default (
             routeName: initialRouteName,
             params: initialRouteParams,
           }));
+        } else {
+          route.params = initialRouteParams;
```
satya164: do `route = { ...route, params: initialRouteParams }` instead of mutating

grabbou:
I believe we could generally do some refactoring here if anyone is up to.

E.g. we could move this https://github.com/valu-digital/react-navigation/blob/c73482135558440a4d2bf8748a13f5caeb21e9d6/src/routers/StackRouter.js#L98 down the first `if` because it's not used.

Then, instead of using `let route` and if/else we could use just tenary operator.

juhasuni:
Actually, I think that my original commit has a bigger flaw. If there IS an `initialChildRouter` the initial route params are not set at all. This is because I couldn't quite follow the logic of all this, but now when I took a closer look I realized, that the `initialChildRouter.getStateForAction` returns a state of form `{index: number, routes: Route[]}`, which doesn't set the route params for the main route. I'm sending an update that fixes this issue.

juhasuni:
I changed what used to be `let route ...` to `let childState ...` as it better describes the logic. I didn't use tenary operators here, because I think that `if` statement is more readable. 

After child state is merged, the `route` is something like:
```
{
  index: 0, // this comes from the child state
  routes: [...], // this comes from the child state
  routeName: 'myroute',
  params: {...},
  key: 'Init'
}
```

This is how it used to work (except that `params` is now set), I hope you know what it's supposed to do :)

satya164:
cc @ericvicenti

ruimgoncalves:
I don't know if this is the place to do this comment, but here it goes.

I'm using version 1.0.0-beta.5 and for me this problem still persists, I cannot manage to get params from initialRouteParams.

I'm not using the StackNavigator as the root component, but rather, I only show this multi-step login Component if a user is not loggedIn.

despreston:
any update on this? seeing this in beta v6 as well.

juhasuni:
There you go. Tests (with one new) pass now.

satya164:
@juhasuni there are conflicts. can you rebase it against master?

juhasuni:
Conflicts? I don't see any.

satya164:
image

joshuapinter:
Hey @juhasuni, thanks for identifying this issue and then creating a PR to resolve it.

I wasn't able to get this to work, even when pulling from your master branch of the library.

What I'm trying to do is pull in a prop set in `launchOptions` when the app starts and then have that accessible in my initial scene. 

I created an issue to tackle this directly as well at #634.

Any help would be really great. 

Thanks!

Joshua

Issue #136, opened 2017-02-01
Hide TabBar


microwavesafe:
With DrawerNavigation we get two nice routes
```
this.props.navigation.navigate('DrawerOpen'); // open drawer
this.props.navigation.navigate('DrawerClose'); // close drawer
```
In my app I would like to hide the TabBar while the keyboard is shown and the above functionality would be very useful. Is it something you would consider a pull request for, or is there a better way of doing it that I haven't spotted?

ericvicenti:
We currently have a `header.visible` navigation option for the `StackNavigator`.

I think it would make sense to add a `tabBar.visible` option for the `TabNavigator`. A PR would be accepted for that.

@satya164

satya164:
@ericvicenti am I right when I say that `tabBar.visible` won't be animated?

microwavesafe:
In my particular case I don't need the tabBar to animate.

ericvicenti:
Yeah I think we can skip the animation for it. Maybe somebody will want it one day, but I think we can help @microwavesafe's use case pretty easily

satya164:
Adding good first task label then. Should be easy to implement.

microwavesafe:
@satya164 @ericvicenti 

I've got this roughly working, I've implemented the visible check in the render function of TabView.js.
```
  render() {
    const {
      navigation,
      tabBarComponent,
      tabBarPosition,
      animationEnabled,
      lazyLoad,
    } = this.props;

    let renderHeader;
    let renderFooter;

    const { state } = this.props.navigation;
    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].routeName], 'tabBar');

    if ((typeof tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
      if (tabBarPosition === 'bottom') {
        renderFooter = this._renderTabBar;
      } else {
        renderHeader = this._renderTabBar;
      }
    }

    let configureTransition;

    if (animationEnabled === false) {
      configureTransition = this._configureTransition;
    }

    return (
      <TabViewAnimated
        style={styles.container}
        navigationState={navigation.state}
        lazy={lazyLoad}
        renderHeader={renderHeader}
        renderFooter={renderFooter}
        renderScene={this._renderScene}
        renderPager={this._renderPager}
        configureTransition={configureTransition}
        onRequestChangeTab={this._handlePageChanged}
      />
    );
  }
}
```
[...]

PR #132, opened 2017-02-01
Add SetParams handling in TabRouter.js

microwavesafe:
Following on from #126 setParams in TabNavigator was not working. I believe this is because it is not handled in the getStateForAction function in TabRouter.js

The following code works with my changes
```js
import React from 'react';
import {
    AppRegistry,
    Text,
    View,
    Button,
    StyleSheet
} from 'react-native';
import { TabNavigator, TabView } from 'react-navigation';
import Ionicons from 'react-native-vector-icons/Ionicons';
import IconBadge from './IconBadge'

class Home extends React.Component {

    static navigationOptions = {
        tabBar: ({ state }) => ({
            label: 'Home',
            icon: ({ tintColor, focused }) =>
                <IconBadge
                MainElement={
                    <Ionicons
                    name={focused ? 'ios-home' : 'ios-home-outline'}
                    size={26}
                    style={{ color: tintColor }}
                    />
                }
                badgeNumber={typeof state.params === 'undefined' ? 0 : state.params.badgeCount}
                />,
        }),
    }

    render() {
        const {navigation} = this.props;

        return (
            <View>
                <Button
                    onPress={() => navigation.setParams({badgeCount : 1})}
                    title="Set Badge Number"
                />
            </View>
        );
    }
}

class Settings extends React.Component {
    static navigationOptions = {
        tabBar: {
            label: 'Settings'
        },
    }

    render() {
        const { navigate } = this.props.navigation;

        return (
            <Button
                onPress={() => this.props.navigation.goBack()}
                title="Go back"
            />
        );
    }
}

const MyApp = TabNavigator({
    Home: {
        screen: Home,
    },
    Settings: {
        screen: Settings,
    },
}, {
    ...TabNavigator.Presets.iOSBottomTabs,
    tabBarOptions: {
        activeTintColor: '#e91e63',
    },
});

AppRegistry.registerComponent('ReactNavigation', () => MyApp);
```
And fails without them. I'm presuming this is how you intended it to work, so please dismiss if I have misunderstood.

PR #152, opened 2017-02-02
Add visible param to TabBar

microwavesafe:
This implements the ability to hide the tabBar by setting visible = false

It can be turned on in the navigationOptions
```
    static navigationOptions = {
        tabBar: ({ state }) => ({
            label: 'Messaging',
            visible: false,
        }),
    };
```
or accessed through SetParams (when #132 or similar functionality is implemented). See #136 for discussion.

### DIFF
--- a/docs/guides/Screen-Nav-Options.md
+++ b/docs/guides/Screen-Nav-Options.md
@@ -98,21 +98,26 @@ The 2nd argument passed to the function are the default values for the `header` 
 
 ```js
 class TabScreen extends React.Component {
+
   static navigationOptions = {
-    label: 'Tab Label',
-    icon: ({ tintColor }) => (
-      <Image
-        source={require('./tab-icon.png')}
-        style={[styles.icon, {tintColor: tintColor}]}
-      />
-    ),
-    visible: true
-  }
+    tabBar: ({ state }) => ({
+      label: 'Tab Label',
+      icon: ({ tintColor }) => (
+        <Image
+          source={require('./tab-icon.png')}
+          style={[styles.icon, {tintColor: tintColor}]}
+        />
+      ),
+      visible: true
+    }),
+  };
+
+};
 ```
 
 - `label` - can be string or react component
 - `icon` - function that returns icon component
-- `visible` - true or false to show or hide the tab bar
+- `visible` - true or false to show or hide the tab bar, if not set then defaults to true
 
 ## Stack Navigation Options
 
--- a/docs/guides/Screen-Nav-Options.md
+++ b/docs/guides/Screen-Nav-Options.md
@@ -96,7 +96,23 @@ The 2nd argument passed to the function are the default values for the `header` 
 
 ## Tab Navigation Options
 
-Coming Soon
+```js
+class TabScreen extends React.Component {
+  static navigationOptions = {
+    label: 'Tab Label',
+    icon: ({ tintColor }) => (
+      <Image
+        source={require('./tab-icon.png')}
+        style={[styles.icon, {tintColor: tintColor}]}
+      />
+    ),
+    visible: true
+  }
+```
+
+- `label` - can be string or react component
+- `icon` - function that returns icon component
+- `visible` - true or false to show or hide the tab bar
 
 ## Stack Navigation Options
 
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -160,7 +160,7 @@ class TabView extends PureComponent<void, Props, void> {
     const { state } = this.props.navigation;
     const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
 
-    if ((tabBarComponent !== undefined) && (tabBar) && (tabBar.visible !== false)) {
+    if (tabBarComponent !== undefined && tabBar && tabBar.visible !== false) {
       if (tabBarPosition === 'bottom') {
         renderFooter = this._renderTabBar;
       } else {
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -158,7 +158,7 @@ class TabView extends PureComponent<void, Props, void> {
     let renderFooter;
 
     const { state } = this.props.navigation;
-    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].routeName], 'tabBar');
+    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
 
     if ((typeof tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
       if (tabBarPosition === 'bottom') {
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -160,7 +160,7 @@ class TabView extends PureComponent<void, Props, void> {
     const { state } = this.props.navigation;
     const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
 
-    if ((tabBarComponent !== 'undefined') && (tabBar) && (tabBar.visible !== false)) {
+    if ((tabBarComponent !== undefined) && (tabBar) && (tabBar.visible !== false)) {
       if (tabBarPosition === 'bottom') {
         renderFooter = this._renderTabBar;
       } else {
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -160,7 +160,7 @@ class TabView extends PureComponent<void, Props, void> {
     const { state } = this.props.navigation;
     const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
 
-    if ((typeof tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
+    if ((tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
       if (tabBarPosition === 'bottom') {
         renderFooter = this._renderTabBar;
       } else {
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -160,7 +160,7 @@ class TabView extends PureComponent<void, Props, void> {
     const { state } = this.props.navigation;
     const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
 
-    if ((tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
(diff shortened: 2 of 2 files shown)


src/views/TabView/TabView.js
```diff
@@ -157,7 +157,10 @@ class TabView extends PureComponent<void, Props, void> {
     let renderHeader;
     let renderFooter;
 
-    if (typeof tabBarComponent !== 'undefined') {
+    const { state } = this.props.navigation;
+    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].routeName], 'tabBar');
```
satya164: I think this should be `this.props.childNavigationProps[state.routes[state.index].key]`

reference - https://github.com/react-community/react-navigation/blob/master/src/withCachedChildNavigation.js#L51

microwavesafe: Ah yes, it was working on mine because they are the same value. I will update.

microwavesafe:
Requested changes commited

microwavesafe:
I guess we want to add something to the docs as well, but alongside my #132 pull request I can turn the tab bar on and off using this code in the navigationOptions
```
   static navigationOptions = {
        tabBar: ({ state }) => ({
            label: 'Messaging',
                visible: typeof state.params === 'undefined' ? true : state.params.visible,
        }),
    };
```
and these functions to turn them on and off with the keyboard
```
    _keyboardDidShow = () => {
        this.props.navigation.setParams({ visible: false })
    };

    _keyboardDidHide = () => {
        this.props.navigation.setParams({ visible: true })
    };
```

src/views/TabView/TabView.js
```diff
@@ -157,7 +157,10 @@ class TabView extends PureComponent<void, Props, void> {
     let renderHeader;
     let renderFooter;
 
-    if (typeof tabBarComponent !== 'undefined') {
+    const { state } = this.props.navigation;
+    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
+
+    if ((typeof tabBarComponent !== 'undefined') && (tabBar.visible !== false)) {
```
ericvicenti: You can just do `tabBarComponent !== undefined`

jhilden:
Thanks for this PR @microwavesafe   This is exactly what we need.  Let me know if I can help in any way to get this merged.

microwavesafe:
The documentation is a bit sparse but hopefully it gets the point across!

docs/guides/Screen-Nav-Options.md
```diff
@@ -96,7 +96,23 @@ The 2nd argument passed to the function are the default values for the `header`
 
 ## Tab Navigation Options
 
-Coming Soon
+```js
+class TabScreen extends React.Component {
+  static navigationOptions = {
+    label: 'Tab Label',
```
jhilden: I'm not 100% sure, but don't these properties (`lable`, `icon`, `visible`) have to be nested into another property `tabBar`?

satya164: you're right. @microwavesafe can you change this?

microwavesafe:
@jhilden good catch, new docs committed.

microwavesafe:
I've just committed another change. I was playing with a stack navigator underneath a tab navigator. Doing this you get a situation where the tab navigator tries to get the tabBar options from a stack navigator page, leading to a null error. The _renderIcon function already deals with this correctly, but my function caused a null error.

I'm not sure what we want to do as a default behaviour here. My last commit will close the entire tab bar. Previously it would have rendered the tab bar, but the current icon would disappear. Which behaviour would you prefer?

microwavesafe:
I should add to display the tab bar when the page is a stack navigator, you just have to add the tabBar options as you would with a normal tab component.

src/views/TabView/TabView.js
```diff
@@ -157,7 +157,10 @@ class TabView extends PureComponent<void, Props, void> {
     let renderHeader;
     let renderFooter;
 
-    if (typeof tabBarComponent !== 'undefined') {
+    const { state } = this.props.navigation;
+    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
+
+    if ((tabBarComponent !== 'undefined') && (tabBar) && (tabBar.visible !== false)) {
```
satya164: `tabBarComponent !== 'undefined'` isn't correct, do `tabBarComponent !== undefined`

nit: remove unnecessary parenthesis

satya164:
Can you show the stack trace of the error you get?

src/views/TabView/TabView.js
```diff
@@ -157,7 +157,10 @@ class TabView extends PureComponent<void, Props, void> {
     let renderHeader;
     let renderFooter;
 
-    if (typeof tabBarComponent !== 'undefined') {
+    const { state } = this.props.navigation;
+    const tabBar = this.props.router.getScreenConfig(this.props.childNavigationProps[state.routes[state.index].key], 'tabBar');
+
+    if ((tabBarComponent !== undefined) && (tabBar) && (tabBar.visible !== false)) {
```
satya164: Please remove the unnecessary parenthesis so we can merge

microwavesafe:
I realise that strict equality and inequality has a higher operator precedence than logical AND so I can remove all of the internal parenthesis, but I would have thought this would make the code harder to read?

Anyway can you clarify if you want
```
if ((tabBarComponent !== undefined) && tabBar && (tabBar.visible !== false))
```
OR
```
if (tabBarComponent !== undefined && tabBar && tabBar.visible !== false)
```
so I don't have to commit again if I get it wrong.

satya164:
Too many parenthesis are harder to read :D

But anyways, yeah that's correct

microwavesafe:
Which one is correct, the first or second one?

satya164:
`if (tabBarComponent !== undefined && tabBar && tabBar.visible !== false)`

microwavesafe:
Thought so but wanted to be sure! I'll do it now.

starchand:
Can we update docs at https://reactnavigation.org/docs/navigators/tab
I'm confused how to hide the tabbar.

KabirKang:
@microwavesafe what about if you nest a stacknavigator in a tabnavigator and u want to make it so that only some of the screens in the underlying stacknav hide the tabs? this isn't working for me. i set up a function to get the tabnavigator from a context object and the use setparams on it, but that isn't working. whats another way to selectively hide tabs from within the lower stacknav?

microwavesafe:
I have this
```
    static navigationOptions = {
        title: 'Messaging',
        tabBar: ({ state }) => ({
            label: 'Messaging',
            icon: ({ tintColor, focused }) =>
                <MessagingIconBadge
                    MainElement={
                        <Icons
                        name={'message-text'}
                        size={32}
                        style={{ color: tintColor }}
                        />
                    }
                />,
            visible: !state.params ? true : state.params.visible,
        }),
    };
```
and these to set it within the same component

```
this.props.navigation.setParams({ visible: false })
this.props.navigation.setParams({ visible: true })
```

as long as you are accessing the correct navigation, I assume this would work elsewhere.

KabirKang:
@microwavesafe but the navigationOptions on the component would be affecting the stack navigator that's nested in the tab navigator, not the tab navigator itself right? these are the navigation options on the ProfileScreen I have that are within the stack nav:

`static navigationOptions = {
    header: ({state: {params = {}}}) => {
      const title = (typeof params.teacherName === 'undefined') ? '' : params.teacherName;
      return {
        title
      }
    }
  }`

Adding a tabBar property here doesn't seem to work. Trying to explore other ideas like maintaining nav state with redux but i dont know if i'm overthinking it.

KabirKang:
The tabs:

`this.TabNav = TabNavigator(
      HOME_TABS_ROUTER,
      {
        initialRouteName: this.initialTab,
        tabBarPosition: 'bottom',
        tabBarComponent: TabView.TabBarBottom,
        order: [HOME_TABS.FIND_TEACHER_TAB, HOME_TABS.CHAT_TAB, HOME_TABS.DASHBOARD_TAB],
        lazyLoad: true,
        swipeEnabled: false,
        animationEnabled: false,
        tabBarOptions: {
          showIcon: true,
          style: {
            backgroundColor: Colors.headerBackground,
            height: normalizeValue(65),
            padding: normalizeValue(10)
          },
          labelStyle: textStyles.footerLabel,
          activeTintColor: Colors.brandDefault,
          inactiveTintColor: Colors.defaultText
        }
      }
    );`

A child stack nav:

this.StackNavigator = Router(
      {
        headerComponent: CustomHeader,
        headerMode: 'screen',
        initialRouteName: ROUTES.FIND_TEACHER_TAB,
        navigationBarOptions: {
          header: {
            title: () => i18n.t('home.chatHeaderTitle'),
          }
        }
      }
    );

microwavesafe:
Your right, you can't use set tabBar properties from inside the stack page component. But you can use setParams to set parameters from any route to any route.

```
import { NavigationActions } from 'react-navigation'

const setParamsAction = NavigationActions.setParams({
  params: {}, // these are the new params that will be merged into the existing route params
  // The key of the route that should get the new params
  key: 'screen-123',
})
this.props.navigation.dispatch(setParamsAction)
```

as seen on this docs page

https://reactnavigation.org/docs/navigators/navigation-prop

So you'd have the visible param set up on one of your tab components, inside the tabBar options.
```
visible: !state.params ? true : state.params.visible,
```
then from the stack component call ```this.props.navigation.dispatch(setParamsAction)``` with the correct params and route set. I haven't tested this, so I can't be sure.

AlanFoster:
I'm not quite sure where to post this; But I initially couldn't get the tabBar hiding successfully using this method. In my set up I had configured a top-level TabNavigator with icon configuration, and it wasn't making use of the `navigationOptions` provided within a child within a StackNavigator that had `visible: false`

I was finally able to track down a fix for the issue; I had set `navigationOptions` within the top level TabNavigator, which was ignoring the child's tabBar configuration for some reason. This was my previous TabNavigator configuration:

```js
const DefaultTabNavigation = TabNavigator({
  Main: {
    screen: MainScreen,
    navigationOptions: {
      tabBar: () => ({
        label: 'Home',
        icon: tabIconFor('md-stats')
      })
    }
  }
```

I fixed this by making use of the defaultOptions argument:

```js
const DefaultTabNavigation = TabNavigator({
  Main: {
    screen: MainScreen,
    navigationOptions: {
      // Ensure that the defaultOptions are mixed in too, otherwise the 
      tabBar: (navigation, defaultOptions) => ({
        ...defaultOptions,
        label: 'Home',
        icon: tabIconFor('md-stats')
      })
    }
  }
```

I'm not sure is is a small API oversight or not, i.e. the TabNavigator options should have mixed in with the child's options - rather than replacing the entire configuration from a child view.

bhavyasruthi:
But now "tabBar" is deprecated only tabBarIcon is in use. How to hide tab bar now ?

nikasvan:
Any news about this?

Issue #59, opened 2017-01-28
Reset action does not seem to work


markflorisson:
I'm trying to reset the stack in a StackNavigation using:

```javascript
        dispatch({
            type: 'Reset',
            actions: [{
                type: 'Navigate',
                routeName: 'Home',
            }],
        })
```

However, this gives an error: "There should always be only one scene active, not 0".

If I try to dispatch() the reset action, followed by a navigate(), then it works but it does not "forget" any part of the history stack. How can I reset the stack?

JustinNothling:
Hey Mark, I got the same problem on my end so I'll let you know if I come up with a solution.

JustinNothling:
Ok So I have a solution:

You need to add an index of 0 to your reset action:
`dispatch({type: 'Reset', index: 0, actions: [{ type: 'Navigate', routeName:'Home'}]})`

Haven't yet figured out to run this only at the end of a transition - think you may need to use the Transitioner component

markflorisson:
Awesome, that works! Maybe it would be useful to have a 'reset' function on the 'navigation' object?

ericvicenti:
@JustinNothling, good point, we don't expose Transitioner's `onTransitionEnd` through CardStack or `StackNavigator`. That hasn't come up yet! Want to submit a PR?

I'm on the fence about adding a `reset` helper because it doesn't always make sense for a child component to be aware of the application around it. Pretty soon people will want to add `popAndReplace` and `pushTwiceAndJumpBack` or whatever :-) . The replace option won't always be practical, because the screen can't see the whole navigation state (nor should it! you don't want every screen to re-render when one of their state changes). Instead, to support complicated navigation behaviors, you may want to override behavior on the router: https://reactnavigation.org/docs/routers/#Custom-Navigation-Actions

JustinNothling:
@ericvicenti we're in agreement about the `reset` helper.

How would you go about structuring a simple login/logout using react-navigation? At the moment pushing a home route onto login allows users to `goBack` to login using gestures (which isn't ideal). Would you reset the stack or would you prevent the gesture?

nebula-ai:
You might fine this useful.  It is a variation of the tutorial where I have a login screen and then a logout function under the Options right element of the RecentChatsScreen which will get you back to the original login screen.

import React from 'react';
import {
  AppRegistry,
  Text,
  View,
  Button
} from 'react-native';
import { StackNavigator, TabNavigator } from 'react-navigation';

[...]

PR #157, opened 2017-02-02
Fixed reset action docs

mmazzarolo:
Related to #59  

- Added the index
- Actions is now an array 
  
Let me know if I need to fix anything :)


ericvicenti:
Thanks!

PR #158, opened 2017-02-02
Better back button label

grabbou:
This pull request implements a better back button title for a back button on iOS. 

It changes the way components are rendered. Instead of rendering all left, title and right components separately with position `absolute`, it renders them within another View. That makes it possible to use layout, flex, and built-in truncating when width of the container is reached.

In order to achieve that, we listen to `onLayout` to get the real width.

**Todo**
Test on Android. Should have no changes. Will do later today, for now, waiting for review.

**Test plan**

Run on iOS, go to any view, push a screen, observe the header:
screen shot 2017-02-02 at 10 56 27
or go back and see smaller title:
screen shot 2017-02-02 at 10 56 31

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -5,7 +5,6 @@ import React, { PropTypes } from 'react';
 import {
   Animated,
   Platform,
-  Dimensions,
   StyleSheet,
   View,
 } from 'react-native';
@@ -244,7 +243,7 @@ class Header extends React.Component<void, HeaderProps, HeaderState> {
         key={`${name}_${key}`}
         style={[
           width && {
-            width: (Dimensions.get('window').width - width) / 2,
+            width: (props.layout.initWidth - width) / 2,
           },
           styles.item,
           styles[name],
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -246,7 +246,7 @@ class Header extends React.Component<void, HeaderProps, void> {
   }
 
   render(): React.Element<*> {
-    let component = null;
+    let children = null;
 
     if (this.props.mode === 'float') {
       const scenesProps: Array<NavigationSceneRendererProps> = this.props.scenes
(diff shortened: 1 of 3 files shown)


migueloller:
Is truncating the text going to be the default behavior now? Or is this PR more focused on the flexibility of the back button's contents?

Relevant discussion: https://github.com/react-community/react-navigation/pull/109#issuecomment-276420102

grabbou:
@ericvicenti thanks! That's a good idea. Mind explaining how's `initWidth` different from Animated width and whether it's safe to use the former?

grabbou:
@migueloller for now I focused on changing the layouts and making it possible - in next PR I'll focus on more platform-specific things - all in all `left` button is replaceable, we can override it at any point, even in the user land.

ericvicenti:
I know its super confusing how layout includes the init width as well as the animated value. The init width is usually correct, but it will not change dynamically over time. Sometimes the init width is needed for synchronous layout logic, but other times we need the dynamic value.

src/views/Header.js
```diff
@@ -209,11 +219,32 @@ class Header extends React.Component<void, HeaderProps, void> {
...
 
     const pointerEvents = offset !== 0 || isStale ? 'none' : 'box-none';
+
+    // Only measure `title` component
+    const onLayout = name === 'title'
+      ? (e: LayoutEvent) => {
+        this.setState({
+          widths: {
+            ...this.state.widths,
+            [`${index}`]: e.nativeEvent.layout.width,
```
satya164: `[index]` should do the same thing

src/views/Header.js
```diff
@@ -209,11 +219,32 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+      ? this.state.widths[`${index}`]
+      : undefined;
+
     return (
       <Animated.View
         pointerEvents={pointerEvents}
+        onLayout={onLayout}
         key={`${name}_${key}`}
         style={[
+          width && {
```
satya164: What happens before the layout is measured? Do the title and button label jump when layout happens?

grabbou: Haven't noticed that, most likely because we pre render next scenes

satya164: @grabbou how do we pre-render?

grabbou: With mode `float`, we render 2 scenes ahead https://github.com/react-community/react-navigation/blob/c58a99712172cfe09c10ddb52a21249fdc52222f/src/views/Header.js#L202 so that the animation is smooth. That should make it possible to calculate the widths appropriately as far as I understand.

satya164: @grabbou these are previous 2 scenes.

PR #166, opened 2017-02-03
Pass through screenProps to nested navigators

wschurman:
### Motivation

This formalizes passing `screenProps` to nested navigator screens by passing through the `screenProps` parameter to screen component in addition to pulling them out to props individually.

Prior to the change, this was accomplished by doing something like the following on the parent navigator, so that the `screenProps` would be available in nested navigator's screens:

```javascript
<AppNavigator
  screenProps={{
    exampleScreenProp: 'exampleScreenPropValue',
    screenProps: {
      exampleScreenProp: 'exampleScreenPropValue',
      // add additional levels for number of nested navigators
    },
  }}
/>
```

Open to suggestions, changes, etc. If accepted, a follow-up could change the behavior to only pass through the `screenProps` prop into screen component (along with `navigator`) instead of pulling them into separate params on the component (remove `{...screenProps}`).

### Test Plan

- Run `NavigationPlayground` example on both Android and iOS and verify that the `"exampleScreenPropValue"` is displayed on the nested navigator's screens.
- Integrate into own app and verify that the `screenProps` hack mentioned above can now be removed.

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -103,7 +103,13 @@ const AppNavigator = StackNavigator({
   mode: Platform.OS === 'ios' ? 'modal' : 'card',
 });
 
-export default () => <AppNavigator />;
+export default () => (
+  <AppNavigator
+    screenProps={{
+      exampleScreenProp: 'exampleScreenPropValue',
+    }}
+  />
+);
 
 const styles = StyleSheet.create({
   item: {
--- a/src/views/SceneView.js
+++ b/src/views/SceneView.js
@@ -19,6 +19,7 @@ export default class SceneView extends PureComponent<void, Props, void> {
 
   render() {
     const { screenProps, navigation, component: Component } = this.props;
-    return <Component {...screenProps} navigation={navigation} />;
+    // pass through screenProps as well for nested navigators
+    return <Component {...screenProps} screenProps={screenProps} navigation={navigation} />;
   }
 }
--- a/examples/NavigationPlayground/js/StacksInTabs.js
+++ b/examples/NavigationPlayground/js/StacksInTabs.js
@@ -15,9 +15,11 @@ import {
 import Ionicons from 'react-native-vector-icons/Ionicons';
 import SampleText from './SampleText';
 
-const MyNavScreen = ({ navigation, banner }) => (
+// exampleScreenProp passed through recursively from root AppNavigator
+const MyNavScreen = ({ navigation, banner, exampleScreenProp }) => (
   <ScrollView>
     <SampleText>{banner}</SampleText>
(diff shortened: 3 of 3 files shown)


src/views/SceneView.js
```diff
@@ -19,6 +19,7 @@ export default class SceneView extends PureComponent<void, Props, void> {
 
   render() {
     const { screenProps, navigation, component: Component } = this.props;
-    return <Component {...screenProps} navigation={navigation} />;
+    // pass through screenProps as well for nested navigators
+    return <Component {...screenProps} screenProps={screenProps} navigation={navigation} />;
```
satya164: Probably we can just do `<Component screenProps={screenProps} navigation={navigation} />`, will be a breaking change though. what do you think @ericvicenti ?

not a fan of `this.props.screenProps` though

ericvicenti: I'm ok with splatting or non-splatting. But doing both seems strange. Also I'm hopeful we can avoid breaking changes

wschurman: One other option would be to only pass the non-splatted version if Component is a navigator, though that may require type introspection or forking the behavior further up the stack.

My gut says that making the breaking change is probably the best option since it is the cleanest, but I also agree that avoiding breaking changes is probably for the better.

examples/NavigationPlayground/js/App.js
```diff
@@ -103,7 +103,13 @@ const AppNavigator = StackNavigator({
   mode: Platform.OS === 'ios' ? 'modal' : 'card',
 });
 
-export default () => <AppNavigator />;
+export default () => (
+  <AppNavigator
+    screenProps={{
+      exampleScreenProp: 'exampleScreenPropValue',
```
ericvicenti: This is the top-level file for all examples, and I would prefer if it didn't get too bloated with random example stuff. Can we remove this example from `App` and `StacksInTabs`?

Perhaps we can create a new example for this that demonstrates how props can be deeply passed.

satya164:
> One other option would be to only pass the non-splatted version if Component is a navigator, though that may require type introspection or forking the behavior further up the stack.

I don't like this very much

> My gut says that making the breaking change is probably the best option since it is the cleanest, but I also agree that avoiding breaking changes is probably for the better.

I agree that making it a breaking change is the best option. Probably we can just include a warning saying that the behaviour has changed (with a prop to disabled the warning) and remove the warning in future.

wschurman:
> I agree that making it a breaking change is the best option. Probably we can just include a warning saying that the behaviour has changed (with a prop to disabled the warning) and remove the warning in future.

Sounds good, I'll put up a commit soon.

satya164:
Closing in favor of #241

PR #173, opened 2017-02-03
Implement `labelStyle` for `TabBarBottom`

jhilden:
This allows styling of the label component for `TabBarBottom` (analog to how it is already possible for  `TabBarTop`)

### DIFF
--- a/docs/api/navigators/TabNavigator.md
+++ b/docs/api/navigators/TabNavigator.md
@@ -106,12 +106,16 @@ Several options get passed to the underlying router to modify navigation logic:
 - `inactiveTintColor` - label and icon color of the inactive tab
 - `inactiveBackgroundColor` - background color of the inactive tab
 - `style` - style object for the tab bar
+- `labelStyle` - style object for the tab label
 
 Example:
 
 ```js
 tabBarOptions: {
   activeTintColor: '#e91e63',
+  labelStyle: {
+    fontSize: 12,
+  },
   style: {
     backgroundColor: 'blue',
   }
--- a/docs/api/navigators/TabNavigator.md
+++ b/docs/api/navigators/TabNavigator.md
@@ -113,9 +113,7 @@ Example:
 ```js
 tabBarOptions: {
   activeTintColor: '#e91e63',
-  labelStyle: {
-    fontSize: 12,
-  },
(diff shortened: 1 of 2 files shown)


src/views/TabView/TabBarBottom.js
```diff
@@ -39,6 +39,7 @@ type Props = {
   getLabelText: (scene: TabScene) => string;
   renderIcon: (scene: TabScene) => React.Element<*>;
   style: any;
+  labelStyle: any?;
```
ericvicenti: Does this actually pass flow? Usually the `?` comes before the type, but I don't think its needed for `any`.

docs/api/navigators/TabNavigator.md
```diff
@@ -106,12 +106,16 @@ Several options get passed to the underlying router to modify navigation logic:
...
 - `inactiveBackgroundColor` - background color of the inactive tab
 - `style` - style object for the tab bar
+- `labelStyle` - style object for the tab label
 
 Example:
 
 ```js
 tabBarOptions: {
   activeTintColor: '#e91e63',
+  labelStyle: {
```
ericvicenti: Can you revise this to be on one line for consiseness?

src/views/TabView/TabBarBottom.js
jhilden: sorry, but I don't have any experience with flow yet.  I just copied that code from `TabBarTop`

jhilden: According to the docs about [any](https://flowtype.org/docs/quick-reference.html#any) it would also seem to me that the `?` is obsolute.  But I can't be sure.

Should I just remove the `?` here and in `TabBarTop` @ericvicenti ?

satya164: `TabBarTop` has `labelStyle?: any;` not `labelStyle: any?;`

jhilden: oh, totally missed that.  Sorry.  Fixed in my last commit.  Thanks for the catch.

jhilden:
Thanks for your reviews, I made all the requested changes.  Let me know if there is anything else I should change before this can get merged.

PR #176, opened 2017-02-04
Add titleStyle header option

dmhood:

Changing the style of the individual title component required rolling your own title component for every screen.  This way developers can quickly add a single overriding style to their app, without having to define a new component for each case (ie still use the "string" version of the title option).

**Test plan (required)**

100% of current tests passing/flow results the same (122 errors).  No change to current examples.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -97,6 +97,8 @@ export type NavigationScreenOption<T> =
     router?: NavigationRouter) => T
   | T;
 
+export type Style = Object | number | false | void;
+
 export type HeaderConfig = {
   /**
    * Title string used by the navigation bar, or a custom component
@@ -121,7 +123,12 @@ export type HeaderConfig = {
   /**
    * Style passed into navigation bar container
    */
-  style?: Object,
+  style?: Style,
+
+  /**
+   * Style passed into navigation bar title
+   */
+  titleStyle?: Style,
 
   // // Style of title text
   // titleTextStyle?: $NavigationThunk<Object>,
@@ -254,7 +261,7 @@ export type NavigationStackViewConfig = {
   mode?: 'card' | 'modal',
   headerMode?: HeaderMode,
   headerComponent?: ReactClass<HeaderProps>,
-  cardStyle?: Object
+  cardStyle?: Style
 };
 
 export type NavigationStackRouterConfig = {
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -99,10 +99,19 @@ class Header extends React.Component<void, HeaderProps, void> {
     return undefined;
   }
 
+  _getHeaderTitleStyle(navigation: Navigation): ?object {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.titleStyle) {
+      return header.titleStyle;
+    }
+    return undefined;
+  }
+
   _renderTitleComponent = (props: SubViewProps) => {
+    const titleStyle = this._getHeaderTitleStyle(props.navigation);
     const color = this._getHeaderTintColor(props.navigation);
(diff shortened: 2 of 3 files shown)


benvium:
A really useful improvement. One of the first things I tried to do while evaluating this library was to change the title font, and this PR should make that simple.

src/views/Header.js
```diff
@@ -99,10 +99,22 @@ class Header extends React.Component<void, HeaderProps, void> {
...
   _renderTitleComponent = (props: SubViewProps) => {
-    const color = this._getHeaderTintColor(props.navigation);
+    let titleStyle = this._getHeaderTitleStyle(props.navigation);
+    if (!titleStyle) {
+      const color = this._getHeaderTintColor(props.navigation);
+      titleStyle = color && ({ color });
+    }
     const title = this._getHeaderTitle(props.navigation);
-    return <HeaderTitle style={color && ({ color })}>{title}</HeaderTitle>;
+    return <HeaderTitle style={titleStyle}>{title}</HeaderTitle>;
```
satya164: `titleStyle` doesn't need to override `tintColor`. It can just be `style={[ titleStyle, tintColor && { color: tintColor } ]}`

src/views/Header.js
```diff
@@ -99,10 +99,22 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.titleStyle) {
+      return header.titleStyle;
+    }
+    return undefined;
+  }
+
   _renderTitleComponent = (props: SubViewProps) => {
-    const color = this._getHeaderTintColor(props.navigation);
+    let titleStyle = this._getHeaderTitleStyle(props.navigation);
```
satya164: It seems wasteful to `getScreenConfig` for each configuration option once for each element of the screen for each screen. Especially since there is no caching.

cc @grabbou

src/TypeDefinition.js
```diff
@@ -123,6 +123,11 @@ export type HeaderConfig = {
    */
   style?: Object,
 
+  /**
+   * Style passed into navigation bar title
+   */
+  titleStyle?: Object,
```
satya164: styles can be objects, numbers, booleans etc. so these ones look wrong. Maybe define a new type called style and reuse it.

```js
type Style = Object | number | false | void
```

src/views/Header.js
grabbou: Yes, that's true. Could be done once. Will send a PR.

src/views/Header.js
grabbou: I think it `cannot` :)

satya164: @grabbou what do you mean? the author is setting `tintColor` only if `titleStyle` is not given, above

src/views/Header.js
```diff
@@ -99,10 +99,22 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+    if (!titleStyle) {
```
satya164: Always check for `undefined`, it should be valid to specify `titleStyle` as a falsy value, e.g. - `null` to reset the global config

src/views/Header.js
dmhood: @grabbou thanks!

dmhood:
@satya164 made your suggested changes.  I set titleStyle to merge over the color value since it seemed more likely someone would set a navigator-wide tintColor while overriding w/ specific styles, but let me know if you would prefer it the other way around.

satya164:
`TabNavigator` does it the other .way around. Maybe you can change them too? Can be separate PR.

dmhood:
I'd be happy to--are you just referring to renderLabel methods in TabBarBottom and TabBarTop?

satya164:
@dmhood yup

PR #177, opened 2017-02-04
override drawer navigator styles

rafaelmmathias:
Fix to override styles of DrawerView.
For exemple, if we want to remove the default paddingTop (20px) of DrawerNavigator on IOS, the style don't overrides the default because DrawerNavigator component don't pass the custom styles to DrawerView.

**Test plan (required)**

DrawerNavigatorConfig:

{
  contentOptions: {
    activeTintColor: '#e91e63'
  },
  **style: {
    backgroundColor: '#000',
    paddingTop: 0
  }**
}

if put style here, now the style changes, without this fix, don't.


satya164:
Can you change this to be in `navigationOptions.drawer` instead of here?

srdjanprpa:
Hi @satya164 can you explain to me why you want to change to be a navigationOptions.drawer.style?

rafaelmmathias:
Hello guys, i'm back, sorry.
So, @satya164, is semantic? Or what?

raymundtusk7:
The paddingTop is implemented here. Thinking this should be removed and have the contentComponent define its own styles.

https://github.com/react-community/react-navigation/blob/master/src/views/Drawer/DrawerSidebar.js#L94

```javascript
export default withCachedChildNavigation(DrawerSidebar);

  container: {
    flex: 1,
    backgroundColor: '#fff',
    paddingTop: Platform.OS === 'ios' ? 20 : 0,
  },
});
```

satya164:
@rafaelmmathias the header style is in `navigationOptions`

@raymundtusk7 you're right

srdjanprpa:
@satya164 if we changed on `navigationOptions.drawer.style` it will not work if we use `contentComponent`. I don't know why is this solution bad?

satya164:
Why will it not work?

petroslafaz:
Can this be merged?

ericvicenti:
This change looks reasonable. Be sure to update the docs before moving forward

RobertSheaO:
It would be great if it got checked into master. It appears this was just an oversight of passing params when this navigator was created. `this.props.styles` is already being applied inside `DrawerView`, despite them being suppressed in `DrawerNavigation` without this fix.

satya164:
A better approach is to add `paddingTop` to https://github.com/react-community/react-navigation/blob/master/src/views/Drawer/DrawerNavigatorItems.js and remove padding from here.

There's no other reason to pass styles to the drawer really.

BTW @danielmschmidt, if I remeber correctly, we have set the background color manually because the drawer on iOS doesn't have a background color like the android version has.

DanielMSchmidt:
@satya164 Exactly that was the reason 👍

satya164:
@DanielMSchmidt maybe we should add a background by default to iOS version to have same behaviour on both platforms?

srdjanprpa:
How can I change this white color on the header?   
screen shot 2017-03-06 at 08 17 25

DanielMSchmidt:
@satya164 What would you recommend exactly? Setting the background color to white or make it white by default and overwritable?

DanielMSchmidt:
@srdjanprpa You can simply wrap the content of your drawer in a View which has the complete height and width and set a backgroundColor there.

satya164:
@DanielMSchmidt I'd just make it white. Though the android version has a prop to set the background color https://facebook.github.io/react-native/docs/drawerlayoutandroid.html#drawerbackgroundcolor, so we'd want to have white by default and allow overwriting it to match the API.

IMO people can just set style for the view they are using for `renderNavigationView` to set the background color since they always need a wrapper view. So I don't see any use for the prop. But it's there anyway :)

srdjanprpa:
@DanielMSchmidt how to use wrap in the drawer?
In DrawerNavigator I use contentComponent, you can see my code here https://github.com/srdjanprpa/FormulaOne/blob/master/src/app.js

And for this solution the problem is padding
https://github.com/react-community/react-navigation/blob/master/src/views/Drawer/DrawerSidebar.js
```const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: '#fff',
    paddingTop: Platform.OS === 'ios' ? 20 : 0, 
  },
```

cainaleaouk:
👍 to @srdjanprpa's question, could someone please post some details on how to modify styling for the `DrawerNavigator` and friends? Would rather not rely on a custom fork...

satya164:
Closing in favor of #642

hassankhan:
Can you please re-open this issue? Not all of us want to just remove the padding, there's a lot of other styling props and no simple way of doing it.

As far as I can tell, it seems to have been a minor oversight not exposing the props from the child components, so can someone please either (a) recommend The Correct Way™ of achieving this, or (b) merging this PR

satya164:
Did you even read the linked PR?

hassankhan:
The linked PR, if my English is correct, is referring to removing the top margin from the drawer view. This PR adds a `style` prop, which allows for more customization.

If I'm misunderstanding, can you explain? Witty replies aren't useful.

satya164:
> which allows for more customization

What kind of customization? You can replace the whole drawer content which allows you to do anything you want. You don't need a `style` prop for that.

cainaleaouk:
@satya164 : How to? The documentation says nothing about it. Closest I could find was `contentComponent - Component to use to render the navigation items. Receives the navigation prop for the drawer. Defaults to DrawerView.Items.`
Which is not what I need changing, I need to be able to change the sidebar not the items.

satya164:
@cainaleaouk 

```js
contentComponent: props => (
  <View style={style.container}>
    <Image source={require('./cover.png') />
    <DrawerView.Items />
  </View>
)
```

cainaleaouk:
@satya164 : You might want to review your documentation (/english) because that's not what its saying. I got 3 people to check this out and now one understood that contentComponent would change the container but just the Items inside the container. ;)

hassankhan:
Cheers @satya164, that was the answer I was looking for!

satya164:
PRs welcome for improving the documentation.

cainaleaouk:
@srdjanprpa : On it. Thanks.

PR #181, opened 2017-02-05
Allow labelStyle to override tab label color

dmhood:
Per https://github.com/react-community/react-navigation/pull/176#issuecomment-277480911


satya164:
Thank you

PR #191, opened 2017-02-05
Don't allow transitioner to run two transitions at once (Fixes #160)

### LINKED ISSUE #160: StackNavigator: setParams in 2nd screen reroutes the app back to the 1st screen
For some reason the code below inside my 2nd screen causes the app to briefly show the 2nd screen then immediately transition back to the 1st. I'm not sure if this is a bug..

```
componentDidMount () {
      onBack: 'whatever'
    });
  }
```

dmhood:
In order to set dynamic header buttons/titles, we need often need to do something like:

```javascript
componentDidMount () {
    const isDisabled = <someLogicUsingLocalState>;
    const params = {
            right: (
                <Button
                    onPress={() =>!isDisabled ? this.onSave() : null}
                >
                    <Text style={[styles.rightButtonTextStyle, isDisabled ? styles.disabledStyle : null]}>
                        Save
                    </Text>
                </Button>
            ),
        };
    this.props.navigation.setParams(params);
 }

// And our navigation options:

function getCurrentParams (state) {
    if (state.routes) {
        return getCurrentParams(state.routes[state.index]);
    }
    return state.params || {};
}

// ...some setup...

// main stack navigator navigationOptions
navigationOptions: {
    header: ({state}) => {
        // get the "deepest" current params.
        const currentParams = getCurrentParams(state);

        const left = currentParams.left;
        const right = currentParams.right;
        const style = currentParams.style;
        const tintColor = currentParams.tintColor;
        return { left, right, style, tintColor };
    }
 }
```

However, due to https://github.com/react-community/react-navigation/issues/160 this option doesn't work.  I believe the reason is that the quick firing of setParams causes Transitioner to try and run two transitions at the same time.  Specifically, _transitionProps gets updated mid-transition and  causes the app to get into a weird/unresponsive state.

This PR prevents Transitioner from queuing more than one transition at a time.  You can see the error by using https://github.com/react-community/react-navigation/compare/master...sportsaction:setParamsError and navigating to "Stack with Dynamic Header" in the NavigationPlayground example.  PR was tested using the same example/setup.

I'm still not super familiar with the codebase to let me know if this isn't the right track.

### DIFF
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -95,6 +95,8 @@ class Transitioner extends React.Component<*, Props, State> {
     this._prevTransitionProps = null;
     this._transitionProps = buildTransitionProps(props, this.state);
     this._isMounted = false;
+    this._isTransitionRunning = false;
+    this._queuedTransition = null;
   }
 
   componentWillMount(): void {
@@ -121,6 +123,16 @@ class Transitioner extends React.Component<*, Props, State> {
       return;
     }
 
+    const indexHasChanged = nextProps.navigation.state.index !== this.props.navigation.state.index;
+    if (this._isTransitionRunning) {
+      this._queuedTransition = { nextProps, nextScenes, indexHasChanged };
+      return;
+    }
+
+    this._startTransition(nextProps, nextScenes, indexHasChanged);
+  }
+
+  _startTransition(nextProps: Props, nextScenes: Array<NavigationScene>, indexHasChanged: boolean) {
     const nextState = {
       ...this.state,
       scenes: nextScenes,
@@ -162,7 +174,7 @@ class Transitioner extends React.Component<*, Props, State> {
       ),
     ];
 
-    if (nextProps.navigation.state.index !== this.props.navigation.state.index) {
+    if (indexHasChanged) {
       animations.push(
         timing(
           position,
@@ -173,8 +185,8 @@ class Transitioner extends React.Component<*, Props, State> {
         ),
       );
     }
-
     // update scenes and play the transition
+    this._isTransitionRunning = true;
     this.setState(nextState, () => {
       nextProps.onTransitionStart && nextProps.onTransitionStart(
         this._transitionProps,
@@ -239,6 +251,16 @@ class Transitioner extends React.Component<*, Props, State> {
         this._transitionProps,
         prevTransitionProps,
       );
+      if (this._queuedTransition) {
+        this._startTransition(
+            this._queuedTransition.nextProps,
+            this._queuedTransition.nextScenes,
+            this._queuedTransition.indexHasChanged
+        );
+        this._queuedTransition = null;
+      } else {
+        this._isTransitionRunning = false;
+      }
     });
   }
 }


ericvicenti:
Ok, if I understand correctly, this addresses the following race condition, which is currently handled poorly:

```
- State Changes A->B
- Transition A->B starts
- State Changes B->C
- State Changes C->D
- Transition A->B completes
```

I think the code change will avoid errors, but leave the final visual state as `B` instead of `D`.

Instead, when the state changes, I think we should mark the current `A->B` animation as 'dirty', and start a new animation from `B->D`, which will begin after the current dirty animation completes.

dmhood:
If we go that route then I think just a queue of transitions is probably the best way to implement it, although we also need to solve the issue of setParams creating a new transition.  I'll look at refactoring into a queue in the morning, but also changing line 121 (nextScenes === this.state.scenes) to check if the the same key is active instead of just an identity comparison.

Let me know if anyone has other thoughts/suggestions.

dmhood:
@ericvicenti So it turns out this fixes the setParams issue without having to drill down each route to check the active key.  Here it is working when I fire two navigation actions on the "Go To Profile Screen" button 100ms apart:
queuedtransition

Thanks!

src/views/Transitioner.js
```diff
@@ -235,6 +256,12 @@ class Transitioner extends React.Component<*, Props, State> {
     this._transitionProps = buildTransitionProps(this.props, nextState);
 
     this.setState(nextState, () => {
+      if (this._queuedTransition) {
+        this._startTransition(this._queuedTransition.nextProps, this._queuedTransition.nextScenes);
+        this._queuedTransition = null;
+      } else {
+        this._isTransitionRunning = false;
+      }
       this.props.onTransitionEnd && this.props.onTransitionEnd(
```
ericvicenti: I think this call to `onTransitionEnd` should be moved up to the start of this block. That way it is clear that it will be called before the `onTransitionStart` of the next transition

src/views/Transitioner.js
```diff
@@ -48,6 +48,16 @@ type State = {
   scenes: Array<NavigationScene>,
 };
 
+type Scenes = Array<Scene>;
+
+type Scene = {
```
ericvicenti: You should use the existing type from `TypeDefinition` instead of re-declaring it here

dmhood:
Good suggestions, thanks.  I'm not sure how to go about mocking everything involved in creating the race condition, but here are two issues that I know of that people can use to test out the PR:

1.) Calling setParams in componentDidMount/componentWillMount (or anytime a transition is occuring) results in an unresponsive app: https://github.com/react-community/react-navigation/compare/master...sportsaction:setParamsError.  In this case, Transitioner should do nothing on the setParams call;

2.) Navigating to another route while a transition is occurring also results in an unresponsive app: https://github.com/react-community/react-navigation/compare/master...sportsaction:navError.  In this case, Transitioner should complete the first navigation transition (to 'Profile'), and then make a second transition (to 'HeaderTest') once the first transition is complete.

dmhood:
Also, in the case of several navigation actions happening at once, this PR changes the behavior from:
currenttransition

which sort of feels like it is "cancelled", to:
newtransition

where the first animation fully completes.  I feel like the first/current way is a little jarring, but I think this could be argued either way.

lnikkila:
I tried this in an existing codebase and repeatedly managed to get to a state where the app would become unresponsive after resetting StackNavigator a couple of times. This doesn’t happen with master.

I’m not yet sure about what the minimal test case for this is but I’ll give you an update if I can figure it out.

dmhood:
@lnikkila I can reproduce this fairly reliably if I set the transition time very high (>1000ms) and navigate more than once through the example I gave above (with multiple navigation actions).  I'll try to drill down and figure out what's causing it.

dmhood:
Ok, figured out the race condition (navigation state index check)--everything seems to be working now!

ericvicenti:
Ok, this looks like a good fix!

sreucherand:
@ericvicenti @dmhood I haven't been involved in this thread and I don't know every aspects of it but I wonder about the following questions. Is the fact that the `transitioner` is able to run 2 transitions at once, the direct cause of the issue you initially mentioned?

Because, from an UX perspective, I would say that the `transitioner` should be able to run 2 transitions at once. Just to make the app more responsive. Let's take the example: I'm moving from A to B, I wan't to be able to go back to A (by pressing the back button for example) before the transition is actually done.

In fact, the `transitioner` should be able to run 2 transitions at once ONLY when you go back.
A -> B
A <- B

But not the other way around since you don't know what the user's gonna click on next, it might be C like:
A <- B
A -> C

If you take a look at the iOS settings app, you would see the same behaviour. When moving from main screen to General, you can immediately press the back button. But when moving back from General, you can't press any items of the main screen until the transition is finished.

So my point is: can we fix the params issue without having to remove the transition concurrency? Especially since the current pending animation makes the navigation looks non-responsive to me.

ericvicenti:
> Because, from an UX perspective, I would say that the transitioner should be able to run 2 transitions at once. Just to make the app more responsive. Let's take the example: I'm moving from A to B, I wan't to be able to go back to A (by pressing the back button for example) before the transition is actually done.

I think you should be able to request a new transition while a transition is already happening, but I think we should only ever display one transition at a time. So if you quickly tap back twice, you will move back twice, but the first back transition will complete before the second back transition begins.

sreucherand:
I see your point. In the case you tap back twice, you don't wan't two transitions to start at the same time. Sounds correct. But in case you quickly tap the back button before the transition is finished, I think the "back" transition should start immediately. Meaning that technically, it should revert the `Animated.Value` to its previous value. Only in that scenario. It would make the app more responsive to me.

In case you trigger a "push" transition (not a "back" transition) while a transition is running, then it should wait for it to be finished before starting the new one.

satya164:
>  in case you quickly tap the back button before the transition is finished, I think the "back" transition should start immediately

if the animation stops abruptly, it's bad UX IMO

PR #201, opened 2017-02-06
withCachedChildNavigation basically did not cache

### LINKED ISSUE #159: TabNavigator re-renders all tabs on switch
Currently all of my tabs are re-rendered every time you switch a tab, one of the tabs will be displaying a graph so I don't want to re-render the graph every time you switch between two other tabs.

Can we implement a function in the library that prevents rendering the tab if it is not being viewed? Or would this be better as a condition passed to the screen through this.props.params?

EmielM:
Fixes #159. Any ideas on testing this welcome.

Also, I'm unsure if routes can change. If that is the case, we could add logic to clean out the cache of no-longer-used routes.

### DIFF
--- a/src/withCachedChildNavigation.js
+++ b/src/withCachedChildNavigation.js
@@ -42,7 +42,9 @@ export default function withCachedChildNavigation<T: Props>(Comp: ReactClass<T>)
       navigation: NavigationScreenProp<NavigationState, NavigationAction>
     ) => {
       // Update props for each child route
-      this._childNavigationProps = {};
+      if (!this._childNavigationProps) {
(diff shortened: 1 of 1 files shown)


src/withCachedChildNavigation.js
```diff
@@ -53,6 +55,7 @@ export default function withCachedChildNavigation<T: Props>(Comp: ReactClass<T>)
           state: route,
         });
       });
+      // Not clearing out cached values for which no routes exist
```
satya164: nit: remove this comment

satya164:
Thank you :D

EmielM:
Updated!

PR #230, opened 2017-02-08
Expose onTransitionStart/End props to StackNavigator

juhasuni:
Currently onTransitionStart/End props are not exposed by CardStack or StackNavigator. This pull request adds them as props to CardStack and as options to StackNavigator.


ericvicenti:
It looks like you nailed it!

Issue #197, opened 2017-02-06
Header title covers half of back button on iOS


davidskaarup:
I think a recent commit caused this, it used to work fine. Now you can only pop the scene by pressing the left half of the back button.

<img width="487" alt="screen shot 2017-02-06 at 13 56 09" src="https://cloud.githubusercontent.com/assets/9447649/22647972/76ce9572-ec74-11e6-8591-9e471774fa09.png">

satya164:
cc @grabbou

michaelduminy:
I think there may be an issue with the recent header changes on Android.

We're using master at https://github.com/react-community/react-navigation/commit/eece5ad56a074ef2046c36f43ccb3ef66f5cc312. When we go to current this happens:

Main screen before:
![screen shot 2017-02-06 at 5 13 15 pm](https://cloud.githubusercontent.com/assets/4092035/22653022/07c6bdae-ec91-11e6-9312-993bede699e5.png)

Main screen after:
![screen shot 2017-02-06 at 5 11 20 pm](https://cloud.githubusercontent.com/assets/4092035/22653043/12218dce-ec91-11e6-985f-13083b18539b.png)

The title is set to `left: 10`.

Also in our modal screens the header used to look like this:
![screen shot 2017-02-06 at 5 13 22 pm](https://cloud.githubusercontent.com/assets/4092035/22653113/3c3f256c-ec91-11e6-9b5a-d5e2b91a14ef.png)

Now they look like this:
![screen shot 2017-02-06 at 5 11 41 pm](https://cloud.githubusercontent.com/assets/4092035/22653124/41b39226-ec91-11e6-8ff9-4774f1b7051e.png)

Our custom back button is not rendered or clickable.

satya164:
Sorry for the trouble guys. Can you try reverting https://github.com/react-community/react-navigation/commit/677625ef6a4253c5cdd3a02a2b0293eb640358c3 and see if it still happens?

davidskaarup:
Reverting that commit seems to solve the problem. I forgot to mention that the back button label had also disappeared, as can be seen on the screenshot. That also works again after reverting. It would be nice to have an option to hide the back button label and only show the arrow, but that should probably have its own issue.

ericvicenti:
If you add an example with a similar header, we can make sure to test it before landing any more changes to the header.

Usually we would add a test case to prevent this sort of thing, but this is a visual bug and the best way to avoid future breakage is to add this use case to the examples.

[...]

PR #233, opened 2017-02-08
Don't render back button label just yet

grabbou:
The back button label works prefectly fine, however the PR that was implementing truncating of too long titles has been reverted due to issues on Android: https://github.com/react-community/react-navigation/pull/158

Since @satya164 is about to release a new version, please consider merging that one so that people don't end up having issues with back button being too long.

The fix I am working on is to land tomorrow latest, but this is in case one wants to release something in the meantime.

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -119,10 +119,11 @@ class Header extends React.Component<void, HeaderProps, void> {
       return null;
     }
     const tintColor = this._getHeaderTintColor(props.navigation);
-    const previousNavigation = addNavigationHelpers({
-      ...props.navigation,
-      state: props.scenes[props.scene.index - 1].route,
-    });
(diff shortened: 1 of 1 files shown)


satya164:
There are still bugs with back button like https://github.com/react-community/react-navigation/issues/197

Probably we'll have to revert your previous PR before releasing.

grabbou:
#197 has nothing to do with my past-work, please see last comment.

satya164:
Are you talking about https://github.com/react-community/react-navigation/issues/197#issuecomment-278335666 ? Yeah, that's fixed. But what about the original issue? According to https://github.com/react-community/react-navigation/issues/197#issuecomment-277999725, it still happens. Are you sure it's fixed?

PR #234, opened 2017-02-08
enable `showLabel` prop in `tabBarBottom`

kylpo:
This PR adds the `showLabel` option to tabBarBottom for projects like mine that only want icons visible in the bottom tab bar.

It should be safe, since it is an exact copy/paste from `tabBarTop`. See https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabBarTop.js#L66

**Test plan (required)**
- All tests passed, though I don't know if any tests have been written to cover `TabBarBottom`

### DIFF
--- a/docs/api/navigators/TabNavigator.md
+++ b/docs/api/navigators/TabNavigator.md
@@ -105,6 +105,7 @@ Several options get passed to the underlying router to modify navigation logic:
 - `activeBackgroundColor` - background color of the active tab
 - `inactiveTintColor` - label and icon color of the inactive tab
 - `inactiveBackgroundColor` - background color of the inactive tab
+- `showLabel` - whether to show label for tab, default is true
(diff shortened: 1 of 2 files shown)


satya164:
It should probably automatically reduce tabbar height unless specified

kylpo:
Defaults to labels:
defaultlabels

With `showLabel: false`:
showlabelfalse

kylpo:
Thanks for the quick comment @satya164. I personally prefer the height remain unchanged by this prop. I wouldn't want this simple prop to do more than I think it should, which is to just remove the label.

Thoughts?

satya164:
Ok, this looks good enough. I was worried it'll leave blank space below.

ericvicenti:
This LGTM! Thanks @kylpo!

PR #237, opened 2017-02-08
Fixes several back button issues

grabbou:
There are few features/fixes included in this PR:

- Disable back button label as it's not read yet - it renders perfectly fine, but we haven't landed truncating yet, which is crucial for the production apps. I left it in master to iterate faster, but as @satya164 is planning to do the release, I have temporarily turned it off. I am also working on mimicking natural iOS animations as much as possible, hopefully this lands later this week.

- Fix layout issues on iOS. My changes introduced last week that were making an icon looking just like on a native app changed the size of container - the touchable area was no longer 44x44 as recommended by Apple, but 12x20 (size of an icon) + left margin of 10 points. This pull request fixes that by adjusting sizes.
screen shot 2017-02-08 at 18 29 27

- Add additional style for `icon with label` - this is the case for the future, as attached below. In this case, side margins are equal to 10.
screen shot 2017-02-08 at 18 30 12

### DIFF
--- a/src/views/HeaderBackButton.js
+++ b/src/views/HeaderBackButton.js
@@ -27,7 +27,11 @@ const HeaderBackButton = ({ onPress, title, tintColor }: Props) => (
   >
     <View style={styles.container}>
       <Image
-        style={[styles.button, { tintColor }]}
+        style={[
+          styles.icon,
+          title && styles.iconWithTitle,
+          { tintColor },
+        ]}
         source={require('./assets/back-icon.png')}
       />
       {Platform.OS === 'ios' && title && (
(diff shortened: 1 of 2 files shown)


src/views/HeaderBackButton.js
```diff
@@ -74,6 +81,11 @@ const styles = StyleSheet.create({
       resizeMode: 'contain',
       transform: [{ scaleX: I18nManager.isRTL ? -1 : 1 }],
     },
+  iconWithTitle: Platform.OS === 'ios'
+    ? {
+      marginRight: 5,
+    }
+    : {},
```
ericvicenti: What does this look like on Android?

grabbou: Just like before :) We do not render title on Android at all and so, tweaks because of that are not needed. Not sure if that's what you were asking about?

mkonicek: We don't render header title on Android?

mkonicek: Oh you mean back button title?

PR #241, opened 2017-02-08
Pass through screenProps to nested navigators

satya164:
This changes the behaviour of screenProps so that the props no longer get splatted.
Components will receive this as `this.props.screenProps`.

### DIFF
--- a/src/views/SceneView.js
+++ b/src/views/SceneView.js
@@ -11,12 +11,12 @@ import type {
 
 type Props = {
   screenProps?: {};
+  hideScreenPropsWarning?: boolean;
   navigation: NavigationScreenProp<NavigationState, NavigationAction>;
   component: ReactClass<*>;
 };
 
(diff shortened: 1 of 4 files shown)


src/views/SceneView.js
```diff
@@ -29,8 +29,30 @@ export default class SceneView extends PureComponent<void, Props, void> {
     };
   }
 
+  componentWillMount() {
+    if (this.props.screenProps !== undefined && this.props.hideScreenPropsWarning) {
```
ericvicenti: Isnt this logic inverted? I think you want `&& !this.props.hideScreenPropsWarning`

I'm not super happy about this part. I like to warn people, but encouraging them to dismiss the warning in this way feels odd. I would be happier with a thing like this:

```
// This would dismiss all warnings about API changes prior to this version.
ReactNavigation.setWarningLevel('1.0.0-beta.2');
```

PR #244, opened 2017-02-08
Make compatible with webpack

satya164:


package.json
```diff
@@ -7,6 +7,7 @@
     "react-native-v1": "lib-rn/react-navigation.js",
     "web": "lib/react-navigation.web.js"
   },
+  "module": "lib/react-navigation.web.js",
   "webpack": "lib/react-navigation.web.js",
```
skevy: maybe take this out? Probably not needed.

website/config/webpack.config.dev.js
```diff
@@ -80,7 +80,7 @@ module.exports = {
     alias: {
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
-      'react-native': 'react-native-web',
+      // 'react-native': 'react-native-web',
```
skevy: Remove the comment if we're not using it.

website/config/webpack.config.prod.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
     alias: {
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
-      'react-native': 'react-native-web',
+      // 'react-native': 'react-native-web',
```
skevy: ditto ^^

package.json
satya164: @skevy out website doesn't work without it. probably difference between webpack 1 and 2

skevy: Hmm. ok.

PR #254, opened 2017-02-09
Fix style prop for header

### LINKED ISSUE #189: header style is applied to 'left' and 'right' components
for instance with:
```
header: {
     style: {
          borderWidth: 1,
          borderColor: 'red',
    }
}
```
header
 creates a border around the header container but also around the left and right

grabbou:
Fixes #189

When we introduced style for header, it got included with `NavigationSceneRendererProps` causing some glitches for title/left/right components.

It is wise to remove it as we do not use it at all anywhere and each of the components (left/right/title) have its own custom style (e.g. `titleStyle`)

Test plan:

Set #189 repro and compare fixed header with the one posted in the issue.

screen shot 2017-02-09 at 09 44 04

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -45,7 +45,6 @@ export type HeaderProps = NavigationSceneRendererProps & {
   renderTitleComponent: SubViewRenderer,
   tintColor: ?string,
   router: NavigationRouter,
-  style?: any,
 };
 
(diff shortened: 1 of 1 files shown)


cyprusglobe:
@grabbou mind updating the docs for that aswell?

satya164:
@cyprusglobe this was a bug

PR #257, opened 2017-02-09
Back button label

### LINKED ISSUE #276: Backbutton title
How can i set the backbutton title? or leave it blank.

grabbou:
Fixes #276 

> This is a follow-up to https://github.com/react-community/react-navigation/pull/158 with no logical changes. It only contains layout fixes with screnshoots of all combinations on all platforms. Previous PR was pretty good, but the story is it got merged w/o tests on Android. I was planning to fix them in a follow-up PR, but we decided to cut release, so reverted to avoid issues in production.

Note: We have to render empty containers for items even if there's no back button, so that they can take up space. This is needed for proper alignment (e.g. missing left button, but right is defined - title should stay center etc.).

Back button & title, no right:
screen shot 2017-02-09 at 12 42 57
screen shot 2017-02-09 at 12 42 51

Back button & title, right:
screen shot 2017-02-09 at 12 42 38
screen shot 2017-02-09 at 12 42 00

Just title:
screen shot 2017-02-09 at 12 42 35
screen shot 2017-02-09 at 12 41 57

Title, right, no back-button:
screen shot 2017-02-09 at 12 47 39
screen shot 2017-02-09 at 12 47 32


grabbou:
Bonus, too long left title:
screen shot 2017-02-09 at 12 49 04

satya164:
Can we change the text to back instead of truncating the text to match iOS behaviour?

grabbou:
Yeah, I'll hack around that later today with a separate component `HeaderBackButtonTitle`

src/views/Header.js
```diff
@@ -218,16 +223,33 @@ class Header extends React.Component<void, HeaderProps, void> {
       subView = defaultRenderer(subViewProps);
     }
 
-    if (subView === null) {
-      return null;
-    }
-
     const pointerEvents = offset !== 0 || isStale ? 'none' : 'box-none';
+
+    // Only measure `title` component
```
satya164: That's obvious from the logic below, maybe it should say why

src/views/Header.js
```diff
@@ -218,16 +223,33 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+      : undefined;
+
     return (
       <Animated.View
         pointerEvents={pointerEvents}
+        onLayout={onLayoutIOS}
         key={`${name}_${key}`}
         style={[
+          width && {
+            width: (props.layout.initWidth - width) / 2,
```
satya164: I think the `width` variable needs a better name. it's not obvious what this line does

src/views/Header.js
```diff
@@ -306,29 +329,25 @@ const styles = StyleSheet.create({
...
-    top: 0,
+  item: {
+    justifyContent: 'center',
   },
+  title: Platform.OS === 'android'
+    ? {
+      flex: 1,
+      alignItems: 'flex-start',
+    }
+    : {},
```
satya164: Can't it be `null` instead of empty object?

Blinnikov:
I think that usage of hardcoded string constants like 'Back' is not the best solution here. I can use another language in my app other than English. So the text 'Back' I will not understand. I would prefer to make back button's text configurable.

grabbou:
Also, according to this link, let's consider hiding back button label https://github.com/react-community/react-navigation/issues/211#issuecomment-279204975

@satya164 what you think?

grabbou:
FYI this will be soon rebased and I am planning to land it tomorrow with the fixes. Stay tuned.

src/views/Header.js
grabbou: Yeah, maybe `titleWidth` ?

grabbou:
Still WIP - rest will land tomorrow. For now, just rebased and fixed few flow issues.

src/views/HeaderBackButton.js
```diff
@@ -92,4 +141,17 @@ const styles = StyleSheet.create({
     : {},
 });
 
+
+function nextFrameAsync() {
+  return new Promise((resolve: () => void) => requestAnimationFrame(() => resolve()));
```
satya164: Doesn't this do the same thing?

```js
return new Promise(resolve => requestAnimationFrame(resolve));
```

grabbou:
Updated screenshot with a custom `truncatedBackTitle`:
screen shot 2017-02-16 at 10 47 45

satya164:
I think it should be `truncateBackTitle`, verb + noun

grabbou:
@satya164 that suggest a function, doesn't it?

satya164:
`tuncateBackTitle: false`, question, answer. Or `backTitleTruncated`

grabbou:
But the purpose is to provide a replacement for default `Back`. You cannot turn off that feature on iOS. Alternatively, you can provide an empty string here or a short enough title in `backTitle`.

satya164:
Ah so it's a string. I think just `backTitle` is better

grabbou:
Yeah, maybe i over-optimised that. will simplify

grabbou:
Actually, I think both are needed as they serve a different purpose.

Imagine you have a route that displays a title `Welcome to universe!`. When you push another route, you want back button to say `Welcome`, not `Welcome to universe!`. 

Then, you set `backTitle` to `Welcome`. 

Now, when you push a route that has a big enough title that `Welcome` doesn't fit, we replace Welcome with `truncateBackTitle` (Defaults to `Back`). That prop is here just to customi`e (or localize) default `Back`.

grabbou:
Actually I think both are needed, first one, `backTitle` is just to change the title for the back button we select. By default it's the value we got from `title`. You can return `null` to disable title at all.

The latter, `truncatedBackTitle` is just to override `Back` string. People requested ability to localise it. Apart from that, it serves no clear benefit.

I just don't feel that mixing these two in one prop makes sense. Passing custom `backTitle` I still want it to be a subject to truncating when it's too long.

grabbou:
I think for now, I'll get rid of `truncatedBackTitle` to avoid confusion, I imagine people will be happy with `backTitle` (they'll just have to make sure it's small enough).

To much config isn't a good idea.

satya164:
I don't think we even need `truncatedBackTitle`. If it's too long, people can override it with `backTitle`.

satya164:
Yeah.

src/views/Header.js
```diff
@@ -34,9 +34,9 @@ type SubViewProps = NavigationSceneRendererProps & {
   onNavigateBack?: () => void,
 };
 
-type Navigation = NavigationScreenProp<NavigationState, NavigationAction>;
+type Navigation = NavigationScreenProp<*, NavigationAction>;
```
satya164: Not sure using `*` is the correct approach here. Looking at the code below, it looks like this should probably be `NavigationScreenProp<NavigationRoute, NavigationAction>`

src/views/Header.js
```diff
@@ -34,9 +34,9 @@ type SubViewProps = NavigationSceneRendererProps & {
...
 
-type SubViewRenderer = (subViewProps: SubViewProps) => ?React.Element<*>;
+type SubViewRenderer = (subViewProps: SubViewProps) => ?React.Element<any>;
```
satya164: Why `any` and not `*`?

src/views/Header.js
```diff
@@ -232,61 +258,55 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+      >
+        {left}
+        {title}
+        {right}
+      </View>
+    );
+  }
 
-    let children = null;
+  render(): React.Element<*> {
```
satya164: This annotation is probably unnecessary

src/views/Header.js
```diff
@@ -306,30 +326,26 @@ const styles = StyleSheet.create({
     elevation: 4,
   },
   appBar: {
-    flex: 1
+    flex: 1,
+    position: 'relative',
```
satya164: isn't everything `position: 'relative'` by default?

src/views/Header.js
```diff
@@ -306,30 +326,26 @@ const styles = StyleSheet.create({
...
-    alignItems: 'center',
   },
-  title: {
-    bottom: 0,
-    left: 40,
-    position: 'absolute',
-    right: 40,
-    top: 0,
+  item: {
+    justifyContent: 'center',
```
satya164: Probably should have `alignItems: 'center'` too

src/views/Header.js
```diff
@@ -306,30 +326,26 @@ const styles = StyleSheet.create({
...
+      flex: 1,
+      alignItems: 'flex-start',
+    }
+    : null,
   left: {
-    bottom: 0,
-    left: 0,
-    position: 'absolute',
-    top: 0,
+    alignItems: 'flex-start',
```
satya164: You probably want `alignSelf: 'flex-start'`, since everything is `column` be default this will align items to top

src/views/Header.js
```diff
@@ -306,30 +326,26 @@ const styles = StyleSheet.create({
...
   },
   right: {
-    bottom: 0,
-    position: 'absolute',
-    right: 0,
-    top: 0,
+    alignItems: 'flex-end',
```
satya164: Same as above

src/views/HeaderBackButton.js
```diff
@@ -10,54 +10,104 @@ import {
...
+  state = {};
+
+  render() {
+    const { onPress, title, tintColor, truncatedTitle } = this.props;
+
+    const renderTruncated = this.state.containerWidth && this.state.initialTextWidth
+      ? this.state.containerWidth < this.state.initialTextWidth
+      : false;
+
+    const onContainerLayout = Platform.OS === 'ios'
```
satya164: Define this function as a class property instead of defining in render

src/views/Header.js
```diff
@@ -212,16 +220,34 @@ class Header extends React.Component<void, HeaderProps, void> {
...
 
-    if (subView === null) {
-      return null;
-    }
-
     const pointerEvents = offset !== 0 || isStale ? 'none' : 'box-none';
+
+    // On iOS, width of left/right components depends on the calculated
+    // size of the title.
+    const onLayoutIOS = Platform.OS === 'ios' && name === 'title'
```
satya164: Define this function as a class property instead of inside render to avoid creating a new function every render

src/views/Header.js
```diff
@@ -72,6 +78,10 @@ class Header extends React.Component<void, HeaderProps, void> {
 
   props: HeaderProps;
 
+  state = {
+    widths: {},
+  };
+
   shouldComponentUpdate(nextProps: HeaderProps, nextState: any): boolean {
```
satya164: Should switch this to `PureComponent` and remove this `shouldComponentUpdate` block

src/views/Header.js
grabbou: In this case, it's impossible to avoid creating a new function every render as the function needs to know the `index` property.

src/views/Header.js
grabbou: Since everything is `column` by default, justifyContent will center items vertically. That's the desired effect. We don't want title to stick to top.

src/views/Header.js
grabbou: Nope, `alignItems` aligns vertically when flex is `row`. In `column`, it aligns horizontally (see screenshots), so that left children (`renderLeftComponent()`) is always to the left of its Animated container.

grabbou: screen shot 2017-02-17 at 09 13 04

Left button can be smaller than the space allowed for it. That's why we use `flex-start` on the container (`red` background). `alignSelf` would make sense to be applied to back button itself (blue background).

For right container, we have exactly same situation thus we use `flex-end`.

src/views/Header.js
grabbou: It is, removed.

src/views/Header.js
grabbou: I might be not the best one in Flow, but my understanding was that with `*`, it was inferring the type from first renderSubView method (renderLeftComponent) and so, it was complaining that the renderTitleComponent return value is different than in renderLeftComponent (saying prop types don't match). I guess it's because we didn't tell it that renderSubView can return *any*thing?

src/views/Header.js
grabbou: Probably. It was more of a `hack`, I decided to use `*` because:
a) I don't access its values, so it doesn't really matter to me
b) It was causing an error in DrawerItem that it's incompatible with `<T: *>`. Can be observed in past CircleCI logs. Not sure how to solve it w/o touching other files.

PR #279, opened 2017-02-10
Allow icons in TabBar be hidden for iOS

coodoo:
**motivation**

In some use cases it would be handy to have icons hidden in TabBar for iOS, I've figured out the minimum code path to have it implemented, see [screencap here](https://cl.ly/2h0O1M3o0i0k).


satya164:
Can you rebase it against master?

Issue #247, opened 2017-02-09
Typo in contributors guide


DiscoStarslayer:
In the contributors guide, there is a sentence that is abruptly cut off. See below (excuse the blame link, but there is no way to reference a individual markdown line as far as I can tell):

https://github.com/react-community/react-navigation/blame/master/docs/guides/Contributors.md#L71

Is there some known text to go here? Or does something need to be written for this?

grabbou:
CC: @ericvicenti the original blame suggests you authored the file, did you mean to explain some additional things after `But it`? :fire:

PR #280, opened 2017-02-10
Remove uncompleted sentence in Contributors guide.

zergov:
Hello, there is an uncompleted sentence in the Contributors guide.
@skevy wrote that line, if you remember what you wanted to write, ill add it :) 
Otherwise, I just removed the floating words.


grabbou:
There's already #247 open that tracks that.

PR #281, opened 2017-02-10
[FLOW] Fix missing "type" in flow import.

admmasters:
Fixes: Named import from module `../TypeDefinition` `ContextWithNavigation` is a type, but not a value. In order to import it, please use `import type`.

**Flow complains that we aren't using the import type syntax.**

**Flow requires import type syntax up until the current version (at least).**

**Test plan (required)**
Flowtype change - not tested as far as I can see.


satya164:
Thank you

PR #291, opened 2017-02-10
Add `showIcon` to `tabBarOptions` so that icons can be hidden on iOS

coodoo:
See sample and use case [here](https://cl.ly/2h0O1M3o0i0k).


coodoo:
@satya164 just rebased against master, please have a look, thanks!

PR #292, opened 2017-02-10
Add gesturesEnabled prop to StackNavigatorConfig

thomasbruketta:
Currently you can't disable gestures in the StackNavigatorConfig. This PR adds the ability to enable or disable gestures in the StackNavigatorConfig which then passes that prop to the CardStack component.


satya164:
I think it's useful to be able to do this per screen. So navigation options is a better place.

thomasbruketta:
this is already being set at the StackNavigatorConfig level by default. When you create a new StackNavigator it uses the CardStack component which has a default gesturesEnabled={true} for iOS and gesturesEnabled={false} for Android. See CardStack.js for reference.

thomasbruketta:
Or are you saying this is already included in navigationOptions?

gracelauren:
Would like this as a feature

Ashoat:
Literally composed the same PR in my sandbox before finding out it already exists :)

Looks like the test is failing because of Flow. If it's helpful, here's what I have in my sandbox (which includes Flow types as well as updating the docs to list the new config param): https://gist.github.com/Ashoat/54a4b0efd8ef1a2ea040dfcd1579819d

Agree it would be more useful on a per-screen basis. I went ahead and got that working on my sandbox, so I'm gonna put up a separate PR with the changes.

Ashoat:
See #385 for my attempt

ericvicenti:
I think @satya164 is right and we want to toggle gestures on a per-screen basis. Can we add this to the screen's `navigationOptions`?

joeferraro:
Is this considered a static value or one that can be modified on-the-fly (e.g. temporarily disabling gestures for the active screen)?

PR #305, opened 2017-02-11
Fix flow

satya164:
This PR does few things,

1. Upgrade React Native
2. Enable Flow on CI
3. Fixes a lot of flow errors, and suppresses rest

You can search for suppressed warnings with `/* FlowFixMe */`. Most are in `StackRouter`. This is not perfect, but at least now we will be able to run flow on CI and not worry about introducing more flow errors.

Hopefully someone can take up the task of fixing rest of the warnings.

One important change I made is change the signature of `NavigationState` to,

```js
type NavigationState = {
  index: number,
  routes: Array<NavigationRoute | (NavigationRoute & NavigationState)>,
};
```

This is because most of the time routes also have same structure as navigation state with nested navigators.

**Test plan**

Ran the app on Android and iOS and everything runs fine


circle.yml
```diff
@@ -17,3 +17,6 @@ deployment:
     commands:
       - yarn run build-docs
       - ./scripts/deploy-website.sh
+test:
+  pre:
+    - yarn run flow
```
ericvicenti: Nice! Can we get flow feedback to show in PRs?

src/TypeDefinition.js
```diff
@@ -92,12 +93,11 @@ export type NavigationRouter = {
...
 export type NavigationScreenOption<T> =
+  | T
   | (navigation: NavigationScreenProp<NavigationRoute, NavigationAction>,
-    config: NavigationScreenOptionConfig,
-    router?: NavigationRouter) => T
-  | T;
+    config: T) => T;
 
-export type Style = Object | number | false | void;
+export type Style = { [key: string]: any } | number | false | null | void | Array<Style>;
```
ericvicenti: Nice! StyleSheets are crazy but I think you may have nailed it. Maybe we could get RN core to export this type?

src/TypeDefinition.js
```diff
@@ -290,23 +290,18 @@ export type NavigationAction =
...
 };
 
-export type NavigationRouteConfig =
-  | NavigationScreenRouteConfig
-  | NavigationLazyScreenRouteConfig;
+export type NavigationScreenRouteConfig = NavigationScreenRouteConfig<{
+  // React component or navigator for this route */
+  screen: NavigationComponent,
+} | {
+  // React component to lazily require and render for this route */
```
ericvicenti: So, the component deferred with `getScreen` is not supposed to be a navigator/NavigationComponent (have a router).

This type makes that unclear. If this is needed to make flow pass, then I think its fine for now, but otherwise we should probably change this to be a normal react component that receives a `NavigationScreenProp`

satya164: @ericvicenti yeah, couldn't get flow to pass without this. can revisit later

src/TypeDefinition.js
satya164: @ericvicenti haha yeah, it'll be useful

circle.yml
satya164: @ericvicenti you mean comments on code like the react native repo does? there's a bot for that afaik. maybe we can copy it.

satya164:
So this is ready to merge? I'd assume yes since you approved the changes?

ericvicenti:
Yes, I just was wondering if the `getScreen` thing could be fixed. Nevermind, we'll get to it later!

satya164:
@ericvicenti great. I'll try to fix it tonight. shouldn't be hard.

PR #306, opened 2017-02-11
Merge action params into navigator's child screens

jamesisaac:
Without this change, I'm having trouble using react-navigation in real world apps.  I feel like it's a very common pattern to want to bring up a navigator (either Tab or Stack) which is focused around a single entity.

To take an example: say the main navigator of my app is a StackNavigator, with the main screen being a ListView of contacts.  When I click a contact, I want to navigate to a screen with sliding tabs (TabNavigator), where each tab displays a different piece of info about that contact.

With this change, I can simply call `naviagion.navigate('ContactTabNavigator', { contactId: id })`, and have that param automatically passed down to all the nested screens.

Without this change, I can't think of any way to achieve this same effect without some kind of global state to keep track of which contact had been selected.  I've seen some talk of dispatching a Navigate action, with a sub-action to set the params (https://github.com/react-community/react-navigation/issues/143#issuecomment-276829914), but on a TabView with lazyLoad=true that wouldn't help, as it would only set params for the initial tab, so the app crashes.

This is just a quick proof of concept, and only currently works on StackNavigator -> TabNavigator, but I'd like to hear other's thoughts on whether this behaviour makes sense.

#134 might be trying to achieve the same thing here, but I'm not sure - that patch didn't work for me.


ericvicenti:
I think you may be on to something here. I think passing params down may be a pretty sane default for the reducers.

But I'm afraid the proposed change to the reducer may be a bit risky. Generally I'd like to keep the logic for each action under the corresponding `if` block, to make it easy to see what each action does. In your case, I think you may only need to add this logic to the `NavigationActions.NAVIGATE` area of the code.

Also, it would be super helpful if you add a few test cases for this, that demonstrates the desired behavior your change achieves.

jamesisaac:
It will have to target both `NavigationActions.NAVIGATE` and `NavigationActions.INIT` I believe.  The way that actions delegate down is a bit tricky to follow, but I think if I call `naviagion.navigate('ContactTabNavigator', { contactId: id })` and `ContactTabNavigator` has never been shown, it gets passed an INIT action (see my change to StackRouter like 145) and no NAVIGATE action.

Given the above, would you still prefer the param merging duplicated in both the action `if` blocks, or does it make more sense in the current `if (action.params)` standalone block?

And yeah sure, just wanted to make sure this approach makes sense before expanding too much on the solution.

src/routers/__tests__/TabRouter-test.js
```diff
@@ -141,6 +141,42 @@ describe('TabRouter', () => {
     });
   });
 
+  test('Handles passing params to nested tabs', () => {
+    const ChildTabNavigator = () => <div />;
+    ChildTabNavigator.router = TabRouter({ Foo: BareLeafRouteConfig, Bar: BareLeafRouteConfig });
+    const router = TabRouter({ Foo: BareLeafRouteConfig, Baz: { screen: ChildTabNavigator } });
+    const action = router.getActionForPathAndParams('Baz', { foo: '42' });
+    const navAction = { type: NavigationActions.NAVIGATE, routeName: 'Baz', params: { foo: '42' } };
+    expect(action).toEqual(navAction);
```
jamesisaac: I'm having trouble getting this to pass (`action` doesn't contain the params field) - any suggestions?

jamesisaac:
Ok, I've taken this a bit further.  Added a coupe of tests (although issue with `getActionForPathAndParams` - see inline comment), and moved the logic into an `if` block.  It actually seems to be only `NavigationActions.INIT` that needs to be targeted, not `NavigationActions.NAVIGATE`.

Also, in my original post I said this could be useful with a child StackNavigator as well as TabNavigator, but looking more closely at how Stack stores state, I think it only makes sense for Tabs.  With Stack you're explicitly calling Navigate for each screen anyway, so not too hard to pass the params along manually.  It's just Tab which spawns a load of children with no context (params).

src/routers/__tests__/TabRouter-test.js
ericvicenti: That sounds like a separate bug. Maybe you can fix it in a separate PR? If you write test cases like this for it, then we can be sure it won't be a problem from here on out!

satya164:
So when someone does `navigation.navigate` to another tab without params, do these params get overwritten or preserved?

jamesisaac:
Given the following nav hierarchy...

```
- MyStackNav
  - MyTabNav
    - Tab1
    - Tab2
```

...are you asking what happens if someone executes:

```js
naviagion.navigate('MyTabNav', { foo: 'bar' });
naviagion.navigate('Tab2');
```

If so, the only behaviour this PR should change is that when MyTabNav is navigated directly to with params, those params get cascaded down to all the tabs.  So to answer your question, the same thing would happen as if you ignored this PR and did:

```js
naviagion.navigate('Tab2', { foo: 'bar' });
naviagion.navigate('Tab1');
naviagion.navigate('Tab2');
```

Which is, I'm guessing, that the params get preserved (and merged with any new params).

satya164:
> are you asking what happens if someone executes

yes

> Which is, I'm guessing, that the params get preserved (and merged with any new params)

Not sure what the current behaviour is, but I think we need to ensure that they don't change/get removed.

Also what happens when you have a stack navigator inside one of the tabs? Does the initial screen of the stack navigator receive the params? I think it should.

satya164:
cc @ericvicenti do you think this is good to go?

jamesisaac:
> Not sure what the current behaviour is, but I think we need to ensure that they don't change/get removed.

Just tried writing a test for this, and looks like NAVIGATE on a TabRouter doesn't support params to begin with?  Which makes this PR even more important I suppose as it's the only way to pass any params to a TabNavigator.  I've added a test to confirm they don't get changed/removed with NAVIGATE.

> Also what happens when you have a stack navigator inside one of the tabs? Does the initial screen of the stack navigator receive the params? I think it should.

Yeah I just ran into a need for this in my app.  I had `StackA -> StackB -> ScreenA`.  While in StackA, calling `navigate('StackB', params)` didn't pass the params to ScreenA.  I suppose I could have solved this with a sub-action (maybe... not sure it works if you're navigating to the initial screen which may already be mounted?), but this way seems more intuitive.

jamesisaac:
I tried writing a test with `StackA -> TabA -> StackB -> ScreenA` as that's what you asked about, but StackB didn't seem to be initialising... the corresponding leaf on TabA didn't contain a routes key.  Couldn't really figure it out.

satya164:
Hey! Would love to get this merged. Can you address the flow errors in CI? Happy to help if you need help with fixing them.

almadsen:
Lots of linking to this issue, is it being worked on, and is there a temp workaround for this?

jamesisaac:
Yes it's being worked on... this was opened on Saturday and there's been discussion/commits pretty much daily since.  The temp workaround is that you can link to my repo in your package.json which contains the fix:

`"react-navigation": "git@github.com:jamesisaac/react-navigation.git",`

@satya164 I should have time to sort the Flow issues tonight or tomorrow, but feel free to amend yourself if needed.

jamesisaac:
@satya164 Flow tests should now all be passing.

Had to do a couple of `$FlowFixMe`s in the StackRouter tests - I see that's been done elsewhere too when trying to check deep parts of state object.  Can't check the whole thing as StackRouter gives a random key.

satya164:
Hey @ericvicenti , do you think we can merge this?

src/routers/__tests__/TabRouter-test.js
```diff
@@ -139,6 +139,44 @@ describe('TabRouter', () => {
     });
   });
 
+  test('Handles passing params to nested tabs', () => {
+    const ChildTabNavigator = () => <div />;
+    ChildTabNavigator.router = TabRouter({ Boo: BareLeafRouteConfig, Bar: BareLeafRouteConfig });
+    const router = TabRouter({ Foo: BareLeafRouteConfig, Baz: { screen: ChildTabNavigator } });
+    const navAction = { type: NavigationActions.NAVIGATE, routeName: 'Baz', params: { foo: '42', bar: '43' } };
+    // const action = router.getActionForPathAndParams('Baz', { foo: '42' });
+    // expect(action).toEqual(navAction);
```
jamesisaac: Ok I've removed it.  It's an odd behaviour, not caused by this PR, which I wasn't sure how to go about fixing.

PR #316, opened 2017-02-13
Fixes TabBar upperCaseLabel option

jeremistadler:
The TabBar has a upperCaseLabel option, default as true on android but the labels were never uppercased. 

This uppercases them in the render method.


satya164:
I've fixed this in another PR. I'll close this since otherwise it'll cause conflicts, and that's a pretty big PR :D

Thank you for the PR>

PR #317, opened 2017-02-13
Export Header component

kennydee:
Hi,

This Pull Request allow the `Header` component to be imported as it was with `NavigationExperimental` 

```javascript
import { Header } from 'react-navigation';
```

See previous NavigationExperimental exports
https://github.com/facebook/react-native/blob/master/Libraries/NavigationExperimental/NavigationExperimental.js#L31

P.S: i know, the syntax is not the same as the others exports, because the Header component don't use `default` export.

If you have suggestion or remarks, please don't hesitate.

Thanks ;)

Kenny


satya164:
Pretty sure you can do `import { CardStack } from 'react-navigation'` and then use `CardStack.Header`

kennydee:
Oh s... :)
Thanks ;)

Issue #256, opened 2017-02-09
Jest fails with "SyntaxError: Unexpected reserved word" everytime


half-shell:
Hello,
I'm working with react-native and when I run the basic `index.android.js` test file, it systematically fails with the following:
```
SyntaxError: Unexpected reserved word
      
  at transformAndBuildScript (node_modules/jest-runtime/build/transform.js:320:12)
  at Object.StackNavigator (node_modules/react-navigation/src/react-navigation.js:15:33)
  at Object.<anonymous> (src/appNavigator.js:13:37)
```
It looks like I'm doing something wrong, but I can't figure out what. I've tried fiddling around with it (mostly changing variables name), but I keep getting the same error.

The application works though, it doesn't show any error. It is the simpliest navigation app I could make, following the react-navigation doc, using redux.
I was wondering if this happened to anyone else, or if anyone has a clue of the dumb mistake I made.

grabbou:
cc: @ferrannp

ferrannp:
You need to use `transformIgnorePatterns`, see from the [docs](https://facebook.github.io/jest/docs/tutorial-react-native.html#transformignorepatterns-customization):

> The transformIgnorePatterns option can be used to whitelist or blacklist files from being transformed with babel. Many react-native npm modules unfortunately don't pre-compile their source code before publishing.
> 
> By default the jest-react-native preset only processes the project's own source files and react-native. If you have npm dependencies that have to be transformed you can customize this configuration option by whitelisting modules other than react-native.

So you can add in your `package.json`:

```
  "jest": {
    "preset": "react-native",
    "transformIgnorePatterns": [
      "node_modules/(?!(jest-)?react-native|react-navigation)"
    ]
  }
```

Unfortunately, it seems that we will still have an error related to [PlatformHelpers.native.js](https://github.com/react-community/react-navigation/blob/master/src/PlatformHelpers.native.js):

```
 Invariant Violation: Native module cannot be null.
      at invariant (node_modules/fbjs/lib/invariant.js:44:15)
      at Linking.NativeEventEmitter (node_modules/react-native/Libraries/EventEmitter/NativeEventEmitter.js:32:1)
      at new Linking (node_modules/react-native/Libraries/Linking/Linking.js:119:141)
      at Object.<anonymous> (node_modules/react-native/Libraries/Linking/Linking.js:191:16)
      at Object.Linking (node_modules/react-native/Libraries/react-native/react-native.js:91:22)
[...]

Issue #236, opened 2017-02-08
Hide tabbar label completely (labelStyle not working?)


flavordaaave:
Is there an easy way to just hide the Tabbar Labels completely?
I tried to apply some labelStyle (width height: 0) but labelStyle does not seem to work at all...

```
const Tabs = TabNavigator({
  Home: {
    screen: HomeScreen,
    path: 'home',
    navigationOptions: {
      tabBar: {
        label: 'TEST',
      },
    },
  },
...
}, {
  swipeEnabled: false,
  animationEnabled: true,
  backBehavior: 'none',
  tabBarOptions: {
    labelStyle:{ backgroundColor: 'yellow', color: 'red', fontSize: 30, }
  },
})
```

The labelStyle in the above code does NOT change anything.

kylpo:
https://github.com/react-community/react-navigation/pull/234 might be what you're looking for. Just merged in.

manavbhikadiya:
try this,

tabBarOptions={{
showLabel : false
}}

PR #331, opened 2017-02-14
add jest config for react-navigation in docs (#256)

half-shell:
This pull request concerns issue #236. It adds the jest configuration necessary to make react-navigation pass the basic jest's rendering test.

I wasn't sure where to put it since there is no test-related section in the guides. I decided to put it in the redux integration because I was following it when I noticed the issue.

**Test plan (required)**
No test required since this is for docs.


ferrannp:
I think it would be nice to have a Test section in the docs rather than having this as part of Redux integration.

grabbou:
+1 for that suggestion, what you think @ericvicenti ?

ericvicenti:
Yeah I think it would be great to have a guide for testing.

Not related to this exact issue but Ideally the testing doc would discuss how you can test your app's navigation logic by writing tests for `AppNavigator.router.getStateForAction`

half-shell:
@satya164 @ferrannp I sure agree. But I don't think I'm qualified enough to get the test guide started. I'll add the issue related snippet if it is needed.
Then again, what about the `Invariant Violation: Native module cannot be null.` discussed [here](https://github.com/react-community/react-navigation/issues/256)?

tugorez:
@half-shell @ferrannp  Hi guys, I managed all this test-concerns (with jest) using this in my package.json 
```javascript
{
   .
   .
   .
  "jest": {                                                                      
    "preset": "react-native",                                                    
    "setupFiles": ["<rootDir>/jest/setup.js"],                                   
    "transformIgnorePatterns": [                                                 
      "node_modules/(?!(jest-)?react-native|react-navigation)"                   
    ]                                                                            
  },
  .
  .
  .                                                                              
}  
```

and this mock in **<rootDir>/jest/setup.js**
```javascript
jest.mock('Linking', () => ({                                                    
  addEventListener: jest.fn(),                                                   
  removeEventListener: jest.fn(),                                                
  openURL: jest.fn(),                                                            
  canOpenURL: jest.fn(),                                                         
  getInitialURL: jest.fn(),                                                      
})); 
```

an this in my .flowconfig (in case you're using flow too)
```
[ignore] 

.
.
.
; react-navigation isn't compatible with the newest flow version, ignore it         
.*/node_modules/react-navigation*
.
.
.                                                     
 
```

ericvicenti:
Lets get this in for now, and break out a testing section of the docs later on

Issue #320, opened 2017-02-13
Navigation Actions Doc


ericvicenti:
The docs for navigation actions are hard to find. There is currently an incomplete list of actions under the `dispatch` section of [the navigation prop doc](https://reactnavigation.org/docs/navigators/navigation-prop).

We should remove the list from that page and instead link to a new page for the actions, that can provide more thorough documentation for `NavigationActions`. Its not clear exactly where in the sidebar this doc should go, but probably in the Navigators section.

ElliottJRo:
hey @ericvicenti I can do this but one problem I ran into with dispatching actions is the additional `action` param that is the "sub-action to run in the child router". When I add in a `RESET` action as a sub-action (inside a `NAVIGATE`) I expected the `Reset` action to be handled by the new router given by the `NAVIGATE` action by it was handled by the router before transition. Is this expected?

Also I don't see `NavigationActions.URI` handled in `StackRouter` or `TabRouter`.

ElliottJRo:
@ericvicenti  Can be closed now!

guigrpa:
@elliotjro what about your question on the embedded reset subaction being handled by the current router (before the transition) instead of by the future one? Did you figure it out?

ElliottJRo:
@guigrpa no sorry I did not. I ended up writing my own custom actions to deal with that certain scenario.

Issue #319, opened 2017-02-13
Docs: Navigation Prop doc should discuss the navigator's navigation prop


ericvicenti:
The top of the [Navigation Prop doc](https://reactnavigation.org/docs/navigators/navigation-prop) should be rewritten to explain that the navigation prop is passed to every navigation-aware component, including navigators. The big exception is that a navigator's navigation prop may not have the helper functions on it, it may only have `state` and `dispatch`. 

Also, people don't always hook things up to redux correctly, because they mis-understand the navigator's top-level API, where the navigation prop is optional. The navigator will maintain its own state if it doesn't get a navigation prop, but this is not a feature you generally want to use when hooking your app up with redux. For navigators that are nested inside of your main navigator, you *always* want to pass the screen's navigation prop down. This allows your top-level navigator to communicate and provide state for all the children navigators. Only your top-level router needs to be integrated with redux, because all the other routers are inside it.

Who would like to take on this docs improvement work? I've seen a lot of confusion around these things, so I hope we can make it more clear to the community what the difference is between a navigator's navigation prop in comparison to a navigation screen prop.

grabbou:
Closing as changes were implemented already.

PR #338, opened 2017-02-14
Navigation Actions Doc

ElliottJRo:
Related: #320, #319

1. Separated Navigation actions from Navigation prop docs
2. Added `Init` and `Back` action docs
3. Added some documentation on what params can be set on actions (eg: routeName on `navigate`)

TODO:

- [x] Add in actual link to Navigation Actions from Navigation Prop doc
- [x] Find all params that can be set on actions and what they do
- [x] Add in explanation for `RESET` action `index` param
- [x] Add in explanation of navigator's navigation prop
- [x] ~~Add examples on how some nested actions (sub-action) are handled~~ This belongs somewhere else I think


burdiuz:
Could you add an explanation what means index passed to reset action

machadogj:
I have some special concerns about the documentation of the "goBack" helper (or back action). The difference between `goBack()` and `goBack(null)` is not clear, and neither is the behavior of `goBack('screen-123')`.

What I suggest, is to display for a given stack, how each of these will affect the stack.

The other concern I have, is with the `goBack('screen-123')`, which apart from not being 100% clear that the `screen-123` will also be dismissed, when I look at the code, it is using the route's `key` to compare. What I think would be most helpful on the sample code, is to show how to obtain the key for a given route (ie: if I do `navigate('home')`, then the key ends up being a weird string that has nothing to do with `home`, at least when used with `redux`).

ElliottJRo:
Hey @machadogj thanks for the feedback! I don't think there is a way to grab the key without using `StateUtils` or some other internal function. Also the keys are auto generated using current time I believe. My understand of the Back action is that it pops the described route (and anything on top) from `state.routes` array - if that described route is found. By default `goBack` simply pops last route in `state.routes` and decrements the index.. I think.

`goBack(null)` vs `goBack()` is unclear to me as well, looking at the [source](https://github.com/react-community/react-navigation/blob/master/src/routers/StackRouter.js#L204) seems like `goBack(null)` isn't handled but I just copied what was there before.

@satya164 can you confirm?

ericvicenti:
I'm trying to remember why `goBack()` would have different behavior from `goBack(null)`, and it makes no sense to me. We should probably make them behave the same

ElliottJRo:
Did some digging in [source](https://github.com/react-community/react-navigation/blob/master/src/addNavigationHelpers.js#L18) and found that only does a undefined check.

ElliottJRo:
ps I am not sure how the doc website is actually done so I don't know what the Navigation Actions link would be

ericvicenti:
@ElliottJRo, For links you'd just use a domain-absolute path, like `/docs/routers/api`

ElliottJRo:
I mean since Navigation Actions is a new doc, I don't know where it would go on the side bar / the actual link since the file is located in `/docs/guides/`.
I see that screen navigation prop doc is also located in `/docs/guides` but the url is actually `/docs/navigators/navigation-prop`. I am having trouble understanding how this is done.

ericvicenti:
Have you seen the website section of the contributors docs?

https://reactnavigation.org/docs/guides/contributors#4.-Developing-Docs

It explains that everything is set up in `App.js`

ElliottJRo:
hey @ericvicenti Ive added the new doc into above mentioned file and was able to navigate around the doc site locally

docs/guides/Screen-Navigation-Prop.md
```diff
@@ -1,16 +1,28 @@
...
+* `dispatch` - send an action to router
 
+*NOTE:* The `navigation` prop is passed down to every navigation-aware component including navigators. The big exception is that a navigator's `navigation` prop may not have the helper functions (`navigate`, `goBack`, etc); it may only have `state` and `dispatch`. In order to `navigate` using the navigator's `navigation` prop, you will have to `dispatch` using an [action creator](navigation-actions).
+
+*Notes regarding hooking things up with Redux*
+
+> People don't always hook things up to redux correctly, because they mis-understand the navigator's top-level API, where the navigation prop is optional. The navigator will maintain its own state if it doesn't get a navigation prop, but this is not a feature you generally want to use when hooking your app up with redux. For navigators that are nested inside of your main navigator, you always want to pass the screen's navigation prop down. This allows your top-level navigator to communicate and provide state for all the children navigators. Only your top-level router needs to be integrated with redux, because all the other routers are inside it.
```
ericvicenti: Thanks for writing this note, the concept is super important! Do you think you could combine this note with the one above? Also, it might be better to find a different place to explain the distinction between a navigator navigation prop and a screen navigation prop, and link to that from here.

Or, we could theoretically expand this page and rename it to 'Navigation Props' or something. I'll let you make the call.

ElliottJRo: hey @ericvicenti I just left this part unedited because I didn't quite understand what was going on with the redux integrations. According to the [redux integration guide](https://reactnavigation.org/docs/guides/redux) I read it as: if you pass in a navigation prop to a `Navigator`:
1. it effectively replaces the router -> meaning you need to implement `getStateForProps`, `getStateForAction` not that they will actually be called since `react-navigation`'s actions aren't being used
2. All the built in actions like `Back`, `Reset` need to be re-implemented or copied inside your own app when creating your redux store/connecting to app

According to your notes when hooking up with Redux, you can hook up just the top level Navigator to Redux and still have the child Navigators' Routers with `react-navigation`'s router api..?

There must be something I am missing, could you please point me in the right direction?

ElliottJRo:
Hey @ericvicenti, just a reminder -> waiting on your input regarding the redux integration note

docs/guides/Screen-Navigation-Prop.md
ericvicenti: @ElliottJRo, apologies for the mega-delay on this!

If I understand you correctly, you're wondering how the app can know what state to provide to the top-level navigator, and how to respond to actions that come out of the top level navigator's dispatch channel.

Say I have an `App` defined with `const App = StackNavigator({ ... });`. To determine what state to provide to `<App navigation={{state}} />`, you can use `getStateForAction` on the App's static router: `const initState = App.router.getStateForAction(initAction);`

ericvicenti:
This looks pretty great. Lets get it in and keep on iterating!

ElliottJRo:
Hey @ericvicenti sorry for the delay, been working on API side of my project so i've been neglecting the client side. I still dont quite understand the concept that I copied your notes on :/ I will create a new issue requesting for help!

docs/guides/Screen-Navigation-Prop.md
JulioC: Hello @ericvicenti!

I have my root navigator storing its state in redux, but I'm having issues connecting a child navigator that is inside a component used as `screen` for a navigator route.

What should I pass to the navigator so I can refer to its routes when using the root `Navigator.router.getStateForAction`?

PR #341, opened 2017-02-15
Remove 'appBar' child component to fully support header styling

juhasuni:
This PR is related to #105 but instead of introducing two new props, we're adding full support for Header layout customisation using `style` prop.

Now, the header size can be also configured on per-screen basis as demonstrated here:

```
class FullScreen extends Component {
  static navigationOptions = {
    header: {
      style: {
        paddingTop: 0, // clears the default Header margin for the status bar
        height: 70 // sets new height for the Header
      }
    }
  }
}
```


src/views/Header.js
```diff
@@ -279,26 +277,23 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+        titleComponent,
+        leftComponent,
+        rightComponent
+      ];
     }
 
     return (
-      <Animated.View {...rest} style={[styles.container, style]}>
-        {appBar}
+      <Animated.View {...rest} style={[styles.container, {marginTop: STATUSBAR_HEIGHT, height: APPBAR_HEIGHT}, style]}>
```
satya164: This should be padding, othewise the statusbar color won't match appbar

juhasuni: Okay, I submitted a fix for this, which uses additional child component with `flex: 1`.

src/views/Header.js
```diff
@@ -279,18 +277,16 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+        titleComponent,
+        leftComponent,
+        rightComponent
+      ];
     }
 
     return (
       <Animated.View {...rest} style={[styles.container, style]}>
-        {appBar}
+        <View style={{flex: 1}}>{children}</View>
```
satya164: Can you move `{ flex: 1 }` to `StyleSheet.create` below?

juhasuni: Done

grabbou:
I'd say we should wait for #257 to be merged, I've changed a lot in `render` method. Otherwise, I can rebase.

satya164:
@grabbou since we're doing a release today, will be good to have this

grabbou:
Cool, then just go ahead. I was busy with other OSS today, so haven't managed to finish on time. Let me know when the release gets cut and I'll follow up with a changelog.

satya164:
Only if this CI things works. lol xD

juhasuni:
Yup... it is nerve-wracking to work without CI and it is equally frustrating to try to keep them working.

PR #345, opened 2017-02-15
Header: Use static properties `Title` and `BackButton`

flochtililoch:
...instead of imported `HeaderTitle` and `HeaderBackButton`

**Motivation**

Allow overriding the HeaderTitle and HeaderBackButton component easily.
The header API already exposes these properties but does not use them to render the underlying components title and back button.

**Test**

Tested with the navigation playground project:

Added the following to `SimpleStack.js`:

```diff
diff --git a/examples/NavigationPlayground/js/SimpleStack.js b/examples/NavigationPlayground/js/SimpleStack.js
index 3a0a3d7..d85c48e 100644
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -9,9 +9,14 @@ import {
 } from 'react-native';
 import {
   StackNavigator,
+  CardStack,
 } from 'react-navigation';
 import SampleText from './SampleText';

+CardStack.Header.BackButton = ({ onPress }) => (
+  <Button title="Back" onPress={onPress}/>
+);
+
 const MyNavScreen = ({ navigation, banner }) => (
   <ScrollView>
     <SampleText>{banner}</SampleText>
```

produces the following experience:

| iOS | Android |
|-----|--------|
| ios mov | android mov |

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -111,7 +111,7 @@ class Header extends React.Component<void, HeaderProps, void> {
     const titleStyle = this._getHeaderTitleStyle(props.navigation);
     const color = this._getHeaderTintColor(props.navigation);
     const title = this._getHeaderTitle(props.navigation);
-    return <HeaderTitle style={[color ? { color } : null, titleStyle]}>{title}</HeaderTitle>;
+    return <Header.Title style={[color ? { color } : null, titleStyle]}>{title}</Header.Title>;
   };
 
   _renderLeftComponent = (props: SubViewProps): ?React.Element<HeaderBackButton> => {
@@ -121,7 +121,7 @@ class Header extends React.Component<void, HeaderProps, void> {
     const tintColor = this._getHeaderTintColor(props.navigation);
     // @todo(grabobu):
     // We have implemented support for back button label (which works 100% fine),
-    // but when title is too long, it will overlap the <HeaderTitle />.
+    // but when title is too long, it will overlap the <Header.Title />.
     // We had to revert the PR implementing that because of Android issues,
     // I will land it this week and re-enable that for next release.
     //
@@ -131,7 +131,7 @@ class Header extends React.Component<void, HeaderProps, void> {
     // });
     // const backButtonTitle = this._getHeaderTitle(previousNavigation);
     return (
-      <HeaderBackButton
+      <Header.BackButton
         onPress={props.onNavigateBack}
         tintColor={tintColor}
       />


satya164:
You can use `left` and `title` properties in `navigationOptions`  to use your custom elements. Overwriting static properties in core views is not supported.

flochtililoch:
@satya164 it seems like the title view is not customizable via `navigationOptions`, only its content string - am I missing something here?
Regarding the back button, I can indeed customize it via the `header.left` prop, but I then need to re-implement the decision as whether the button should be displayed or not, i.e. right now I'm doing something like this:

```javascript
...
header: (navigation) => {
  return {
    left: navigation.state.index === 0 ? null : (<MyBackButton/>)
  }
}
```
but it seems providing a way to customize the view itself, without having to decide when to render it, would be a nice to have feature.

Since the core `Header` view already provides "read only" hooks to `HeaderBackButton` and `HeaderTitle` views, via the static property, why not allowing these to be "read/write"?

satya164:
> it seems like the title view is not customizable via navigationOptions, only its content string - am I missing something here?

You can pass a react element as title.

> but it seems providing a way to customize the view itself, without having to decide when to render it, would be a nice-to-have feature.

You can specify `undefined` to render the default button. If it doesn't work, then it's bug and we need to fix it.

> Since the core Header view already provides "read only" hooks to HeaderBackButton and HeaderTitle views, via the static properties, why not allowing these to be "read/write"?

Same reason you don't overwrite things in the React Native API or any library. Besides, there's no reason to overwrite them.

flochtililoch:
> You can pass a react element as title.
Can you iterate on that? I have not see a single example that documents it, and [this code](https://github.com/react-community/react-navigation/blob/02a388073cd520d94c30ffc5f01d2e1754e93c9c/src/views/Header.js#L91) seems to filter `header.title` prop to be of string type only.

doing:

```javascript
...
static navigationOptions = {
    title: () => (<Text>My Title</Text>),
   ...
```

or

```javascript
...
static navigationOptions = {
    title: (<Text>My Title</Text>),
   ...
```

does not actually render a title view.

> You can specify undefined to render the default button. If it doesn't work, then it's bug and we need to fix it.
I want to render a custom button without having to decide _when_ to render it - how does setting `left: undefined` help?

> Besides, there's no reason to overwrite them.
If you can show how to achieve what I'm trying to do then I can agree with that...

satya164:
You're trying to set header's title, so you'd pass `title` element to header

```js
header: {
  title: <MyComponent />
}
```

> I want to render a custom button without having to decide when to render it - how does setting left: undefined help?

No reason you cannot use a conditional

```js
left: navigation.state.routeName === 'Home' ? <HamburgerMenu /> : undefined
```

> I want to render a custom button without having to decide when to render it - how does setting left: undefined help?

Even if it was not possible, overwriting library API is never the right API.

flochtililoch:
> You're trying to set header's title, so you'd pass title element to header

That did work indeed, thanks for pointing that out!

> No reason you cannot use a conditional

I suppose it fine but it becomes verbose to re-implement the visibility logic while the only thing I want to do is change the appearance of the button.

> Even if it was not possible, overwriting library API is never the right API.

Cool.

PR #373, opened 2017-02-17
Support back nav on Apple TV with menu button

dlowder-salesforce:
First Apple TV commit... support back nav for navigators that use `src/views/Header.js`

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -9,6 +9,8 @@ import {
   View,
 } from 'react-native';
 
+import TVEventHandler from 'react-native/Libraries/Components/AppleTV/TVEventHandler';
+
 import ReactComponentWithPureRenderMixin from 'react/lib/ReactComponentWithPureRenderMixin';
 
 import HeaderTitle from './HeaderTitle';
@@ -78,6 +80,24 @@ class Header extends React.Component<void, HeaderProps, void> {
       nextProps,
       nextState
     );
+  }
+
+  _tvEventHandler: TVEventHandler;
+
+  componentDidMount(): void {
+    this._tvEventHandler = new TVEventHandler();
+    this._tvEventHandler.enable(this, function(cmp, evt) {
+      if (evt && evt.eventType === 'menu') {
+        cmp.props.onNavigateBack && cmp.props.onNavigateBack();
+      }
+    });
+  }
+
+  componentWillUnmount(): void {
+    if (this._tvEventHandler) {
+      this._tvEventHandler.disable();
+      delete this._tvEventHandler;
+    }
   }
 
   _getHeaderTitle(navigation: Navigation): ?string {
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -9,6 +9,7 @@ import {
   View,
 } from 'react-native';
 
+      /* $FlowFixMe */
 import TVEventHandler from 'react-native/Libraries/Components/AppleTV/TVEventHandler';
 
 import ReactComponentWithPureRenderMixin from 'react/lib/ReactComponentWithPureRenderMixin';


dlowder-salesforce:
@satya164 @ericvicenti this change allows StackNavigator to work as expected on Apple TV

src/views/Header.js
```diff
@@ -9,6 +9,9 @@ import {
   View,
 } from 'react-native';
 
+      /* $FlowFixMe */
```
satya164: nit: remove extra extra space before comment

src/views/Header.js
```diff
@@ -80,6 +83,24 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+      if (evt && evt.eventType === 'menu') {
+        cmp.props.onNavigateBack && cmp.props.onNavigateBack();
+      }
+    });
+  }
+
+  componentWillUnmount(): void {
+    if (this._tvEventHandler) {
+      this._tvEventHandler.disable();
+      delete this._tvEventHandler;
```
satya164: This is unnecessary

src/views/Header.js
```diff
@@ -80,6 +83,24 @@ class Header extends React.Component<void, HeaderProps, void> {
     );
   }
 
+  _tvEventHandler: TVEventHandler;
+
+  componentDidMount(): void {
+    this._tvEventHandler = new TVEventHandler();
+    this._tvEventHandler.enable(this, function(cmp, evt) {
```
satya164: If I'm understanding this correctly, the first parameter is `this` and second parameter is event listener. Am I right? looks weird

src/views/Header.js
```diff
@@ -80,6 +83,24 @@ class Header extends React.Component<void, HeaderProps, void> {
...
+      if (evt && evt.eventType === 'menu') {
+        cmp.props.onNavigateBack && cmp.props.onNavigateBack();
```
satya164: If you use arrow function instead of `function` keyword, you can just do `this.props.onNavigateBack`

ericvicenti:
The PR description doesn't make it clear what behavior this changes. (I'm not quite sure what TVEventHandler does). I think direct integration into the header view may not be the best idea.

Can you verify this doesn't cause any bugs on iOS or Android? Please include a test plan in future PRs.

I have an AppleTV myself so I would be curious on how to test this. Is there a getting started guide published anywhere?

dlowder-salesforce:
@ericvicenti you can run this on Apple TV today... just go through the first StackNavigator example at https://reactnavigation.org/docs/intro/ , but in Xcode select the SimpleApp-tvOS target instead of the SimpleApp target.  When you start the app on Apple TV simulator, you will see that the "Chat With Lucy" button is highlighted, and you can press the TV remote button to push the next screen on the navigation stack.  With my change from this PR, you will also be able to use the remote's Menu button to navigate back to the home screen.  The TVEventHandler doesn't do anything on iOS or Android, so this change will have no effect on those platforms.  I'll have a look to see if there is any way to add automation for Apple TV.

src/views/Header.js
dlowder-salesforce: That's correct.

src/views/Header.js
dlowder-salesforce: The delete might not be necessary, but the disable() is definitely a good idea....

satya164: yes, my comment was referring to `delet`

src/views/Header.js
satya164: let's remove the `function` and use an arrow function then?

satya164:
@dlowder-salesforce back button handling logic lives here https://github.com/react-community/react-navigation/blob/a7b124305393a8e392621a32e49fa96a3a14c6c9/src/createNavigationContainer.js#L89

The `BackAndroid` module is exported from here - https://github.com/react-community/react-navigation/blob/master/src/PlatformHelpers.native.js

I think we should rename it to `BackHandler` or something, and implement same interface for tvOS

dlowder-salesforce:
Thanks @satya164 for the tip, I'll take a look and rework this PR.

skevy:
@dlowder-salesforce I'm pretty sure this will break things when using react-native-web and using this lib in a web-based context. We should resolve that in some fashion before we merge this.

dlowder-salesforce:
Closing this pull request and adding the TV event handler in react-native's BackAndroid, as suggested.

Issue #383, opened 2017-02-18
Undismissable log-in modal in StackNavigator


Ashoat:
I am looking to implement a design where when you log out, a log-in modal pops in, and when you log in, the log-in modal disappears. A StackNavigator with `{ headerMode: 'none', mode: 'modal' }` serves this purpose quite well on both iOS and Android! When I add the route (`navigate`) for the log-in modal, it gracefully swipes/fades in, and when I remove that route (`goBack`) it graceful swipes/fades out.

My only issue, however, is that it's still possible for the logged out user to return to the logged in view. On iOS they can pan the modal away; on Android, they can use the hardware back button.

I tried simply reseting the state between a single main app route and a log-in modal route, but that means I can't use those beautiful swipe/fade animations.

I also tried combining the animation effects of `navigate`/`goBack` with `reset`. The plan was that on log-out, I would first `navigate` to the log-in modal, and then get rid of the app route with a `reset` action (probably through some callback triggered in `onTransitionEnd`). On log-in, I would first `reset` so there are two routes (the first being the app, and the second being the modal), and then `goBack` to animate the modal disappearing. However, this doesn't work, in part because it seems `reset`s can trigger animations too, and in part because I couldn't get a `goBack` after a `reset` to actually do anything.

It seems that the simplest solution here would be a simple mechanism to disable navigation functionality. I can't seem to find an option like that in the docs, though. I had a thought that I could simply override the reducer to drop any `goBack`-type actions when I'm on the log-in modal, but that doesn't address the visual behavior of the pan gesture, which happens outside the context of any actions.

**What do you guys think of a navigation option that would disable the "pan handlers" in Card for a given Screen, as well as the hardware back button behavior?**

Note: I have react-navigation set up to go through my Redux store. Not sure if that might be relevant.

Ashoat:
Did some digging through the source code and came up with a patch to propagate `gesturesEnabled` to the right place. Then discovered #292, which is basically what I have in my sandbox.

The hardware back button has to be handled separately, but this is definitely doable (especially in my case, since I'm handling that myself given the Redux integration).

Closing this issue in favor of discussing further on the PR.

Ashoat:
(Actually, ended up creating my own PR in #385)

PR #385, opened 2017-02-19
Make it possible for a screen to set gesturesEnabled

Ashoat:
Context for the problem I was facing #383. A related `gesturesEnabled` PR with some additional context in #292.

In StackNavigator, sometimes it's useful to prevent somebody from dismissing a route. My use case is a log-in screen that I display as a modal above the rest of my app.

Test plan: I hooked this into my own app, and verified that I was able to set gesturesEnabled via navigation options. flow/jest pass

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -125,6 +125,7 @@ All `navigationOptions` for the `StackNavigator`:
   - `style` - Style object for the navigation bar
   - `titleStyle` - Style object for the title component
   - `tintColor` - Tint color for the header
+- `gesturesEnabled` - Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
 
 ### Navigator Props
 
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -125,7 +125,8 @@ All `navigationOptions` for the `StackNavigator`:
   - `style` - Style object for the navigation bar
   - `titleStyle` - Style object for the title component
   - `tintColor` - Tint color for the header
-- `gesturesEnabled` - Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
+- `cardStack` - a config object for the card stack:
+  - `gesturesEnabled` - Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
 
 ### Navigator Props
 
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -185,6 +185,11 @@ export type NavigationScreenOptions = {
    * Options passed to the drawer for this screen.
    */
   drawer?: NavigationScreenOption<DrawerConfig>;
+  /**
+   * Whether you can use gestures to dismiss this screen.
+   * Defaults to true on iOS, false on Android.
+   */
+  gesturesEnabled?: bool,
 };
 
 export type NavigationScreenConfig = {
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -167,6 +167,14 @@ export type DrawerConfig = {
   label?: string;
 };
 
+export type CardStackConfig = {
+  /**
+   * Whether you can use gestures to dismiss this screen.
+   * Defaults to true on iOS, false on Android.
+   */
+  gesturesEnabled?: bool;
+};
+
 export type NavigationScreenOptions = {
   /**
    * Title is rendered by certain navigators, e.g. the tab navigator,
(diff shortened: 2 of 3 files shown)


Ashoat:
Note that I removed the `gesturesEnabled` prop from `CardStack`. The thinking was that it would be set via navigationOptions (either as a static member of the screen, or the defaults specified in the StackNavigator config), and there wasn't any existing use case for setting it as a prop.

docs/api/navigators/StackNavigator.md
```diff
@@ -125,6 +125,7 @@ All `navigationOptions` for the `StackNavigator`:
   - `style` - Style object for the navigation bar
   - `titleStyle` - Style object for the title component
   - `tintColor` - Tint color for the header
+- `gesturesEnabled` - Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
```
satya164: nit: indent

src/TypeDefinition.js
```diff
@@ -185,6 +185,11 @@ export type NavigationScreenOptions = {
    * Options passed to the drawer for this screen.
    */
   drawer?: NavigationScreenOption<DrawerConfig>;
+  /**
+   * Whether you can use gestures to dismiss this screen.
+   * Defaults to true on iOS, false on Android.
+   */
+  gesturesEnabled?: bool,
```
satya164: can you put it inside another object `cardStack`? we'll likely add more options in future and it'll be useful to access navigation state and params.

Ashoat:
Okay, I encapsulated the new config param with a `cardStack` config object.

(The indentation nit you pointed out on the documentation is on purpose; check the preview)

src/views/CardStack.js
```diff
@@ -343,7 +331,16 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
     let panHandlers = null;
 
-    if (this.props.gesturesEnabled) {
+    const cardStackConfig = this.props.router.getScreenConfig(
+      props.navigation,
+      'cardStack'
+    ) || {};
+    const gesturesEnabledConfig = cardStackConfig.gesturesEnabled;
+    const gesturesEnabled =
+      (gesturesEnabledConfig === true || gesturesEnabledConfig === false)
```
satya164: `typeof gesturesEnabledConfig === 'boolean'`

Ashoat: Okay, done. Also updated the styling of the ternary expression to match what was used elsewhere in the file

satya164:
cc @ericvicenti

src/views/CardStack.js
```diff
@@ -343,7 +331,15 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
 
-    if (this.props.gesturesEnabled) {
+    const cardStackConfig = this.props.router.getScreenConfig(
+      props.navigation,
+      'cardStack'
+    ) || {};
+    const gesturesEnabledConfig = cardStackConfig.gesturesEnabled;
+    const gesturesEnabled = typeof gesturesEnabledConfig === 'boolean' ?
+      gesturesEnabledConfig :
+      Platform.OS === 'ios';
```
ericvicenti: This logic looks good to me, but can you write a comment describing the platform defined default?

Ashoat:
Okay, added a comment.

One thought I had - would this config parameter be better named `dismissGesturesEnabled`? I think it's a bit more descriptive of what it actually controls, although if there are future plans to add non-dismiss gestures to `Card`/`CardStack` then it wouldn't really be helpful.

satya164:
I think `gesturesEnabled` is fine.

RichardLindhout:
If you want to do this on your login screen. You can do something like this.
``` 
import {
  StackNavigator as stackNavigator,
  

} from 'react-navigation'


import Login from './routes/Login'
import Home from './routes/Home'

import EditTimer from './routes/EditTimer'

const MainNavigator = stackNavigator(
  {
    Home: { screen: Home },
    EditTimer: { screen: EditTimer },
  },
  {
    cardStyle: {

      opacity: 1,
    },
    navigationOptions: {


      header: {
        tintColor: 'rgba(255,255,255,0.8)',
        titleStyle: {
          color: 'rgba(255,255,255,0.8)',
          fontWeight: '400',
        },
        style: {
          backgroundColor: 'rgba(69,37,200,0.8)',
          shadowColor: 'rgba(69,37,200,0.8)',
          shadowOpacity: 1,
          shadowRadius: 3,
          shadowOffset: {
            height: 3,
            width: 0,
          },
        },
      },
    },
    headerMode: 'float',
  }
)

const Router = stackNavigator(
  {
    Main: { screen: MainNavigator },
    Login: { screen: Login },
  },
  {
    headerMode: 'none',
    mode: 'modal',
    cardStyle: {
      opacity: 1,
    },
    navigationOptions: {
      cardStack: {
        gesturesEnabled: true,
      },
    },

  }
)


export default Router
```

tsjason:
@RichardLindhout  I believe you meant `gesturesEnabled: false` :)

RichardLindhout:
@tsjason Good catch, I disabled it later that day ;)

aqnaruto:
does it is work on react-native 0.44,android? why i test it ,it is not work? @RichardLindhout

PR #397, opened 2017-02-20
Make compatible with webpack

rjkmurray:

**Test plan (required)**

**Code formatting**


ericvicenti:
The intention here looks good, but you haven't filled in any details in the PR description. Why is this change needed, what does it fix? How have you tested this?

satya164:
This is my commit from long back which is already in master.

Issue #402, opened 2017-02-20
Screen tracking with TabNavigator?


markflorisson:
There is a great example here for how to do screen tracking:

    https://reactnavigation.org/docs/guides/screen-tracking

But this doesn't seem to work with TabNavigator: `null is not an object (evaluating 'this.state.nav')`

ericvicenti:
Its not clear from the information provided what the issues is

Is the problem broken documentation, or is there a bug in the code? What is the expected behavior in comparison to what is happening?

Also, you can check the history of that document, and ping the authors: https://github.com/react-community/react-navigation/commits/master/docs/guides/Screen-Tracking.md

grabbou:
@markflorisson it is because `TabNavigator` doesn't have `nav` in its state. It's simply a different `Navigator`.

You can try that code:
```js
const screenName = (navigation) => {
  return navigation.state.routes[navigation.state.index];
};

SimpleTabs.prototype.componentDidUpdate = function(prevProps, curProps) {
  const currScreen = screenName(this.props.navigation);
  const prevScreen = screenName(prevProps.navigation);
  if (!!currScreen && currScreen != prevScreen) {
    // the line below uses the Google Analytics tracker
    // change the tracker here to use other Mobile analytics SDK.
    console.log('changed', currScreen);
  }
}
```

grabbou:
@satya164 @ericvicenti any ideas why we've decided to use `this.state.nav` instead of `this.props.navigation.state` in the docs? The above seems to work universally for both components https://github.com/react-community/react-navigation/pull/107/files#diff-1a19ce8a5a58e51bde21c6798edcbbd3R18

satya164:
Sent a PR removing this part. Sorry I didn't notice this in the PR, but overriding prototype is a hack we shouldn't document. https://github.com/react-community/react-navigation/pull/449/files

satya164:
I think a the correct way is to have `onNavigationStateChange` prop on the navigator.

grabbou:
@satya164 I think opening a new issue makes sense.

Issue #347, opened 2017-02-15
Unable to observer TabNavigator route change 


thomasbruketta:
In `react-native-tab-view` there was the ability to know when a tab was going to change via `onRequestChangeTab` is there any way to know that a tab has changed? Or are there any hooks / events exposed from the API? Thanks.

ElliottJRo:
hey @thomasbruketta you want to do something similar to [this](https://github.com/react-community/react-navigation/issues/348#issuecomment-280164995) which is an example implementation of [custom routers](https://reactnavigation.org/docs/routers/api).

Basically you want to add your hook to the `TabNavigator` router `getStateForAction` method.. At least I think that is how you would do it

grabbou:
Duplicate of #51

Issue #198, opened 2017-02-06
TabView: have way of knowing which tab is focused


EmielM:
To "mark messages as read" I need to know if my tab view is currently being focused.

It looks like `react-native-tab-view` actually passes a `focused` prop to `props.renderScene({focused})`, but `react-navigation` does not pass this to the scene (neither in `TabView._renderScene` nor `SceneView.render`), presumably because it's not part of the generic scene API.

What would the approach for `TabNavigator` route components be to know if they are currently focused? Is there an alternative approach, perhaps subscribe directly to some events on the `navigation` prop and determine the current navigation state?

EmielM:
I would suggest a `visible` prop could fit the generic API, because the same would apply for `StackNavigator` scenes that are in the history and currently not visible.

Adding such a property would break caching pure scenes, as they will have to be re-rendered when they become invisible. Quite stupid.

Any ideas, @satya164 ?

satya164:
@EmielM We can add lifecycle hooks to the screens, and provide a HOC to pass the `focused` prop to the underlying component. Check https://github.com/react-community/react-navigation/issues/51#issuecomment-275866099

PR welcome if you want to implement it. Let's continue the discussion in the other issue.

PR #405, opened 2017-02-20
[WIP] Add an onChange callback prop to the navigationOptions.tabBar

Vlad-Zhukov:
Currently there is no proper way to know when a user moved to another tab. Existing methods to track such changes involve custom routers or Redux integration, both have their own downsides and are an overkill for such a simple task. 

There is also an undocumented *hack* to add `onTabPress` callback to `tabBarOptions` as follows:
```javascript
const MyTabs = TabNavigator({
    TabOne: {screen: TabOne},
    TabTwo: {screen: TabTwo},
    TabThree: {screen: TabThree},
}, {
    initialRouteName: 'TabOne',
    tabBarOptions: {
        onTabPress: route => console.log(route),
    },
});
```
It will be passed down to the [react-native-tab-view/TabBar.js](https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabBar.js) and called on every tab press; not very practical either.

---

This PR is a proposal to add an `onChange` prop to the `navigationOptions.tabBar` that will be called on `onRequestChangeTab`, i.e. when one presses the tab header or swipes to a tab.

The change worked fine for me in a real app with the TabBarTop on Android and iOS. If you are OK with this change I will do more tests, fix flow issues and add related docs. Also I am not sure if `onChange` is a good name, but I couldn't find a better one for now.

Related issues: #347, #402, #198

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -117,6 +117,34 @@ contentOptions: {
 }
 ```
 
+### Screen Navigation Options
+
+Usually you define static `navigationOptions` on your screen component. For example:
+
+```jsx
+class ProfileScreen extends React.Component {
+
+  static navigationOptions = {
+
+    title: ({ state }) => `${state.params.name}'s Profile!`,
+
+    drawer: {
+      icon: (
+        <Image src={require('./my-icon.png')} />
+      ),
+    },
+  };
+  ...
+```
+
+All `navigationOptions` for the `DrawerNavigator`:
+
+- `title` - a title (string) of the scene
+- `drawer` - a config object for the drawer:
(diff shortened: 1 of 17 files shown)


ericvicenti:
It should be possible to notice the change of state/ the index changing of the active tab.

Vlad-Zhukov:
I was thinking about renaming `onChange` to `onToggle`, or something else if someone has a better name for it.

satya164:
We'll unlikely accept such a prop. It solves a very specific purpose.

GalFriD:
it works only when pressed. what about swipe?

Vlad-Zhukov:
**@GalFriD**: works for me on swipes

ericvicenti:
There are a lot of changes in the PR, so I can't see what is actually happening. Can you simplify this PR to have the minimal change, and update the title to reflect it?

PR #410, opened 2017-02-21
Update 2017-01-Introducing-React-Navigation.md

flyingace:
I think it's important to include dates in blog posts so the reader can have a better understanding of how recent the information contained therein is.

Since all I've done is add a date I haven't run any tests 😛 

I wish there was more I could add to reduce the spacing between the `<h1>` and the date but I couldn't locate the css files in the repo.

### DIFF
--- a/docs/blog/2017-01-Introducing-React-Navigation.md
+++ b/docs/blog/2017-01-Introducing-React-Navigation.md
@@ -1,4 +1,5 @@
 # Introducing React Navigation for React Native
+_January 26, 2017_
 
 Today we're excited to introduce React Navigation, a flexible navigation library for React Native and web, including customizable views for React Native, routers for any platform, and navigators that make it super easy to get started. We aim to provide a simple and extensible solution which enables developers to share one navigation paradigm for all of their React apps.
 


satya164:
How does this look on the website? Can you post a screenshot?

flyingace:
@satya164 I have never in my life had to work so hard to make such a simple change. Here is a screenshot: 
screen shot 2017-02-21 at 2 46 44 pm

satya164:
> I have never in my life had to work so hard to make such a simple change

You have an easy life then.

We require a test plan for every PR. Which means a screenshot or assurance that you tested it. Without a screenshot it's not possible for me to review this PR, because I don't know if it breaks the layout of the website. This is not a simple typo fix, since this changes a visual element on the screen, it needs a screenshot.

flyingace:
I am loathe to get into an argument with someone who has contributed so much to projects I respect but I am not keen on being chastised either or condescended to. The PR involved the addition of an italicized date to a markup file. There was nothing to test.

> Without a screenshot it's not possible for me to review this PR, because I don't know if it breaks the layout of the website. 

The highlighted line in the screenshot below represents the PR in its entirety. Can you not tell whether or not it will break the layout of the website from that? 

screen shot 2017-02-21 at 8 34 44 pm

skevy:
@flyingace I think it's just because the website is fully custom, and thus it's very possible that when @ericvicenti built the first version of this, he didn't test with dates. It's not that it would have broken the website horribly, in any case...but just better to be safe than sorry.

Let's just move on.

Thanks for your contribution!

PR #412, opened 2017-02-21
[RFC] Remove confusing header.title API

grabbou:
Currently, `header.title` can be either `string` or a `React.Element` to render in the middle of the header.

I believe that letting users specify `header.title`as a string introduces unnecessary confusion (already asked about differences [here](https://exponentjs.slack.com/archives/general/p1487650298022645)) as it's not clear how it's different from `title` property. Also, we don't use `header.title` strings anywhere in the docs.

The difference is conceptual, as `header.title` is a custom React.Element to be rendered in the middle of the header. The problem is that we implicitly wrap `strings` so that user can pass exactly same values he passes for `title`. W/o that implicit behavior, the difference would be: "Use `title` to render title of a scene, pass custom `header.title` component to e.g. achieve some more advanced interactions etc"

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -119,8 +119,8 @@ All `navigationOptions` for the `StackNavigator`:
 - `title` - a title (string) displayed in the header
 - `header` - a config object for the header bar:
   - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
-  - `title` - Title string used by the navigation bar, or a custom React component
   - `backTitle` - Title string used by the back button or `null` to disable label. Defaults to `title` value by default
+  - `title` - Custom React Element to display in the middle of the header
   - `right` - Custom React Element to display on the right side of the header
   - `left` - Custom React Element to display on the left side of the header
   - `style` - Style object for the navigation bar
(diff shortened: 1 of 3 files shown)


grabbou:
Note: this is likely a breaking change, how to update:

1) move `header.title` to `title`
2) wrap `header.title` with a `<Text>Your old title</Text>`

satya164:
The confusion is because docs make it pretty unclear. And don't say what is the difference.

The docs say `title` to display in `header`. But it's not `title` to display in header. It's meant to be the document title, not header title. Like the what's in `title` tag on a web page, the title used by native navigation components etc.

grabbou:
Yeah, exactly. That would be one thing to improve. However, I still don't see a point of implicitly allowing `header.title` to be a string. Aligning with `left`/`right` makes a lot of sense, otherwise, why not allow `left` to be a string as well? (instead of `backTitle`)

satya164:
`header.title` is the standard way to set header title like `tabBar.label` and `drawer.label`. All accept both strings and react elements (perhaps drawer doesn't yet)

grabbou:
With that in mind, it makes much more sense.

grabbou:
> Using a string for left doesn't make sense, because you're are supposed to replace the content of left

Not quite, you could use it to define a `backButton` title, the logic would be the same, if string, use default renderer and treat it as a back button title. Just styling for a text. Otherwise, render.

Anyways, the way I see it now is that `title` is somewhat shared screen title, whereas `header.title` would be exclusive to a particular view. Default renderer for title looks in both places.

satya164:
I don't think extra styling is the same as a full fledged component. I don't mind `backTitle` to accept react elements. but when `left` is just a string I think doesn't make much sense to render a back button along with the string + extra styling.

grabbou:
Yeah, I agree. It was just an example to prove a point that turned out to be no longer valid after I have figured out the real connection between document title and other parts.

Issue #295, opened 2017-02-11
Navigate and delete previous route... possible?


aroth:
Screen structure:
```
Home
  --> Login         (back button goes to Home)
  --> Sign Up       (back button goes to Home)
     --> Login      (back button goes to Sign Up, but I'd like it to go Home)
```

- From the 'Home' screen a user can navigate to 'Sign Up' or 'Login'
- From 'Sign Up' a user can also navigate to 'Login'
  - The back button will return the user to the 'Sign Up' screen, but I'd like to return to 'Home' instead (essentially removing the 'Sign Up' route from the stack).

This code works, but without animation as it's a reset: 
```
//
// SignUpScreen
//
  navigateToLogin() {
    const resetAction = NavigationActions.reset({
      index: 1,
      actions: [
        NavigationActions.navigate({ routeName: 'Home' }),
        NavigationActions.navigate({ routeName: 'Login' })
      ]
    });

    this.props.navigation.dispatch(resetAction);
}
```

What may be the ideal way to achieve this?

ericvicenti:
To hack it with the current API, you could override the router getStateForAction and do some extra logic to remove it from the stack.

But ideally you would be able to configure login and sign up as routes that can replace each-other. Then, when navigating between them, you would actually replace but the card stack would make it look like a push. There is no existing mechanism for this, but if you want to post a more detailed proposal of how routes could be configured to replace eachother, I'd be in support of it.

satya164:
I think we could maybe document this particular example. It's a very common case.

ericvicenti:
Agreed, maybe somebody could add it to [this area of the routers doc](https://reactnavigation.org/docs/routers/#Custom-Navigation-Actions).

MichalZalecki:
Is this acceptable solution? It works perfectly fine for me although `newIndex = newRoutes.length - 1;` somehow feels wrong.

```js
const prevGetStateForAction = MyApp.router.getStateForAction;

MyApp.router.getStateForAction = (action, state) => {
  // Do not allow to go back from Login
  if (action.type === "Navigation/BACK" && state && state.routes[state.index].routeName === "Login") {
    return null;
  }
  // Do not allow to go back to Login
  if (action.type === "Navigation/BACK" && state) {
    const newRoutes = state.routes.filter(r => r.routeName !== "Login");
    const newIndex = newRoutes.length - 1;
    return prevGetStateForAction(action, { index: newIndex, routes: newRoutes });
  }
[...]

PR #418, opened 2017-02-21
Fix glitches in Header when calling `replace` 

grabbou:
When having a StackNavigator with a route `A` and active index 0, calling `reset` action with a route `B` (leaving the same index) produces a series of glitches that this PR fixes.

Real use-case is when e.g. having a `Login` route and after clicking `Login` button, you want to replace it.

In order to fix those issues, I decided to change the following:

1) Transitioner shouldn't perform any animations when we do `reset` on the same index. Animations that rely on `index` will produce inaccurate effects as the index never changes, but the route does. What I experienced was e.g. after replacing a route, the title of a previous route stayed on screen for the duration of the animation, because `progress` was animating, but `position` wasn't. I cannot think of a use-case where it would be useful, but I might be wrong. 

2) We don't store computed layout for a title component under index, but unique scene key instead. That ensures even if we change the scene at the same index, it will work flawlessly and produce no races.

Before fix: https://drive.google.com/file/d/0B8qD3OEkGnJ9VDEtQ3JLMWU3RGc/view?usp=sharing
After fix: https://drive.google.com/file/d/0B8qD3OEkGnJ9aUFqLWhEWUhPQ2s/view?usp=sharing

Likely related: #295 
Likely related NavigationExperimental issue: https://github.com/facebook/react-native/issues/11114

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -52,7 +52,7 @@ type SubViewName = 'left' | 'title' | 'right';
 
 type HeaderState = {
   widths: {
-    [key: number]: number,
+    [key: string]: number,
   },
 };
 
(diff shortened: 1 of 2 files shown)


grabbou:
Related #295

grabbou:
NavigationExperimental issue: https://github.com/facebook/react-native/issues/11114

grabbou:
CC: @ericvicenti are you okay with landing these changes?

ericvicenti:
Great, I think this makes sense

Issue #421, opened 2017-02-21
iOS Header.js style throws warning


wschurman:
Warning is: `Invalid style with id '106'. Skipping ...`

Blame: 79f21277cbec2a684edc6f2073ff36530d093f41 (see Header.js null style ternary condition).

I'll work on a fix now, probably just going to change the `null` to an empty object.

PR #422, opened 2017-02-21
Fix null style warning in Header component

wschurman:
Fixes invalid style warning: #421

**Test plan**
Before Change: Open NavigationPlayground on iOS, open stack example, see warning.
After Change: No warning. Also verify no change on Android for stack example.


grabbou:
Thanks! That was annoying log!

PR #431, opened 2017-02-21
Updated the reducer

godlevel:
When the drawer opens / close, the result of AppNavigator.router.getStateForAction(action, state); is null. So the nav state in the store is null which causes errors. This solves that.


nihgwu:
Shouldn't we change the reducers to return the default state instead of hacking the docs?

godlevel:
@nihgwu we should, but I thought this is a quick fix and might help people.

PR #437, opened 2017-02-22
Fix left button in header on Android

### LINKED ISSUE #432: Android back button icon not  displayed properly.
I am using `react-nativation` master in my app 
screen shot 2017-02-22 at 4 31 14 am

grabbou:
Fixes #432 

For some reason, `onLayoutIOS` was missing Platform check.

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -220,7 +220,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
 
     // On iOS, width of left/right components depends on the calculated
     // size of the title.
-    const onLayoutIOS = name === 'title'
+    const onLayoutIOS = Platform.OS === 'ios' && name === 'title'
       ? (e: LayoutEvent) => {
         this.setState({
           widths: {


ayalpani:
Thanks, I npm updated but this didn't fix the issue for me.

ayalpani:
But maybe my case is different. I am adding back button through navigationOptions:

```
  static navigationOptions = {
    title: 'Details',

    header: ({ goBack }) => {
      return {
        visible: true,
        left: <BackButton onPress={ () => goBack() } />
      }
    },
  }
```

grabbou:
Not sure what is your case as I don't see your comment in the original issue. What exactly is broken for you?

ayalpani:
hi @grabbou 

In my case, it also looks like this, the arrow is covered in parts:

grabbou:
Make sure you are running latest master (what is your react-navigation dependency in package.json)? Remove node_modules and reinstall. You likely have a stale version.

Also, I don't understand this part: `left: <BackButton onPress={ () => goBack() } />`. Where is that component coming from? After removing it, does the problem occur?

ayalpani:
I had checked the dependency before. This is from my package.json:

`"react-navigation": "git+https://github.com/react-community/react-navigation/",`

Anyway, I deleted node_modules but nothing changed.

This is where I got the BackButton:

```
import { StackNavigator, CardStack }     from 'react-navigation'
const { BackButton } = CardStack.Header
```

I am trying to manually add it to the header, so if I remove the component, there will be no back button at all ...

grabbou:
Makes sense.

Anyways, I've added:

```js
  title: 'Welcome',
  header: {
    left: <CardStack.Header.BackButton />,
  },
};
```

to the SimpleStack.js file in examples/NavigationPlayground, just like you
did and still, I don't see the glitch.

Be sure to restart the packager as well.

ayalpani:
Yes, I tried what you tried too and it works perfectly. I will further investigate and let you know when I have found the issue. Thanks for your time

PR #449, opened 2017-02-23
Remove screen tracking without redux

satya164:
This is a hack and we shouldn't put it in official documentation.


grabbou:
Still might be worth documenting somewhere?

satya164:
@grabbou there's no way to do it unless we implement `onNavigationStateChange` prop for navigators afaik.

PR #450, opened 2017-02-23
Fix redux documentation

javiercr:
I think the navigation key for the redux store should be `nav` and not `navReducer`.

### DIFF
--- a/docs/guides/Redux-Integration.md
+++ b/docs/guides/Redux-Integration.md
@@ -13,7 +13,7 @@ const navReducer = (state, action) => {
 };
 
 const appReducer = combineReducers({
-  navReducer,
+  nav: navReducer,
   ...
 });
 


RobertSheaO:
Is there a reason to have a specific reducer named `nav`?

satya164:
No reason. You can name it whatever you want.

RobertSheaO:
Cool, I was confused if it was a requirement, the documentation seems to imply so.

javiercr:
@RobertSheaO just keep in mind that the name of the key you use for the reducer needs to be the same than the one you use here (this.props.**nav**):

```es6
      <AppNavigator navigation={addNavigationHelpers({
        dispatch: this.props.dispatch,
        state: this.props.nav,
      })} />
```

PR #453, opened 2017-02-23
Add onNavigationStateChange

grabbou:
Sometimes it is useful to know when `navigation` did change, so that we can use e.g. Google Analytics to track page change.

This improves over past example that was hacking `componentDidUpdate` prototype.

Test plan:
add `onNavigationStateChange` prop and console log args. Should be called every dispatch with prev/next state.


ericvicenti:
I think this feature will come in handy. Great work!

PR #459, opened 2017-02-23
[Docs] Link every heading automatically

### LINKED ISSUE #434: Ability to link to specific sections and config options in docs
It'll be easier to link to documentation when someone asks a question.

grabbou:
[docs] Fixes #434 

How it looks:
screen shot 2017-02-23 at 16 36 07

With `:hover`
screen shot 2017-02-23 at 16 36 11

URL:
screen shot 2017-02-23 at 16 36 15

Tested on Safari.

### DIFF
--- a/website/src/MDPage.js
+++ b/website/src/MDPage.js
@@ -62,7 +62,7 @@ const MDPage = ({navigation, docPath}) => (
         const Header = getHeadingForLevel(level);
         return (
           <Header id={id} className="md-header">
-            <a href={`#${id}`} title={children}>{children}</a>
+            {children} <a href={`#${id}`} title={children}>#</a>
           </Header>
         );
       },
--- a/website/src/MDPage.js
+++ b/website/src/MDPage.js
@@ -9,6 +9,26 @@ import CodeBlock from './CodeBlock';
 
 const safeString = s => slugify(s).replace(/\)/g, '-').replace(/\(/g, '-').replace(/^-/,'').replace(/-$/,'');
 
+const getHeadingForLevel = (level) => {
+  switch (level) {
+    case 2:
+      return 'h2';
+    case 3:
+      return 'h3';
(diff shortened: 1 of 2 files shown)


grabbou:
Note: this still needs some loving. Navigating to an URL with hash covers the header of section, because it's rendered below the heading. We might need to hack CSS to solve it.

satya164:
I think showing a "#" before the heading is nicer than underline. Underline feels like it's link that'll take me somewhere.

grabbou:
Not really sure. Note how Github does that when you click the time of your comment. It's the same. Adding `#` requires additional work that I believe can be carried in next PR.

satya164:
React Native docs show a pound sign https://facebook.github.io/react-native/docs/getting-started.html#content

grabbou:
screen shot 2017-02-23 at 16 53 30
screen shot 2017-02-23 at 16 53 27

Okay, fixed :) Used a different approach to avoid un-semantic use of `a` and absolute positions.

website/src/MDPage.js
```diff
@@ -39,23 +59,12 @@ const MDPage = ({navigation, docPath}) => (
...
-            return <h5 id={id}>{children}</h5>;
-          case 6:
-            return <h6 id={id}>{children}</h6>;
-          case 7:
-            return <h7 id={id}>{children}</h7>;
-          default:
-          case 1:
-            return <h1 id={id}>{children}</h1>;
-        }
+        const Header = getHeadingForLevel(level);
```
ericvicenti: Interesting, I'm not used to this pattern with a string variable as a component type, but it looks good if it works!

grabbou: To be honest, I was also surprised that it did work. But I guess it's because of the way `jsx` is transpiled to `React.createElement` and the fact it accepts strings 🗡

PR #463, opened 2017-02-23
Drawer custom item

Fenwil:
**Motivation**

We want to render a custom item on DrawerNavigator but without the constraints that `renderIcon` has so we open this issue: https://github.com/react-community/react-navigation/issues/416
So we'd added an option called `item` under the navigationOptions on the DrawerNavigator to render a custom component.

**Test plan (required)**
We added an example under `examples/NavigationPlayground`.
[Run the native playground](https://reactnavigation.org/docs/guides/contributors) and go to the Custom Drawer item

screen shot 2017-02-23 at 4 19 25 pm

--------------------

screen shot 2017-02-23 at 4 19 34 pm

--------------------

**Inbox item focused:**

screen shot 2017-02-23 at 4 22 21 pm

**Drafts items focused:**

screen shot 2017-02-23 at 4 22 34 pm

**Code formatting**
We use the same code formatting that the project has. It's different only on the example on `examples/NavigationPlayground/js/CustomDrawer.js`

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -15,6 +15,7 @@ import { StackNavigator } from 'react-navigation';
 import Banner from './Banner';
 import CustomTabs from './CustomTabs';
 import Drawer from './Drawer';
+import CustomDrawer from './CustomDrawer';
 import ModalStack from './ModalStack';
 import StacksInTabs from './StacksInTabs';
 import SimpleStack from './SimpleStack';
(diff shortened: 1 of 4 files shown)


grabbou:
Have you tried passing a custom [`contentComponent`](https://github.com/underscopeio/react-navigation/blob/2e40da887cd7524cbf115d6f3ed67c5f747848c6/src/navigators/DrawerNavigator.js#L33)? I believe you can achieve exactly these results.

Default `contentComponent` is [DrawerNavigatorItems](https://github.com/underscopeio/react-navigation/blob/2e40da887cd7524cbf115d6f3ed67c5f747848c6/src/views/Drawer/DrawerNavigatorItems.js#L50). You can just copy it, change what you need and specify like above.

Fenwil:
Great! Thanks @grabbou we will try that

grabbou:
Awesome! In case it works, feel free to close the PR! However, please paste here your final component so maybe others in the future that google the solution can find it!

If it gets popular enough, we might consider documenting it

satya164:
Closing since you can achieve it with `contentComponent`

PR #470, opened 2017-02-24
Bring back better `header.bar`

### LINKED ISSUE #53: Support for hiding navigation header when `headerMode: float`
Currently it's supported only with `headerMode: screen`. We should support it on `float` also and hide the header with animation.

### LINKED ISSUE #100: add ability to specify headerComponent per screen
currently we can do this for navigator, but very useful would be if we can do this for specific screens.

example:
```js
StackNavigator({
  profile: { screen: MyScreen, headerComponent: MyCustomHeader },
})
```

### LINKED ISSUE #570: Use a custom component to render the Header
I know it's possible to use a custom React Element for the left, right and title properties.
But is it possible to fully render the Header with a custom element?

I want my header to be a search bar with my logo:

screen shot 2017-03-04 at 12 50 12

grabbou:
Fixes #570
Fixes #100 
Fixes #53

Brings back `header.bar` instead of `headerComponent`. Removes `header.visible`. Use `null` value to hide it for a specific screen.

Works for `headerMode: float` (with a small glitch we can tackle in next PR).

Tested the following:
```js
const SimpleStack = StackNavigator({}, {
  navigationOptions: {
    header: {
      bar: CardStack.Header,
    },
  },
});
```
no-issues, that's the default effect.

Another case:
```js
navigationOptions: {
    header: {
      bar: null,
    },
  },
```
on per screen basis.

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -118,7 +118,7 @@ All `navigationOptions` for the `StackNavigator`:
 
 - `title` - a title (string) of the scene
 - `header` - a config object for the header bar:
-  - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
+  - `bar` - React Element to display as a header or `null` to hide it. Only works when `headerMode` is `screen`
   - `title` - String or React Element used by the header. Defaults to scene `title`
   - `backTitle` - Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
   - `right` - React Element to display on the right side of the header
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -118,7 +118,7 @@ All `navigationOptions` for the `StackNavigator`:
 
 - `title` - a title (string) of the scene
 - `header` - a config object for the header bar:
-  - `bar` - React Element to display as a header or `null` to hide it. Only works when `headerMode` is `screen`
+  - `bar` - React Element or a function that given `HeaderProps` returns a React Element, to display as a header. Only works when `headerMode` is `screen`. Setting to `null` hides header.
   - `title` - String or React Element used by the header. Defaults to scene `title`
   - `backTitle` - Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
   - `right` - React Element to display on the right side of the header
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -118,7 +118,7 @@ All `navigationOptions` for the `StackNavigator`:
 
 - `title` - a title (string) of the scene
 - `header` - a config object for the header bar:
-  - `bar` - React Element or a function that given `HeaderProps` returns a React Element, to display as a header. Only works when `headerMode` is `screen`. Setting to `null` hides header.
+  - `bar` - React Element or a function that given `HeaderProps` returns a React Element, to display as a header. Setting to `null` hides header.
   - `title` - String or React Element used by the header. Defaults to scene `title`
   - `backTitle` - Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
   - `right` - React Element to display on the right side of the header
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -133,10 +133,7 @@ export type HeaderConfig = {
   /**
    * React element to display as a header
    */
-  bar?: React.Element<*>
-    | (
-      transitionProps: NavigationTransitionProps & { router: NavigationRouter }
-    ) => React.Element<*>,
+  bar?: React.Element<*> | HeaderProps => ReactElement<*>
 
   // // Style of title text
   // titleTextStyle?: $NavigationThunk<Object>,
@@ -281,7 +278,6 @@ export type NavigationContainerConfig = {
 export type NavigationStackViewConfig = {
   mode?: 'card' | 'modal',
   headerMode?: HeaderMode,
-  headerComponent?: ReactClass<HeaderProps<*>>,
   cardStyle?: Style,
   onTransitionStart?: () => void,
   onTransitionEnd?: () => void
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -133,7 +133,10 @@ export type HeaderConfig = {
   /**
    * React element to display as a header
    */
-  bar?: React.Element<*>,
+  bar?: React.Element<*>
+    | (
+      transitionProps: NavigationTransitionProps & { router: NavigationRouter }
+    ) => React.Element<*>,
 
   // // Style of title text
   // titleTextStyle?: $NavigationThunk<Object>,
@@ -177,7 +180,7 @@ export type CardStackConfig = {
    * Whether you can use gestures to dismiss this screen.
    * Defaults to true on iOS, false on Android.
    */
-  gesturesEnabled?: bool;
+  gesturesEnabled?: boolean;
 };
 
 export type NavigationScreenOptions = {
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -130,6 +130,11 @@ export type HeaderConfig = {
    */
   titleStyle?: Style,
 
+  /**
+   * React element to display as a header
+   */
+  bar?: React.Element<*>,
+
   // // Style of title text
   // titleTextStyle?: $NavigationThunk<Object>,
   // // Tint color of navigation bar contents
--- a/src/navigators/StackNavigator.js
+++ b/src/navigators/StackNavigator.js
@@ -24,7 +24,6 @@ export default (routeConfigMap: NavigationRouteConfigMap, stackConfig: StackNavi
     initialRouteName,
     initialRouteParams,
     paths,
-    headerComponent,
     headerMode,
     mode,
     cardStyle,
@@ -42,7 +41,6 @@ export default (routeConfigMap: NavigationRouteConfigMap, stackConfig: StackNavi
   return createNavigationContainer(createNavigator(router)(props => (
     <CardStack
       {...props}
-      headerComponent={headerComponent}
       headerMode={headerMode}
       mode={mode}
       cardStyle={cardStyle}
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -45,7 +45,6 @@ const NativeAnimatedModule = NativeModules && NativeModules.NativeAnimatedModule
 type Props = {
   screenProps?: {};
   headerMode: HeaderMode,
-  headerComponent?: ReactClass<*>,
   mode: 'card' | 'modal',
   navigation: NavigationScreenProp<*, NavigationAction>,
   router: NavigationRouter,
@@ -62,7 +61,6 @@ type Props = {
 
 type DefaultProps = {
   mode: 'card' | 'modal',
-  headerComponent: ReactClass<*>,
 };
 
 class CardStack extends Component<DefaultProps, Props, void> {
@@ -90,11 +88,6 @@ class CardStack extends Component<DefaultProps, Props, void> {
      * is `screen` on Android and `float` on iOS.
      */
     headerMode: PropTypes.oneOf(['float', 'screen', 'none']),
-
-    /**
-     * Custom React component to be used as a header
-     */
-    headerComponent: PropTypes.func,
 
     /**
      * Style of the cards movement. Value could be `card` or `modal`.
@@ -141,7 +134,6 @@ class CardStack extends Component<DefaultProps, Props, void> {
 
   static defaultProps: DefaultProps = {
     mode: 'card',
-    headerComponent: Header,
   };
 
   componentWillMount() {
@@ -211,7 +203,7 @@ class CardStack extends Component<DefaultProps, Props, void> {
     }
 
     return (
-      <this.props.headerComponent
+      <Header
         {...transitionProps}
         router={this.props.router}
         style={headerConfig.style}
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -202,7 +202,12 @@ class CardStack extends Component<DefaultProps, Props, void> {
         headerMode === 'screen',
         'header.bar is only supported with headerMode: screen',
       );
-      return headerConfig.bar;
+      return typeof headerConfig.bar === 'function'
+        ? headerConfig.bar({
+          ...transitionProps,
+          router: this.props.router,
+        })
+        : headerConfig.bar;
     }
 
     return (
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -187,7 +187,10 @@ class CardStack extends Component<DefaultProps, Props, void> {
       'header'
     ) || {};
 
-    if (headerConfig.bar === null || typeof headerConfig.bar !== 'function') {
+    if (
+      headerConfig.bar === null
+      || (headerConfig.bar && typeof headerConfig.bar !== 'function')
+    ) {
       return headerConfig.bar;
     }
 
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -195,7 +195,7 @@ class CardStack extends Component<DefaultProps, Props, void> {
       return headerConfig.bar;
     }
 
-    const renderHeader = headerConfig.bar || ((props: HeaderProps) => <Header {...props} />);
+    const renderHeader = headerConfig.bar || ((props: *) => <Header {...props} />);
 
     return renderHeader({
       ...transitionProps,
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -189,26 +189,19 @@ class CardStack extends Component<DefaultProps, Props, void> {
       'header'
     ) || {};
 
-    if (typeof headerConfig.bar !== 'undefined') {
-      invariant(
-        headerMode === 'screen',
-        'header.bar is only supported with headerMode: screen',
-      );
-      return typeof headerConfig.bar === 'function'
-        ? headerConfig.bar({
-          ...transitionProps,
-          router: this.props.router,
-        })
-        : headerConfig.bar;
+    if (headerConfig.bar === null || typeof headerConfig.bar !== 'function') {
+      return headerConfig.bar;
     }
 
+    const HeaderComponent = headerConfig.bar || Header;
+
     return (
-      <Header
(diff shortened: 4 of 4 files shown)


grabbou:
Next PR: Figure out how to handle animations for mode: `float`

docs/api/navigators/StackNavigator.md
```diff
@@ -118,7 +118,7 @@ All `navigationOptions` for the `StackNavigator`:
 
 - `title` - a title (string) of the scene
 - `header` - a config object for the header bar:
-  - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
+  - `bar` - React Element to display as a header or `null` to hide it. Only works when `headerMode` is `screen`
```
satya164: Not sure how this is gonna work. headers usually need more info than just the current screen state. so this feels very limiting. Maybe this should be just a function which receives the props header usually receives, like tab icons.

src/views/CardStack.js
```diff
@@ -189,11 +190,19 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
+    const headerConfig: HeaderConfig = this.props.router.getScreenConfig(
       transitionProps.navigation,
       'header'
     ) || {};
 
+    if (typeof headerConfig.bar !== 'undefined') {
+      if (headerMode === 'screen') {
+        return headerConfig.bar;
+      } else {
+        console.warn('header.bar is only supported with headerMode: screen. Ignoring');
```
satya164: Probably should just throw. there's no reason to support invalid config

src/views/CardStack.js
```diff
@@ -283,22 +292,18 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
-    const header = this.props.router.getScreenConfig(props.navigation, 'header');
     const headerMode = this._getHeaderMode();
     if (headerMode === 'screen') {
-      const isHeaderHidden = header && header.visible === false;
-      const maybeHeader =
-        isHeaderHidden ? null : this._renderHeader(props, headerMode);
       return (
         <View style={styles.container}>
-          <View style={{flex: 1}}>
+          <View style={{ flex: 1 }}>
```
satya164: This should be in `StyleSheet.create`

src/views/CardStack.js
grabbou: Then I think `invariant` is enough?

docs/api/navigators/StackNavigator.md
grabbou: For that purpose, you have `headerComponent` that receives all the things. If it needs to change based on screen, I believe you can read that from `navigation` that gets passed. This is just to render something screen specific, just like `left` and `right`.

satya164: `header.bar` and `headerComponent` are redundant at global level. I think there's no reason to have both.

The `navigation` that gets passed is only for current screen. You don't even know if there's something before this screen in the stack to decide if you want to render a back button or not.

grabbou: I agree that it could also be a function (which is exactly what we do for `icon` in `TabBar` and others), what you think? What could be the props that cannot be read from the outer scope?

grabbou: Good use-case was demonstrated in #100. Anyways, I suggest we implement the required by the community features first (or at least make them consistent) and support `more info` when people request it.

satya164: The parent navigation state is most important. You can't have a full-featured header without knowledge of routes around the screen.

grabbou: > You can't have a full-featured header without knowledge of routes around the screen.

Why? If I just want to render: `<MyHeader title="Scene title" onBackPress={() => navigation.goBack()} />`

satya164: What happens when there's nothing to go back to? Do you render a non-functional button? Also how'd you render something like previous screen title in back button>

grabbou: This is esp. useful to e.g. render an empty, transparent header on a screen that displays a full-screen image with just `X` icon.

satya164: Yes, totally agree this is useful. It's pretty similar to `headerComponent` and I don't think we should have two configs doing very similar things.

> This is esp. useful to e.g. render an empty, transparent header on a screen that displays a full-screen image with just X icon.

That's just a single use case. What if you want to render previous screen title or something like that?

grabbou: > What happens when there's nothing to go back to?

You have an access to `navigation` so you can read the state and decide what's the index. It's not different from what we currently do. 

> Also how'd you render something like previous screen title in back button

Same as above? I think in that case, I think it could be either a function that receives `{...transitionProps, router}` (just like header) or an element for those who need a simple use-case.

grabbou: @satya164 problem with `headerComponent` is that it's more on a low-level side of things. One should be able to override header for a single scene without implementing entire header once again (there's a lot of stuff, like renderXComponent that has to be implemented).

I believe that most people will not even read from `router` or `navigation` as their use-cases will be rather simple. However I agree we can leave it to avoid limiting too much.

grabbou: Still, it's easy to implement later, we are in beta stage, so I don't see any reason why we shouldn't start with minimal API and see what happens.

grabbou: However, if we let people go with `header.bar`, there's no need to keep `headerComponent` I believe, isn't it?

grabbou:
Based on my discussions with @satya164 , I decided to deprecate `headerComponent` because:
1) what it provides can be achieved by a `header.bar` 
2) it can be per-screen as well, not only global to navigator
3) by default, we pass props to `headerComponent` that are specific to our internal implementation. That may confuse users. `header.bar` improves over that and it's similar to what we already do for icon etc.

docs/api/navigators/StackNavigator.md
satya164: > One should be able to override header for a single scene without implementing entire header once again

You're overriding entire header in either case. So it's not much different. It's just that you don't need to deal with animations when `mode` is float, so it's simpler.

> However, if we let people go with header.bar, there's no need to keep headerComponent I believe, isn't it?
     
Yes, that's what I'm saying

satya164:
> by default, we pass props to headerComponent that are specific to our internal implementation

`headerComponent` is a public API though. I think the props passed to it are reasonable. But I guess can be toned down if you think the props aren't correct.

src/TypeDefinition.js
```diff
@@ -273,7 +278,6 @@ export type NavigationContainerConfig = {
 export type NavigationStackViewConfig = {
   mode?: 'card' | 'modal',
   headerMode?: HeaderMode,
-  headerComponent?: ReactClass<HeaderProps<*>>,
   cardStyle?: Style,
```
satya164: We should also move this to `navigationOptions.cardStack.style`

grabbou: Can it be done in a separate PR? 🏑

satya164: yes, that's why I didn't requested a change

grabbou:
@satya164 is there anything else you want me to change here?

src/views/CardStack.js
```diff
@@ -189,18 +182,27 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
+      || (headerConfig.bar && typeof headerConfig.bar !== 'function')
+    ) {
+      return headerConfig.bar;
+    }
+
+    const HeaderComponent = headerConfig.bar || Header;
+
     return (
-      <this.props.headerComponent
+      <HeaderComponent
```
satya164: Can you call this as a function please? As I said previously, this will result in unmounting and re-mounting the header everytime something changes when config is specified as following:

```
header: () => ({
  bar: props => <MyComponent {...props} />,
})
```

src/views/CardStack.js
```diff
@@ -189,18 +182,27 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
         {...transitionProps}
         router={this.props.router}
         style={headerConfig.style}
         mode={headerMode}
-        onNavigateBack={() => this.props.navigation.goBack(null)}
+        onNavigateBack={() => { this.props.navigation.goBack(null); }}
```
satya164: The braces seem unnecessary

src/views/CardStack.js
```diff
@@ -189,18 +182,27 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
     headerMode: HeaderMode
   ): ?React.Element<*> {
-    const headerConfig = this.props.router.getScreenConfig(
+    const headerConfig: HeaderConfig = this.props.router.getScreenConfig(
       transitionProps.navigation,
       'header'
     ) || {};
 
+    if (
+      headerConfig.bar === null
```
satya164: I think this logic should be `typeof headerConfig.bar !== 'undefined' && typeof headerConfig.bar !== 'function'`

src/navigators/StackNavigator.js
```diff
@@ -42,7 +41,6 @@ export default (routeConfigMap: NavigationRouteConfigMap, stackConfig: StackNavi
   return createNavigationContainer(createNavigator(router)(props => (
     <CardStack
       {...props}
-      headerComponent={headerComponent}
```
satya164: Also since we're removing this API, we should warn/error when people use this.

src/views/CardStack.js
grabbou: Required, otherwise flow complains that `goBack` returns `boolean` instead of `void` expected. Turns out it started working after I annotated `HeaderConfig` few lines above.

src/navigators/StackNavigator.js
grabbou: Shall we? It's beta and breaking changes are expected. Not sure that was a broadly used feature enough to deserve a special treatment in the codebase. I believe people at this point read release notes anyway.

src/views/CardStack.js
grabbou: Now it makes sense!

src/views/CardStack.js
grabbou: I need to investigate it, maybe you got an idea in the meantime:
> [flow] [flow] call of method `goBack` (Method cannot be called on any member of intersection type intersection Member 1: type application of polymorphic type: type `NavigationProp` Error: property `goBack` Property not found in object type Member 2: object type Error: boolean This type is incompatible with the expected param type of)

grabbou: Required as `goBack` has a return type of `boolean` whereas `onNavigateBack` has `void`. Looks like my changes to this method enabled flow for it.

src/navigators/StackNavigator.js
grabbou: Otherwise I think we can use `deprecated` prop. Small change.

grabbou:
up @satya164

nonameolsson:
Great work, looking forward to this!

grabbou:
Flow fixed, I believe this is now ready to land :)

grabbou:
Flow is fixed now @satya164, updated so it's ready to go

grabbou:
Ping @satya164 for a review

satya164:
cc @ericvicenti

satya164:
BTW not very happy with the name `header.bar`, say we do the same for `tabBar`, would we call it `tabBar.bar`? I prefer `header.component`, but since it can also be an `element` it's not correct.

nihgwu:
what about just a `render`?

satya164:
Yeah, `render` is better, though it feels like a function

nonameolsson:
What about `renderComponent`?

satya164:
I think `renderComponent` is same as `render`, it sounds like a function, but it can be an element also

nihgwu:
so we only support a function
`() => null` means `visible: false`
`() => component` or `(params) => {}` for custom

satya164:
I prefer only function too. but we have many other options which support both functions and elements.

nihgwu:
or we can just make header receive a component or function besides config, I think it's more semantic since if we provide a `bar` then the other parts are useless

grabbou:
I am not sure what's wrong with `tabBar.bar`. So far seems like the best option for other alternatives. Also, `tabBar.bar` makes more sense to me than `header.bar` since we actually have a  tab *bar*.

satya164:
`tabBar` already means tab bar, so does `header`, the `bar` is redundant.

Anyway, I'll prefer if @ericvicenti reviews this since this changes the API.

grabbou:
> tabBar already means tab ba

Yes, it does, but since you are passing a component to be rendered as a `tabBar`, that makes sense. 

I don't think this changes the API as it effectively brings back what we used to have, but I agree if someone else gives it a review so we avoid changing it once again from `bar` to whatever in the future.

ecstasy2:
Any chance this will get merged soon?

intpp:
+

grabbou:
@ericvicenti just following up on this after our discussions

Kerumen:
Sad this was not part of beta 7 😞

valeryq:
I think it would be pretty good if by default be `Headers.js` from react-navigation, if I want to custom all my screen headers I define headerComponent in CardStack stackConfig as `headerComponent` and if I want to redefine headerComponent, I do `header.bar = ReactComponent` in Component `navigationOptions`.

kevin-wolf:
What @valeryq says should be a must in my opinion!

chauthai:
I really need the feature suggested by @valeryq.

valeryq:
@chauthai if somebody will review my PR, may be it will be pushed in react-navigation.
You can review https://github.com/react-community/react-navigation/pull/681

chauthai:
@valeryq Ok, I will try to do it later today.

src/views/CardStack.js
```diff
@@ -189,39 +183,48 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
-      />
-    );
+    if (
+      typeof headerConfig.bar !== 'undefined'
+      && typeof headerConfig.bar !== 'function'
+    ) {
+      return headerConfig.bar;
+    }
+
+    const renderHeader = headerConfig.bar || ((props: *) => <Header {...props} />);
```
ericvicenti: if headerConfig.bar is a real component, I'm not sure this will work properly. Below we should be using JSX, no?

docs/api/navigators/StackNavigator.md
```diff
@@ -118,7 +118,7 @@ All `navigationOptions` for the `StackNavigator`:
 
 - `title` - a title (string) of the scene
 - `header` - a config object for the header bar:
-  - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
+  - `bar` - React Element or a function that given `HeaderProps` returns a React Element, to display as a header. Setting to `null` hides header.
```
ericvicenti: Shouldn't this say 'Component' instead of 'Element'? Related to my confusion below

grabbou: Yeah, nice catch! Let me sort this out on monday, update & rebase (if neccessary) and let's ship. Will also test for all use-cases once again.

satya164: This is not a component though

grabbou:
I think the confusion is (honestly, I don't remember - need to
re-investigate once again next week) because docs mention `React Element`
whereas the code that Eric commented on seems to handle functions only.

satya164:
> whereas the code that Eric commented on seems to handle functions only

That code handles both react elements and functions

grabbou:
Yes. For anyone interested, here are the lines that handle such case:
https://github.com/react-community/react-navigation/blob/060e6e21d8faab3434f7cd7c20ace292ccce2630/src/views/CardStack.js#L191-L196

src/views/CardStack.js
grabbou: @ericvicenti we don't support components to align with the rest of the API (see left/right for reference). `React.Element` is accepted (by default), see lines 192-196. Function is accepted when `bar` has to be set based on props (generic implementation or any other use-case).

valeryq:
@grabbou it's not so good, because if user passed the string, for example. It should be checked if typeof is function.
Please, look at the: https://github.com/valeryq/react-navigation/blob/84cc1607133883e3c91350434dbc2d417fdddd8a/src/views/CardStack.js#L287

docs/api/navigators/StackNavigator.md
grabbou: Yes, `Component` is not accepted to align with the rest of API

satya164:
@valeryq I don't think that's a good way to go. Passing a string is an error, and user should get a hard error rather than getting the error ignored.

src/views/CardStack.js
```diff
@@ -189,39 +183,48 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
+      onNavigateBack: () => {
+        this.props.navigation.goBack(null);
+      },
+      renderLeftComponent: (props: NavigationTransitionProps) => {
+        const header = this.props.router.getScreenConfig(props.navigation, 'header') || {};
+        return header.left;
+      },
+      renderRightComponent: (props: NavigationTransitionProps) => {
+        const header = this.props.router.getScreenConfig(props.navigation, 'header') || {};
+        return header.left;
```
Kerumen: ⚠️  `return header.right;` ⚠️

aranda-adapptor:
Looking forward to this PR, but I noticed on testing that header right elements were not rendered anymore. This is how it is set up:

```
  static navigationOptions = {
    title: "Home",
    header: ({ navigate }) => ({
      right: (
        <View
          style={{
            flex: 1,
            flexDirection: "row",
            alignItems: "center"
          }}
        >
          <Icon
            style={{ paddingRight: 15 }}
            name="map-marker"
            size={24}
            onPress={() => navigate("Map")}
          />
          <Icon
            style={{ paddingRight: 15 }}
            name="search"
            size={24}
            onPress={() => navigate("Search")}
          />
        </View>
      )
    })
  };
```

ajonno:
hi folks, when will this get pushed into master? Also will this allow us to use a custom component as the nav bar, so we can style it / size it etc. as we like ?

In our case we need to :

- use a custom height on a screen by screen basis. (already looks do-able) 

- a background image (every screen). currently not possible?

- custom back button. is an arrow icon from our UI team. is this possible ? 

- put a search bar in the (custom height) nav bar on selected screens. 

That's it.

an example of one screen's nav bar looks like this:
image

grabbou:
I am going to close this pull request as it needs to be re-based. I'll submit a new one in a bit as a follow up.

PR #499, opened 2017-02-27
Solves TabBarBottom item layout to match iOS UITabBar HIG

alvaromb:
Instead of `flexGrow`, `TabBarBottom` should use `flex` in order to match iOS HIG for UITabBar. Otherwise, if labels with different length are set, the layout of the tab bar isn't consistent with what iOS users expect (tab highlighted with red border):

screen shot 2017-02-27 at 11 23 27

After the change, this is how the tab bar looks:

screen shot 2017-02-27 at 15 14 35

**Test plan (required)**

Tests pass and both Android and iOS `NavigationPlayground` example looks OK.


satya164:
Thank you for the PR

PR #500, opened 2017-02-27
Ability to view current stack within StackNavigator

### LINKED ISSUE #427: Current index or previous routes in StackNavigator not available
The StackNavigator should have access to the previous routes however when I console log out the navigation state there is no array of routes

stief510:
Added the ability to view the current stack within StackNavigator. Fixes #427

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -52,19 +52,13 @@ type Props = {
   style: Style,
   gestureResponseDistance?: ?number,
   /**
-   * If true, enable navigating back by swiping (see CardStackPanResponder).
-   * TODO move this to TransitionConfig.
-   */
-  gesturesEnabled: ?boolean,
-  /**
    * Optional custom animation when transitioning between screens.
    */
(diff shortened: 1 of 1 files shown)


stief510:
Well the `stack` property should only be updating when the route changes because it is whatever the current stack of routes is. So why would it be out of date? If the route changes, then it will get updated accordingly.

Also, I would be happy to add it to other navigators but I have only been working with the StackNavigator and, therefore, that is where most of my familiarity lies.

grabbou:
I believe we can do it with `onNavigationStateChange`. You can listen to scene changes and use e.g. redux or `setState` to store its value.

nonameolsson:
What's missing to get this merged? @grabbou @stief510

ericvicenti:
Satya's concerns are very valid above, specifically, "The stack property only updates when the current route changes, it means that most of the time it'll be out of date"

For now, people can use onNavigationStateChange as grabbou mentioned

PR #510, opened 2017-02-28
Query string parsing capabilities for deep linking

normanrz:
This PR adds query string parsing capabilities. The items of the query string will be added to the `params` object of a navigation action. When there are same-named params in the query and the path, the params from the path will override the ones from the query.

**Motivation**
In my app I needed this capability, because I have a deep link `/redeem_key?key=<key>&name=<name>`. As with query params in the web, the ordering shouldn't be significant.

**Test plan (required)**
I added a test case that verifies the functionality. There are no UI changes and it shouldn't affect any platform-specific code.

I'm happy to add more test cases or documentation, if you find this PR useful.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -245,7 +245,9 @@ export default (
         });
       }
 
-      // Attempt to match `pathToResolve` with a route in this router's
+      const [pathNameToResolve, queryString] = pathToResolve.split('?');
+
+      // Attempt to match `pathNameToResolve` with a route in this router's
       // routeConfigs
       let matchedRouteName;
       let pathMatch;
@@ -254,7 +256,7 @@ export default (
       for (const routeName in paths) {
         /* $FlowFixMe */
         const { re, keys } = paths[routeName];
-        pathMatch = re.exec(pathToResolve);
+        pathMatch = re.exec(pathNameToResolve);
         if (pathMatch && pathMatch.length) {
           pathMatchKeys = keys;
           matchedRouteName = routeName;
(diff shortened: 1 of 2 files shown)


normanrz:
I added a test with an empty query value as suggested. The current implementation treats empty values as empty string. Would this be your expected behavior?

matthamil:
@normanrz I think that's a good approach. Tagging @grabbou for input

serco-chen:
@grabbou Just wondering when will this PR get merged, it's definitely a must-have feature to implement OAuth flow.

ericvicenti:
Heh, sorry this PR was left unattended for so long!

Your implementation and test looks great. Thank you for this important contribution!

If possible I'd really like to see the docs fleshed out for this, maybe in the deep linking guide and the docs for the router `getActionForPathAndParams`.

Do you think it makes sense to add this capability for TabRouter as well? It might come in handy on websites, where TabRouter gets a bit more usage- like in the `website/src/App.js`

max-mykhailenko:
@skevy When you plan to release to npm this?

skevy:
@max-mykhailenko it's actually on NPM now as 1.0.0-beta.8 (published under the `next` tag)

koenpunt:
Unfortunately this only works for single level navigator stacks. When a `StackNavigator` is nested in another `StackNavigator` the params are not passed to the "nested action".

MovingGifts:
Any solution to the nested `StackNavigator` missing params issue?

PatNeedham:
@koenpunt how can the limitation you bring up be removed?

PR #524, opened 2017-03-01
Accept tabBar label to be a function

hose314:
It could be useful to dynamically change (`focused`) tabBar label default styles for certain tab(s).
This pull request is aim to discuss how to implement it properly. 

Usage example: 
```js
navigationOptions: {
  tabBar: {
    icon: ({ focused }) => <TabItem focused={focused} name="whatshot" activeColor="red" />,
    label: ({ focused, tintColor, defaultStyles }) => {
      return (<Text style={[defaultStyles, { color: focused ? 'red' : defaultStyles.color }]}>Profile</Text>);
    },
  }
},
```

Current implementation provided but without `Animated.Text` color specification

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -89,17 +89,8 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
         </Animated.Text>
       );
     }
-    
     if (typeof label === 'function') {
-      const st = [styles.label, { color: inactiveTintColor }, labelStyle];
-      const params = {
-        ...scene,
-        defaultStyles: StyleSheet.flatten(st)
-      };
-
-      return (
-        label(params)
-      )
+      return label(scene);
     }
 
     return label;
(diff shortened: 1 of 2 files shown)


src/views/TabView/TabBarBottom.js
```diff
@@ -89,6 +89,18 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
         </Animated.Text>
       );
     }
+    
+    if (typeof label === 'function') {
+      const st = [styles.label, { color: inactiveTintColor }, labelStyle];
+      const params = {
+        ...scene,
+        defaultStyles: StyleSheet.flatten(st)
```
satya164: Remove this. You can just return a string if you want to use default styles.

hose314: Sorry i don't get it, i want to make possible to leave all tabs with the same inactive styles but for certain tabs change it color when this tab is active. If i remove this i will lose information about `inactiveTintColor`

satya164: That sounds like very specific usecase. You already receive the `tintColor` as a prop, so you can write your own label component which animates between the colors when it changes. Also there's no reason to flatten the style here.

hose314: Agree. PR updated.

PR #525, opened 2017-03-01
Allow `TabBarBottom` accept `onTabPress` callback (#486)

### LINKED ISSUE #486: Is it possible to catch tab bar press events?
I'm trying to figure out a way to detect tab press event. What I'd like to achieve is:

1. A user presses the first tab and navigates to _tab-1_ screen.
2. Scrolls down a bit.
3. Presses the first tab again --> __scrolls to top__

I saw passing `onTabPress` callback down to the `<TabBar />` (of _react-native-tab-view_) mentioned at PR #405. I tried something like:

```javascript
TabNavigator(
  { /* ...routes */ },
  { tabBarOptions: { onTabPress: () => { /* ... */ } }
)
```

Unfortunately, nothing happened. It seems like `onTabPress` is not passed down to the `<TabBar />`. 

Feels like I'm missing something very obvious. Any comments would be appreciated!

__My question is__: Is it possible to pass `onTabPress` callback down to the `<TabBar />`? If not, are there any workarounds to achieve my goal to catch tab press events?

kimdhoe:
Pressing a tab bar label often requires performing some kind of side effects
other than navigating: _Scroll-to-top_ functionality, for instance.
At the moment `TabView.TabBarTop`, which is the default on _Android_,
handles `onTabPress` callback as expected, but `TabView.TabBarBottom`
ignores it.

This PR addresses the problem so that `TabView.TabBarBottom` can accept
and handle the optional `onTabPress` callback. The callback is invoked
exactly once if present on every tab bar press, accepting as a sole
parameter a `NavigationRoute`, which is the route object the pressed
target represents.

Fixes #486.

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -36,6 +36,7 @@ type Props = {
   navigationState: NavigationState;
   jumpToIndex: (index: number) => void;
   getLabel: (scene: TabScene) => ?(React.Element<*> | string);
+  getTabPressCallback: (scene: TabScene) => Function;
   renderIcon: (scene: TabScene) => React.Element<*>;
   showLabel: boolean;
   style?: Style;
@@ -56,6 +57,13 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
   };
 
   props: Props;
+
+  _onTabPress = (scene: TabScene) => {
+    const { getTabPressCallback, jumpToIndex } = this.props
+    const onTabPress = getTabPressCallback(scene);
+    onTabPress(scene);
+    jumpToIndex(scene.index);
+  };
 
   _renderLabel = (scene: TabScene) => {
     const {
@@ -147,7 +155,7 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
           });
           const justifyContent = this.props.showIcon ? 'flex-end' : 'center';
           return (
-            <TouchableWithoutFeedback key={route.key} onPress={() => jumpToIndex(index)}>
+            <TouchableWithoutFeedback key={route.key} onPress={() => this._onTabPress(scene)}>
               <Animated.View style={[styles.tab, { backgroundColor, justifyContent }]}>
                 {this._renderIcon(scene)}
                 {this._renderLabel(scene)}
--- a/src/views/TabView/TabBarTop.js
+++ b/src/views/TabView/TabBarTop.js
@@ -35,6 +35,7 @@ type Props = {
   position: Animated.Value;
   navigationState: NavigationState;
   getLabel: (scene: TabScene) => ?(React.Element<*> | string);
+  getTabPressCallback: (scene: TabScene) => Function;
   renderIcon: (scene: TabScene) => React.Element<*>;
   labelStyle?: Style;
 };
@@ -50,6 +51,23 @@ export default class TabBarTop extends PureComponent<DefaultProps, Props, void> 
   };
 
   props: Props;
+
+  _onTabPress = (route: NavigationRoute) => {
+    const { navigationState, getTabPressCallback } = this.props;
+    const { routes } = navigationState;
+    const routesLength = routes.length;
+    let index = -1;
+    for (let i = 0; i < routesLength; i++) {
+      if (routes[i] === route) {
+        index = i;
+        break;
+      }
+    }
+    const focused = index === navigationState.index;
+    const scene = { route, index, focused };
+    const onTabPress = getTabPressCallback(scene);
+    onTabPress(scene);
+  }
 
   _renderLabel = (scene: TabScene) => {
     const {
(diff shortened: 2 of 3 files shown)


satya164:
I don't think this is the right way to do it. How would you wire it up to with the correct rendered navigator instance when you have multiple navigators mounted? Seems impossible with this approach.

kimdhoe:
@satya164 I'm thinking dispatching an action to a Redux store from `onTabPress`, so that child-screen can be passed relative information, which is a route's name in my use case.

satya164:
@kimdhoe yes, but say you have the navigator rendered in two places, for example it might just be tabs nested inside a stack, then how do you know which one sent the event?

kimdhoe:
@satya164 Hmm.. I thought I could use route names for that. Wouldn't it be possible if I use unique names for all screens?

satya164:
Route names are unique per navigator. If you have the same navigator rendered in two places, they will have same route names.

kimdhoe:
I hadn't thought that far ahead. It would be tricky to make a correct view to scroll to top in such complicated situations.

In Redux store or context, I'd have to keep track of route names and _refs_ to _ScrollViews_ in the last mounted tab navigator, which would be visible at a moment. The structure will be something like LIFO stack. Whenever a new tab navigator is mounted, new data will be pushed to the stack, and poped from it on unmounting. Then, when a route name is sent from `onTabPress`,  a _ref_ to the correct ScrollView could be retrieved from _top_ of the stack.

But even if such fine control is impossible, IMHO adding `onTabPress` is useful since it can be used for general purposes.

satya164:
I'm sure that fine control is possible if the event listeners could be specified as part of screen configuration rather than being defined when navigator is defined. It's a common pattern and shouldn't require redux or complicated logic to achieve.

stevehollaar:
I have a similar requirement, where I would like to dispatch a navigation action whenever the active screen's tab button is pressed.
I solved it in a similar way as this diff, but by passing the navigation instance, via `this.props.navigation` (which is available, but unused) in `TabBarBottom.js` to the `onTabPress` callback.

@sagivo I believe that would handle the ambiguous case of multiple Navigators, and passing navigation instances to callbacks seems to be a common pattern.

kimdhoe:
Ah, I see. I have come to something like this:

```js
class RecentChatsScreen extends React.Component {
  static navigationOptions = {
    tabBar: () => {
      onTabPress: (scene: TabScene) => {
        if (scene.focused) {
          // scroll to top please
        }
      }
    }
  };
}
```
But I still had to manually manage a stack of ScrollView refs to implement _scroll to top_.

kimdhoe:
Updated the PR. This will allow something like:

```js
class RecentChatsScreen extends React.Component {
  static navigationOptions = {
    tabBar: () => {
      onTabPress: ({ route, index, focused }: TabScene) => {
        if (focused) {
          alert('active tab pressed.');
        }
      }
    }
  };
  // ...
}
```

Ingibjorg:
@satya164 can you review this again? I could really use this functionality in my app 💯

gregorykan:
@satya164 i'd be keen for this too :)

kennydl:
+1

src/views/TabView/TabBarBottom.js
```diff
@@ -57,6 +58,13 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
 
   props: Props;
 
+  _onTabPress = (scene: TabScene) => {
+    const { getTabPressCallback, jumpToIndex } = this.props
+    const onTabPress = getTabPressCallback(scene);
+    onTabPress(scene);
```
leekiernan: Is there a better way to block navigation to a tab than

`
if !!onTabPress {
  jumpToIndex
}
`

superquest:
My use case is to have a tab icon that simply opens a modal (e.g. how camera is opened in Instagram).

For this use case, I'd need to be able to prevent the default [`jumpToIndex(index)`](https://github.com/react-community/react-navigation/pull/525/commits/f120c72af57d1a6b406a32c692deb6d3999d92ef#diff-cf9accb5a3555e3383f8175c7e7c6fd9R65) call.

Thought I'd throw that out there.

rad182:
+1

gislitg:
+1

artembloom:
+1

navid94:
+1

jamesisaac:
@kimdhoe Seeing as the proposed callback doesn't have access to `this.props.navigation`, how would you implement something like resetting the stack when the focused tab is pressed (common pattern)?

cooperka:
Somehow I missed this when searching open PRs... I've implemented onPress consistently for both `TabBarTop` and `TabBarBottom` in this PR: https://github.com/react-community/react-navigation/pull/1335. Curious what your thoughts are.

spencercarli:
Hey @kimdhoe thanks for putting the time and effort into this! 

I'm going to close this in favor of #1335 since that PR is able to be merged immediately and has picked up some more feedback from the community.

Again, thank you for putting the time into this!

PR #545, opened 2017-03-02
Update Screen-Tracking.md

thurt:
updates names in the onNavigationStateChange example function


grabbou:
Why such change? The function is called in [`componentDidUpdate`](https://github.com/react-community/react-navigation/blob/master/src/createNavigationContainer.js#L110) meaning it receives `previous` and `current` state (after update), not `current` and `next`.

thurt:
will update

nihgwu:
I just want to send a PR to revise this 😂 , but there is still another issue so a PR is on the way

nihgwu:
there you go #549

PR #549, opened 2017-03-02
Update Screen-Tracking.md

nihgwu:
when using nested navigators, **getCurrentScreen** always returns the root navigator's routeName and it's totally wrong

### DIFF
--- a/docs/guides/Screen-Tracking.md
+++ b/docs/guides/Screen-Tracking.md
@@ -12,14 +12,14 @@ import { GoogleAnalyticsTracker } from 'react-native-google-analytics-bridge';
 const tracker = new GoogleAnalyticsTracker(GA_TRACKING_ID);
 
 // gets the current screen from navigation state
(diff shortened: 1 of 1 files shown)


docs/guides/Screen-Tracking.md
```diff
@@ -16,7 +16,12 @@ function getCurrentScreen(navigationState) {
   if (!navigationState) {
     return null;
   }
-  return navigationState.routes[navigationState.index].routeName;
+  const route = navigationState.routes[navigationState.index];
+  // dive into nested navigators
+  if (route.routes) {
+    return getCurrentScreen(route);
```
satya164: Maybe we should rename the function to `getCurrentRouteName`

nihgwu: Aye

nihgwu: Done

Issue #531, opened 2017-03-01
DrawerNavigator: unable to scroll through Items


Blackscr33n:
Hey!

As mentioned in the title I am unable to scroll through my list of Screens in the DrawerMenu.

Is there a way I can achieve this? 
Couldn't find an option for.

![image](https://cloud.githubusercontent.com/assets/6410789/23474526/53f0a6e2-feb4-11e6-8b40-0ad6f641fe87.png)

grabbou:
Oh, looks like default contentComponent for Drawer doesn't use `ScrollView` which makes it impossible to scroll. 

For now, what you can do is:
1) copy the contents of [the file](https://github.com/react-community/react-navigation/blob/master/src/views/Drawer/DrawerNavigatorItems.js) into your source code
2) replace container `View` with `ScrollView`
3) pass `contentComponent` to `DrawerConfig`

Generally a quick glance at the available solutions to that problem on Android reveals that it's scrollable when content is too big. 

I'd accept a pull request for that. Please test that it has no difference when there are less items on screen.

satya164:
It's already exported, so it's just `contentComponent: props => <ScrollView><DrawerNavigation.Items {...props} /></ScrollView>`

Blackscr33n:
Thanks both for your reply!

I thought it would be something with `contentComponent`
Thanks!

I'll try it.

Greets

grabbou:
@satya164 neat thing!

Blackscr33n:
May not a bad idea to add it to the docs as example?

grabbou:
If you feel like sending PR - go ahead!

PR #552, opened 2017-03-02
add example for DrawerNavigatorConfig

Blackscr33n:
Related to this issue: #531 

Updated the docs so it is easy to find a solution for a common problem with DrawerNavigator.


docs/api/navigators/DrawerNavigator.md
```diff
@@ -91,6 +91,16 @@ The route configs object is a mapping from route name to a route config, which t
...
 - `contentOptions` - Configure the drawer content, see below.
 
+#### Example:
+
+```js
+{
+  drawerWidth: 200,
+  drawerPosition: 'right',
+  contentComponent: props => <ScrollView><DrawerView.Items {...props} /></ScrollView>
+}
```
grabbou: What this example is demonstrating? Please describe. E.g. `The following example demonstrates how you can set a custom `contentComponent`. In this case, we wrap default `DrawerView` with a `ScrollView` so that we can scroll the items in the sidebar.

grabbou: +1 to what matthamil said

Blackscr33n:
Thanks for your reply!

I will add a description soon!

satya164:
Maybe a throw an image as well there to show how to add a cover image in the drawer.

Issue #540, opened 2017-03-02
Title overlaps back button and pushes right component outside on iOS


lfabreges:
Hi,

I want my header to render like bellow:
![image](https://cloud.githubusercontent.com/assets/2059919/23505563/72c37c30-ff44-11e6-86f7-c658f22c5dcc.png)

But when I provide a long title, it renders like this, with the right button pushed outside :
![image](https://cloud.githubusercontent.com/assets/2059919/23505600/a4fcc508-ff44-11e6-8b00-00e561252ad1.png)

And with a longer title:
![image](https://cloud.githubusercontent.com/assets/2059919/23505656/dc452492-ff44-11e6-80ec-7d2efdae85e2.png)

I want the title to be truncated so that the left and right components are always visible with a minimal margin between them and the title. How can I do it ?

Thanks.

grabbou:
Since you know the title of your component, I believe you can truncate it manually when it's too long. 

Anyways, I've submitted a proof of concept in a PR, please try and let me know if it works.

grabbou:
I am working on a better solution, for now try pointing to this commit https://github.com/react-community/react-navigation/pull/543/commits/0ed203a3996628b1dda933ea438d87b8a3bed965 and see if it works.

lfabreges:
It works:
![image](https://cloud.githubusercontent.com/assets/2059919/23507597/1682be4a-ff4e-11e6-829e-4723013e2aec.png)

Thank you.

grabbou:
Cool, so need to wrap it up into a better implementation :) I'll let you know when it's ready for testing again

lfabreges:
I found an issue with the commit, when changing orientation, the title is flickering:

![flickering](https://cloud.githubusercontent.com/assets/2059919/23518507/f00ce5bc-ff73-11e6-8901-86658ec8b6ab.gif)

grabbou:
Yeah, I need to test it for that use case! I'll. Make sure to send it over
tomorrow 😃

Noitidart:
On Android, if there is a rightHeaderComponent like this:

```
    row: {
        flexDirection: 'row',
        justifyContent: 'center',
        alignItems: 'center'
    },

                <View style={styles.row}>
                    { canEdit && <ButtonHeaderRight icon="edit" lessPadding /> }
                    { hasSocial && <ButtonHeaderRight icon="favorite" lessPadding /> }
                    { hasSocial && <ButtonHeaderRight icon="more_vert" /> }
[...]

Issue #539, opened 2017-03-02
Contents of  header has change offset 


guoliang1206:
when first loading the app,the title and right button ,display on the left of the header and moved to center of header fastly.Anyone find this question and know how to handle this?

satya164:
cc @grabbou 

could it be something due to the delay in `onLayout`?

grabbou:
Yeah, probably, however I haven't experienced any issues because of that. `onLayout` was the only reasonable solution as far as I experimented with that. 

Mind posting a screen recording or/and steps to reproduce? Should be easily fixable.

RobertSheaO:
I just posted about this same issue. This bug was introduced in beta.6 and I think it has to do with dynamically configuring width and converting the top nav area to using flexbox.

https://github.com/react-community/react-navigation/issues/553

thorsonmscott:
I'm also seeing this issue and can confirm I wasn't seeing it before beta.6.  I'll grab a screen recording and upload tomorrow AM.

guoliang1206:
![header bug](https://cloud.githubusercontent.com/assets/12846676/23538230/ab033496-000c-11e7-8599-731dbe3058c3.gif)

see this GIF

Issue #445, opened 2017-02-23
Ability to center Android header when only left/right header button present


dmhood:
Due to one of the recent changes in how the header is rendered, I'm not sure how to go about centering the header title on Android if you have only a left/right header button.

![image](https://cloud.githubusercontent.com/assets/2395637/23240860/9adc8e76-f924-11e6-8f07-437438ff9573.png)

Having a right/left button pushes it left/right.  I can't flex it since it's wrapped in the animated view, and alignSelf: 'center' only centers it inside the view, not the outer animated view itself.

This used to work fine before, so maybe there is another option/setting I'm just not seeing.  Is there a way to do this without having to set a specific margin depending on if a left/right button is present?

grabbou:
That sounds like a valid request. I am not an Android developer, so cannot confirm whether this falls under being a `native` feature and should be implemented in default header or not.

> This used to work fine before

I don't think it used to work in the initial release, see this code: https://github.com/react-community/react-navigation/blob/c37ad8a0a924d13f3897bc72fbda52aac76904b6/src/views/Header.js#L115-L120 in original diff - whenever left/right was missing, the title was pushed to the other direction.

Here's screenshot with `textAlign: center` right after titleStyle was supported:
<img width="384" alt="screen shot 2017-02-23 at 07 52 12" src="https://cloud.githubusercontent.com/assets/2464966/23247942/20934b2c-f99d-11e6-9543-636499df3e39.png">

dmhood:
Hmm maybe there was another change further down the line?  I have it working on my fork (up to the back button label change) https://github.com/sportsaction/react-navigation with the following style:

```javascript
titleStyle: {
      textAlign: 'center'
}
```

![image](https://cloud.githubusercontent.com/assets/2395637/23248342/bea9a5f8-f953-11e6-9f72-462d231ddb65.png)

grabbou:
Actually, that was never meant to work that way. [This commit](https://github.com/react-community/react-navigation/pull/254) removed `style` which made the `style.left/style.right` no-longer take effect. It was shortly after fixed in another pull request (the one you refer to).

The intermediate state caused a couple of glitches and one of the side effects was making `center` to work. It is worth noting that it was never supported officially.

Let me label that as a `feature request` and I'll see what we can do in order to support it the best way.

dmhood:
Sounds good, thanks!

PR #562, opened 2017-03-03
Change <Header /> layout back to `absolute`

grabbou:
This pull request fixes all issues regarding Header that were already filled in by:

1) Reverting previous `absolute` layout - turns `flex` looks nice and was easier to implement, but caused too many issues and it was impossible to keep it going forward
2) `onLayout` is async and it causes flickering as reported in issues - this PR makes sure absence of `onLayout` causes no issues. It does so by not touching layout at all. As soon as width of title is known, we only manipulate back button. No other parts are touched. Right and title stay as they used to be.
3) on iOS, title offset is `70`, whereas on `Android` 40 (as it used to be). This is because on iOS, by default we want to render `< Back` when title is too long. This is to make sure we end up with a well truncated title. Having `40` would cover part of `Back` string.

Fixes: #540 #539 #445

### DIFF
--- a/docs/guides/Redux-Integration.md
+++ b/docs/guides/Redux-Integration.md
@@ -5,6 +5,8 @@ To handle your app's navigation state in redux, you can pass your own `navigatio
 With redux, your app's state is defined by a reducer. Each navigation router effectively has a reducer, called `getStateForAction`. The following is a minimal example of how you might use navigators within a redux application:
 
 ```
+import { addNavigationHelpers } from 'react-navigation';
+
 const AppNavigator = StackNavigator(AppRouteConfigs);
 
 const navReducer = (state, action) => {
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -181,17 +181,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     );
   }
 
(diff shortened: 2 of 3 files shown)


grabbou:
iOS screens:
1) Title & Right
screen shot 2017-03-03 at 11 18 56

2) Left, title & right (no truncating)
screen shot 2017-03-03 at 11 18 59

3) Left, title & right (truncated, title of previous scene too long)
screen shot 2017-03-03 at 11 19 02

4) Super long title on iOS
screen shot 2017-03-03 at 11 19 42

grabbou:
Android:
1) No left icon
screen shot 2017-03-03 at 11 40 14

2) Normal
screen shot 2017-03-03 at 11 37 47

grabbou:
Ability to center Android title:
```js
titleStyle: {
   alignSelf: 'center',
}
```
screen shot 2017-03-03 at 11 43 16

Snorlock:
alignSelf og titleStyle, does not work on react elements, when adding those to the header.
Iv added a simple View, with a SVG logo, but, I have to alignSelf:'center', on the view itself

Ajinkya0509:
Please specify how to extend title size if in specific case we don't have left and right buttons.

PR #563, opened 2017-03-03
correction of a few documentation typo

mbaroukh:
Hi,

Some typos on docs (So I think).

The less obvious is on https://reactnavigation.org/docs/routers/ : 
The sample says that we can override methodes. But I think it is not true, at least on StackNavigator.
On StackNavigator, the router is a constant (https://github.com/react-community/react-navigation/blob/master/src/navigators/StackNavigator.js).
What worked for me is to use 
Object.assign(MyApp.router, { ...}) 
to keep the same instance but change properties.

Also, to override, we need keep a ref on previous methid because we can't call MyApp.router.getStateForAction(action, state) : I think it will result in an infinite loop.

### DIFF
--- a/docs/api/routers/RoutersAPI.md
+++ b/docs/api/routers/RoutersAPI.md
@@ -4,7 +4,7 @@ You can make your own router by building an object with the following functions:
 
 ```js
 const MyRouter = {
-  getStateForAction: (action) => ({}),
+  getStateForAction: (action, state) => ({}),
   getActionForPathAndParams: (path, params) => null,
   getPathAndParamsForState: (state) => null,
   getComponentForState: (state) => MyScreen,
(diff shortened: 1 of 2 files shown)


docs/api/routers/RoutersAPI.md
```diff
@@ -51,7 +51,7 @@ Typically this should return a navigation state, with the following form:
 
 If the router has handled the action externally, or wants to swallow it without changing the navigation state, this function will return `null`.
 
-### `getComponentRouteName(routeName)`
+### `getComponentForRouteName(routeName)`
```
matthamil: I'm not familiar enough with the routers to have input on the other corrections in this PR, but changing `getComponentRouteName` to `getComponentForRouteName` is a good catch.

PR #567, opened 2017-03-03
Custom header style interpolator

cychiuae:
Currently we cannot configure header transition style so this PR is going to address this problem.

Developers can adding `headerStyleInterpolator` to `StackNavigatorConfig` to specify the header style interpolator

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -105,7 +105,7 @@ Visual options:
   - `float` - Render a single header that stays at the top and animates as screens are changed. This is a common pattern on iOS.
   - `screen` - Each screen has a header attached to it and the header fades in and out together with the screen. This is a common pattern on Android.
   - `none` - No header will be rendered.
-- `headerStyleInterpolator` - Use this prop to override the default  Custom header style interpolator when transitioning
+- `headerStyleInterpolator` - Use this prop to override the default header style interpolator when transitioning
 - `cardStyle` - Use this prop to override or extend the default style for an individual card in stack.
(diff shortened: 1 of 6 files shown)


docs/api/navigators/StackNavigator.md
```diff
@@ -84,7 +105,7 @@ Visual options:
   - `float` - Render a single header that stays at the top and animates as screens are changed. This is a common pattern on iOS.
   - `screen` - Each screen has a header attached to it and the header fades in and out together with the screen. This is a common pattern on Android.
   - `none` - No header will be rendered.
-- `headerStyleInterpolator` - Custom header style interpolator when transitioning
+- `headerStyleInterpolator` - Use this prop to override the default  Custom header style interpolator when transitioning
```
matthamil: nit: period after default.

cychiuae: My bad

ericvicenti:
I think we should re-consider this approach. Instead of just adjusting the style/interpolators, I think it makes more sense to **provide an entirely different header component**. That way, people can customize the header more completely (for some header designs, it might not make sense to have a 'center' and 'right' interpolator)

Please re-open if you have a good argument for doing it this way. Otherwise, can you make a new PR for the alternate-header-component API?

strazi:
I think this would be helpful to add. My particular situation has a title that is not center aligned on iOS, so being able to customize the transition to only include opacity is very helpful.

It's also unclear why [#1293](https://github.com/react-community/react-navigation/pull/1293) was accepted, when conceptually this feels very similar.

PR #568, opened 2017-03-03
Adjust slide distance of card in navigation transition

davidskaarup:
Set slide distance of the underlying card in the stack to 30% of the
screen width to match the transition on iOS

### DIFF
--- a/src/views/CardStackStyleInterpolator.js
+++ b/src/views/CardStackStyleInterpolator.js
@@ -65,9 +65,10 @@ function forHorizontal(props: NavigationSceneRendererProps): Object {
   // Add ~30px to the interpolated width screens width for horizontal movement. This allows
   // the screen's shadow to go screen fully offscreen without abruptly dissapearing
   const width = layout.initWidth + 30;
+  const slideDistance = width * 0.3;
   const outputRange = I18nManager.isRTL ?
-    ([-width, 0, 10, 10]: Array<number>) :
-    ([width, 0, -10, -10]: Array<number>);
+    ([-width, 0, slideDistance, slideDistance]: Array<number>) :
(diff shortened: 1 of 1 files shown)


ahanriat:
@davidskaarup thanks for this PR 👍  What do you think about setting the opactity to 1 ( on iOS )to match the iOS transition ?

davidskaarup:
@ahanriat well the standard iOS transition does not have the opacity effect, but apps like facebook, twitter and airbnb use it. I don't know which should be the default here, but I don't think the opacity functionality should be removed entirely since I could imagine quite a few people would want to use it. @grabbou ?

ahanriat:
@davidskaarup I've checked and neither Facebook or Airbnb use opacity, they use a shadow (the twitter one is an overlay) 😉

davidskaarup:
@ahanriat you're right, I can see that facebook doesn't use the opacity/overlay as of now - I'm pretty sure they used to though ;) I think airbnb does have a very subtle opacity effect in addition to the shadow, but it's hard to tell.
Are you sure the twitter effect is not opacity with a dark background? I can achieve the same effect in my app by wrapping my stacknavigator in a container with dark background.

ahanriat:
@davidskaarup Yeah I don't really know for twitter, I tought it was an overlay but yeah it may be an opacity with a dark background ;)

ericvicenti:
I think we should make it easier for people to customize this in their app, without changing the default look.

If we do ever change the default look, I expect the PR to have lots of photos/videos of how it looks!

hnryjms:
Can we revisit this PR? On iPad, the `10px` animation when pushing and popping views looks extremely static when compared to the animation speed of the new view that is sliding over to cover what is behind it.

PR #585, opened 2017-03-06
Send destructured screenProps to Component

anback:
Hi!

I have used NavigationExperimeltal a lot with naviagation in Redux tree. I just started to migrate to this awesome library but one show stopper was this. I really dont understand why you are hiding the props sent down from my reducer function in an object called screenProps. My goal is to be able to navigate to normal components. Now when the props are inside the screenProps object i need to split up components into "go To"-able components and normal components => probably not what was intended.

Anyways i use a HOC function for now to solve this, let me know what you guys think and if i missed you something. Ill continue to test the lib if you guys get back to me with some small feedback. 

Let me know if you are interested in a minimalistic repo with this issue explained.

**Test plan (required)**

**Code formatting**

### DIFF
--- a/src/views/SceneView.js
+++ b/src/views/SceneView.js
@@ -37,7 +37,7 @@ export default class SceneView extends PureComponent<void, Props, void> {
 
     return (
       <Component
-        screenProps={screenProps}
+        {...this.props.screenProps}
         navigation={navigation}
       />
     );


satya164:
#166 and #241

anback:
@satya164 can i just get a quick clarification, after reading through #166 and #241  i get the impression that this will not be supported and that i need to use a HOC on my side. IS this correct ?

satya164:
They outline the reason why we don't destructure screen props. `screenProps` are meant to be a last resort anyways, you should connect your individual screens to redux instead of connecting the navigator component for non-navigation state. Otherwise it'll be bad for performance.

anback:
Alright thx.

PR #645, opened 2017-03-10
Make header component containers' styles configurable

cychiuae:
In beta 6, header components' styles are absolute positioned and I hate the fact that title component has left/right offset so I make the style configurable to override the default style

Since we now have a way to set component styles, I also deprecated the `titleStyle`

Demo:

```js
const MyHomeScreen = ({ navigation }) => (
  <MyNavScreen
    banner="Home Screen"
    navigation={navigation}
  />
);
  title: 'Welcome',
  header: () => ({
    componentStyle: {
      title: {
        backgroundColor: 'red',
        left: 0,
        right: 0,
      },
    },
  }),
};
```
screen shot 2017-03-10 at 11 13 53 pm

```js
const MyProfileScreen = ({ navigation }) => (
  <MyNavScreen
    banner={
      `${navigation.state.params.mode === 'edit' ? 'Now Editing ' : ''
      }${navigation.state.params.name}'s Profile`
    }
    navigation={navigation}
  />
);
MyProfileScreen.navigationOptions = {
  header: ({ state, setParams }) => ({
    title: `${state.params.name}'s Profile!`,
    // Render a button on the right side of the header.
    // When pressed switches the screen to edit mode.
    right: (
      <Button
        title={state.params.mode === 'edit' ? 'Done' : 'Edit'}
        onPress={() => setParams({ mode: state.params.mode === 'edit' ? '' : 'edit' })}
      />
    ),
    componentStyle: {
      title: {
        backgroundColor: 'red',
        left: 0,
        right: 0,
      },
      left: {
        backgroundColor: 'blue',
      },
      right: {
        backgroundColor: 'yellow',
      },
    },
  }),
};
```

screen shot 2017-03-10 at 11 13 59 pm

```js
const MyProfileScreen = ({ navigation }) => (
  <MyNavScreen
    banner={
      `${navigation.state.params.mode === 'edit' ? 'Now Editing ' : ''
      }${navigation.state.params.name}'s Profile`
    }
    navigation={navigation}
  />
);
MyProfileScreen.navigationOptions = {
  header: ({ state, setParams }) => ({
    title: `${state.params.name}'s Profile!`,
    // Render a button on the right side of the header.
    // When pressed switches the screen to edit mode.
    right: (
      <Button
        title={state.params.mode === 'edit' ? 'Done' : 'Edit'}
        onPress={() => setParams({ mode: state.params.mode === 'edit' ? '' : 'edit' })}
      />
    ),
    componentStyle: {
      title: {
        backgroundColor: 'red',
        left: 0,
        right: 0,
      },
      right: {
        backgroundColor: 'yellow',
      },
    },
  }),
};
```
screen shot 2017-03-10 at 11 15 42 pm

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -125,6 +125,10 @@ All `navigationOptions` for the `StackNavigator`:
   - `left` - React Element to display on the left side of the header
   - `style` - Style object for the header
   - `titleStyle` - Style object for the title component
+  - `componentStyle` - a config object for styling header components
+    - `title` - Style object for the title component
+    - `left` - Style object for the left component
+    - `right` - Style object for the right component
   - `tintColor` - Tint color for the header
 - `cardStack` - a config object for the card stack:
   - `gesturesEnabled` - Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -124,7 +124,6 @@ All `navigationOptions` for the `StackNavigator`:
   - `right` - React Element to display on the right side of the header
   - `left` - React Element to display on the left side of the header
   - `style` - Style object for the header
-  - `titleStyle` - Style object for the title component
   - `componentStyle` - a config object for styling header components
     - `title` - Style object for the title component
     - `left` - Style object for the left component
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -130,6 +130,16 @@ export type HeaderConfig = {
    */
   titleStyle?: Style,
 
+
+  /**
+   * Styles passed into navigation bar components
+   */
+  componentStyle?: {
+    left?: Style,
+    title?: Style,
+    right?: Style,
+  },
+
   // // Style of title text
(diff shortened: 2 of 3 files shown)


satya164:
What's the motivation? What kind of UI do you want to achieve? Setting background color for left and right components seem very specific. 

There's `titleStyle` for customising the title's style.

cychiuae:
@satya164 
Actually, I have to remove the title component's, which is a custom view,  left and right offset in my current project and `titleStyle` is only applied when we set `title` in `navigationOptions` but not `title` in `header`

satya164:
> titleStyle is only applied when we set title in navigationOptions but not title in header

That looks like a bug.

cychiuae:
> That looks like a bug.

Oh I see. Anyway, I think it is more flexible if developers can set the components' styles.

satya164:
> I think it is more flexible if developers can set the components' styles.

You can set back buttons color with `tintColor`. I don't think people rarely need to set the background color. You can replace whole back button with your implementation if you want heavy customization.

Regarding right component, it's always provided by the user, so there's no need for having a config option for that.

cychiuae:
I see your point. Will close PR and submit a new one for fixing the titleStyle not applying to custom title view.

PR #649, opened 2017-03-11
Fix header title style not applied to custom title component

cychiuae:
Consider this small example:

```js
const MyHomeScreen = ({ navigation }) => (
  <MyNavScreen
    banner="Home Screen"
    navigation={navigation}
  />
);
  header: () => ({
    title: (
      <SampleText>
        Welcome
      </SampleText>
    ),
    titleStyle: {
      backgroundColor: 'red',
    },
  }),
};
```

Before the fix:
screen shot 2017-03-11 at 12 53 51 pm

After the fix:
screen shot 2017-03-11 at 12 54 06 pm


ericvicenti:
This will cause the same style to be added to the View and the Text components. Can we take a different approach and add a new option that configures only the header `titleViewStyle`?

satya164:
Why not pass style to your custom view directly instead of adding a configuration option?

cychiuae:
@satya164 

> Why not pass style to your custom view directly instead of adding a configuration option?

That's because I want to remove the left and right margin actually instead of changing the background color of that view.

Do you have a better solution for that?

satya164:
You could use negative margins. Though in this case you are essentially replacing the header. Which is @grabbou working on

grabbou:
I am going to close this pull request in favour of the one that allows using custom header component. Thank you @cychiuae for suggesting these changes, but we decided to take the direction as mentioned by @satya164. Looking forward to your next PR!

Issue #659, opened 2017-03-12
can't set tab bar icon height properly


zhaoyi0113:
I am using `react-nativation` on an Android app. Below is my tab navigation configuration:

```
const Tabs = TabNavigator({
  Home: {
    screen: HomeContainer,
  },
  Patient: {
    screen: PatientContainer,
  },
  User: {
    screen: UserContainer,
  }
}, {
  tabBarOptions: {
    showIcon: true,
    style: {backgroundColor: 'lightgray', height: 60, padding:0, margin:0},
    showLabel: false,
    tabStyle: {margin: 0, padding: 0, height: 60},
    indicatorStyle: {backgroundColor: 'lightgray'}
  },
  tabBarPosition: 'bottom',
});
```

I set the height to be `60` on the tab bar options. It works fine. But when I place an image icon on the tab bar. The height of the icon can't be set to 60. Below is navigationOptions set on my component:

```
static navigationOptions = {
    tabBar: {
      label: '',
      // Note: By default the icon is only shown on iOS. Search the showIcon option below.
      icon: (obj) => {
        const image = obj.focused ? require('../../../images/home_pre.png') : require('../../../images/home_nor.png');
        return <Image style={{resizeMode:'contain', height:50, width:40}}
                      source={image}
        />
      },
    },
  }
```

From below screenshot you can see that the icon image is clipped and there are spaces on the top and bottom of the image. How to set the icon image to take full height of the tab bar?

![screen shot 2017-03-12 at 11 55 05 pm](https://cloud.githubusercontent.com/assets/3355579/23831955/6a7affe2-077f-11e7-9fc1-aacc00c793fb.png)

ldipasquale:
The tab icon is wrapped by the [TabBarIcon component](https://github.com/react-community/react-navigation/blob/0c8f9f7424e4af92f7ed9b9e564d930d43157bba/src/views/TabView/TabBarTop.js), whose style has 24x24.

`tabBarOptions` object should have a `iconStyle` extending this properties (width, height, etc)

cooltoo:
Has `iconStyle` been implemented? I didn't find it on the source code

ldipasquale:
No, it hasn't. But the PR should be _easy_.

zhaoyi0113:
I have made the changes and it works fine. Can I get the permission to submit a pull request?

rgabs:
It is implemented now in the latest version of react-navigation

kelset:
[...]

PR #676, opened 2017-03-14
can't set tab bar icon height properly #659

zhaoyi0113:

I am using react-nativation on an Android app. Below is my tab navigation configuration:

```
const Tabs = TabNavigator({
  Home: {
    screen: HomeContainer,
  },
  Patient: {
    screen: PatientContainer,
  },
  User: {
    screen: UserContainer,
  }
}, {
  tabBarOptions: {
    showIcon: true,
    style: {backgroundColor: 'lightgray', height: 60, padding:0, margin:0},
    showLabel: false,
    tabStyle: {margin: 0, padding: 0, height: 60},
    indicatorStyle: {backgroundColor: 'lightgray'}
  },
  tabBarPosition: 'bottom',
});
```
I set the height to be 60 on the tab bar options. It works fine. But when I place an image icon on the tab bar. The height of the icon can't be set to 60. Below is navigationOptions set on my component:

```
static navigationOptions = {
    tabBar: {
      label: '',
      // Note: By default the icon is only shown on iOS. Search the showIcon option below.
      icon: (obj) => {
        const image = obj.focused ? require('../../../images/home_pre.png') : require('../../../images/home_nor.png');
        return <Image style={{resizeMode:'contain', height:50, width:40}}
                      source={image}
        />
      },
    },
  }
```
From below screenshot you can see that the icon image is clipped and there are spaces on the top and bottom of the image. How to set the icon image to take full height of the tab bar?

**Test plan (required)**

**Code formatting**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-navigation",
-  "version": "1.0.0-beta.6",
+  "version": "1.0.0-beta.7",
   "description": "React Navigation",
   "main": "src/react-navigation.js",
   "sources": {
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "hoist-non-react-statics": "^1.2.0",
     "path-to-regexp": "^1.7.0",
     "react-native-drawer-layout-polyfill": "^1.0.4",
-    "react-native-tab-view": "^0.0.56"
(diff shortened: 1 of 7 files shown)


zhaoyi0113:
sorry made a mistake

PR #681, opened 2017-03-14
Added flexible option *component* to navigationOptions for redefine StackNavigatorConfig.headerComponent

valeryq:
That's PR follow my opinion.
https://github.com/react-community/react-navigation/pull/470#issuecomment-286036663


grabbou:
Thanks for the PR! I am going to close it as the CardStack and Header did change in a significant way since then. The #470 has been also closed and is waiting for a new PR.

PR #731, opened 2017-03-18
Possibility to overwrite label's style if defined as string.

alimek:
Added option to overwrite label style if label is defined as string in navigationOptions in DrawerNavigation.


ericvicenti:
Can you add this feature to the documentation, so people can see how to use it?

alimek:
@ericvicenti sure, already added :)

PR #732, opened 2017-03-18
truncatedBackTitle navigationOptions for StackNavigator

Freddy03h:
`truncatedBackTitle` `navigationOptions` for `StackNavigator`

This is to redefined the `'Back'` string default props on the `HeaderBackButton` component. But keeping the truncate behavior on iOS (that is impossible with the `backTitle` option).

It's necessary for translation.

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -121,7 +121,7 @@ All `navigationOptions` for the `StackNavigator`:
   - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
   - `title` - String or React Element used by the header. Defaults to scene `title`
   - `backTitle` - Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
-  - `truncatedBackTitle` - Title string used by the back button on iOS when there is not enough space. Default is `'Back'`
+  - `truncatedBackTitle` - Title string used by the back button on iOS when there is not enough space to display the previous screen's title. Default is `'Back'`
   - `right` - React Element to display on the right side of the header
   - `left` - React Element to display on the left side of the header
   - `style` - Style object for the header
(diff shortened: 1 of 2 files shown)


docs/api/navigators/StackNavigator.md
```diff
@@ -121,6 +121,7 @@ All `navigationOptions` for the `StackNavigator`:
   - `visible` - Boolean toggle of header visibility. Only works when `headerMode` is `screen`.
   - `title` - String or React Element used by the header. Defaults to scene `title`
   - `backTitle` - Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
+  - `truncatedBackTitle` - Title string used by the back button on iOS when there is not enough space. Default is `'Back'`
```
ericvicenti: "when there is not enough space to display the previous screen's title"

src/views/Header.js
```diff
@@ -149,6 +157,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
       return null;
     }
     const tintColor = this._getHeaderTintColor(props.navigation);
+    const truncatedBackTitle = this._getHeaderTruncatedBackTitle(props.navigation);
```
grabbou: `truncatedBackTitle` should be defined on the screen that also defines `backTitle`.

grabbou:
Mostly looking good. Needs re-basing.

src/views/Header.js
Freddy03h: I'm sorry but I don't understand what should I supposed to do

Freddy03h:
Not easy to rebase, project change a lot since a month and I'm not aware of the changes. I think it will be easier to re-start from a fresh clone.

grabbou:
Yeah. We already got @fson PR into master #1167 👍

PR #735, opened 2017-03-19
Fix flow in Navigation Playground

agrcrobles:
flow fixes for the NavigationPlayground


examples/NavigationPlayground/js/SampleText.js
```diff
@@ -10,7 +10,13 @@ import {
 /**
  * Used across examples as a screen placeholder.
  */
-const SampleText = ({ children }) => (
+ import type { Children } from 'react';
```
cyprusglobe: nit: Indentation looks off for import

PR #743, opened 2017-03-20
Merge navigationOptions defined in component and routeConfigs

qftgtr:
`navigationOptions` can be defined in two places:

1. In `StackNavigator`

``` js
StackNavigator(RouteConfigs, {
  navigationOptions: {
    header: {
      style: { backgroundColor: '#ffffff' },
      titleStyle: { color: '#00ff00' },
    },
  },
});
```

2. In Component as a `static` property

``` js
class MainScreen extends React.Component {
  static navigationOptions = {
    title: 'Main',
    header: ({ state, setParams }) => ({
      // Render a button on the right side of the header
      right: (
        <Button
          title={'Edit'}
          onPress={() => setParams({ editing: state.params.editing ? false : true })}
        />
      ),
    }),
  };
  ...
```

It will be convenient to define some global properties (such as global header style) when implementing `StackNavigator` and leave by-screen customizations (like right button) in each component. However, the `header` object in `StackNavigator` will *replace* the `header` in the screen component. For example, the right button above (and those in all the other screen components) will be discarded. This PR will merge the two `header` defined in different place.

### DIFF
--- a/src/routers/createConfigGetter.js
+++ b/src/routers/createConfigGetter.js
@@ -66,6 +66,7 @@ export default (
             }
             return acc;
           }
+          return option;
         }
         return acc;
       },
--- a/src/routers/createConfigGetter.js
+++ b/src/routers/createConfigGetter.js
@@ -57,9 +57,15 @@ export default (
     ].reduce(
       (acc: *, options: NavigationScreenOptions) => {
         if (options && options[optionName] !== undefined) {
-          return typeof options[optionName] === 'function'
+          const option = typeof options[optionName] === 'function'
             ? options[optionName](navigation, acc)
             : options[optionName];
+          if (acc && typeof acc === 'object') {
+            for (var attr in option) {
+              acc[attr] = option[attr];
+            }
+            return acc;
+          }
         }
         return acc;
       },


satya164:
@grabbou 😂

rafaelmotta:
I have the same problem... I can't extend navigation options per screen when I define navigationOptions on StackNavigator. The only way to show right button in specific screens  is setting navigation options screen per screen

grabbou:
cc: @ericvicenti

ericvicenti:
We already do merge navigation options between the route and screen. But I agree, it gets confusing because the options are two levels deep. 

The documentation explains all of this, and demonstrates how you can merge options like `header` manually: https://reactnavigation.org/docs/navigators/navigation-options#Default-Navigation-Options (see "Extending defaults").

I think the API is more complex than it needs to be, which is why people are having troubles even though the behavior is documented. In order to avoid this sort of confusion, I think we should have a flat list of navigation options. We wouldn't run in to this problem if we had `headerStyle` and `headerRight` instead of `header.style` and `header.right`.

tomav:
There are 2 or 3 configuration levels?

I see:
- StackNavigator RouteConfigs navigationOptions (per screen definition)
- StackNavigator StackNavigatorConfig navigationOptions (global configuration)
- Component static navigationOptions

grabbou:
Closing as options are now `flat` and there's no need to magically merge them :)

Issue #715, opened 2017-03-16
Q: How would one decorate the children component in a DrawerNavigator?


PuercoPop:
For context, I need to combine the drawer navigator with a nav-bar + tab-bar. To do so I've currently have a forked version of DrawerNavigator that takes to extra config options, topBar and bottomBar, which are the components I want to inject. This works 'ok', using react-redux I can add the navigation actions I need to to the injected components. However forking is not an ideal solution.

Would it be reasonable to ask that for DrawerScreen to be exported in a similar way than DrawerNavigatorItems and for DrawerNavigator to receive a prop DrawerScreenComponent to allow customizing it? (I can make the PR if it seems reasonable to y'all). Am I missing a path of least resistance?

ericvicenti:
This sounds reasonable, but just to make it clear what you want to change, would you mind documenting the proposal a bit more clearly?

Pretend you're writing docs for the new feature (and then if we add the functionality, we can use it as the docs)

PR #749, opened 2017-03-20
Export drawer screen

PuercoPop:
Hi!

The motivation is to be able to decorate the screens of a DrawerNavigator. With this PR one can wrap whatever one wants around DrawerScreen and pass the resulting component to a DrawerNavigator drawerScreenComponent.  See #715


PuercoPop:
@satya164 You are right. I was using HOC to inject them with the navigation properties. Should have thought of that way sooner. Thanks! I'll close this PR.

PR #772, opened 2017-03-22
Navigator which is not front is reacting to action by back processing…

nisin:
… when using multiple navigators.

If you multiplex this with "Stack -> Stack" and you tap the "Back" button on the navigation bar, I want you to operate only on the front Stack which is displaying the navigation bar originally , It solves the problem that "Back" is executed for all Stacks.

```
Stack 0
    TopScene
    Stack 1
       Page 0
       Page 1 <- This is displayed
```

If you create a navigation in the above state and tap the displayed Back button, you can check the difference in behavior.

### DIFF
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -15,9 +15,16 @@ import NavigationActions from './NavigationActions';
 export default function<S: *> (navigation: NavigationProp<S, NavigationAction>) {
   return {
     ...navigation,
-    goBack: (key?: ?string): boolean => navigation.dispatch(NavigationActions.back({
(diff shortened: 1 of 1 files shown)


nisin:
I found it a problem in operation and withdraw it

nisin:
fix problem. reopen this PR.

nihgwu:
I think this is the wrong way, what if I have a nested navigators like 
```
StackNavigator
  StackNavigaotr
    StackNavigator
```
Although it's unlikely to happen in a real case, but tackling nested navigators in `react-navigation` is always tricky.

ericvicenti:
Can you describe in more detail what the bug is, including some example code that others can use to reproduce it?

PR #775, opened 2017-03-22
Option to add numberOfLines inside header in navigationOptions

jakallergis:
#### Motivation

Inside my app I can't always know the title, and sometimes the title can be 4-5 words long. Since I don't want cropping in my header titles, i can make the size of the header and the font of the title change dynamically but I cant change the numberOfLines attribute.

This forces me to drop the font size very low if I want to fit a long title in one line. Enabling the option to manually set the numberOfLines gives me the ability to combine changing the header height, title fontSize and numberOfLines and make my header look and behave in a more flexible way.

#### Example

Here is an example of the default navigationOptions with long title:

one-line

and the code:

```javascript
 static navigationOptions = {
	header : (options) => ({
		right: (<OKButton onPress={() => {}}/>),
		left: (<CancelButton onPress={() => options.state.params.onCloseModal()}/>),
		style: {
			backgroundColor : Colors.navigationBarColor
		},
		tintColor: Colors.defaultTextColor,
		titleStyle: {
			paddingLeft: 20
		}
	}),
	title : 'A Somewhat Long Title That Could Not Fit In One Line Without My Change'
}
```

and the same navigationOptions with numberOfLines option:

three-line

and the code:

```javascript
 static navigationOptions = {
	header : (options) => ({
		right: (<OKButton onPress={() => {}}/>),
		left: (<CancelButton onPress={() => options.state.params.onCloseModal()}/>),
		style: {
			height: 100,
			backgroundColor : Colors.navigationBarColor
		},
		tintColor: Colors.defaultTextColor,
		numberOfLines: 3,  // <------------------------------------ THE CHANGE
		titleStyle: {
			paddingLeft: 20
		}
	}),
	title : 'A Somewhat Long Title That Now Sits In Three Lines Because Of My Change'
}
```

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -118,10 +118,19 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     return undefined;
   }
 
+  _getHeaderTitleLines = (navigation: Navigation): number => {
+    const header = this.props.router.getScreenConfig(navigation, 'header');
+    if (header && header.numberOfLines) {
+      return header.numberOfLines;
+    }
+    return undefined;
+  }
+
   _renderTitleComponent = (props: SubViewProps) => {
     const titleStyle = this._getHeaderTitleStyle(props.navigation);
     const color = this._getHeaderTintColor(props.navigation);
     const title = this._getHeaderTitle(props.navigation);
+    const titleLines = this._getHeaderTitleLines(props.navigation);
 
     // On iOS, width of left/right components depends on the calculated
     // size of the title.
@@ -138,6 +147,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
 
     return (
       <HeaderTitle
+        numberOfLines={titleLines}
         onLayout={onLayoutIOS}
         style={[color ? { color } : null, titleStyle]}
       >
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -123,7 +123,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     if (header && header.numberOfLines) {
       return header.numberOfLines;
     }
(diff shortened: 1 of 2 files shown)


ericvicenti:
What happens during transitions between screens with different header heights?

jakallergis:
The header changes as well, according to the values you provide.

ericvicenti:
Have you tested on iOS? What does it look like when navigating between a screen with different header heights there? What happens when swiping back?

ericvicenti:
Please post a screenshot/gif of the results on iOS

jakallergis:
new project

@ericvicenti Please forgive the uglyness of this screencast. I don't have access to my computer at the moment so I made a real quick demo with create-react-native-app and a trial app on a friends iPhone/Mac.
I'll try to provide a better looking example with my initial project wen i get back to office.
Thanks for your time

jakallergis:
Any news on this one? thanks

chandu0101:
+100 , i need this feature in my app ;)

satya164:
That jump is pretty ugly.

jakallergis:
@satya164  Yeap I totally aggree! But what I'm proposing is something optional. I'm not saying it should be done like this everywhere. But having the option, we can make for example a modal that has a bigger header in which case ugly transition is not a problem.

What I'm proposing doesn't change the state of the current header. It just adds the extra flexibility some of us need for whatever good or bad reasons.

satya164:
In case of a modal you can use a custom header anyways if you are not concerned about ugly animation. But you should really rethink the design, because animation and the header look pretty ugly. Adding this to the library even as an option means make it super easy to create ugly designs, which I'm not a fan of.

jakallergis:
Fair enough. :)
(You may close it then i guess)

PR #784, opened 2017-03-22
WIP - Use Lerna to manage repo

skevy:
** DO NOT MERGE **

Just testing out using Lerna for managing react-navigation as a monorepo.

We have the want/need to separate packages out of `react-navigation` core to encourage better code reuse and sharing throughout the community.

cc @ericvicenti


skevy:
Update: got stuck because symlink fixes in the packager aren't landed in RN stable. Need to do some crazy hacks, and that's just bad.

Will come back to this when I have a better story on how to deal with the packager.

satya164:
Ref https://github.com/react-community/react-navigation/issues/321#issuecomment-279458003

ericvicenti:
> Will come back to this when I have a better story on how to deal with the packager.

Maybe with @satya164's help we can unblock ourselves with [haul](https://github.com/callstack-io/haul) on this repo 😄

skevy:
Hmm. That's an interesting idea.

grabbou:
@ericvicenti I think now that `haul` keeps stabilising, I think we can try to ship an initial take to see if it works.

grabbou:
I am going to close this for now - will send a follow-up tomorrow that tries to set it up with Haul.

@ericvicenti mind describing what's the desired packages structure?

satya164:
Before we do that, do we have an answer to this question?

> Recently at callstack we needed to use a new unpublished commit in jest and were unable to install it from GIT because it uses lerna. I think this will be a problem too, unless there's a solution I'm not aware of.

ericvicenti:
Lerna can be problematic with unpublished modules :-/

I don't really care how we structure things, as long as we can put several packages in the repo

Issue #723, opened 2017-03-17
Road to v1 stable


ericvicenti:
For the first release, lets focus on API sanity and stability, rather than adding every feature that people have requested.

- [ ] Navigate key for idempotent pushes https://github.com/react-community/react-navigation/issues/135
- [ ] Go back to a screen: https://github.com/react-community/react-navigation/issues/1213
- [x] Header bar overriding https://github.com/react-community/react-navigation/pull/1220
- [ ] Screen lifecycle hooks https://github.com/react-community/react-navigation/issues/51
- [x] Simplified Nav Options https://github.com/react-community/react-navigation/issues/722
- [x] Resetting non-deepest navigator https://github.com/react-community/react-navigation/issues/691
- [ ] Explore `<NavigationOptions />` for a less confusing configuration method
- [ ] Fix issues with drawer navigator - https://github.com/react-community/react-navigation/issues/374

After v1:

- [ ] Accessibility Review https://github.com/react-community/react-navigation/issues/58

Lets aim to reach stability by mid-April!

chandu0101:
any chance to get this  https://github.com/react-community/react-navigation/issues/691 in v1 ..

ericvicenti:
@chandu0101, yep, just added it! I'm still searching around for high-pri issues

migueloller:
@ericvicenti is https://github.com/react-community/react-navigation/issues/175 a high priority for v1?

jakecraige:
I'd second #175 / #124 being part of 1.0. There's a large amount of ceremony around having an individual screen within a `StackNavigator` present from the bottom up that seems like a fairly core feature that many would find useful.

nihgwu:
@ericvicenti what about #271 debounce action reduce
and as I've said in #271, we shouldn't override custom **key**, or we can't find specific screen via **key** as it could be override, per #691 and NavigationActions.back('key')

migueloller:
@nihgwu, I believe #135, which @ericvicenti included in the list above, would solve that problem.

cychiuae:
I believe #527 is important too

nihgwu:
@migueloller sorry I didn't notice that before

nihgwu:
@cychiuae but #527 has been resolved, forget to close it though

cychiuae:
@nihgwu someone said he still got the issue with master

TomAtterton:
[...]

Issue #691, opened 2017-03-14
Specify reset key to define what state should be reset


Kerumen:
This is a call to support. The number of issues which mention this is countless.
Currently there is no other way to build a modern app with a Splashscreen or a Login / Register process without this. And it seems everyone try to find workarounds but each has its downsides.
I would be happy to set up a PR but the source code of react-navigation is too cryptic for me yet.

Relevant issues:
#395 
#199 
#452
#156 
#335 
#325
#59
...

Thanks!

Jacse:
👍  With all the open issues related to this it is pretty clear it is a major missing feature.

ericvicenti:
Thanks for filing this, this certainly seems to be a top-priority issue.

I think this could be fixed by adding a `key` to the `reset` action, which specifies which node is getting reset.

Then we could hardcode a special key for the top-level nav state, like 'root', so that you can fire a reset action with the 'root' key to reset the nav of the whole app.

ralkan:
@Kerumen I've opened a [Pull Request](https://github.com/react-community/react-navigation/pull/1384) regarding this issue.

If this PR gets approved you will be able to specify a `stateName` parameter and then dispatch a reset action for a specific `StackNavigator`.

Cheers!

0x6e6562:
@Kerumen Did #789 resolve your issue? Just curious, since I had no luck with that patch but the fix from @ralkan #1384 does appear to work.

vendramini:
People, is there a solution for this issue? I'm really tired to try to solve this. I tried a lot of things that i can't even put some code here to ask for a objective help. So I will try to show my current situation simple as possible.

Basically: I have a `DrawerNavigator` with some nested `StackNavigator`.
```javascript
export const MainNavigator = DrawerNavigator(
    {
        OrderNew: {
            screen: OrderNew,
            navigationOptions: {
                drawerLabel: DrawerLabel('New Order')
            }
        },
        Orders: {
            screen: OrderNavigator, //a StackNavigator
            navigationOptions: {
                drawerLabel: DrawerLabel('Orders')
            }
        },
        MenuNavigator: {
            screen: MenuNavigator, //a StackNavigator
            navigationOptions: {
                drawerLabel: DrawerLabel('Menu')
            }
        },
    //more screens...
    {
        initialRouteName: 'OrderNew',
        drawerPosition: 'right',
[...]

Issue #395, opened 2017-02-19
Problem with RESET with parallel nested routers.


TomiHiltunen:
I'm having problems understanding how to use `RESET` action when I want to take user to **ListMyMaskedItems** screen under a separate navigator after completing login. `RESET` prints to console the names of the available screens which only include the ones under **OnboardingNavigator**

My navigator/screen structure is as follows:
```
- RootNavigator
    - OnboardingNavigator
        - Intro
        - Login (user is here)
        - Register
    - MainNavigator
        - ListMyMaskedItems (user should go here)
```

Piece from log (I know I'm requesting a view with routeName Onboarding in this excerpt):
```
<Error>: There is no route defined for key Onboarding.
	Must be one of: 'Intro','Login','Register'
```

Any ideas on this?

chandu0101:
I am also facing similar issue #389

ericvicenti:
Have you tried simply using navigate for this? It should work to call `props.navigation.navigate('ListMyMaskedItems', {...myParams});`

TomiHiltunen:
@ericvicenti simply navigating to the parallel router works but the effect isn't really as I hoped. It pushes the next view in but if I drag the screen from left, I am brought back to login screen. My hope is to replace the previous state so that there wouldn't be anything to go back to but rather the first view should be the ListMyMaskedItems.

Preferred history/stack:
```
- RootNavigator
    - ListMyMaskedItems
```

actual history/stack:
```
- RootNavigator
    - Intro
        - Login
            - ListMyMaskedItems
```

dggriffin:
Running in to the same exact problem. Any insight would be wonderful.

ericvicenti:
@dggriffin, can you check out the repo and add a unit test for the routers that demonstrates the issue? The contributors guide should help you set up the tests.

This way it will be super clear to see what the issue is.

dggriffin:
@ericvicenti I could write a failing unit test, but in this case it seems less like a bug and more like a library assumption (which you acknowledge here: https://github.com/react-community/react-navigation/issues/199).

#199, #395, #335 all seem to be a similar use case were developers are attempting to navigate between parallel navigators managed by one parent app navigator whilst also dumping the existing navigation stack when transitioning between those parallel navigators.

In your eyes, does this demonstrate an uncovered use case for react-navigation, or are we approaching the the solution wrong conceptually?

[...]

PR #789, opened 2017-03-23
Reset specific StackRouter with key

scbrady:
**Motivation** 

There is no way to build a Login/Register flow without the ability to reset the router to the top-level nav state without hacky workarounds.

#691 #723 

By allowing a key to be passed along with a reset action, you can specify what level of state you want to reset to.

**Test plan (required)**

I added a test case that verifies the functionality.

I'm happy to add more test cases or documentation, if you find this PR useful.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -253,6 +253,7 @@ export type NavigationInitAction = {
 export type NavigationResetAction = {
   type: 'Navigation/RESET',
   index: number,
+  key?: string,
   actions: Array<NavigationNavigateAction>,
 };
 
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -253,7 +253,7 @@ export type NavigationInitAction = {
 export type NavigationResetAction = {
   type: 'Navigation/RESET',
   index: number,
-  key?: string,
+  key?: ?string,
   actions: Array<NavigationNavigateAction>,
 };
 
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -126,14 +126,14 @@ export default (
         };
         state = {
           index: 0,
-          key: 'Root',
           routes: [route],
         };
       }
 
       // Check if a child scene wants to handle the action
-      if(action.key !== 'Root') {
-        const childIndex = action.key ? StateUtils.indexOf(state, action.key) : state.index;
+      if(action.key !== null) {
+        const keyIndex = action.key ? StateUtils.indexOf(state, action.key) : -1
+        const childIndex = keyIndex >= 0 ? keyIndex : state.index;
         const childRoute = state.routes[childIndex];
         const childRouter = childRouters[childRoute.routeName];
         if (childRouter) {
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -130,8 +130,8 @@ export default (
         };
       }
 
-      // Check if a child scene wants to handle the action
-      if(action.key !== null) {
+      // Check if a child scene wants to handle the action as long as it is not a reset to the root stack
+      if(action.type !== NavigationActions.RESET && action.key !== null) {
         const keyIndex = action.key ? StateUtils.indexOf(state, action.key) : -1
         const childIndex = keyIndex >= 0 ? keyIndex : state.index;
         const childRoute = state.routes[childIndex];
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -131,7 +131,7 @@ export default (
       }
 
       // Check if a child scene wants to handle the action as long as it is not a reset to the root stack
-      if(action.type !== NavigationActions.RESET && action.key !== null) {
+      if(action.type !== NavigationActions.RESET || action.key !== null) {
         const keyIndex = action.key ? StateUtils.indexOf(state, action.key) : -1
         const childIndex = keyIndex >= 0 ? keyIndex : state.index;
         const childRoute = state.routes[childIndex];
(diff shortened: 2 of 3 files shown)


bcarroll22:
Can we get some input on this? It would be really nice to get this merged in if there's no other plans to implement this functionality.

cc @satya164 @ericvicenti

darleikroth:
Thanks for contributing @scbrady.

For tests, I have installed `react-navigation` from your master. And it really works fine, solving the issue I have, like reported here [#395](https://github.com/react-community/react-navigation/issues/395).

But, I have noticed another issue. I don't know if it has relation with your modifications.
In my present `screen`, the `this.props.navigation.setParams({ foo: 'test' })` doesn't works. This doesn't update the `state.params`.

An example:

```js
class MyScreen extends React.Component
{
  static navigationOptions = {
    title: ({ state }) => state.params.title,
    // others props
  }

  componentDidMount()
  {
      title: 'Super Title',
    });
  }
}
```
If I put a `console.log` here: 
```js
static navigationOptions = {
  title: ({ state }) => {
    console.log('the state', state);
    return 'Super Title'
  },
  // others props
}
```
There is no `params` inside the `state`. Just this `{ routeName: 'MyScreen', key: 'Init' }`.

As I said, I don't know if this issue is related with yours modifications. But, I hope it will helpful.

despreston:
any reason this hasn't been reviewed yet?

drop-dan:
This is working great for my purposes, I ended up implementing a custom StackNavigator with this change added to it.  Would love to see this merged.

ericvicenti:
This looks pretty great! I'll land it now. Can somebody go and add this to the docs?

stevenleeg:
Does this PR include the ability to specify the key as `root`, as described in @ericvicenti's [comment](https://github.com/react-community/react-navigation/issues/691#issuecomment-287418546)?

As of now this PR is *mostly* solving my issues, but I'm having a problem resetting all the way back to the root navigator because there are multiple routes with the same key (`Init0` seems to be a default somewhere that keeps getting used, but I can't find where it's being set in the `StackRouter` source).

I suppose another solution to this issue would be the ability to specify a route's key as it's been navigated to, but I haven't been able to find documentation/code that does this. Is it possible?

scbrady:
@stevenleeg The root navigator doesn't actually have a key if I remember correctly, so I made it so you can pass the key `null` into it and it will reset to the root navigator.

`NavigationActions.reset({ actions, index, key: null })`

stevenleeg:
Interesting. Sending `null` to the action seems to work, but I'm still having trouble navigating in my particular scenario. 

Is there a way to set the key of a screen? If so that would help me immensely, as my issue would go from needing to figure out how to reset the navigation to a certain state to dispatching `back` given the view's key (I can't do that now because I have no idea what that screen's key will be when it's placed on the stack).

radicalnerds:
What version will this be built into? I am on beta.7 but I don't see the change.

scbrady:
@radicalnerds It looks like this PR just landed in the newest beta version (v1.0.0-beta.9)

felixus95:
Can somebody please add this to the docs with an example. I think this would be really helpful to a lot of people. @ericvicenti 
Thanks for the work you're doing!

MrHubble:
@felixus95 did you figure out how to use this feature? I'm trying to use it for `replace` like functionality (replace a screen in the stack). However, it always routes to the `index` and doesn't seem to take `key` into account:

````js
let resetAction = NavigationActions.reset({ 
    index: 1,
    key: 'Home',
    actions: [
    NavigationActions.navigate({ routeName: 'Home' }),
    NavigationActions.navigate({ routeName: 'SecondScreen' })
    ]
})
this.props.navigation.dispatch(resetAction)
````

I would expect this to go to the *Home* screen using the `key`, however, it goes to the *SecondScreen* using the `index` (I assume).

**Comment by felixus95** [comment]
@MrHubble 
We are using custom Navigation Actions instead. You should be able to manipulate the index in a way to make it reset/swap like behaviour.

It's documented well in the docs now. But if you have problems hit me up again.

**Comment by FMCorz** [comment]
I'm new to React Navigation, and to this issue, but like many others I'm trying to navigate from a "not logged in" to a "logged in" state. I've managed to pull it off using the `key`, however I really dislike that this drops any form of animation (tested on Android).

```
// From not logged-in state.
// Animations aren't shown.
        dispatch(NavigationActions.reset({
          index: 0,
          key: null,
          actions: [
            NavigationActions.navigate({
              routeName: 'LoggedIn'
            })
          ]
        }));
```

What I want is not a _replace_ like others, I want a _navigate and reset_ option, so that the animation works, but once the user has landed on the page, they can't go back to the previous element in the stack. I did try the following, but the result is really odd... The stack ends up being Login -> Home -> Home.

```
// From Login state.
      dispatch(NavigationActions.navigate({
          routeName: 'Home',
          actions: [
            NavigationActions.reset({
              index: 0,
              key: null
            })
          ]
        }));
```

msageryd:
This seems like an interesting approach. Now I wonder if my approach is flawed. I like to separate the login/signup stack completely from the app stack, so I have two separate stacks which are selectively rendered in the main render function, like this:

    renderInitialRoute() {
      if (this.props.isLoggedInOrAnonymous) {
        return <AppRouter />;
      } else {
        return <LoginRouter />;
      }
    }

Of course I don'e get any animations between the stacks, but I'm ok with that.

scbrady:
@MrHubble 
It has been a while since I've looked at the code, but I believe the keys are something that are auto generated when you make a new stack. Usually it is something like `Init` or `Init-{id}` if I remember correctly. I'm not sure if you can set the keys manually. If you are using Redux to store the navigation state, you can take a look at what the keys are.

stiang:
My app is structured as an outer TabNavigator, with StackNavigators inside each tab. I’m really struggling to reset the StackNavigator state in another tab than the current one. I had hoped that this PR would do the trick, but I see that it only touches StackNavigator files. Could the logic in this PR be used in TabNavigator too, so that I can reset StackNavigators across TabNavigator boundaries, so to speak?

I might take a stab at a PR if someone can confirm that the functionality I need is just missing, not impossible to implement.

PR #809, opened 2017-03-24
remove ReactComponentWithPureRenderMixin

nihgwu:
`ReactComponentWithPureRenderMixin` is useless in this file, and it has been removed from the core of latest `react`

build fails on master branch of `react-native` for this error @ericvicenti


zackify:
When will a release be out for this?

PR #819, opened 2017-03-25
refactored Redux Example

samijaber:
While learning how to use react-navigation with Redux, I attempted to follow the Redux Example. It was however too condensed and the declarations inlined to the point of being illegible.

The goal of this PR is to refactor the redux example into multiple folders/files, and trying to adhere to redux app conventions more for the sake of clarity.


matthamil:
Small but great change.

PR #828, opened 2017-03-26
remove redundant animation when swipe back

nihgwu:
Right now the `_goBack` logic is redundant, after swiped back, we need to wait an extra `ANIMATION_DURATION` time before we can perform other interaction, actually the slide animation is already done in `onNavigateBack` ( navigation.goBack )


nihgwu:
ping @satya164

nihgwu:
ping @satya164 @grabbou

ericvicenti:
CardStackPanResponder is currently not used by CardStack, so this PR is a bit confusing. Could we just delete this file entirely?

nihgwu:
@ericvicenti thanks for your feedback, seems `CardStackPanResponder` is used by `Card`, and I can confirm this PR solves my problem

https://github.com/react-community/react-navigation/blob/master/src/views/Card.js#L59
https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L571

https://github.com/react-community/react-navigation/blob/master/src/views/CardStackStyleInterpolator.js#L67 here we add extra 30px when interpolate so we don't notice the delay, but you can see that when using it in modal mode, and I'm using `StackNavigator` with custom transformation, and removed the extra 30px, then it's very easy to see the delay.

With this PR, the following workaround can be removed because the shadow disappears immediately after wipe back with this fix
 
    // Add ~30px to the interpolated width screens width for horizontal movement. This allows
    // the screen's shadow to go screen fully offscreen without abruptly dissapearing

If needed, I will send another PR to remove this workaround

nihgwu:
@ericvicenti any chance to review this again?

nihgwu:
I get it now, I'm working on beta.6 instead of master branch

nihgwu:
since the remove has been done in https://github.com/react-community/react-navigation/commit/93976d358e27580e568ef016f820dc0253423aa7, I'll close this PR and create a new one to address the goBack animation stuff

PR #844, opened 2017-03-28
feat(header): allow hiding floating headers

blackxored:
`header.visible` has only effect when `headerMode` is `screen`, this effectively makes it
unintuitive to use this prop. In addition, allowing floating headers to be hidden supports a
scenario when you would like to have a splash screen, then your main screen (with a Tabs navigator)
and subsequent non-tab screens being rendered with the floating header so it transitions properly
and in iOS' native style.

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -344,7 +344,9 @@ class CardStack extends Component<DefaultProps, Props, void> {
     let floatingHeader = null;
     const headerMode = this._getHeaderMode();
     if (headerMode === 'float') {
-      floatingHeader = this._renderHeader(props, headerMode);
+      const header = this.props.router.getScreenConfig(props.navigation, 'header') || {}; 
+      const isHiddenHeader = header && header.visible === false;
+      floatingHeader = isHiddenHeader ? null : this._renderHeader(props, headerMode);
     }
 
     const responder = PanResponder.create({


satya164:
If I am correct, this will have a janky effect when navigating between two screens with header and without header.

blackxored:
I'm not entirely sure. I do have exactly that, a no header screen (splash)
and a header screen (tabs) that navigates to all header screens (non-tabs
and tabs as well). I'm using a custom transitioner though so I'm not 100%
sure you're not correct.

—
Best Regards,

Adrian Perez
https://adrianperez.codes

satya164:
Post a video/gif

blackxored:
I'm not sure I'm able to post a vid/gif since this is from a client's app I'm working on.
However this is the setup:

AppNavigator -> TabNavigator -> TabScreen/AppNavigatorScreen

Simplified code follows...


```javascript

const TabsNavigator = TabNavigator({
  /* tabRoutes */
}, {
  header: {
    visible: true,
  },
});

const mainRoutes = {
  splash: { screen: SplashScreen },
  login: { screen: Login },
  home: { screen: TabsNavigator }
}

const otherRoutes = {
  settings: { screen: Settings },
};

const AppNavigator = createCustomNavigator({ // or StackNavigator
  ...mainRoutes,
  ...otherRoutes,
}, {
  headerMode: 'float',
});

SplashScreen.navigationOptions = {
  header: {
    visible: false,
  }
};

```

What this patch allows is switching from any of the tab routes in TabNavigator to one of the routes in AppNavigator (otherRoutes in my snippet, e.g Settings) with a floating header that stays in place and transitions back and forth the way iOS does.

satya164:
> that stays in place and transitions back and forth the way iOS does.

How does it transition when there is no code present to animate the hide/show?

ericvicenti:
I think we should go with an approach that animates the header smoothly, and this will cause the header to flash on and off.

PR #861, opened 2017-03-29
Add loggingEnabled option to navigation container

mhagmajer:
This fixes https://github.com/react-community/react-navigation/issues/851.

**Test plan**

See that no information is logged to console unless option `loggingEnabled` is set to true in a navigator.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -264,6 +264,9 @@ export type NavigationUriAction = {
 export type NavigationContainerOptions = {
   // This is used to extract the path from the URI passed to the app for a deep link
   URIPrefix?: string,
+
+  // This is used to log useful debugging information about scene transitions and url handling
+  loggingEnabled: boolean,
 };
 
 export type NavigationContainerConfig = {
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -51,6 +51,8 @@ export default function createNavigationContainer<T: *>(
     };
   }
 
+  const { loggingEnabled } = containerConfig || {};
+
   class NavigationContainer extends React.Component {
     state: State;
     props: Props;
@@ -93,7 +95,9 @@ export default function createNavigationContainer<T: *>(
         Linking.addEventListener('url', this._handleOpenURL);
         Linking.getInitialURL().then((url: string) => {
           if (url) {
-            console.log('Handling URL:', url);
+            if (loggingEnabled) {
+              console.log('Handling URL:', url);
+            }
             const parsedUrl = urlToPathAndParams(url);
             if (parsedUrl) {
               const { path, params } = parsedUrl;
@@ -127,7 +131,9 @@ export default function createNavigationContainer<T: *>(
     }
 
     _handleOpenURL = ({ url }: { url: string }) => {
-      console.log('Handling URL:', url);
+      if (loggingEnabled) {
+        console.log('Handling URL:', url);
+      }
       const parsedUrl = urlToPathAndParams(url);
       if (parsedUrl) {
         const { path, params } = parsedUrl;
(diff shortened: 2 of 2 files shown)


csotiriou:
@satya164 I disagree.

Perhaps the prop isn't the most efficient solution, but when debugging, I find the logs extremely useful. It's only during production that we have a performance problem that we need to address. Therefore, I believe the flag should be present.

satya164:
When you are debugging it's not a lot of work to put logs where you want them, or put a breakpoint. You are going to do that for other parts of code anyways. Also when you are using Redux, you can use the logger middleware. I don't see the point of the prop.

mhagmajer:
@skevy, would you mind weighing in on the value of those console statements? Seems like they were introduced by https://github.com/react-community/react-navigation/commit/c37ad8a0a924d13f3897bc72fbda52aac76904b6. Do you think they're useful in general development of the library?

If they are indeed we could perhaps debate on other possible ways to guard them.

It would be good to provide users of the library with an easy option to opt-in to more logging information.

satya164:
So what's wrong with putting console logs yourself or using redux logger?

mhagmajer:
@satya164 Nothing wrong, it's just inconvenient. Also we can't expect people to use Redux.

In my humble opinion, until react-navigation becomes the official navigation solution for React Native we should do what we can to encourage people to easily debug the library and improve it. Also, I believe there is a reason why those statement were landed and not removed for 2 months.

mhagmajer:
@satya164, I'm fine with either really. If @skevy doesn't reply or some other contributor doesn't oppose by mid-week next week, I'll go ahead and apply your suggestion to remove the logs altogether.

satya164:
> it's just inconvenient

How is using `<MyNavigator onNavigationStateChange={(...args) => console.log(args)} />` or using `redux-logger` (which you probably use if you use redux) more inconvenient than `<MyNavigator loggingEnabled />`?

> we should do what we can to encourage people to easily debug the library and improve it.

Sure, but how is adding logging encourages anyone to improve the library? And people are looking into the code to add a simple console log where they need it? There are so many other places which you also want to debug, why have logs only here and not everywhere?

> I believe there is a reason why those statement were landed and not removed for 2 months.

Or the reason could be just that nobody bothered to remove them.

ericvicenti:
I think the logging should be the default behavior of the navigation state event. If you provide an empty handler, the logging will stop.

The logging was put there intentionally, and I think it helps for the average user to understand what is happening with the navigation state.

mhagmajer:
@ericvicenti can you give a code example of what needs to be done to disable the default logging?

ericvicenti:
Instead of loggingEnabled, can you add an `onNavigationState` option to the container options?

This event will fire any time the navigation state changes, or when it is initially known.

The default onNavigationState handler is the logger. This way, logging can be disabled by setting an empty function as the onNavigationState option.

satya164:
@ericvicenti 

> The logging was put there intentionally, and I think it helps for the average user to understand what is happening with the navigation state.

I think the same can be said for whole application state. I don't think navigation state is any different. As a user, I've used `redux-logger` or redux devtools in every project because of this reason.

> Instead of loggingEnabled, can you add an onNavigationState option to the container options?

Isn't it same as the `onNavigationStateChange` prop already exists? That's what I've suggested above to use for logging. Also Redux users have `redux-logger`.

> The default onNavigationState handler is the logger. This way, logging can be disabled by setting an empty function as the onNavigationState option.

I don't think we should enable logging by default. It affects performance negatively https://facebook.github.io/react-native/docs/performance.html#console-log-statements

ericvicenti:
> I think the same can be said for whole application state. I don't think navigation state is any different. As a user, I've used redux-logger or redux devtools in every project because of this reason.

Yep, but if you're using redux for your navigation state, then the container does not log.

As for performance, maybe we should disable it when `__DEV__===false`

The reason we can't use `onNavigationStateChange` is because this is a container feature so it needs to be in `containerOptions`

joenoon:
It looks like this PR solves a real and immediate problem since we are stuck with these expensive console.logs as it stands now.  Long term maybe it makes sense to have the user listen/react to changes if they want to console.log on their own, but for now I think step 1 is to let the user prevent these console.logs.

grabbou:
I am going to close this PR in favour of changes suggested by @ericvicenti and @satya164

mhagmajer:
Mike, are there any PRs with those suggested changes?

grabbou:
Yes, please see
https://github.com/react-community/react-navigation/pull/1091 which is
likely going to land later today.

PR #862, opened 2017-03-29
Fix TabViewAnimated screenProps caveat

stochris:
Accordin do the react-native-tab-view, `TabViewAnimated` is a pure component, and as such, will not be rerendered unless it's props change.
This prevents rerenders when `screenProps` on `TabNavigator` change, and as such, is a major limitation when using `TabNavigator`.
The proposed solution is to pass the `screenProps` to the `TabViewAnimated`, even if does not use this in any way.
Any change to the `screenProps` of `TabNavigator` will trigger a rerender in `TabViewAnimated`.

Allows using `screenProps` with `TabNavigator`

**Test plan (required)**

Don't know how much this could be tested, since I've only added `screenProps={this.props.screenProps}` to the `TabViewAnimated` component.

**Code formatting**

### DIFF
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -201,6 +201,7 @@ class TabView extends PureComponent<void, Props, void> {
         renderPager={this._renderPager}
         configureTransition={configureTransition}
         onRequestChangeTab={this._handlePageChanged}
+        screenProps={this.props.screenProps}
       />
     );
   }


satya164:
I'll merge this, but you should really just connect individual screens instead of passing data in `screenProps`.

stochris:
I encountered the problem when using GraphQL with the router. Normally, I'd' just skip the screenProps, but in my case, I load some data when on app, then render navigator ( and pass it the recieved data )

PR #867, opened 2017-03-30
remove workaround for screen shadow

nihgwu:
This depends on #828, see https://github.com/react-community/react-navigation/pull/828#issuecomment-290239321

@ericvicenti


ericvicenti:
I think this workaround is still needed. If not, please make a new commit on top of master, and please post a GIF of the new behavior so we can verify it looks good

nihgwu:
I will update #828 ,then this PR could be merged, so please reopen it, I will show you the difference later

PR #873, opened 2017-03-31
fix naming consistency in DrawerNavigator

dulmandakh:
Rename configuration parameters to DrawerNavigator for naming consistency and resolve mis understanding.

IIRC, both contentComponent and contentOptions applies to Drawer/Sidebar of the DrawerNavigator, but name contains **content** and it's confusing that it applies for CONTENT side of DrawerNavigator. So rename configuration for naming consistency.

contentComponent -> drawerComponent, contentOptions -> drawerOptions


satya164:
The content component is the child of the drawer component and they are not the same thing. Also this is an unnecessary breaking change.

If the docs are not clear, please send a PR to update the docs.

PR #874, opened 2017-03-31
Show icons in tab navigations in playground in android

agrcrobles:
As it is mentioned here https://github.com/react-community/react-navigation/issues/516 there is need to set `showIcon: true` for tab navigators to show the icons.

This PR fixes navigation playground on android platform.


satya164:
The example is working as intended. Android tabs with icons it not very common.

PR #889, opened 2017-04-02
Add option to back navigate to previously open tab

eduplus:
Implements https://github.com/react-community/react-navigation/issues/475

Adds optional back button behavior in TabRoutes, which remembers the previously opened tabs and will go back in reverse order.

**Test plan**

Simplest way to test is add to SimpleTabs config in Playground:

  backBehavior: 'previousRoute', 

Then select some tabs on Android and press back. It should go back in reverse order.

Added Jest tests for initialRoute and previousRoute back behaviors to verify that didn't break existing option and new one works as expected.


satya164:
Does this deduplicate tabs in the stack? doesn't seem like it does. What we ideally want is something like what instagram does - https://github.com/react-community/react-navigation/issues/678

eduplus:
No it currently doesn't. I was going by how it was implemented in Spotify. 

Hadn't noticed https://github.com/react-community/react-navigation/issues/678 which explains what is wanted nicely. I will make the appropriate changes. I agree that it is most likely better UX.

eduplus:
@satya164 This PR should cover both behavior now. I made an extra option for the deduplicate use case, since it was so close as to really only need one additional if-check. 

I think it is possible for developers want either one, depending on their use case. Precedence is there as Spotify uses the pure history approach.

satya164:
I'm noticing sure about having the pure history approach. I'm which case someone will need it? Can you give an use case where the de-duplicated history is not enough? Why would you want to force the user to go through the same tabs again and again?

src/routers/TabRouter.js
```diff
@@ -32,6 +32,8 @@ export default (
   const initialRouteIndex = order.indexOf(initialRouteName);
   const backBehavior = config.backBehavior || 'initialRoute';
   const shouldBackNavigateToInitialRoute = backBehavior === 'initialRoute';
+  const shouldBackNavigateToLastActiveTab = backBehavior === 'previousRoute';
+  const backstack = [];
```
ericvicenti: This stores the state in the router, which is not what we want!

Instead we should store the back stack as routes within the state that we return from `getStateForAction`, (which also comes in through inputState)

Please keep the routers stateless! Sometimes the router needs to get re-constructed (like if the app is refreshed), but we still don't want to loose the back stack!

eduplus: Okay, makes sense to keep router stateless. 

So just to make sure I understand, is this what you would like to have after selecting a tab:

{ 
 index: 1
 routes: [ route array as is it now ],
 backstack: [ array of routes (key and routename) of tabs previously selected ]
}

So an extra property for it in the state?

eduplus: @ericvicenti @satya164 or can someone please confirm what is required, so I don't go the wrong way on this. I am new to this library. Thank you.

eduplus: ok I did an update where backstack is in NavigationState, I have them as indexes, let me know if that is fine or if route objects are needed.

satya164:
@ericvicenti what do you think?

lprhodes:
"backstack" should probably be "backStack"?

augustoabreu:
Any news about this PR?

eduplus:
I would love to work on this and get this in, but I need some input whether this is already ok, or whether some changes need to be done.

alexkuc:
Any chance to get this merged? I have tested the new TabRouter.js on my project and navigation seems to work as expected i.e. when `backBehavior` is set to `previousRoute`, it returns to previously selected tab. Didn't run into any errors or issues whatsoever.  @eduplus @ericvicenti @skevy

hutsi:
Also want to see it merged

hoscarcito:
Is there a problem with this PR? We're waiting since June for a review
Thanks!

e2o:
Any update on this? Review is pending since june...
@kelset, can someone from the team have a look at this please?

e2o:
@eduplus I'll gladly resolve the merge conflicts if you don't have the time.
Just add me as a collaborator to your fork and I'll sort it out.

eduplus:
@e2o Added you, thanks for the help. Would be nice to finally get this in.

e2o:
There, I updated the PR to be up-to-date with the target branch. PR is ready to be reviewed. @ericvicenti @eduplus @kelset

eduplus:
@e2o all looking good to me!

e2o:
@kelset Is there any chance this can be reviewed? I'll gladly resolve the merge conflicts if I know there is gonna be follow up. No point in periodically fixing this PR if nobody from the team reviews afterwards 😕

satya164:
cc @brentvatne @ericvicenti

brentvatne:
hey @eduplus and @e2o! this is really great, thanks so much for sticking with it. I will absolutely review this, but I want us to push through the RFC before we merge in case anything else comes up in discussion: https://github.com/react-navigation/rfcs/pull/2

can anyone pick up on where the RFC left off? the template: https://github.com/react-navigation/rfcs/blob/master/0000-template.md

ericvicenti:
Lets hold off on this until the RFC lands. If anybody wants to help on this, we could use some help finalizing the spec here: https://github.com/react-navigation/rfcs/pull/2

Issue #875, opened 2017-03-31
TabRouter uses inner action before outer routeName has been found


monteithpj:
I have a case where I have a TabRouter with nested StackRouters inside and when I send an action with an inner action the inner action is unpacked and sent to the children before it's reached the target of the outer action. eg, given Navigator setup:

```
MyNav = TabNavigator({
   A: StackNavigator({
      initialScreen: Screen,
      B: StackNavigator({...})
  })
})
```
when sending action:
```
{type: NAVIGATE, routeName: "B", action: SOME_INNER_ACTION}
```

MyNav sees that its current child is a router, so it delegates the action - however it delegates the inner action, even though it's current child is A, not B (the target of the outer action).

This is problematic when B isn't the current child of A, ie A is in it's initial state. The intention is that the outer action causes A to select B as the current child and delegate, but A has only been given the inner action, thus it won't select B at all.

I propose that when delegating to the current tab, the outer action should be sent unless the current tab is itself the target of the outer action.

If people agree that this is unintended behaviour, i'm happy to stick a PR up

EDIT: Corrected failure case

kelset:
Pinging OP @monteithpj  since this issue is quite old and related to an old version of the lib. If no answer I'll close in 7 days.

monteithpj:
@kelset This is still a problem - would be great to get this looked at =)

kelset:
Hi @monteithpj, can you let me know if this PR #2674 fixed the issue? They seem to be somewhat related. (you can test it out via pointing to master, just be careful 'cause there are some minor BC)

monteithpj:
Yup, this problem is still happening

kelset:
@monteithpj we finally merged #895, can you try pointing to master to see if it's finally fixed?

monteithpj:
Yes, this is working correctly now, thanks!

PR #895, opened 2017-04-03
Only delegate inner action to current tab if the outer action matches

monteithpj:
**Motivation**
As described in [#875](https://github.com/react-community/react-navigation/issues/875) TabRouters seem to incorrectly pass down the inner action when delegating to the current tab (If delegating the action to other tabs the original action is always passed)

**Test plan**

I've added a test case to demonstrate a minimal case of the issue I was having and the change fixes this.

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -61,11 +61,9 @@ export default (
         const routes = order.map((routeName: string) => {
           const tabRouter = tabRouters[routeName];
           if (tabRouter) {
-            const childAction =
-              action.action ||
-              NavigationActions.init({
-                ...(action.params ? { params: action.params } : {}),
-              });
+            const childAction = NavigationActions.init({
+              ...(action.params ? { params: action.params } : {}),
+            });
             return {
               ...tabRouter.getStateForAction(childAction),
               key: routeName,
@@ -106,7 +104,7 @@ export default (
       const activeTabRouter = tabRouters[order[state.index]];
       if (activeTabRouter) {
         const activeTabState = activeTabRouter.getStateForAction(
-          action.action || action,
+          action,
           activeTabLastState
         );
         if (!activeTabState && inputState) {
--- a/src/routers/__tests__/TabRouter-test.js
+++ b/src/routers/__tests__/TabRouter-test.js
@@ -3,8 +3,11 @@
 
 import React from 'react';
 import TabRouter from '../TabRouter';
+import StackRouter from '../StackRouter';
 
 import NavigationActions from '../../NavigationActions';
+
+import type { NavigationRoute, NavigationState } from '../../TypeDefinition';
 
 const INIT_ACTION = { type: NavigationActions.INIT };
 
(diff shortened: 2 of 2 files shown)


monteithpj:
Any chance of some feedback on this?

kelset:
LGTM - @Ashoat or @vonovak can you guys take a quick look at it? I think we can merge.

src/routers/TabRouter.js
```diff
@@ -105,8 +105,12 @@ export default (
       const activeTabLastState = state.routes[state.index];
       const activeTabRouter = tabRouters[order[state.index]];
       if (activeTabRouter) {
+        const actionForChild =
```
Ashoat: Wait, I'm confused... why is `action.action` *ever* passed in here?

The only case where it's relevant is `NavigationActions.NAVIGATE`, and in that case it's handled separately (see `if (action.action) {` on line 157). I think the point of this block here is for allowing the current child router to handle actions that are targeted to it, like a simple `{ type: NavigationActions.NAVIGATE,  routeName: "anotherChildRoute" }`. In cases where the action is targeted at the parent router, I think the expected result should be that `activeTabState === activeTabLastState`, and the logic below in `NavigationActions.NAVIGATE` handles the action. By passing the sub-action in here we circumvent that logic entirely.

As far as I understand, the right solution should be to never use `action.action` here, but I'm not as familiar with the code here as I'd like to be. Curious to here others' opinions.

monteithpj: Interesting... yeah, that makes sense to me, but if I implement it I get another test failing. Not completely sure why - will investigate further

monteithpj:
@Ashoat Sorry it's taken me so long to look at this - I've changed it as suggested, hope this looks better!

src/routers/TabRouter.js
```diff
@@ -61,11 +61,9 @@ export default (
         const routes = order.map((routeName: string) => {
           const tabRouter = tabRouters[routeName];
           if (tabRouter) {
-            const childAction =
-              action.action ||
-              NavigationActions.init({
-                ...(action.params ? { params: action.params } : {}),
-              });
+            const childAction = NavigationActions.init({
+              ...(action.params ? { params: action.params } : {}),
```
Ashoat: I also wonder if we should ever pass `action.params` in here... we’re just supposed to be initializing a blank navigation state, right? That should have nothing to do with the action we were passed.

Ashoat: Can you actually go ahead and remove `action.params` here? (just pass an empty object)

monteithpj: I'm not totally sure why this is here, but removing params here breaks a test (TabRouter › Handles passing params to nested tabs)

Ashoat:
Got it, thanks for clarifying. I’ll try to take a look later today at that, but I’ll merge this now.

src/routers/TabRouter.js
Ashoat: I think the test case is broken. #3094

PR #900, opened 2017-04-04
Fix the ReduxExample app

cooperka:
I'm happy to make more tweaks if needed, but this works for me now on both Android and iOS.

1. For some reason `yarn.lock` had resolved to a package registry that is unable to serve HTTPS; this doesn't work at all on the latest version of Yarn
2. Add the directory name to the `examples` array in `rn-cli.config.js` to fix `@providesModule naming collision` errors (likely fixes https://github.com/react-community/react-navigation/issues/75)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,5 @@
+# Examples
+
+## Usage
+
+Please see the [Contributors Guide](https://github.com/react-community/react-navigation/blob/master/docs/guides/Contributors.md#development) for instructions on running these example apps.
--- a/examples/ReduxExample/yarn.lock
+++ b/examples/ReduxExample/yarn.lock
@@ -2,41 +2,48 @@
 # yarn lockfile v1
 
 
-abab@^1.0.0:
+abab@^1.0.3:
   version "1.0.3"
-  resolved "https://modules.verse.link/abab/-/abab-1.0.3.tgz#b81de5f7274ec4e756d797cd834f303642724e5d"
+  resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.3.tgz#b81de5f7274ec4e756d797cd834f303642724e5d"
 
 absolute-path@^0.0.0:
   version "0.0.0"
-  resolved "https://modules.verse.link/absolute-path/-/absolute-path-0.0.0.tgz#a78762fbdadfb5297be99b15d35a785b2f095bf7"
+  resolved "https://registry.yarnpkg.com/absolute-path/-/absolute-path-0.0.0.tgz#a78762fbdadfb5297be99b15d35a785b2f095bf7"
 
 accepts@~1.2.12, accepts@~1.2.13:
   version "1.2.13"
-  resolved "https://modules.verse.link/accepts/-/accepts-1.2.13.tgz#e5f1f3928c6d95fd96558c36ec3d9d0de4a6ecea"
+  resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.2.13.tgz#e5f1f3928c6d95fd96558c36ec3d9d0de4a6ecea"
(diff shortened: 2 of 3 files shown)


AlanFoster:
@cooperka I ran into some issues recently with running the redux example, and it turned out that you're supposed to run `npm run start` from the root directory of `react-navigation` in a separate process, then run your individual example app with `react-native run-android`

In the contributor's guide there's an example of how to [run the native playground example](https://reactnavigation.org/docs/guides/contributors#1.-Run-the-native-playground):

```shell
cd examples/NavigationPlayground
npm install
cd ../..
npm start

# In a seperate terminal tab:
npm run run-playground-android
npm run run-playground-ios
```

It also turns out that ` npm run run-playground-android` is just an alias for `cd examples/NavigationPlayground && react-native run-android`

Therefore to get the redux example running I just used

```
cd examples/ReduxExample
npm install
cd ../..
npm start

# In a separate terminal tab:
cd examples/ReduxExample && react-native run-android
``` 

Let me know if that works for you too, and if so - I think we should update the docs to be more obvious on how a user should be running the examples :+1:

cooperka:
@AlanFoster thanks for the info! That's all very unintuitive, the docs should definitely be made more clear. Even just a README file in /examples would help a lot.

Unfortunately I can't get your solution to work because of all the DependencyGraph errors:

```
This error is caused by a @providesModule declaration with the same name across two different files.
Error: @providesModule naming collision:
  Duplicate module name: react-native-vector-icons
  Paths: /Users/cooperka/Dev/cribspot/forks/react-navigation/node_modules/react-native/local-cli/core/__fixtures__/files/package.json collides with /Users/cooperka/Dev/cribspot/forks/react-navigation/examples/ReduxExample/node_modules/react-native/local-cli/core/__fixtures__/files/package.json
```

Regardless, the changes to url endpoints in `yarn.lock` here should probably be kept. If you can come up with a solution to the collisions, I can revert a few of these commits and just leave the yarn.lock changes. For now, running the example as a standalone app is the only way I can make it work.

AlanFoster:
@cooperka Checkout https://github.com/react-community/react-navigation/issues/75 - there was a bunch of folk having the same issue too. I was in the exact same position ~10 days ago it seems! 

I fixed the issue by nuking the `node_modules`, then following the exact install steps above and it worked.

> These always occur for me when there are nested projects that both have react-native installed in node_modules. Seems to be a common problem.

TIL - I wonder if there's a fix for this 🤔

eqyiel:
The `examples` array in `rn-cli.config.js` (in the root) needs `ReduxExample` added as well in order to fix the naming collision errors.

cooperka:
@eqyiel thank you very much! Such an easy fix. I had no idea such a file existed, I'll have to do some more research on it. PR has been updated, hopefully it's mergeable now :)

AlanFoster:
Nice! Is it worth adding some details on a readme file to let developers know how to run the examples, even if it's just a link to the contributors guide

And maybe even a note that if you're contributing an example to update that extra file! :+1:

eqyiel:
@cooperka no worries, for reference I noticed because of this issue and the associated Stack Overflow discussion:
https://github.com/facebook/react-native/issues/12078#issuecomment-276490953

cooperka:
@eqyiel: thanks for the link!

@AlanFoster: added basic README, that should suffice.

ericvicenti:
Thank you! If you can improve the contributors doc or add info elsewhere about running the examples, please do!

cooperka:
@ericvicenti I typically check for READMEs before running the examples, so the doc added in this PR is perfect for me, but there may be other places we can improve docs. Any specific suggestions anyone? I'm happy to add another PR.

Issue #246, opened 2017-02-09
When using DrawerNavigator and connecting to redux the store gets cleared when the drawer opens


rmevans9:
Today I was starting to put together navigation in a app I am beginning working on and wanted to use a drawer. In addition I want to keep the nav state in my redux store. After setting up what I think should be a working drawer navigation I noticed that when I slide open the drawer menu the reducer ends up returning null for its state which then exceptions out complaining that it can't find index on the navigation state object (which is true because the navigation state object is now null). I was able to put together a simple example here:

https://github.com/rmevans9/react-navigation-drawer-redux-test/blob/master/index.ios.js

If you run this and just slide out the drawer you will get a RSOD complaining about the index. I did a little bit of digging in the code and found that the null is being returned from here:

https://github.com/react-community/react-navigation/blob/master/src/routers/TabRouter.js#L158

I am still trying to learn the ins and outs of how React Navigation internals work so I am not sure if a null return is valid (or why it would be valid... typically if you don't want to change state then you return the current state). I am planning to dig more into this tomorrow but wanted to post this up here and see if anyone could either point out where I messed up in my example or if this is actually a bug in React Navigation.

rmevans9:
So digging into why this issue shows in a redux setup but does not if using the internal state management I came across this bit of code:

```
dispatch = (action: NavigationAction) => {
    const { state } = this;
    if (!this._isStateful()) {
        return false;
    }
    const nav = Component.router.getStateForAction(action, state.nav);

    if (nav && nav !== state.nav) {
        if (console.group) {
            console.group('Navigation Dispatch: ');
            console.log('Action: ', action);
            console.log('New State: ', nav);
            console.log('Last State: ', state.nav);
            console.groupEnd();
        } else {
            console.log('Navigation Dispatch: ', { action, newState: nav, lastState: state.nav });
        }
        this.setState({ nav });
        return true;
    }
    
    return false;
};
```
from: https://github.com/react-community/react-navigation/blob/master/src/createNavigationContainer.js#L132

Looking at how it only updated the actual state if the `getStateForAction` returned a value I tried that in a redux reducer that basically just calls the `getStateForAction` method and only returns the updated state if a state is returned:

```
function navWorkaroundReducer(action, state) {
    const newState = {...}.getStateForAction(action, state);

    if (newState) {
        return newState;
    }

    return state;
}
```

And it worked. So the way I see it is there are two options:

1. Update the documentation for redux integration calling this out
[...]

PR #910, opened 2017-04-04
Docs: Update redux-integration guide #246

half-shell:
This is the smallest pull-request just to reflect the solution found in the issue.

**Test plan (required)**

**Code formatting**

### DIFF
--- a/docs/guides/Redux-Integration.md
+++ b/docs/guides/Redux-Integration.md
@@ -11,7 +11,7 @@ const AppNavigator = StackNavigator(AppRouteConfigs);
 
 const navReducer = (state, action) => {
   const newState = AppNavigator.router.getStateForAction(action, state);
-  return newState || state;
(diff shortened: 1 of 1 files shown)


peterp:
@half-shell `a || b` and `a ? a : b` are technically identical as far as I can tell? Did this resolve an issue you had whilst following the redux-guide?

half-shell:
@peterp Oh you're right... 
I thought it did solve the issue for me, but I guess I got lost changing react-native version or whatnot because I just tried again and they both work. I actually didn't know the `a || b` syntax and didn't bother to look since it seemed to work for me.
I'm really sorry for this useless PR, I'll think twice about it next time.

peterp:
@half-shell It's not a problem, we're all in this software thing together! Nothing is useless, please keep contributing in any way that you can.

PR #919, opened 2017-04-05
Clamps interpolation of x and y values on Card animations

drop-dan:
Fixes: https://github.com/react-community/react-navigation/issues/559

By clamping the extrapolation, output values will never go below `min` or above `max`, which is the intended effect.

I wondered if this change was necessary, as it seems like there may have been a pseudo clamp with the input and output ranges.  Is there a reason that method was chosen instead of using extrapolate clamp?

I tested this after making the change, and I was no longer able to reproduce the behaviour described in issue 559.  However, I am unable to reproduce this issue on master currently.  Either way, a clamp seems desirable.

Thanks for the fantastic work on this repo!


drop-dan:
Misunderstood the implementation here, move along

PR #926, opened 2017-04-05
Add pressColor to StackNavigator header options

jamieparkinson:
This resolves https://github.com/react-community/react-navigation/issues/925, adding the option to change the back button's ripple colour. For example:

screenshot_20170405-124107_01

All tests and flow checks are passing. Test plan:

- Don't set `pressColor` in `navigationOptions.header` on a `StackNavigator`. Everything is the same as before.
- Set `pressColor` to anything you want as above. Ripple colour has changed.

Do let me know if there's anything you'd prefer to be done differently :slightly_smiling_face:

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -126,8 +126,8 @@ All `navigationOptions` for the `StackNavigator`:
   - `style` - Style object for the header
   - `titleStyle` - Style object for the title component
   - `tintColor` - Tint color for the header
-  - `pressColor` - Color for material ripple (Android >= 5.0 only)
-- `cardStack` - a config object for the card stack:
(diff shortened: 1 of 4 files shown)


jamieparkinson:
@ericvicenti Agreed, although worth remembering that this means `pressColor` will be inconsistent between StackNavigator and TabNavigator. Will do this later today.

ericvicenti:
Is `pressColor` on TabNavigator also used on Android only? If so, then maybe we should rename it

jamieparkinson:
Yeah, it is Android only - I copied the explanation of 'my' `pressColor` in the docs from TabNavigator. That said, it's only an option for `TabBarTop`, so I guess some level of Android-specificity is implied already

ericvicenti:
Looks great, thanks for the contribution!

jamieparkinson:
@ericvicenti thanks for the great library!

PR #928, opened 2017-04-05
[Playgrond] Add StacksOverTabs example

cooperka:
`StacksOverTabs` is exactly the same as `StacksInTabs`, except that when navigating, the routes get pushed on top of the tabs instead of inside of them. This is a common paradigm.

Also fixes `StacksInTabs` so it navigates to `SettingsTab` instead of pushing a new `Settings` screen to the stack.


ericvicenti:
Looks good, thanks for the contribution!

PR #929, opened 2017-04-05
Fix setParams with nested routers

taiki-t:
**motivation** 

setParams with nested routers doesn't work well with https://github.com/react-community/react-navigation/pull/789 so this PR fixes it.

**Test plan (required)**

Add tests


ericvicenti:
This test makes sense I think, and I'm not sure why the `delete` is in there. Merging now!

PR #930, opened 2017-04-05
[Playground] Add TabsInDrawer example

cooperka:
This example puts `SimpleTabs` and `StacksOverTabs` into a drawer layout with custom labels and icons. Shows how simple it is to nest the two like this :)

This is how my app works, and I needed to write this example anyway in order to test some integration logic with redux.

(should be merged after https://github.com/react-community/react-navigation/pull/928 since they're dependent)

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -15,6 +15,7 @@ import { StackNavigator } from 'react-navigation';
 import Banner from './Banner';
 import CustomTabs from './CustomTabs';
 import Drawer from './Drawer';
+import TabsInDrawer from './TabsInDrawer';
 import ModalStack from './ModalStack';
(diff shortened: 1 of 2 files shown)


kevinold:
@cooperka This looks awesome and is a current need of mine in an app I'm building.  

Do you have a branch that has this and #928 in the same branch as I'd love to test it?

I'm currently solving this with a `customComponent` for the DrawerContents and it does work, but having issues with params being passed between tabs when navigating from one to the other.

cooperka:
@kevinold Here's a branch that has both examples combined: https://github.com/cribspot/react-navigation/commits/redux-demo. It also has some changes to the ReduxExample app but you can ignore those.

ericvicenti:
Nice, looks good to me!

Issue #175, opened 2017-02-04
Support customizable, complex transitions that animate individual views on a screen


lintonye:
Update March 8, 2017: This RFC is based on an experimental implementation at https://github.com/lintonye/react-navigation/compare/master...multi-transitions

## Motivation
So far we only support simple transitions that animate the entire screen. It's desirable to allow highly customizable transitions that animate individual views on the screen. The [shared elements transition](www.reactnativediary.com/2017/01/23/react-navigation-shared-element-transition-1.html) is an example of this. It'd be fun if we could easily implement [this](https://material.uplabs.com/posts/introductory-app-animation) or [this](https://capptivate.co/2014/11/23/goji-2/).

The last two examples are perhaps difficult to implement even in native code where the APIs are not always straightforward to use. It'd be great if we could have a declarative API in RN to simplify that!

## Proposed API
I'm hoping to minimize the impact on the code unrelated to transitions. We should be able to code our components as usual, and add some special markups to pick out the views we want to animate.

## A simple example
```jsx
const sharedImages = Transitions.sharedElement(/image-.+/);                    // ==> 1
const crossFadeScenes = Transitions.crossFade(/\$scene-.+/); 

class PhotoGrid extends Component {
  static navigationOptions: {
    cardStack: {
      transitions: [                                                           // ==> 2
        { 
          to: 'PhotoDetail', 
          transition: sequence(sharedImages(0.8), crossFadeScenes(0.2)),       // ==> 3
          config: { duration: 580 },
        },
      ]
    }
  }
  ....
  renderCell({url, title}) {
    return (
      <View>
        <Transition.Image id={`image-${url}`} source={{uri: url}}/>           // ==> 4
        <Text>{title}</Text>
      </View>
    );
  }
}

const PhotoDetail = ({photo}) => (
  <View>
    <Transition.Image id={`image-${photo.url}`} source={{uri: photo.url}}/>
    <Text>{photo.title}</Text>
    <Text>{photo.description}</Text>
  </View>
);
PhotoDetail.navigationOptions = {
  cardStack: {
    transitions: [
      { 
        to: 'PhotoGrid', 
        transition: together(crossFadeScenes(0.2), sharedImages(1))
        config: { duration: 580 },
      },
    ]
  }
};

```

### 1. Apply a regex to "bind" a transition

```jsx
const sharedImages = Transitions.sharedElement(/image-.+/);
```

[...]

PR #941, opened 2017-04-06
Shared element transition

lintonye:
## Motivation
See #175, this PR is the first part: built-in support for shared element transition. The second part would be an API for custom transitions, which would come after the implementation of shared element transition becomes solid.

Here are some videos: [on Android (Nexus 5X, Nougat)](https://youtu.be/0gadt95GW54), [on iOS simulator](https://youtu.be/jvWhUNFyqvo).

## Usage Example
```jsx
import { Transition } from 'react-navigation';
// PhotoGrid.js
<Transition.SharedElement.Image id='image1' ... />
// PhotoDetail.js
<Transition.SharedElement.Image id='image1' ... />
```

If the "from" and "to" routes contain elements with matching ids, shared element transition will be played. The actual animation is a combination of cross fade and shared element transition.

Otherwise, the platform default transition (as implemented in `CardStyleInterpolator`) will be used. 

Similar to Animated, built-in components include `Transition.SharedElement.Image`, `Transition.SharedElement.View` and `Transition.SharedElement.Text`. `Transition.SharedElement.createComponent()` can be used to create a shared-element component from another component.

## Quick guide to the code
This PR includes changes in the following areas:

- `CardStack` is slightly modified to wrap the original `CardStack` component with a HOC `withTransition`. Code related to transitions is moved to `withTransition.js`.
- `Card` is modified to render to `<Transition.View>` instead of `<Animated.View>`.  A couple of context fields are added to propagate route and transition information to its descendants.
- `src/views/Transition`: This is the "meat", which includes main classes and functions that add custom transition support.  
- `src/views/__tests__`: Added some unit tests
- `examples/NavigationPlayground`: An example app is added here. All the images are from unsplash.com which should not introduce any license issues.

## Test plan
- Test on both platforms with and without `Transition.SharedElement` items
- Test share element transition using different components, including `Image`, `View`, `Text` and custom components
- Verify that the platform default transition is used when there are no shared elements with matching ids
- Perform normal user interaction to the interface, such as scrolling, and verify that the shared element transition behaves correctly
- Test the performance impact on the UI after adding shared elements

## Known issues
- Animating fontSize of `<Text>` is implemented but not working yet.
- The frame rate on iOS seems to be lower than on Android (Nexus 5X Nougat).


examples/NavigationPlayground/js/Transitions/SharedElements/PhotoDetail.js
```diff
@@ -0,0 +1,101 @@
...
+      <View style={styles.titleContainer} onLayout={this._onLayout}>
+        <Transition.SharedElement.Text id={`title-${url}`}
+          style={[styles.text, styles.title]}>{title}</Transition.SharedElement.Text>
+      </View>
+      <Text style={[styles.text]}>{description}</Text>
+    </View>
+  );
+  return (
+    <View>
+      <ListView dataSource={dsComments.cloneWithRows(comments)}
```
browniefed: Maybe use FlatList/SectionList since it's the future?

src/views/Transition/withTransition.js
```diff
@@ -0,0 +1,542 @@
...
+      // Force the component to update when moving to a new scene and the new scene is laid out.
+      // This is necessary since we need to wait for the transition items on the 
+      // new scene to register themselves before creating transitions. Otherwise
+      // the resulting transition will be incomplete/incorrect.
+      //
+      // Once the scene is laid out, we can assume the transition items on it have
+      // already registered.
+
+      const onLayout = async () => {
+        const toRouteName = that._toRoute && that._toRoute.routeName;
```
ericvicenti: why do you need `that` here? Shouldn't `this` work?

src/views/CardStack.js
```diff
@@ -571,6 +523,8 @@ class CardStack extends Component<DefaultProps, Props, void> {
     const SceneComponent = this.props.router.getComponentForRouteName(
       props.scene.route.routeName
     );
+    const extraProps = this.props.createExtraSceneProps &&
```
ericvicenti: I think we should avoid calling this callback from the render path. Instead I think it belongs in componentDidUpdate or componentWillUpdate. That way, we can clear old scenes that are gone

lintonye: This callback requires the `NavigationSceneRendererProps` of the current scene that is being rendered. The props is used to interpolate styles. Since `NavigationSceneRenderProps` is only available in the `render` callback of the Transitioner, I'm not sure how to move that to componentDidUpdate or componentWillUpdate?

What are the downsides of using it in the render path? Performance issue?

lintonye:
@browniefed @ericvicenti Thanks for the feedback. I'll fix those flow errors and ListView in a few days.

A few things that I'd appreciate further clarification:

> For example, do we really need 'compositions' to implement the shared element functionality? 
> I see that there is additional logic to coordinate the CrossFade

CrossFade is needed because simply doing shared element transition and then immediately switching to the new scene doesn't look good IMHO. I'll upload a video for your comparison later.

This PR is based on my previous work on custom transitions. Yes it contains extra features, i.e. the support for custom transitions. This is the easiest way that I can see compared to trying to extract shared-element only code. After all, most of the code in this implementation is required for shared-element transition, such as creating an overlay, cloning items, and measuring view bounds. I just generalized them and made them usable by other transition types. If we have any plan to support custom transitions, trying to make the code specific to shared element seems counterproductive to me.

> Also it seems there is more indirection than needed with extra classes like TransitionItems.

How so? Would renaming it to `SharedElements` or something address the indirection issue? Or did you mean something else?

>  I think the existing CardStack transition should work with shared elements.

Did you mean the `CardStyleInterpolator` etc? This PR actually uses those classes to create default platform transitions. It's more of a "shell" of existing transition though since we need a lot more logic to implement shared element transition. Does this make sense?

> lets avoid using lodash

I could replace lodash. I used it because it's convenient and it's in the dependency list of react-native: https://github.com/facebook/react-native/blob/master/package.json#L178. Will it eventually be removed there?

ericvicenti:
So I think that a more advanced transition API would be nice, but we seem to have a long way to go before the API is decided on. If we try to implement it before the API is fully designed, we are likely to wind up with a very confusing implementation.

> If we have any plan to support custom transitions, trying to make the code specific to shared element seems counterproductive to me.

I am concerned by adding all this extra code for generic custom transitions, when we aren't even sure what that API will look like. Instead, maybe this PR could focus on adding shared transitions, while leaving the card transition intact? So that the card will still come from the right or bottom, but the shared elements will float to the destination location.

> Did you mean the CardStyleInterpolator etc? This PR actually uses those classes to create default platform transitions. It's more of a "container" of existing transition though since we need a lot more logic to implement shared element transition. Does this make sense?

I'm not sure exactly why the card style needs to change at all in this diff. We should just be adding shared elements.

> lodash

Ah, the lodash usage seems OK. I didn't notice that RN already uses it. Generally it is best to use language features when possible, but this usage actually seems fine.

> Also it seems there is more indirection than needed with extra classes like TransitionItems.

My main question here is: why do we need a full class for `TransitionItems`? Couldn't this data be represented with a bit of state in an existing module?

lintonye:
To everyone who's watching this PR, as discussed with @ericvicenti, we've decided to do another round of refinement of the custom transitions API, before moving on with shared element transition. I'll post updates to the proposal on #175. Please stay tuned!

tlvenn:
Thanks for the update and your hard work on this @lintonye ! Much appreciated.

chrsk:
Are you currently working on the open tasks? Do you need any help?

nonameolsson:
I would love to see this one merged! What's missing to get it merged?

JCMais:
Can we please get an update on the status of this PR?

ronal2do:
Any update?

nonameolsson:
What is the state on this?

behzad888:
What is the state on this?

brentvatne:
this is an incredible effort! it seems like now it's a bit out of date -- @lintonye if you'd like to pick up where you left off here, @ericvicenti and I are working on react-navigation close to full-time now and we can support you in this with quick reviews and discussions. let's move over to https://github.com/react-navigation/rfcs/issues/17 and turn this into a RFC, then circle back to a PR once that's ready.

sorry to everyone who is waiting for this, feel free to weigh in on the rfc thread mentioned above with your ideas around this API, or to vote on https://react-navigation.canny.io/feature-requests/p/shared-element-transitions if this is important to you.

PR #944, opened 2017-04-07
fix - redux example

gsaandy:
fixing redux example


ericvicenti:
Whoops, I noticed #945 first so I landed that one

PR #945, opened 2017-04-07
Fix the ReduxExample app (2)

cooperka:
The [awesome refactor](https://github.com/react-community/react-navigation/pull/819) that was done recently (thanks so much for that!) had a few import/export bugs preventing it from running that are fixed here.

I also disabled persistence to AsyncStorage; though it's a cool feature, it isn't necessary to demonstrate how to use react-navigation and may cause more confusion than it's worth. For example, if you change the name of a route and then refresh the app, everything will break because it can't find the old route that had been persisted, even though your code is perfectly fine.

cc @samijaber @ericvicenti

### DIFF
--- a/examples/ReduxExample/index.js
+++ b/examples/ReduxExample/index.js
@@ -3,20 +3,15 @@
  */
 
 import React from 'react';
-import { AppRegistry, AsyncStorage } from 'react-native';
+import { AppRegistry } from 'react-native';
 import { Provider } from 'react-redux';
 import { createStore } from 'redux';
-import { persistStore, autoRehydrate } from 'redux-persist';
 
 import AppReducer from './src/reducers';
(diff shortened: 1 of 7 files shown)


samijaber:
Agh, thought I was thorough and tested the example before making the PR. Thanks for the fix!

cooperka:
No worries mate, simple fix :)

cooperka:
Looks like @gsaandy beat me by 40 minutes! https://github.com/react-community/react-navigation/pull/944

I can rebase if that gets merged first; it would still be nice to remove AsyncStorage.

ericvicenti:
And yep, I think it makes sense to remove AsyncStorage.

Eventually we can add persistence features to NavigationContainer.

Issue #717, opened 2017-03-16
Hide item from Tabbar


aleks63:
How to hide an item from tabbar?

ericvicenti:
We should add a `showTab` option to TabNavigator: https://reactnavigation.org/docs/navigators/tab

ferrannp:
@aleks63 do you want to hide/show it dynamically? Because the problem with `static navigationOptions` is that once is set up, it cannot dynamically change so even if we add `showTab`, that value will be constant.

satya164:
You can wrap the default tab bar inside another component and filter the routes from `navigation.state.routes`

bmcmillan:
Like @satya164 said, to dynamically add options to my DrawerNavigator, I wrapped it in a component, and construct a new DrawerNavigator in the render() method, passing in routes as the first parameter from that component's state.

bmcmillan:
I'm doing this with DrawerNavigator, which should be pretty much the same thing:

`
class MainNavigatorContainer extends Component {
    state: State = {
      Routes: {    
      }
    };

    componentWillMount() {
        User.getNavOptions().then((options) => {          
            this.setState({
               Routes: {
                 ... this.state.Routes, 
                 Home: { screen: Home }                 
               }
             });
            if (options.foo != undefined) {
              this.setState({
                Routes: { ... this.state.Routes, Foo: { screen: Foo }} 
              });              
            }
            if (options.bar != undefined) {
              this.setState({
                Routes: { ... this.state.Routes, Bar: { screen: Bar }}
              });
            }
            this.setState({
              OptionsLoaded: true,
              Routes: { ... this.state.Routes, SignOut: { screen: SignOut }}
            });
        }).catch((err) => {
            console.warn(err);
        })
    }

    render() {           
        if (this.state.Routes == undefined || this.state.Routes.Home == undefined) 
        {
           return (<View></View>)
        } 
        else 
        {
          var MainNavigator = DrawerNavigator(this.state.Routes, {
              contentComponent: Drawer,
              drawerWidth: 250,
              initialRouteName: 'Home'              
            });     
          return (          
              <MainNavigator screenProps={{ rootNavigation: this.props.navigation }} />
          )
        }
    }
}

const AppNavigator = StackNavigator({
  Login: { screen: Login },  
  HomeMain: { screen: ({ navigation }) => <MainNavigatorContainer navigation={navigation} /> } 
}, {
  headerMode: "none"
});
`

[...]

PR #950, opened 2017-04-07
[WIP] Feature show tab

markusguenther:
With the tabBar option, showTab people can have hidden tabs in the tab navigation.

**Test plan (required)**

Actually, I need to fix the flow errors and add tests.

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -41,6 +41,7 @@ type Props = {
   style?: Style;
   labelStyle?: Style;
   showIcon: boolean;
+  showTab: boolean;
 };
 
 export default class TabBarBottom extends PureComponent<DefaultProps, Props, void> {
@@ -53,6 +54,7 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
     inactiveBackgroundColor: 'transparent',
     showLabel: true,
     showIcon: true,
+    showTab: true,
   };
 
   props: Props;
@@ -138,6 +140,7 @@ export default class TabBarBottom extends PureComponent<DefaultProps, Props, voi
         {navigationState.routes.map((route: NavigationRoute, index: number) => {
           const focused = index === navigationState.index;
           const scene = { route, index, focused };
+          const shouldRenderTab = this.props.shouldRenderTab(scene);
           const outputRange = inputRange.map((inputIndex: number) =>
             (inputIndex === index ? activeBackgroundColor : inactiveBackgroundColor)
           );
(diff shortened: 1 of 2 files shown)


markusguenther:
@satya164 I am still new to react-native and the whole ecosystem. I found the issue #717 and needed this for my app.

So I planned to implement this and this was the first try. I actually have in my app a nested Navigation (StackNavigator with one nested tabNavigator) and I used this adjustment to hide a tab and navigate directly to the hidden tab.

So this works in the bottom navigation. I used bottom navigation tabs on Android and iOS. Need to test this also in the top tabbed navigation on android. But if it is not the wish of the project to hide the tabs. I can stop here the development process.

I know that my code is not the perfect solution but the issue was tagged as first task and I thought it would be a nice try.

brentvatne:
this is a pretty desired feature! https://react-navigation.canny.io/feature-requests/p/hiding-tab-from-the-tabbar

alas I think this PR is pretty out of date, if you'd like to update and re-submit I'd be happy to have a look. should be easier now that we removed flow -- you only need to update the flow interface file in the `/flow` directory

satya164:
@brentvatne do people really want to hide a tab item from the tab bar or they want to dynamically remove a tab? it's impossible to know without knowing the use case.

regarding the ability to dynamically remove a screen, people have wanted dynamic tabs in the past and the design of react navigation doesn't really address that use case. a `hiddenTabs: ['SomeTab']` feels like a hacky workaround.

regarding hiding a tab item from the tab bar, but keeping the route rendered (which what this PR did), I don't get the use case. and it'll be impossible to have with animated/sliding tabs.

markusguenther:
I had a project in the past where I used only a Tab Navigation and needed one Scene to be there and accessible in the navigation. But it was not a visible Tab Item. So for me it was the easiest way to use a hidden Tab in the single navigation.

This scene was always invisible in the navigation. It was only a view for some special information.
Don't remember the details anymore.

satya164:
Is there a reason that screen couldn't be a screen in a child stack navigator?

rimzici:
Hi, This feature seem to be useful. And a use case could be, having three tabs and each tab being a source/channel of news(sports/politics/tech). And a settings screen in which user can choose to show/hide certain channel(tab), could be like unsubscribing, and the tab would need to show/hide on navigating back to main tabs from settings screen, according to visibility set in there.

It is not based on user privileges or anything, which can be handled in better way by having different routeConfigs for different privileges.

Do agree with @satya164  on issues, if what this PR does is hiding a tab item keeping the route rendered.

I tried filtering out route from the navigation state of 'tabBarComponent', Not fully sure though, that I have implemented the right way suggested by @satya164 

There are some issue that I face in the implementation.
1. will need do disable swipe , to prevent showing hidden tab.
2. when the hidden tab index is 0, the whole route is messed up.

Here is the [sample](https://github.com/rimzici/navigation/tree/show-hide-tabs)

An option to set screens/tabs dynamically would be helpful, something like an argument for `RouteConfigs` in 
```
TabNavigator(RouteConfigs, TabNavigatorConfig)
```
using which the set of screens can be controlled.

Pardon me if this was the wrong place to ask this, was just trying to emphasize the need of this feature.
Thanks.

PR #965, opened 2017-04-08
add iconStyle prop to tabBarOptions for TabBarTop

navid94:

I created a tab bar using TabNavigator and I wanted to place icons instead of labels.
This is my TabNavigator configuration:

```
var MainTabNavigator = TabNavigator({
   JobList: {
        screen: JobList
    },
    SearchPieces: {
        screen: SearchPieces
    },
    SideMenu: {
        screen: SideMenu
    }
}, {
    lazyLoad: true,
    animationEnabled: false,
    tabBarOptions: {
        activeTintColor: '#3b5998',
        inactiveTintColor: '#cccccc',
        showIcon: true,
        showLabel: false,
        style: {
            backgroundColor: 'white'
        },
        pressColor: '#d5dcea',
        indicatorStyle: {
            backgroundColor: 'white'
        }
    }
});
```

and this is one of the tabs with the icon that gives me problems:

```
static navigationOptions = {
        header: {
            visible: false
        },
        tabBar: {
            icon: ({ tintColor }) => {
                return (
                    <FontAwesome
                        name='cubes'
                        size={30}
                        color={ tintColor }
                    />
                );
            }
        }
    };
```

The problem is that the icon is being rendered as clipped. So I added a prop for styling the TabBarIcon component. The behaviour is the same as labelStyle prop.

**Test plan (required)**

I ran the following commands to test the code:
```
npm run test
npm run flow
```

They both gave me no errors.
I also tested the code using the NavigationPlayground example and it works.

**Code formatting**

### DIFF
--- a/docs/api/navigators/TabNavigator.md
+++ b/docs/api/navigators/TabNavigator.md
@@ -136,6 +136,7 @@ tabBarOptions: {
 - `tabStyle` - style object for the tab
 - `indicatorStyle` - style object for the tab indicator (line at the bottom of the tab)
 - `labelStyle` - style object for the tab label
+- `iconStyle` - style object for the tab icon
 - `style` - style object for the tab bar
 
 Example:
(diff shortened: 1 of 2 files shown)


navid94:
I know it's a duplicate of #677 but that pull request it's 25 days old and got some flow errors.

navid94:
Done, i have updated the docs.

PR #969, opened 2017-04-09
Add support for passing params on navigate to tabs

mgtitimoli:
- Closes: #80 
- Closes: #134 

(for more info see [this comment](https://github.com/react-community/react-navigation/issues/80#issuecomment-292352817) I left in the issue)

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -136,12 +136,27 @@ export default (
           }
           return false;
         });
-        if (didNavigate && action.action) {
+        if (didNavigate) {
+          const childState = state.routes[activeTabIndex];
+          let newChildState;
+
           const tabRouter = tabRouters[action.routeName];
-          const newChildState = tabRouter
-            ? tabRouter.getStateForAction(action.action, state.routes[activeTabIndex])
-            : null;
-          if (newChildState && newChildState !== state.routes[activeTabIndex]) {
+
+          if (action.action) {
+            newChildState = tabRouter
+              ? tabRouter.getStateForAction(action.action, childState)
+              : null;
+          } else if (!tabRouter && action.params) {
+            newChildState = {
+              ...childState,
+              params: {
+                ...(childState.params || {}),
+                ...action.params,
+              },
+            };
+          }
+
+          if (newChildState && newChildState !== childState) {
             const routes = [...state.routes];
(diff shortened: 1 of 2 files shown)


satya164:
Can you describe the API changes a bit more? For example, what was the behaviour earlier and what is the behaviour now.

I'm assuming it allows passing params to another tab in the same tab router with `navigate('Tab2', params)`. I see few issues with the approach,

Unlike stack navigator, screens in tab navigator are not rendered when you navigate to them, they are rendered in advance when the tab navigator gets rendered. Also you can navigate to other screens by swipe, or pressing the icon/label on the tab bar, and you cannot pass params this way. So the params need to be available when all the screens render. 

You should be able to do `navigate('TabsScreen', params)`, and those params will be available to all screens.

To change the params after rendering, as you might need sometimes, you should be able to use `updateParams(params)` to update the params.

mgtitimoli:
Hi @satya164, let me go through your questions one by one:

> Can you describe the API changes a bit more? For example, what was the behaviour earlier and what is the behaviour now.

Sure:
- **earlier**: there was no way to pass params to tabs, that this not only included TabNavigator, but also DrawerNavigator as it is made by TabRouters
- **now**: you can pass params to tabs  

> Unlike stack navigator, screens in tab navigator are not rendered when you navigate to them, they are rendered in advance when the tab navigator gets rendered. Also you can navigate to other screens by swipe, or pressing the icon/label on the tab bar, and you cannot pass params this way. So the params need to be available when all the screens render.

This can easily be fixed by adding an onNavigate handler to the route definition so you could receive a navigation event, that would let you do the following that currently none of this is supported
- cancel navigation
- transform params (if there is no params and you need some, here is where you can set the defaults)

I find it crucial to widely support the ability of passing params no matter which kind of navigator/router are you using, as otherwise forces us to share information by using an external module with state, that as you know this should be avoided as much as possible, since complicates testing, and other stuff.

I also thing that in the future we should try to introduce a change that let us separate route params and route state as I described [here](https://github.com/react-community/react-navigation/issues/225) since IMHO this concepts are different, but this is a subject of a different conversation.

ericvicenti:
Ok, this looks like a well-documented fix to a pretty common problem. Thanks for the contribution!

nonameolsson:
Awesome! Is this somewhere in the documentation? Couldn't find it.

PR #970, opened 2017-04-09
Fix flowtype errors on React Native >= 0.43

joemckie:
When using `react-native>=0.43` (`flow ^0.40.0`), flow errors have been introduced, which breaks build steps. I've bumped the version to match what React Native installs by default and fixed these errors.

I'm unsure as to whether the flow errors are actually a result of a flow bug, however. There is [this](https://github.com/facebook/flow/issues/3493) issue on the flow repository, but unfortunately no one has chimed in to say whether it's intended or not. In any case, the error was stemming from two different types of React elements being conditionally returned.

Maybe the deps version bump would be better off as a separate pull request entirely. Happy to tweak things if necessary.


DanielMSchmidt:
@satya164 Could you do me a favor and have a look at this? This currently blocks our team :/

grabbou:
Closed in https://github.com/react-community/react-navigation/commit/2bbfc03cd0669e01cfd8f7f2d22ccae23e4dfb88

PR #971, opened 2017-04-09
use gestureResponseDistance in CardStack internal panresponder

vijairamcharan:
The CardStack has an internal PanResponder that was not taking into account the passed in gestureResponseDistance property.

I basically copied the code from CardStackPanResponder into the CardStack.

**Usage example**
When creating a StackNavigator like this:

```
const MainNavigator = StackNavigator({
  Home: { screen: HomeScreen },
  Chat: { screen: ChatScreen}
});
```

And then using it like this:

```
const app = () => {
  return (
    <MainNavigator
      gestureResponseDistance={100}
    />
  );
}
```

will result in an increased drag range (from 35 to 100) for the pop gesture.

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -378,11 +378,17 @@ class CardStack extends Component<DefaultProps, Props, void> {
           : layout.width.__getValue();
         const axisHasBeenMeasured = !! axisLength;
 
-        // Measure the distance from the touch to the edge of the screen
-        const screenEdgeDistance = currentDragPosition - currentDragDistance;
-        // GESTURE_RESPONSE_DISTANCE is about 30 or 35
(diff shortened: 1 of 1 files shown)


vijairamcharan:
@ericvicenti i just went with the current code in the repo where

1. The StackNavigatorConfig does not expose a gestureResponseDistance field
2. The CardStack *does* already expose the gestureResponseDistance field

It was also internally automatically used I guess by the CardStackPanResponder, however a recent commit https://github.com/react-community/react-navigation/commit/11cab8eab62b6ca1c2a0445f30a9d0dc75458e2e moved away from the CardStackPanResponder however did not include a number of features like the gestureResponseDistance.

See also issue https://github.com/react-community/react-navigation/issues/818 which talks about an omission in the mentioned commit.

Would you rather prefer an issue to discuss?

PR #984, opened 2017-04-10
Introducing flat options

### LINKED ISSUE #48: router.getScreenConfig is broken
Currently the second argument received by a navigation options item is the previous config returned from a parent navigation options, and the options passed to `getScreenConfig`.

IMO it's a confusing behaviour, because fundamentally they are different types of objects, one is the configuration for navigator, other is arbitrary options the navigator can pass to the configuration function. For example, the tab navigator can provide a `focused` argument for the icon, which doesn't make sense in a config.

The way this works is, only the global navigator config receives the options passed to the `getScreenConfig`, and it has to explicitly merge it with, say, header config and return it. Which means if you have a static header configuration for screens, and then add a header configuration at the global level to be used as default, it breaks the screen configuration because it doesn't receive the options anymore and you've to explicitly merge the options in the global navigator config. Same for the route navigation config, which will break if you add a screen config.

image

Personally I think the confusing behaviour and the learning curve associated with it is not worth getting rid of one extra function.

global config - the `navigationOptions` passed in the second argument to the navigator
screen config - the static `navigationOptions` in the screens
route config - the `navigationOptions` specified for individual routes in the first argument to the navigator

Flow types for both,

```js
type NavigationOptionItem<T> =
  | T
  | (navigation: NavigationScreenProp<*,*>, T) => T

type IconOptions = {
  focused: boolean;
  tintColor: boolean;
}
```

These are the current api and the proposals we discussed:

## Current API

Options is the result of previous header function/previous header object

```js
header: (navigation, options) => ({
  ...options,
  icon: ({ focused, tintColor }) => <Icon name={'settings' + focused ? '' : '-outline'} color={tintColor} />
});
```

Flow type,

```js
type HeaderOptionItem = NavigationOptionItem<{
  title: string;
  icon: IconOptions => React.Element<*>;
}>
```

How it is used,

```js
const { icon } = router.getScreenConfig(...);
return icon({ focused, tintColor });
```

### Pros

- `router.getScreenConfig` doesn't deal with options passed to icon, so it's pretty generic, navigators can do whatever they want
- the icon function can be declared inline, so much simpler to use
- we can support both function and plain element, so no one needs to use a function if not needed

### Cons

- Function inside object inside function, though it's function inside object if you don't need the outer function

## Proposed API #1

Options is the result of previous header function/previous header object merged with params passed to `getScreenConfig`

```js
header: (navigation, options) => ({
  ...options,
  icon: <Icon name={'settings' + options.focused ? '' : '-outline'} color={options.tintColor} />
});
```

Flow type,

```js
type HeaderOptionItemConfig = {
  title: string;
  icon: React.Element<*>;
}

// not sure if `HeaderOptionItemConfig & ?IconOptions` is valid, but wanted to denote that the merged object may or may not contain icon options even if the navigator always passes it
type HeaderOptionItem = NavigationOptionItem<HeaderOptionItemConfig & ?IconOptions>
```

How it is used,

```js
return router.getScreenConfig(.., { focused, tintColor }).icon;
```

### Pros

- Only two arguments to learn

### Cons

- Automatically merging 2 different kinds of objects can be confusing, for example `focused` doesn't make sense inside `navigationOptions.header`. For some things like `tintColor`, it could be  irrelevant if `tintColor` might make sense in the config, but see next one for that
- `getScreenConfig` has to handle the options, limited to objects only
- What takes precedence? `navigationOptions` or the params passed to `getScreenConfig`? Can I override the other?

## Proposed API #2

Options is the result of previous header function/previous header object

```js
header: (navigation, options) => ({
  ...options,
  activeTintColor: 'orange',
  inactiveTintColor: 'blue',
  activeIcon: <Icon name='settings' color='blue' />,
  inactiveIcon: <Icon name='settings-outline' color='orange' />
});
```

Flow type,

```js
type HeaderOptionItem = NavigationOptionItem<{
  title: string;
  activeTintColor: string;
  inactiveTintColor: string;
  activeIcon: React.Element<*>;
  inactiveIcon: React.Element<*>;
}>
```

How it is used,

```js
return router.getScreenConfig(..).activeIcon;
return router.getScreenConfig(..).inactiveIcon;
```

### Pros

- `router.getScreenConfig` doesn't deal with options passed to icon

### Cons

- I can configure the colors of text with `activeTintColor`, but for the icon, I have to type it again, and I made a typo
- I have to write almost similar component twice, can make it a function to simpify, but then it's no better than current API where I can use a function, and even write it inline
- This assumes that the navigator doesn't need to pass options, and this assumption will likely break in future

## Proposed API #3

Previous result is the result of previous header function/previous header object

```js
header: (navigation, params, previousResult) => ({
  ...previousResult,
  icon: <Icon name={'settings' + params.focused ? '' : '-outline'} color={params.tintColor} />
});
```

Flow type,

```js
type NavigationOptionItem<T, U> =
  | T
  | (navigation: NavigationScreenProp<*,*>, T, U) => T

type HeaderOptionItem = NavigationOptionItem<{
  title: string;
  icon: React.Element<*>;
}, IconOptions>
```

How it is used,

```js
return router.getScreenConfig(.., { focused, tintColor }).icon;
```

### Pros

- No automatic merging between incompatible types

### Cons

- 3 arguments to learn (should be an object most prolly)
- `getScreenConfig` has to handle the options, limited to objects only, e.g. - `getScreenConfig(..., () => {})` is not possible (likely not a significant issue)

cc @skevy @mkonicek

### LINKED ISSUE #722: RFC: Simplified Navigation Options
To simplify nav option configuration, what if we always provided it as a function?

The results of several navigationOptions will be merged. Navigator options will be overridden by static screen options, which will be overridden by route config options. Because this merging happens only at one level, we should attempt to flatten the options:

```js
static NavigationOptions = props =>({
  // these components will be rendered with screen nav props
  header: MyHeaderComponent,
  headerRight: MyRightComponent,
  headerTintColor: 'red',
  
  // this can either be a string, element, or component
  title: <TitleText>{props.navigation.state.params.name}</TitleText>,
  tabBarTintColor: 'blue',
});
```

We can add `navigationOptions` into the `screenNavigationProp`, which will allow screens to see its own computed and flattened navigationOptions. Also it will allow us to kill `HeaderProps` entirely.

Hopefully this has enough details to start off discusion. cc @grabbou @satya @skevy

grabbou:
This pull request is breaking. 

Fixes #48 
Fixes #722 

This PR introduced flat `navigationOptions` that, when defined as a function, receive the same props as screen (`navigation` and `screenProps`). That makes it easier to remember and eliminates a lot of confusion. Also, thanks to flat options, we no longer need to manually merge / extend configs with Object.assign.

It's a first PR from a series that focuses on simplifying the API and making it easier for the community to work with.

For what's changing, please see the diff, especially documentation and examples.

**Todo**
- [x] Rebase
- [ ] Wrap it up
- [X] Find issues that this PR fixes and automatically close
- [x] Go over master merge and adapt Redux example and other changes as well
- [x] Request review
- [X] Split into smaller PRs #985 

**Misc**

- Much improved `Flow` integration with fully typed `nav` options 
- Split out CardStackTransitioner from CardStack


src/views/CardStack.js
```diff
@@ -119,166 +119,60 @@ class CardStack extends Component<DefaultProps, Props, void> {
...
-        configureTransition={this._configureTransition}
-        navigation={this.props.navigation}
-        render={this._render}
-        style={this.props.style}
-        onTransitionStart={this.props.onTransitionStart}
-        onTransitionEnd={this.props.onTransitionEnd}
-      />
-    );
+  componentWillReceiveProps(props: Props) {
+    console.log('Controlled Card stack recieving props ',props.navigation.state);
```
grabbou: @evv we most likely want to get rid of `console.log`. There's already discussion that there's should be a way to configure it.

grabbou:
FYI 3eb7b12f855e4f25448674573edfc37059f81a3a has been reverted as it needs more work, I'll re-apply it tomorrow.

grabbou:
FYI 2 https://github.com/react-community/react-navigation/pull/984/commits/aa4871d06f56355e9f50640681de6ca69c8f1e90 merge was clean but diff needs to be adapted to new API - will likely break Flow now. I'll adjust it tomorrow.

docs/api/routers/RoutersAPI.md
```diff
@@ -72,30 +72,32 @@ Based on the routeNames in the state, the router is responsible for returning va
...
-### `getScreenConfig`
+### `getScreenOptions(navigation, screenProps)`
 
-Used to retrieve the navigation options for a route. Must provide the screen's current navigation prop, and the name of the option to be retrieved.
+Used to retrieve the navigation options for a screen. Must provide the screen's current navigation prop, and the name of the option to be retrieved.
 
 - `navigation` - This is the navigation prop that the screen will use, where the state refers to the screen's route/state. Dispatch will trigger actions in the context of that screen.
-- `optionName` - What named option is being fetched, such as 'title'
+- `screenProps` - Other props that your navigation options may need to consume
+- `lastConfig` - The previous set of options that are default or provided by the previous configurer
```
ericvicenti: hmm, I think this should be renamed to `navigationOptions`, which we will hopefully now pass down to the screen as a prop

docs/guides/Screen-Nav-Options.md
```diff
@@ -17,21 +17,31 @@ class MyScreen extends React.Component {
...
 
-Or, each option can be a function that takes the following arguments, and returns the value of the option.
+Or, the options can be a function that takes the following arguments, and returns an object of navigation options that will be override the route-defined and navigator-defined navigationOptions.
 
-- `navigation` - the [navigation prop](/docs/navigators/navigation-prop) for the screen, with the screen's route at `navigation.state`
-- `childRouter` - The child router, if the screen is a navigator
+- `props` - The same props that are available to the screen component
+  - `navigation` - The [navigation prop](/docs/navigators/navigation-prop) for the screen, with the screen's route at `navigation.state`
+  - `screenProps` - The props passing from above the navigator component
+  - `prevOptions` - The default or previous options that would be used if new values are not provided
```
ericvicenti: this should also be called `navigationOptions` I think

examples/NavigationPlayground/.flowconfig
```diff
@@ -35,13 +35,8 @@ suppress_type=$FlowIssue
 suppress_type=$FlowFixMe
 suppress_type=$FixMe
 
-<<<<<<< ours
-suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(>=0\\.\\(3[0-2]\\|[1-2][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
-suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(>=0\\.\\(3[0-2]\\|1[0-9]\\|[1-2][0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
-=======
 suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(>=0\\.\\(3[0-6]\\|[1-2][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
 suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(>=0\\.\\(3[0-6]\\|1[0-9]\\|[1-2][0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
->>>>>>> theirs
```
ericvicenti: Yikes. Doesn't look like we test the flow config in the playground very well

src/TypeDefinition.js
```diff
@@ -54,42 +56,68 @@ export type NavigationRoute = {
...
+  params?: NavigationParams,
 
-export type NavigationRouter = {
+
+  index: number,
+  routes: Array<NavigationRoute>,
+};
+
+
+export type NavigationRouter<State, Action, Options> = {
```
ericvicenti: This extensible router type seems much closer to the real thing! I'm glad to see that the options are a part of the router definition.

One concern: NavigationState is now customizable, but NavigationRoute is still used inside `getScreenOptions`. So I don't think this type is perfect yet.

src/TypeDefinition.js
```diff
@@ -54,42 +56,68 @@ export type NavigationRoute = {
...
    *  {routeName: 'Foo', key: '123'}
    */
-  getScreenConfig: (
-    navigation: NavigationScreenProp<NavigationRoute, NavigationAction>,
-    optionName: string,
-  ) => ?any, // todo, fix this any type to become a key of NavigationScreenConfig
+  getScreenOptions: (
+    navigation: NavigationScreenProp<NavigationRoute, Action>,
+    screenProps?: {},
+  ) => Options,
```
ericvicenti: Aha, good to see the `any` go away! This signature and name makes a lot of sense I think.

website/src/PageWithSidebar.js
```diff
@@ -62,15 +62,14 @@ class PageWithSidebar extends Component {
...
+              const options = router.getScreenOptions(childNavigation, {});
               const isActive = state.index === i;
               return (
                 <span key={i}>
                   <LinkableLi
                     to={route.routeName}
-                    className={"pt-menu-header "+icon+' '+(isActive?'active':'')}>
-                    <h6>{linkName}</h6>
+                    className={"pt-menu-header "+options.icon+' '+(isActive?'active':'')}>
+                    <h6>{options.linkName}</h6>
```
ericvicenti: Here's a nice example of when a router will produce an option that is non-standard according to our other navigators. I wish the website had better flow typing so we could actually verify that the flow types work properly.

docs/api/navigators/DrawerNavigator.md
```diff
@@ -157,25 +153,22 @@ Usually you define static `navigationOptions` on your screen component. For exam
...
-    },
-  };
+    drawerIcon: (
+      <Image src={require('./my-icon.png')} />
+    ),
+  });
   ...
 ```
 
 All `navigationOptions` for the `DrawerNavigator`:
```
ericvicenti: Ugh, so we need to document every navigation option on the navigator page AND the `Screen-Nav-Options` page? Maybe we can remove the options from the navigators page, especially as there is some overlap like `title`.

grabbou:
> Also, maybe we should somehow detect when the user is trying to use the old API, and provide a graceful warning for it?

The API change from `getScreenConfig` to `getScreenOptions` is rather internal and shouldn't affect end-users. Also, since options are now flattened, it's impossible to make them working and display a warning.

I think that what we could do is to provide a value `state` to the object that is passed to `navigationOptions` along `navigation` and `screenParams` and proxy it to `navigation.state` & print `console.warn`.

docs/api/routers/RoutersAPI.md
grabbou: Actually this should be `prevOptions`

docs/guides/Screen-Nav-Options.md
grabbou: We actually pass it as `prevOptions`. We could rename it to `prevNavigationOptions`.

docs/api/navigators/DrawerNavigator.md
grabbou: Yeah, I'll take a closer look at the documentation

grabbou: I think all options specific to given navigator belong to its `api` documentation (as on this page). I have removed example from that page and added links from `Screen-Nav-Options`. I think that's a good first try.

examples/NavigationPlayground/.flowconfig
grabbou: It would be nice to turn it on to make `Playground` a demo of how PropTypes can be consumed with a library.

grabbou:
Before we merge this, we also need to satisfy `propTypes` and `eslint` because currently, `Flow` has different types.

website/src/PageWithSidebar.js
grabbou: We can now make it more specific (`Options` are separate for each navigator and not shared). That means we can tell Flow the specific type we are expecting.

grabbou: One could simply define `props` for that component to be:
```js
type Props = {
  router: NavigationRouter<NavigationState, NavigationAction, {
    linkName: string,
    title: string,
  }>,
};
```

docs/guides/Screen-Nav-Options.md
ericvicenti: We want these arguments to match up with the screen's props. When we render the screen there is no 'previous' options, just the current ones. So that is why I think we should just call it `navigationOptions`

ericvicenti:
> The API change from getScreenConfig to getScreenOptions is rather internal and shouldn't affect end-users. Also, since options are now flattened, it's impossible to make them working and display a warning.

It is not really an internal change because router API is public, so people may already be using `getScreenConfig`. So I think we should provide an error if it gets called.

> I think that what we could do is to provide a value state to the object that is passed to navigationOptions along navigation and screenParams and proxy it to navigation.state & print console.warn.

You're right that it isn't perfect, but the easiest things we can do is:

- Do not warn if nav options are provided as a function
- Throw a useful error if they provide a function for the `title`
- Throw a helpful error if they provide a `header` or `tabBar` object

grabbou:
> The API change from getScreenConfig to getScreenOptions is rather internal and shouldn't affect end-users

You are right. I've overlooked the change. Will adapt warnings later today.

src/routers/createConfigGetter.js
```diff
@@ -17,8 +17,8 @@ import type {
 } from '../TypeDefinition';
 
 function applyConfig(
-  configurer: ?NavigationScreenConfig,
-  prevOptions: *,
+  configurer: ?NavigationScreenConfig<Object>,
+  prevOptions: Object,
```
ericvicenti: can we use a more detailed type than Object? Even if we just import the same thing from typeDefinition

website/src/App.js
```diff
@@ -19,6 +19,11 @@ import type {
   NavigationScreenComponent,
 } from 'react-navigation';
 
+export type ScreenOptions = {
```
ericvicenti: Cool, does this file pass flow now?

src/TypeDefinition.js
```diff
@@ -143,17 +143,18 @@ export type NavigationScreenConfigProps = {
...
-export type NavigationScreenConfig = NavigationScreenOptions | NavigationScreenConfigProps => NavigationScreenOptions;
+export type NavigationScreenConfig<Options> = Options
+  | NavigationScreenConfigProps => Options;
 
-export type NavigationComponent = NavigationScreenComponent<*> | NavigationNavigator<*, *, *, *>;
+export type NavigationComponent = NavigationScreenComponent<*, *> | NavigationNavigator<*, *, *, *>;
 
-export type NavigationScreenComponent<T> = ReactClass<T> & {
-  navigationOptions?: NavigationScreenConfig,
+export type NavigationScreenComponent<T, O> = ReactClass<T> & {
```
ericvicenti: can we write out the full `Options`?

src/routers/__tests__/createConfigGetter-test.js
```diff
@@ -91,7 +91,7 @@ test('should throw if the route does not exist', () => {
   ];
 
   expect(
-    () => getScreenOptions(addNavigationHelpers({ state: routes[0], dispatch: () => false }), {})
+    () => getScreenOptions(addNavigationHelpers({ state: routes[0], dispatch: () => false }), {}),
```
ericvicenti: Looking forward to prettier!

docs/guides/Screen-Nav-Options.md
grabbou: Yeah, I think it makes more sense given that it's `the default or previous options`. I've renamed it to `navigationOptions`.

website/src/App.js
grabbou: Yeah, however separate scenes itself are not yet flow-typed. Will be useful when we actually do it!

src/routers/createConfigGetter.js
```diff
@@ -86,5 +87,7 @@ export default (
     outputConfig = applyConfig(componentScreenConfig, outputConfig, configOptions);
     outputConfig = applyConfig(routeScreenConfig, outputConfig, configOptions);
 
+    validateScreenOptions(outputConfig, route);
```
grabbou: @ericvicenti we should probably `if` on node_env here so that this can get removed when bundling?

grabbou:
Nice deprecation errors (I think) were added in this commit https://github.com/react-community/react-navigation/pull/984/commits/21f6595e04ac9502c4771ba1058d6e7fabac4ef2

ericvicenti:
> Not really sure we should warn instead of throwing. When broken navigationOptions form is provided, we most likely want to inform user as soon as we detect such case, otherwise it may result in glitches in the UI that are not discovered at first glance.

Ah, agreed, we should throw an error!

src/views/CardStackTransitioner.js
```diff
@@ -0,0 +1,181 @@
...
+  _render = (props: NavigationTransitionProps): React.Element<*> => {
+    const {
+      headerComponent,
+      headerMode,
+      mode,
+      router,
+      style,
+    } = this.props;
+    return (
+      <CardStack
```
shrutic: As per this http://facebook.github.io/react-native/docs/navigation.html, CardStack is deprecated as of React native 0.43

ericvicenti: Its not deprecated, it was moved from react-native to this repo

ericvicenti: "The CardStack and other NavigationExperimental views live on as a part of the React Navigation project"

shrutic: "As of React Native 0.43, NavigationExperimental is deprecated. It will be removed from the codebase in a later version."

ericvicenti: Yes, NavigationExperimental is deprecated and has been removed from React Native.

CardStack is now a part of React Navigation, and it is not deprecated.

shrutic: Thanks ... Good to know

ericvicenti:
Ok, this is super exciting. Even though its a breaking change I think it will address a ton of issues that people are facing. Lets merge it and then keep iterating!

JCMais:
When should we expect this to be available on npm?

BrennanRoberts:
Please correct me if I'm wrong, but I don't think this patch provides access to `props` from Redux `mapStateFromProps`.
```javascript
class FooScreen extends Component {

  static navigationOptions = props => ({
    title: props.email,
  }

  static propTypes = {
    email: proptypes.String.isRequired,
  }

  ...

}

export default connect(state => { email: state.email })(FooScreen);
```

bithavoc:
@grabbou I'm running master and this used to work with a previous commit(fb2a0ad33d476807c9380967c2570f944c4345d0) but now that this got merged I'm no longer able to receive `screenProps` in the screen component via `this.props.screenProps`, now `this.props.screenProps` is always `undefined`.

You can see the at some point in the state tree `screenProps` is actually set but not available via `this.props.screenProps`.

**Tree**
```
'render props', { screenProps: undefined,
  navigation: 
   { dispatch: [Function: bound ],
     state: { routeName: 'Home', key: 'Init' },
     goBack: [Function: goBack],
     navigate: [Function: navigate],
     setParams: [Function: setParams] },
  navigationOptions: 
   { title: 
      { '$$typeof': {},
        type: 
         { [Function]
           displayName: 'Text',
           propTypes: 
            { ellipsizeMode: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              numberOfLines: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              textBreakStrategy: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              onLayout: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              onPress: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              onLongPress: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              pressRetentionOffset: { [Function: chainedCheckType] isRequired: [Function: bound checkType] },
              selectable: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              selectionColor: { [Function: bound colorPropType] isRequired: [Function: bound colorPropType] },
              suppressHighlighting: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              style: [Function],
              testID: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              allowFontScaling: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              accessible: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              adjustsFontSizeToFit: { [Function: bound checkType] isRequired: [Function: bound checkType] },
              minimumFontScale: { [Function: bound checkType] isRequired: [Function: bound checkType] } },
           getDefaultProps: { [Function: getDefaultProps] isReactClassApproved: {} },
           childContextTypes: { isInAParentText: { [Function: bound checkType] isRequired: [Function: bound checkType] } },
           contextTypes: { isInAParentText: { [Function: bound checkType] isRequired: [Function: bound checkType] } },
           defaultProps: 
            { accessible: true,
              allowFontScaling: true,
              ellipsizeMode: 'tail' } },
        key: null,
        ref: null,
        props: 
         { style: undefined,
           children: 'Hello 2',
           accessible: true,
           allowFontScaling: true,
           ellipsizeMode: 'tail' },
        _owner: 
         { _currentElement: 
            { '$$typeof': {},
              type: [Function: CardStack],
              key: null,
              ref: null,
              props: 
               { headerComponent: 
                  { [Function: Header]
                    HEIGHT: 64,
                    Title: [Function: HeaderTitle],
                    BackButton: 
                     { [Function: HeaderBackButton]
                       defaultProps: 
                        { pressColorAndroid: 'rgba(0, 0, 0, .32)',
                          tintColor: '#037aff',
                          truncatedTitle: 'Back' } } },
                 headerMode: undefined,
                 mode: 'card',
                 router: 
                  { getComponentForState: [Function: getComponentForState],
                    getComponentForRouteName: [Function: getComponentForRouteName],
                    getStateForAction: [Function: getStateForAction],
                    getPathAndParamsForState: [Function: getPathAndParamsForState],
                    getActionForPathAndParams: [Function: getActionForPathAndParams],
                    getScreenOptions: [Function],
                    getScreenConfig: [Function] },
                 style: undefined,
                 layout: 
                  { height: 
                     { _children: [],
                       _value: 568,
                       _offset: 0,
                       _animation: null,
                       _listeners: {} },
                    initHeight: 568,
                    initWidth: 320,
                    isMeasured: true,
                    width: 
                     { _children: [],
                       _value: 320,
                       _offset: 0,
                       _animation: null,
                       _listeners: {} } },
                 navigation: 
                  { dispatch: [Function: bound ],
                    state: { index: 0, routes: [ { routeName: 'Home', key: 'Init' } ] },
                    goBack: [Function: goBack],
                    navigate: [Function: navigate],
                    setParams: [Function: setParams] },
                 position: 
                  { _children: 
                     [ { _children: [ { _children: [Object], _style: [Object] } ],
                         _parent: [Circular],
                         _config: 
                          { inputRange: [ -1, 0, 0.99, 1 ],
                            outputRange: [ 1, 1, 0.3, 0 ] },
                         _interpolation: [Function] },
                       { _children: [ { _children: [Object], _transforms: [Object] } ],
                         _parent: [Circular],
                         _config: 
                          { inputRange: [ -1, 0, 0.99, 1 ],
                            outputRange: [ 350, 0, -10, -10 ] },
                         _interpolation: [Function] },
                       { _children: [ { _children: [Object], _style: [Object] } ],
                         _parent: [Circular],
                         _config: { inputRange: [ -1, 0, 1 ], outputRange: [ 0, 1, 0 ] },
                         _interpolation: [Function] },
                       { _children: [ { _children: [Object], _transforms: [Object] } ],
                         _parent: [Circular],
                         _config: { inputRange: [ -1, 1 ], outputRange: [ 200, -200 ] },
                         _interpolation: [Function] } ],
                    _value: 0,
                    _offset: 0,
                    _animation: null,
                    _listeners: { '2': [Function: bound _onPositionChange] } },
                 progress: 
                  { _children: [],
                    _value: 1,
                    _offset: 0,
                    _animation: null,
                    _listeners: {} },
                 scenes: 
                  [ { index: 0,
                      isActive: true,
                      isStale: false,
                      key: 'scene_Init',
                      route: { routeName: 'Home', key: 'Init' } } ],
                 scene: 
                  { index: 0,
                    isActive: true,
                    isStale: false,
                    key: 'scene_Init',
                    route: { routeName: 'Home', key: 'Init' } },
                 index: 0 },
              _owner: 
               { _currentElement: 
                  { '$$typeof': {},
                    type: [Function: Transitioner],
                    key: null,
                    ref: null,
                    props: 
                     { configureTransition: [Function],
                       navigation: 
                        { dispatch: [Function: bound ],
                          state: { index: 0, routes: [ [Object] ] },
                          goBack: [Function: goBack],
                          navigate: [Function: navigate],
                          setParams: [Function: setParams] },
                       render: [Function],
                       style: undefined,
                       onTransitionStart: undefined,
                       onTransitionEnd: undefined },
                    _owner: 
                     { _currentElement: 
                        { '$$typeof': {},
                          type: { [Function: CardStackTransitioner] defaultProps: { mode: 'card', headerComponent: [Object] } },
                          key: null,
                          ref: null,
                          props: 
                           { screenProps: { AuthToken: '52orYXvTvH7flWz+AKQ6Wb/vD6mubZzpxbLYA/c7ZIOZwW+2MQjhnJ4hC8BUEQzPtbqJsviJ/4KU037ozgb1PQ==' },
...
```

PR #985, opened 2017-04-10
Remove HybridExample

grabbou:
This pull request removes `HybridExamples` due to ongoing work on the way it's set up.


ericvicenti:
Thanks! I think this will help avoid confusion about integration with `native-navigation`

PR #999, opened 2017-04-12
bugs fix

leoJFeng:
navigation.goBack('key') not work and navigation.reset method lack of params.


ericvicenti:
Closing until the comments are addressed

PR #1002, opened 2017-04-12
Expose Card through the API surface

ptomasroos:
We use Card directly without CardStack hence it would be nice to be able to import it through the api surface.


ericvicenti:
I'm generally in support of this, but I think it would be nice to fold in `CardStackStyleInterpolator` to the `Card` component. Because that will change the API, I think it might make sense to wait for it.

Also, it would be nice to have docs for the Card component! Same with the Header!

grabbou:
@ericvicenti I believe we can merge it now and optionally follow up with adding `interpolator` when people request such import?

ptomasroos:
👍

grabbou:
@ptomasroos can you also export `Header` and `HeaderBackButton` ?

ptomasroos:
Sure. We have no use for it but def makes sense.

satya164:
The API is inconsistent now because `TabBar`, `DrawerNavigationItems` etc are static properties on `TabView` and `DrawerView` etc. :(

ptomasroos:
let's fix it then. Either expose those or move these to belong to card. I don't have a strong opinion since I don't use the underlying components.

grabbou:
Yeah, we will take care of it! `Card` will stay, so no worries @ptomasroos

ptomasroos:
Thanks! Let me know whenever you need a hand

ericvicenti:
Yeah I'm in support of flattening everything out by removing things like `DrawerView` from `DrawerNavigator` and instead exposing it on the top level, along with `Card` as you've done here.

Fixing the exports is easy, but the non-easy part is documenting all of the views so that they can be used individually. There may be some glaring inconsistencies between the different views, which we had not addressed. These may be awkward to document, and we may prefer to make them consistent before we document everything and encourage people to use it. Also once we change the export, we should be more careful about breaking changes.

@ptomasroos, do you want to take charge on this? Fixing the exports, and then improving documentation to describe how the views can be used independently? And also even fixing inconsistencies in the View's API?

grabbou:
I have a working diff that fixes exports. I believe we can sneak it into next release and then, keep working on documentation. I'd prefer to include that breaking change now and avoid it for post v8 releases.

PR #1009, opened 2017-04-12
Setup prettier

grabbou:
Adds `prettier` with `eslint` compatible config. All runs on `precommit` hook.


grabbou:
Eslint is currently turned off as there are numerous proptype issues. I am going to close this PR until we make it green (first).

PR #1017, opened 2017-04-13
allow headerLeft on 0 index

pvinis:
Very small change to allow headerLeft on 0 index screens.

Basically, I want a left button on my first screen. And since there is no reason that this is not allowed, I enabled it.

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -128,9 +128,6 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
   };
 
   _renderLeftComponent = (props: SceneProps) => {
-    if (props.scene.index === 0) {
-      return null;
-    }
     const details = this.props.getScreenDetails(props.scene);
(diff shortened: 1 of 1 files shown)


src/views/Header.js
```diff
@@ -128,9 +128,6 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
   };
 
   _renderLeftComponent = (props: SceneProps) => {
-    if (props.scene.index === 0) {
-      return null;
-    }
```
grabbou: Can you undelete these lines and place them below `if (headerLeft)` ? That way, when no custom headerLeft implementation is given, we apply our default that doesn't render on 0 index.

grabbou:
I will send a PR including this fix as I've noticed some other changes to Header that we made in previous PR. Sorry about that!

Thanks for keeping an eye on changes!

grabbou:
See #1027

PR #1027, opened 2017-04-13
Fix regressions in Header

### LINKED ISSUE #1015: allow headerLeft on index 0
i see no reason why i shouldnt be able to put a left button on the screen at 0 index.
its currently hardcoded to return null. is there a good reason for that choice besides to hide an "easy" back button?

### LINKED ISSUE #1021: Can't hide header back button anymore
I was using this code to hide the header back button and place the title in center:

```
static navigationOptions = {
        header: {
            title: 'Login',
            left: null,
            tintColor: 'white',
            style: {
                backgroundColor: '#3F51B5'
            },
            titleStyle: {
                alignSelf: 'center'
            }
        }
    };

```

now with the new syntax I tried:

```
static navigationOptions = {
        headerTitle: 'Login',
        headerLeft: null,
        headerTintColor: 'white',
        headerStyle: {
            backgroundColor: '#3F51B5'
        },
        headerTitleStyle: {
            alignSelf: 'center'
        }
    };
```

but it doesn't work anymore. So I tried the following:

```
headerLeft: <View></View>,
```
it works, but the title is being rendered not in the center of the header and I can't center it.

grabbou:
After reviewing #1017, I've discovered other changes that have to be applied. I also decided to bring back documentation.

Changes:
- `headerLeft` on scene with index `0` works. Fixes #1015 
- `backButtonTitle` is taken from a proper scene
- `headerLeft` different than `undefined` will take prio over default impl. Fixes #1021

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -95,6 +95,10 @@ Visual options:
 
 Generic title that can be used as a fallback for `headerTitle` and `tabBarLabel`
 
+#### `headerBackTitle`
+
+Title string used by the back button on iOS or `null` to disable label. Defaults to `title`.
+
 #### `headerVisible`
 
 True or false to show or hide the header. Only works when `headerMode` is `screen`. Default value is `true`.
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -128,7 +128,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
 
   _renderLeftComponent = (props: SceneProps) => {
     const options = this.props.getScreenDetails(props.scene).options;
-    if (options.headerLeft) {
+    if (typeof options.headerLeft !== 'undefined') {
       return options.headerLeft;
     }
     if (props.scene.index === 0) {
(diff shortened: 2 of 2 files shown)


docs/api/navigators/StackNavigator.md
```diff
@@ -95,6 +95,10 @@ Visual options:
 
 Generic title that can be used as a fallback for `headerTitle` and `tabBarLabel`
 
+#### `headerBackTitle`
```
ericvicenti: We also need to document this in the screen navigaiton options doc.

grabbou: Note we don't have any of options listed there. My past commit just changed the list to references to this page. We might want to change that in the future if it's not the perfect UX https://reactnavigation.org/docs/navigators/navigation-options#Navigation-Option-Reference

EmielM:
I've tried these changes, but I'm having trouble sizing the (custom) title element to the proper width (`screenWidth - 2*max(leftWidth/rightWidth)` I guess). In one of the previous implementations, just adding `flex: 1` would do the trick. How should I do this today?

davidlee1435:
I'm having a similar problem, except with `headerRight`. The View is at index 0 and I have no nagivationOptions defined when I instantiate the StackNavigator; each navigationOptions object is defined per screen. Might these possibly be related?

PR #1033, opened 2017-04-13
Automatically generate prop-types from Flow

grabbou:
This pull request removes prop-types entirely from `react-navigation` and uses babel plugin to automatically generate them from `Flow`.

Why?

We wanted to keep `prop-types` for non-Flow users, however this library was always developed with Flow in mind. That caused issues of prop-types either missing on components are not staying in sync. 

Since things couldn't been worse than already, we now removed them and replaced with generated ones.

Despite babel plugin limitations, that should give us better prop-types that are in sync with Flow and also encourage us to define better Flow types.

As a side effect, we also now use `prop-types` package in `src` which should be future-proof.


dantman:
This broke requiring `react-navigation` as a dependency:
```
error: bundling: UnableToResolveError: Unable to resolve module `prop-types` from `{mobile-app}/node_modules/react-navigation/src/views/withNavigation.js`: Module does not exist in the module map or in these directories:
  {mobile-app}/node_modules/react-navigation/node_modules
,   {mobile-app}/node_modules
```

`prop-types` is a dev dependency, so it's not installed when you depend on react-navigator.

grabbou:
@dantman See #1042

faceyspacey:
Have you guys looked at the propTypes generated. I've used `babel-plugin-flow-react-proptypes` and my guess is it's not generating half of what you're expecting.

Here's an issue I posted on that repo a while back:

https://github.com/brigand/babel-plugin-flow-react-proptypes/issues/71

Basically I've found that only if you define the type in a given file *where you use it*, and you **don't use intersections** it works reliably. For basically everything else nothing is generated. This is a real problem because it leads to tons of redundancy in your types. 

Check what's being generated. What you find likely won't be pleasing.

PR #1035, opened 2017-04-13
Feat/better lint

grabbou:
This pull request turns on prettier with pre-commit hook that will fix and format small issues.

It also fixes all remaining Eslint issues and turns it on to run on CI by default.


grabbou:
This is not ready for merge yet. I need to re-apply few things and squash commits.

ericvicenti:
Wow this is awesome!!

I ran it locally and it works great

skevy:
I want to raise an issue here.

I LOVE Prettier. Never will I say to not use it on a project. :) In fact, as you may know, we just Prettier-ed the whole Expo codebase.

But, `lint-staged`, and running Prettier and actually _changing_ code automatically is a bad idea, I think, with the reason being that it can break people's Git workflows in strange ways.

As an example -- it's very common to use `git add -P` (git add patch), to stage pieces of files. Even if the user doesn't use this command directly -- if they use a Git GUI, it will be using this functionality behind the scenes.

With the way `lint-staged` and running Prettier (or in this case, `eslint --fix`) on pre-commit works, what would happen in this scenario is:

- I don't use Prettier/Eslint autofix in my editor
- I stage a piece of my file to commit it (perhaps because I'm rebasing on a branch and making logical commits), and it has formatting issues inside of it.
- I commit, and this pre-commit hook runs.
- `lint-staged` will see the filename of my partially committed file in the Git index, and run Prettier on the whole file, and then run `git add` and add the _entire_ file, instead of simply my staged piece.
- I've now lost my staged changes, and lots of my file may have been modified

One option that I've heard people discuss is stashing with `--keep-index` first, which would keep ONLY the changed piece of the file, running Prettier, running `git add`, popping the stash, and then committing. It sounds great in theory, but unfortunately is also flawed -- the changes for merge conflicts and leaving people in a hard to resolve state is common.

What I would recommend here is that instead of actually running `prettier --write` on pre-commit, we run a function that essentially checks if `contentsOfFile === prettier(contentsOfFile)`. This way, we can still alert the user that they have a badly formatted file, keep code quality high, and provide instructions on how to run Prettier on that file (we can have an NPM script (`npm run format`) that just formats the whole repo that we can recommend they run), all without breaking contributors Git workflows.

skevy:
(sorry I didn't mean to close)

ericvicenti:
Ok, so instead of a precommit hook we should have `npm run format`, and add a test step that verifies that every file is prettier'd.

tychota:
At bam.tech we are using prettier with https://github.com/not-an-aardvark/eslint-plugin-prettier.

We had the same problems that what @skevy rised, and enforcing it via eslint was a good move.
The errors message can be optimized but it is pretty easy to set-up.

skevy:
Indeed @tychota we are doing the same thing @ Expo -- we have Prettier configured with eslint, and then most of us have "Fix Eslint Errors on Save" turned on in our editors, which in-turn runs Prettier.

PR #1036, opened 2017-04-14
skip Transitioner animation after swiped back

### LINKED ISSUE #1028: gesture regression
That's a long story, https://github.com/react-community/react-navigation/commit/6aff0ac36617219d1c9cbd06710fe1e7aa6011d6 introduced an issue #527, then was fixed by https://github.com/react-community/react-navigation/commit/11cab8eab62b6ca1c2a0445f30a9d0dc75458e2e, but those two commits introduced more issues when I'm trying to make the PR #828, 

**a**. https://github.com/react-community/react-navigation/blob/master/src/views/TransitionConfigs.js#L32 duration is an invalid props for `Animated.spring`, and the change from `Animated.timing` to `Animated.spring` raised a shadow issue

**b**. https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L220-L221 `velocity` and `bounciness` are invalid props for `Animated.timing`, so `velocity` and `GESTURE_ANIMATED_VELOCITY_RATIO` have no effect at all

**c**. The two changes in **a** and **b** above combined to introduce a issue which leads to a workaround here https://github.com/react-community/react-navigation/blob/master/src/views/CardStackStyleInterpolator.js#L65-L67

**d**. **gesture** support for **modal** is removed https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L261 

actually the workaround in **c** is not needed if we do the following works:
1. revert **a**, `Animated.spring -> Animated.timing`
2. remove redundant animations in `_goBack` https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L208-L232 which mentioned in #828

the workaround introduces a really annoying issue, when we swipe back, we need to wait for the `Animated.spring` animation before the app response to the coming interaction, although is hard to notice after this workaround, but it only hides the shadow, we still need to wait until finish, because this animation is started after the `Animated.timing` in `_goBack`, you can remove the **30px offset** workaround, change the `Animated.spring` to `Animated.timing` in https://github.com/react-community/react-navigation/blob/master/src/views/TransitionConfigs.js#L33 and enlarge the **duration**, then the **none response** will be much obvious

the purpose mentioned in https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L112-L118 doesn't be done in the right way, if we want a smooth swipe back animation, the animation duration should be calculated according to the release position but not a invalid **velocity**

@grabbou @ericvicenti

nihgwu:
fix #1028 with #1110 

Test plan (before #1110 landed):
1.  change `DefaultTransitionSpec` as the following in https://github.com/react-community/react-navigation/blob/master/src/views/TransitionConfigs.js
``` js
const DefaultTransitionSpec = ({
  // The following options are meant to mimic the nav animations of iOS 10
  duration: 2500,
  timing: Animated.timing,
  bounciness: 0,
  speed: 9,
} : NavigationTransitionSpec);
```

2. remove the **30px offset workaround** in https://github.com/react-community/react-navigation/blob/master/src/views/CardStackStyleInterpolator.js#L67
`const width = layout.initWidth + 30;` => `const width = layout.initWidth;`

3. test the `Stack Example` in the `NavigationPlayground` example, press the `Go go profile screen`

4. swipe back with gesture

you will need to wait for an extra 2500ms before the the screen shadow will disappears, and the navigator will become non-responsible

actually if you change the `StatusBar`'s `barStyle`, you will need to wait for the extra time before the `barStyle` change too

after applied this change, the extra waiting time will be gone


nihgwu:
@ericvicenti You can see the difference if you remove the **30px offset workaround** in the master branch

grabbou:
> there are still lots of useless code in the gesture logic

This is a very bold statement. Could you describe intentions behind this PR, what is really fixing and why the changes? What each deletion is doing? Thanks!

nihgwu:
@grabbou see #1028  for detail

nihgwu:
Perhaps the `Animated.spring => Animated.timing` change was introduced by mistake in https://github.com/react-community/react-navigation/commit/11cab8eab62b6ca1c2a0445f30a9d0dc75458e2e

src/views/CardStack.js
```diff
@@ -181,41 +164,24 @@ class CardStack extends Component {
...
-        const backFromScene = scenes.find(s => s.index === toValue + 1);
-        if (!this._isResponding && backFromScene) {
-          navigation.dispatch(
-            NavigationActions.back({ key: backFromScene.route.key })
-          );
-        }
-      });
+    const backFromScene = scenes.find(s => s.index === toValue + 1);
+    if (!this._isResponding && backFromScene) {
+      navigation.dispatch(
```
ericvicenti: This dispatches an action in the middle of the gesture! This causes us to loose a bunch of frames at the time when the touch is released. That is why this optimization was made. We wait to fire the back action until the gesture actually completes.

nihgwu: I know that, so the current change is to resolve the **none-response** issue and the **30px offset walkaround**

This is the first step as you asked me to take in #1028, I'll try to find a more proper way to achieve this

nihgwu: I think the right way is using `Animated.spring` after gesture to achieve smooth transition, but we should take care the following screen transition animation, if we swipe back, the following screen transition should be done in 0ms, then the two issues would be resolved in the same time

nihgwu:
I think the right way is using `Animated.spring` after gesture to achieve smooth transition, then we should take care of the following screen transition animation, if we swipe back, the following screen transition should be done in 0ms, then the two issues **non-responsiveness** and **30px offset workaround**  would be resolved in the same time

ericvicenti:
Ok @nihgwu, I think your approach makes sense. This commit is a good first step, but I don't want to land it to master until the transition feels smooth again. In your branch, can you show me the rest of the fix that you recommended? Specifically, the spring transition and the 0ms back action transition that happens after the swipe-back animation completes?

Once I can test your full fix then we can start landing it in several pieces.

Thanks again for your work on this, it is greatly appreciated!

nihgwu:
@ericvicenti sorry I'm a bit busy these days, I will try to make it out tomorrow or the day after that

nihgwu:
@ericvicenti tried to minimize the changes, should be safe to be merge to master, if you think it's needed I'd love cleanup the useless `duration` for `Animated.spring`
BTW, the **30px offset workaround** is still needed in this approach, as the screens transition will be done abruptly , unless we dispatch the back action before gesture animated finish

nihgwu:
actually the **30px offset workaround** is needed for **Modal mode** too, I'll create a new PR the remove this workaround, get any time to review this one? @ericvicenti

src/views/Transitioner.js
```diff
@@ -155,7 +155,11 @@ class Transitioner extends React.Component<*, Props, State> {
     const { timing } = transitionSpec;
     delete transitionSpec.timing;
 
-    const animations = indexHasChanged
+    const toValue = nextProps.navigation.state.index;
+    const positionHasChanged = position.__getValue() !== toValue;
```
ericvicenti: This `__getValue()` function will not provide the correct value when the animated value is on the native thread.

Instead, we can use the asynchronous `animatedValue.stopAnimation(value => doSomethingWith(value));`

nihgwu: but actually, the former animation has already completed, and the following animation is on the go, so I think it's safe to do it this way https://github.com/react-community/react-navigation/blob/master/src/views/CardStack.js#L208
BTM, mind to review #1110 now?

nihgwu:
@ericvicenti updated test plan in the first comment

src/views/Transitioner.js
nihgwu: If you still think it's needed, I'll make the change, I just don't want to make it complicated

nihgwu:
@ericvicenti can we make this be merged soon? This pr is really important to achieve smooth animation and interaction
I think there is only one thing we should concern, if we swiped too fast, the whole StackNavigator will bounce a little, but I think it's OK, and I know how to make a workaround, that's to limit the following spring animation's speed to a certain value

So there are two things to address out:
1. Is `animatedValue.stopAnimation` really needed
2. Should we limit the speed of the spring animation

nihgwu:
@ericvicenti what about the second problem? don't know why extreme speed will lead to bouncing although bounciness is set to 0, perhaps @janicduplessis would know

janicduplessis:
@nihgwu @ericvicenti `__getValue` will work with native animated if there is a listener on the animated value, kind of a hack but from looking at the code quickly I don't see a better way. I'm considering changing it so the JS value stays in sync with the native one even without listener but I want to make sure to coalesce these events properly first to avoid affecting perf.

As for the bounciness thing I have no idea :(

nihgwu:
@janicduplessis In this scenario we are concerning will `AnimatedValue.__getValue` get the correct value we expected on a native animated value, I think it should do because the animation has already finished completely

ericvicenti:
Perhaps we can limit the maximum speed?

nihgwu:
OK, I will do that, I've already find a safe speed while inspecting in this issue, extreme speed is not necessary needed

PR #1039, opened 2017-04-14
Allow the routes outputted by DrawerView.Items to be overridden

dantman:
Among other things, this allows sections and dividers to be implemented by providing multiple DrawerView.Items with partial versions of `navigation.state.routes`.

`getScreenOptions` is also exposed so contentComponent can make use of its own navigation properties.

## Test plan (required)
Here's a screenshot of an app with a custom contentComponent using these changes to add its own divider/subheading handling.
screenshot_20170413-190353

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -130,6 +130,7 @@ const styles = StyleSheet.create({
 
 ### `contentOptions` for `DrawerItems`
 
+- `items` - an array of routes to use in place of `navigation.state.routes`
 - `activeTintColor` - label and icon color of the active label
 - `activeBackgroundColor` - background color of the active label
 - `inactiveTintColor` - label and icon color of the inactive label
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -130,11 +130,13 @@ const styles = StyleSheet.create({
 
 ### `contentOptions` for `DrawerItems`
 
-- `items` - an array of routes to use in place of `navigation.state.routes`
+- `items` - the array of routes, can be modified or overridden
+- `activeItemKey` - key identifying the active route
 - `activeTintColor` - label and icon color of the active label
 - `activeBackgroundColor` - background color of the active label
 - `inactiveTintColor` - label and icon color of the inactive label
 - `inactiveBackgroundColor` - background color of the inactive label
+- `onItemPress(route)` - function to be invoked when an item is pressed
 - `style` - style object for the content section
 - `labelStyle` - style object to overwrite `Text` style inside content section, when your label is a string
 
--- a/src/views/Drawer/DrawerNavigatorItems.js
+++ b/src/views/Drawer/DrawerNavigatorItems.js
@@ -12,17 +12,19 @@ import type {
   NavigationRoute,
   Style,
 } from '../../TypeDefinition';
-import type { DrawerScene } from './DrawerView.js';
+import type { DrawerScene, DrawerItem } from './DrawerView.js';
 
 type Props = {
   navigation: NavigationScreenProp<NavigationState, NavigationAction>,
-  items?: Array<NavigationRoute>,
+  items: Array<NavigationRoute>,
+  activeItemKey?: string,
   activeTintColor?: string,
   activeBackgroundColor?: string,
   inactiveTintColor?: string,
   inactiveBackgroundColor?: string,
   getLabel: (scene: DrawerScene) => ?(React.Element<*> | string),
   renderIcon: (scene: DrawerScene) => ?React.Element<*>,
+  onItemPress: (info: DrawerItem) => void,
   style?: Style,
   labelStyle?: Style,
 };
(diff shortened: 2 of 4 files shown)


docs/api/navigators/DrawerNavigator.md
```diff
@@ -128,6 +128,7 @@ const styles = StyleSheet.create({
 
 ### `contentOptions` for `DrawerView.Items`
 
+- `routes` - an array of routes to use in place of `navigation.state.routes`
```
ericvicenti: Can we name this something else? These aren't really routes

src/views/Drawer/DrawerNavigatorItems.js
```diff
@@ -22,6 +22,7 @@ import type {
 
 type Props = {
   navigation: NavigationScreenProp<NavigationState, NavigationAction>;
+  routes?: Array<*>;
```
dantman: As in use a type other than `*`?

The original code doesn't define a type for the items `navigation.state.routes.map((route: *, index: number) => {`.

docs/api/navigators/DrawerNavigator.md
dantman: What would you suggest? The original code called them routes everywhere, so I called the prop `routes`.

ericvicenti: I would call them 'items'. Of course the routes are currently used as the source of item data, but I see you want to use an alternative list of items other than the routes.

dantman: I could change the name to items. Though I should note that the actual use case is not passing arbitrary data to `DrawerView.Items`, it's splitting `navigation.state.routes` into groups and passing a subset of `navigation.state.routes` to multiple `DrawerView.Items`.

Very crudely:
```
<DrawerView.Items routes={navigation.state.routes.slice(0, 1)} ... />
{some divider}
<DrawerView.Items routes={navigation.state.routes.slice(1, 2)} ... />
{some heading}
<DrawerView.Items routes={navigation.state.routes.slice(3, 4)} ... />
```

You probably could pass some data not part of navigation.state.routes in. However the implementation is still relies on `navigation.state` for a few things such as what item is focused. So you can't actually pass in 100% custom data and have complete control over the items view.

dantman:
The requested changes should be ready

dantman:
@ericvicenti ?

dantman:
> Remove usage of navigation.state.routes entirely, and pass the items prop from DrawerSidebar with the default value of navigation.state.routes

Done.

>  Remove usage of navigation.state to determine the active item, and pass a prop activeItemIndex to indicate the active item, the default value passed from DrawerSidebar should be navigation.state.index

I've added a `activeItemKey`. `index` is not reliable, when `navigation.state.routes` was used directly and `items` never existed, `index` just coincidentally matched up with `state.index`. As soon as you are separately providing `items` which may be filtered or have a modified order that unreliable index completely falls apart. And falls apart further when you try to pass `activeItemIndex`.

>  Instead of calling navigation.navigate, pass a prop onItemPress which should receive { route, index } as the argument (similar to tab bar), the default value should do the navigate

Done, but the function signature is `onItemPress(route)` for the aforementioned issue and uselessness of `index`.

>  Remove the getScreenOptions prop as we are thinking of providing an utility function/HOC to do this in future

I'll welcome this when you do add this – and at that point in the future would happily remove `getScreenOptions`. However at the moment `getScreenOptions` is the **only** way for a `customComponent` to get custom `navigationOptions` options besides `drawerItem` and `drawerLabel` – which are necessary to implement just about any special behaviour including sections and dividers.

>  Remove the screenProps prop since it's supposed to be screen's props, anything else should be in params, it should probably be possible to pass dynamic params as prop in future

`screenProps` is not actually exclusively passed to screens, it's also passed to other places like `navigationOptions`. And there are a variety of situations where props passed to screens may also apply to the drawer.

* The height of a translucent StatusBar isn't actually static, there are conditions it can change. And if an app handles translucent status bars the only way to apply it dynamically is to pass it through `screenProps` to use it in `headerStyle`. This screenProp is also needed in `customContainer` to apply required padding to the top of the drawer.
* Apps may pass state down to scenes such as the logged in user in `screenProps`. A custom drawer can require this state – it can affect the state of drawer items, change how you filter the `items` (show/hide the items for a login/logout route), be required for avatars in the drawer.

So this is also immediately required to be able to implement a `customComponent`.

dantman:
> > Done, but the function signature is onItemPress(route) for the aforementioned issue and uselessness of index.
>
> I don't think we need to change signatures for different components regardless of the argument about index. Also maybe in future we might wanna pass more info. { route, index } looks good to me.

I can change the signature to `{ route }` to leave room for future info. But `index` should still really not be passed here, it's not the same thing as it is in other components so pretending that it is and can be used in the same way by passing it is actively harmful.

In TabBarBottom I only see the `jumpToIndex: (index: number) => void` prop which doesn't use the `{ route, index }` signature, so that's already different. In the router `state.index` is really the `routesIndex` (index in `state.routes` of the active route). In Tab*, `index` is really a `tabIndex` (index of the tab that matches up with the index in `state.routes`). But in `DrawerNavigatorItems` `index` is really `unrelatedIterationIndex` (it cannot be compared to `state.index` – but people will still try and be confused when their app breaks – and it will change and reset depending on what `items` are passed – if you use multiple `DrawerNavigatorItems` you'll get multiple `onItemPress` that'll say `index: 0` for their route).

satya164:
`TabBarTop` has a `onTabPress` prop which accepts `{ route, index, focused }`. `TabBarBottom` should too, but it's just unimplemented. Anyway. it's a really small thing and I don't want to bikeshed on this.

dantman:
>> I'll welcome this when you do add this – and at that point in the future would happily remove getScreenOptions. However at the moment getScreenOptions is the only way for a customComponent to get custom navigationOptions options besides drawerItem and drawerLabel – which are necessary to implement just about any special behaviour including sections and dividers.
>
> You should be able to get options as long as you get the router and navigation props with some typing. It's better not to introduce temporary APIs only to deprecate them later.

Ok, I've modified my code to not require `props.getScreenOptions` and removed it from the PR. However doing what you describe is actually pretty hacky. In order to get screen options from my own code I had to:

Replicate the `getScreenOptions` method in my own code:
```js
	_getScreenOptions(routeKey) {
		const DrawerScreen = this.props.router.getComponentForRouteName('DrawerClose');
		return DrawerScreen.router.getScreenOptions(
			this.props.childNavigationProps[routeKey],
			this.props.screenProps
		);
	}
```

Import the **internal** `@withCachedChildNavigation` decorator directly from its module (it's not an officially exported interface) in order to get ahold of the `childNavigationProps` prop required by `getScreenOptions`:
```js
import {NavigationActions, DrawerItems} from 'react-navigation';
import withCachedChildNavigation from 'react-navigation/src/withCachedChildNavigation';

@withCachedChildNavigation
export default MyDrawerComponent ...
```

Also because `DrawerScreen.router.getScreenOptions` requires `screenProps` this means that removing the `getScreenOptions` prop makes it impossible to remove the `screenProps` prop.

dantman:
@satya164 Should be ready for re-review

satya164:
We probably should export `withCachedChildNavigation` since it's useful for custom navigators.

PR #1045, opened 2017-04-14
Upgrade `babel-plugin-flow-prop-types`

grabbou:
This PR switches `babel-plugin-flow-prop-types` to our fork temporarily as the npm version is broken for higher order components.

We will switch back to `npm` when this gets merged: https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/76

That unbreaks website build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,4 @@
 {
-  "presets": ["react-native"],
   "env": {
     // For RN example development
     "development": {
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -2,7 +2,8 @@
 
 import React from 'react';
 
-// @todo figure out how to use Animated.Value type in web context
+// @todo when we split types into common, native and web,
+// we can properly change Animated.Value to its real value
 type AnimatedValue = *;
 
 export type HeaderMode = 'float' | 'screen' | 'none';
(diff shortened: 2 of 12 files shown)


src/TypeDefinition.js
```diff
@@ -2,14 +2,16 @@
...
-import {
-  Animated,
-} from 'react-native';
+// @todo figure out how to use Animated.Value type in web context
+type AnimatedValue = *;
 
-import type {
-  HeaderMode,
-  HeaderProps,
-} from './views/Header';
```
grabbou: That file is required by prop-types plugin everywhere and so, it cannot have any React Native related types in it (it will try to bundle react-native for web target otherwise).

.babelrc
```diff
@@ -1,5 +1,4 @@
 {
-  "presets": ["react-native"],
```
grabbou: Unnecessary, also gets merged with `env` presets field. That causes `web` target to build with `react-native` preset.

xavierlefevre:
@grabbou I think it has been merged with master https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/76

Came to see because I have an error from the package : 
screen shot 2017-04-17 at 10 03 02

grabbou:
Can you submit a PR that bumps the version? Thanks :)

PR #1064, opened 2017-04-16
Remove unused import

RWOverdijk:
Import not needed.


przemkow:
In `line 16` of `AppNavigator.js` you have JSX code so I think this React import is required :) otherwise you will find "Can't find variable: React" error during runtime

RWOverdijk:
I didn't know that... I didn't see it being used anywhere. I hate React more every day haha

PR #1073, opened 2017-04-17
Fix StackRouter's back navigation to pop routes up until the route which is identified by the key

tjahyaf:
I fixed the behavior of StackRouter's back navigation when used with a route key. Before the fix, the route identified by the key is also popped from the route stack. After the fix, the identified route is at the top of the stack.

Consider issuing the back navigation with key 'C' in the following example:
Route stack: [A B C D E]
State routes before fix: [A B]
State routes after fix: [A B C]

I feel it is more natural to read it as 'back to the screen identified by the key'. In most cases, your application knows the target route to go back to. Not having to search through the route stack again for the index of the next route saves a lot of complexity in the code.

**Test plan (required)**
I updated the test 'Handle goBack identified by key' in src/routers/__tests__/StackRouter-test.js with the new expectation.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -246,21 +246,16 @@ export default (
       }
 
       if (action.type === NavigationActions.BACK) {
-        let backRouteIndex = null;
-        if (action.key) {
-          /* $FlowFixMe */
-          const backRoute = state.routes.find((route: *) => route.key === action.key);
-          /* $FlowFixMe */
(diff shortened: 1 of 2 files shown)


ericvicenti:
Sorry, this is already the desired behavior. If you change this, lots of things will break.

When you provide a key to 'Back', you are telling the router which key you want to go back *from*, not what key you are going *to*. Could you submit a pull request to clarify the docs and make this more apparent to the next person who gets confused by it?

PR #1083, opened 2017-04-17
add headerBackTitleStyle screen option

koenpunt:
I'm currently migrating an existing iOS app to React Native, where I use customized titles and back buttons.

The titles currently can be styled using `headerTitleStyle`, but for the back button there was no such option.

So this change adds a `headerBackTitleStyle` screen option, which makes it possible to set the style for the back button title ~, and when not set it inherits the `headerTitleStyle`~.

**TODO**

- [x] Add documentation

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -81,15 +81,13 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
 
   _getBackButtonTitleStyle(scene: NavigationScene): ?Style {
     const lastScene = this._getLastScene(scene);
+    const { headerBackTitleStyle } = this.props.getScreenDetails(scene).options;
     if (!lastScene) {
-      return null;
+      return headerBackTitleStyle;
     }
-    const {
-      headerBackTitleStyle,
-      headerTitleStyle,
-    } = this.props.getScreenDetails(lastScene).options;
+    const { headerTitleStyle } = this.props.getScreenDetails(lastScene).options;
 
-    return headerBackTitleStyle || headerTitleStyle;
+    return [headerTitleStyle, headerBackTitleStyle];
   }
 
   _renderTitleComponent = (props: SceneProps) => {
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -15,7 +15,6 @@ import type {
   NavigationStyleInterpolator,
   LayoutEvent,
   HeaderProps,
-  Style,
 } from '../TypeDefinition';
 
 type SceneProps = {
@@ -79,17 +78,6 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     ).options.headerTruncatedBackTitle;
   }
 
-  _getBackButtonTitleStyle(scene: NavigationScene): ?Style {
-    const lastScene = this._getLastScene(scene);
-    const { headerBackTitleStyle } = this.props.getScreenDetails(scene).options;
-    if (!lastScene) {
-      return headerBackTitleStyle;
-    }
-    const { headerTitleStyle } = this.props.getScreenDetails(lastScene).options;
-
-    return [headerTitleStyle, headerBackTitleStyle];
-  }
-
   _renderTitleComponent = (props: SceneProps) => {
     const details = this.props.getScreenDetails(props.scene);
     const headerTitle = details.options.headerTitle;
@@ -136,7 +124,6 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     const truncatedBackButtonTitle = this._getTruncatedBackButtonTitle(
       props.scene,
     );
-    const backButtonTitleStyle = this._getBackButtonTitleStyle(props.scene);
     const width = this.state.widths[props.scene.key]
       ? (this.props.layout.initWidth - this.state.widths[props.scene.key]) / 2
       : undefined;
@@ -149,7 +136,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
         tintColor={options.headerTintColor}
         title={backButtonTitle}
         truncatedTitle={truncatedBackButtonTitle}
-        titleStyle={backButtonTitleStyle}
+        titleStyle={options.headerBackTitleStyle}
         width={width}
       />
     );
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -15,6 +15,7 @@ import type {
   NavigationStyleInterpolator,
   LayoutEvent,
   HeaderProps,
+  Style,
 } from '../TypeDefinition';
 
 type SceneProps = {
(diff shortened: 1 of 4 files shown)


src/views/Header.js
```diff
@@ -71,6 +76,24 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
...
+    if (!lastScene) {
+      return null;
+    }
+    const {
+      headerBackTitleStyle,
+      headerBackTitle,
+      headerTitleStyle,
+    } = this.props.getScreenDetails(lastScene).options;
+
+    if (headerBackTitle === null) {
```
grabbou: I don't think we need to return `null` if headerBackTitle is `null`. Passing `style` down should be rather independent, right?

src/views/Header.js
```diff
@@ -71,6 +76,24 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
     return this._getHeaderTitleString(lastScene);
   }
 
+  _getBackButtonTitleStyle(scene: NavigationScene): ?Style {
+    const lastScene = this._getLastScene(scene);
+    if (!lastScene) {
+      return null;
+    }
+    const {
+      headerBackTitleStyle,
```
grabbou: I believe that `headerBackTitleStyle` and `headerTitleStyle` you use for styling back button should be taken from a "current" scene, not the previous one. I think as far as styling is concerned, it's responsibility of currently displayed scene to define it.

CC: @satya164

src/views/Header.js
koenpunt: My idea was that there's no use in forwarding the style if the title isn't there anyway. But I can remove the check.

src/views/Header.js
koenpunt: I had it like that, but the `headerBackTitle` is also taken from the previous screen, so changing this would make it inconsistent.

grabbou: The `title` is taken from a previous scene as the scene cannot control what scenes are behind it in the stack. However, it's in charge of its own styling and how the status bar appear. 

Existing approach makes it impossible to set back button style of the same scene different on two screens.

src/views/Header.js
grabbou: I think it's better if that logic remains in `BackButton` component

src/views/Header.js
koenpunt: True, but the purpose of this change is also that it makes the style of the button match while transitioning from one scene to the other. So IMO it should take the style from the same place as where the text is coming from.

src/views/Header.js
koenpunt: Sure, I'll remove that

src/views/Header.js
grabbou: Makes sense. 

Curious to hear what @satya164 and @ericvicenti think about that

koenpunt: Me too, haha. I'll just keep my branch up-to-date with master.

satya164: If you have a dark header in one screen and a light one in previous screen, it doesn't make much sense to inherit styles from previous screen. 

IMO each element in a screen should be themable by that screen only and not by other screens.

grabbou: > True, but the purpose of this change is also that it makes the style of the button match while transitioning from one scene to the other. So IMO it should take the style from the same place as where the text is coming from.

Shouldn't we then assume that style of a back button equals style of a previous scene title?

koenpunt: > If you have a dark header in one screen and a light one in previous screen

Not sure if there's even a usecase for that. What if we apply both the styles? Than the defaults are defined in the last screen, and the next screen can optionally override those.

koenpunt: Because I think in most cases you want the font of the button to match with the title, to have a more-or-less seamless transition.

koenpunt: I've now added [a commit](https://github.com/react-community/react-navigation/pull/1083/commits/54f7eab34f7760d0349dbd0af67184125a0fbf6c) that applies both styles.

satya164: It's pretty common to have different header colors (while/colorful/dark) between different screens.

src/views/Header.js
```diff
@@ -80,6 +81,17 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
...
 
+  _getBackButtonTitleStyle(scene: NavigationScene): ?Style {
+    const lastScene = this._getLastScene(scene);
+    const { headerBackTitleStyle } = this.props.getScreenDetails(scene).options;
+    if (!lastScene) {
+      return headerBackTitleStyle;
+    }
+    const { headerTitleStyle } = this.props.getScreenDetails(lastScene).options;
+
+    return [headerTitleStyle, headerBackTitleStyle];
```
satya164: I'm not sure why anyone will want to style their back button according to the styles of the previous screen's title. You always want the font family and color to be consistent with the current screen. Also merging the previous styles makes it very difficult to override all of them, you never know what the previous screen's style was, so you literally have to override all style properties supported if you want to gurrantee that it'll look like you intended it to.

koenpunt: > You always want the font family and color to be consistent with the current screen.

IMO it's not for nothing that the titles moves to the position of the back button; it's supposed to match up with the title of the last screen, and doing it like this will prevent the need of supplying anything.

What do you think about an option which defines whether or not the style of the previous scene should be used?

koenpunt: Oh and reading back makes me realise that @grabbou and you both have different opinions too?

> Inheriting things like fontFamily from `headerTitleStyle` makes sense.

But I understand that every possible additions have to be thought through. So please speak out your wishes concerning the matter and hopefully can come to something we're all happy with :)

satya164: I haven't used many iOS apps to say how the transition is supposed to happen, but I know that lots of apps have different styling for back button and title. A common example is how the title color is black and the back button has accent color in many apps.

To have a seamless transition, you need to ensure that all your titles and back button title have similar styles, if you don't, either
1. The back button in the current screen will look inconsistent with rest of the items
2. During the transition, two titles with different styles will cross-fade

Now, when you actually have two different styles between screens, and want to have a seamless transition, merging doesn't help, since you'll want to animate previous styles to current styles, not overwrite them.

@grabbou what does normal iOS apps do when 2 different screens have different styles? or is this very uncommon?

satya164: Also font family probably makes sense, but there are so many other styles you don't wanna inherit. I'm interested in what do other iOS apps do

koenpunt: https://www.dropbox.com/s/k11exzy8c370app/nav-transition.mov?dl=0

satya164: @koenpunt app store has the same font family between the 2 screens, so it works fine without merging any styles. I want to see an app with different styles (e.g. different font family) between two screens

also the back button doesn't exactly transition from the title, there seems to be a disconnect between them.

koenpunt: >  I want to see an app with different styles (e.g. different font family) between two screens

I actually don't have any example of such an iOS app.

> also the back button doesn't exactly transition from the title, there seems to be a disconnect between them.

Not exactly, but it fades over, making it look like seamless. But making the transition/animation like the native one is a different issue I guess.

satya164: > I actually don't have any example of such an iOS app.

In that case, I think we don't need to worry about using styles from previous screen's title.

koenpunt:
@grabbou @satya164 I came to the conclusion that inheriting or merging the styles isn't that logical, since the title and the backbutton already have different styles. I overlooked that fact because in the application I'm working on I have the title and the button title styled the same.

That being said, I've reverted it to taking `headerBackTitleStyle` from the current scene, and accept that I have to explicitly set those styles on the backbutton when necessary.

zywj:
So, it also cannot customi headerBackButton or the headerBackButton style ?

PR #1084, opened 2017-04-17
Update flow-react-proptypes to ^1.2.0

joemckie:
Not much info necessary on this one. Fixes the issue discussed in #1045.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "babel-core": "^6.24.0",
     "babel-eslint": "^7.0.0",
     "babel-jest": "^19.0.0",
-    "babel-plugin-flow-react-proptypes": "grabbou/babel-plugin-flow-react-proptypes#release",
+    "babel-plugin-flow-react-proptypes": "^1.2.0",
     "babel-preset-es2015": "^6.24.0",
     "babel-preset-react": "^6.23.0",
     "babel-preset-react-native": "^1.9.0",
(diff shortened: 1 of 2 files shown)


AlanFoster:
@joemckie Just tried this out locally and I'm getting the issue:

>  error: bundling: ReferenceError: Unknown plugin "flow-react-proptypes" specified in "/Users/dev/Documents/foo/node_modules/react-navigation/.babelrc.env.development" at 0, attempted to resolve relative to "/Users/dev/Documents/foo/node_modules/react-navigation"

joemckie:
@AlanFoster you're right - it looks like the dependency isn't being installed correctly. I tested and moving it to `dependencies` works, but it should work as it was in `devDependencies` before.

I don't think the other issue you're having is related. That looks like you've updated react, as they've removed `PropTypes` from the latest version and moved it to another dependency.

For me, using this PR as my `react-navigation` dependency and installing the `babel-plugin-flow-react-proptypes` dependency worked.

@grabbou, any idea why it isn't being installed correctly?

AlanFoster:
Thanks for the investigation @joemckie :+1: Pointing this to master after the recent changes seems more stable.

I can now also confirm that the `PropTypes` error was unrelated, I had copied the code samples from the redux examples but missed out the line where it explicitly imports `PropTypes`:

```
import React, { PropTypes } from 'react';
```

I was able to debug this by stepping through the module loading/factory logic manually :+1:

PR #1091, opened 2017-04-18
Fix logger

grabbou:
Reimplements #861, as suggested in comments -> https://github.com/react-community/react-navigation/pull/861#issuecomment-292598320

### DIFF
--- a/docs/api/navigators/Navigators.md
+++ b/docs/api/navigators/Navigators.md
@@ -48,9 +48,11 @@ When rendering one of the included navigators, the navigation prop is optional. 
 
 For the purpose of convenience, the built-in navigators have this ability because behind the scenes they use `createNavigationContainer`. Usually, navigators require a navigation prop in order to function.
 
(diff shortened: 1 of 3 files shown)


grabbou:
CC: @ericvicenti we have agreed with @satya164 that it might make more sense to eventually `abandon` containerOptions in favour of defining props in-place. 

All in all, `navigator` can be rendered either by user directly (which promotes it to a top-level navigator) or by us (which on the other hand makes it controlled navigator that receives `navigation` prop from a parent). 

That scenario gives us an ability to define all properties directly on a component as we control *when* navigator is controlled internally (we can place invariants that will guide advanced users what to set / what not when doing it manually with e.g. Redux)

ericvicenti:
Yeah, I agree and think it makes sense to deprecate the static `navigationOptions` in favor of props. It is pretty confusing to have both. I agree we can still provide the same quality of validation even when we have several navigation container props.

grabbou:
Don't merge before #1100 lands.

PR #1093, opened 2017-04-18
Upgrade React Native

grabbou:
That upgrades React Native for `NavigationPlayground` and fixes not working debugger menu.


grabbou:
Next PR will upgrade to latest React Native and fix remaining flow issues.

PR #1096, opened 2017-04-18
Use the title text style for the back button text.

JodiWarren:
The back text should match the Header title. While it was possible to set custom styles typeface (i.e. typeface) for the title, the back button text was impossible to style.

**Test plan (required)**

Add the following after line 70 of `examples/NavigationPlayground/js/SimpleStack.js`:
```js
    headerTitleStyle: {
      textDecorationLine: 'line-through',
      letterSpacing: 2,
      fontWeight: '500',
    },
```

You should then have something a little like this:
styled-back-button


koenpunt:
This is already being tracked in #1083

PR #1100, opened 2017-04-18
Breaking: Replace `containerOptions` with just `props`

grabbou:
This is a breaking change but eventually, the API will become simpler.


EmielM:
@grabbou This breaks our app at places where we insert a (stateless) navigator between components with the "both navigation/container props set" assertion @ https://github.com/react-community/react-navigation/blob/master/src/createNavigationContainer.js#L84.

That's because the `router` prop is set, but that's perfectly valid for a stateless container, no?

PR #1115, opened 2017-04-19
Add an ability to disable built-in logger

grabbou:
This PR moves logger to `onNavigationStateChange` as a default implementation. You can turn it off by passing your own function or a `null`.

Breaking: `onNavigationStateChange` cannot be set on a controlled component anymore. If you are passing `navigation` prop yourself to a Navigator, you can utilise `componentDidUpdate` of an outer component. This prop is now reserved for `top-level` navigators that are stateful and control their state internally.

### DIFF
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -19,7 +19,7 @@ import type {
 
 type NavigationContainerProps = {
   uriPrefix?: string,
-  onNavigationStateChange?: (NavigationState, NavigationState) => void,
+  onNavigationStateChange?: (NavigationState, NavigationState, NavigationAction) => void,
 };
 
 type Props<T> = NavigationContainerProps & NavigationNavigatorProps<T>;
@@ -75,7 +75,7 @@ export default function createNavigationContainer<T: *>(
       }
 
       const {
-        navigation, screenProps, navigationOptions, onNavigationStateChange,
+        navigation, screenProps, navigationOptions,
         ...containerProps
       } = props;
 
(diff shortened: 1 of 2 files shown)


src/createNavigationContainer.js
```diff
@@ -160,16 +175,7 @@ export default function createNavigationContainer<T: *>(
       }
       const nav = Component.router.getStateForAction(action, state.nav);
```
satya164: We should use the `setState(updater)` pattern here since the new state depends on the old state.

grabbou: Good point, worth refactoring when as we work on it. I'll get back to that later today.

grabbou: I am suggesting to defer this change to a separate issue as there are few concerns raised, including the synchronous return value of `dispatch` that is used in e.g. `navigate` as a return value.

Not sure how to handle such case. Rather than that, this should be good to go.

grabbou:
@satya164 @ericvicenti there's nothing more we can do in this PR, can we move it forward?

cooperka:
@grabbou It seems this PR [removed `onNavigationStateChange`](https://github.com/react-community/react-navigation/pull/1115/files#diff-cbc20970cef83ca314bfdb0359f01115L78) from the navigation props and now I get the invariant warning saying I can't use both `navigation` and `onNavigationStateChange` props on master. Was this an intentional change?

cooperka:
Nevermind, I read your PR description again and I understand.

Issue #709, opened 2017-03-16
iOS: DrawerNavigation inside stackNavigator doesn't work properly


aarjithn:
I have an app with below structure

Mainscreen (Stacknavigator)
  - Homepage  
 - Chatpage - (DrawerNavigator)
       - Homescreen
       - Notifications
  - Infopage

In iOS, there are couple of issues
 - during transition to/from Chatpage, the drawer is visible, even if it is in closed state
 - when drawer is open, the entire screen gets dimmed apart from the header. Setting headerMode to none in the stacknavigator doesn't solve this.

Everything works correctly in Android

![drawernav](https://cloud.githubusercontent.com/assets/10304259/23995701/b333aede-0a70-11e7-8aac-0e06e2a8afd3.gif)

satya164:
cc @DanielMSchmidt

prescottprue:
Also seeing this

DanielMSchmidt:
@satya164 I just found the time to reproduce / debug this. From the DrawerLayout perspective the current behavior is as expected. From the react-navigation perspective I see that this is a problem. I have not that much experience with the architecture of this project, so it is hard for me to debug and pin the problem. On a high level perspective we would need to wrap the DrawerLayout as outermost navigation component. I don't know if there is a good and clean way with the current pluggable architecture to really solve this problem. IMHO the following steps would be the best to "resolve" this issue:

1) Add a note in the docs, I guess in the Drawer part that states that the Drawer should be the outmost navigation component to achive a classic drawer behavior
2) Add an example in the playground for a Stack navigation nested in a Drawer
3) If we wish to encourage both ways also provide the opposite example

DanielMSchmidt:
The other problem with the drawer being shown on navigation may be a real issue, though I could not verify it with the "correct" order. I will try to hunt this bug down in https://github.com/react-native-community/react-native-drawer-layout/issues/26

ng-hai:
I have to set the transition on StackNavigator to `{ mode: 'modal' }` to hide this bug in my app while waiting for it fixed.

SamMatthewsIsACommonName:
I wanted to keep the modal behaviour for our actual modals so I put the drawer on the right side in the meantime... It works but I think will be strange for our users when it shifts ;)

DanielMSchmidt:
[...]

PR #1121, opened 2017-04-19
docs: add remark about nesting

DanielMSchmidt:
This relates to #709, adding a remark about the consequences of nesting the `DrawerNavigation`  may help users in the future.


grabbou:
Great! Is this going to fix the 709? To me, it is not an error but a consequence of nesting, so a simple note should be enough.

DanielMSchmidt:
@grabbou Yeah, it will fix #709. I share your opinion, stating it may help users to understand this consequence.

PR #1137, opened 2017-04-20
Add a config to override the truncated back title

### LINKED ISSUE #1136: Override the default truncated back button title
Introduced in #257, the truncated back button title is displayed when the back button title is too long to fit in the header. This is useful because it might not be possible to predict the length of the title, e.g. if it's translated to many languages or if the title is created dynamically from data (e.g. name of the item displayed). It might also only be truncated on devices with smaller screens.

Currently `truncatedTitle` is hard-coded to "Back" and there is no configuration option to override that default. Because apps can be translated to other languages than English, it's necessary to be able to replace the text with a localized string.

I propose we add a new configuration option `headerTruncatedBackTitle` that can be used to configure the truncated back title.

fson:
Fixes #1136.

Currently `truncatedTitle` is hard-coded to `"Back"` and there is no configuration option to override that default. Because apps can be translated to other languages than English, it's necessary to be able to replace the text with a localized string.

**Test plan (required)**

I tested the config by adding a translated  `headerTruncatedBackTitle` in the stack navigation example:
```diff
diff --git a/examples/NavigationPlayground/js/SimpleStack.js b/examples/NavigationPlayground/js/SimpleStack.js
index 3ab204f..d10648e 100644
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -89,6 +89,10 @@ const SimpleStack = StackNavigator({
     path: 'photos/:name',
     screen: MyPhotosScreen,
   },
+}, {
+  navigationOptions: {
+    headerTruncatedBackTitle: 'Takaisin',
+  },
 });

 export default SimpleStack;
```
The back title was displayed correctly:
simulator screen shot 20 apr 2017 15 15 32

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -123,6 +123,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
         pressColorAndroid={options.headerPressColorAndroid}
         tintColor={options.headerTintColor}
         title={backButtonTitle}
+        truncatedTitle={options.headerTruncatedBackTitle}
         width={width}
       />
     );
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -107,6 +107,10 @@ String or React Element used by the header. Defaults to scene `title`
 
 Title string used by the back button on iOS or `null` to disable label. Defaults to scene `title`
 
+#### `headerTruncatedBackTitle`
+
+Title string used by the back button when `headerBackTitle` doesn't fit on the screen. `"Back"` by default.
+
 #### `headerRight`
 
 String or React Element to display on the right side of the header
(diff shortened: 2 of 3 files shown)


src/views/Header.js
```diff
@@ -123,6 +123,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
         pressColorAndroid={options.headerPressColorAndroid}
         tintColor={options.headerTintColor}
         title={backButtonTitle}
+        truncatedTitle={options.headerTruncatedBackTitle}
```
grabbou: Actually I believe we should read it from precious scene, just like we do for back title. Didnt notice that while reviewing. Mind creating a new PR to update it?

grabbou: *previous

fson: Ah, that's right. I'll fix this in a new PR.

fson: Fixed in #1167.

PR #1144, opened 2017-04-20
Fix the ReduxExample app (3)

cooperka:
This simply upgrades dependencies to fix the ReduxExample and LinkingExample apps. The react-native versions were out of sync, causing the errors described in https://github.com/react-community/react-navigation/issues/550 (`Expected array with count X, but count is Y`).

Perhaps we should add example app testing to the CI to prevent these apps from continually breaking? I'm not familiar with CircleCI but I'd be happy to try if nobody has objections.

Fixes https://github.com/react-community/react-navigation/issues/550.


grabbou:
`LinkingExample` is going to be integrated with the `NavigationPlayground`, so that should be less visible.

I agree it would be great to have some sort of a check to make sure Redux example also works. 

Thanks for keeping an eye on it!

PR #1147, opened 2017-04-20
Respect `gesturesEnabled`

maxkomarychev:
`gesturesEnabled` is broken since #984 (https://github.com/react-community/react-navigation/pull/984/files#diff-1bb911cc75b3e921fd5d4db86802c58cL450).


maxkomarychev:
I'm fairly new to this project please suggest better way to retrieve options (if any) because original piece via `this.props.router.getScreenConfig` is not working in current codebase. Thanks

alvaromb:
This used to work for changing `gesturesEnabled`

```js
const AuthNavigator = StackNavigator(
  {
    Auth: { screen: Auth },
    AppSelector: { screen: AppSelector }
  },
  {
    initialRouteName: 'Auth',
    headerMode: 'none',
    navigationOptions: {
      cardStack: {
        gesturesEnabled: false
      }
    }
  }
)
```

But it doesn't under `1.0.0-beta.8`.

nihgwu:
I just planned to send a PR to fix this yesterday 👍 

But I have to ask that what's happened with reviews, PRs from the others need to wait for weeks to be viewed or shipped, but too easy if it's from core contributor, I'm not saying it's unfair, I just want to say it's really ridiculous  that this kind of regressions happen again and again and again

https://github.com/react-community/react-navigation/commit/11cab8eab62b6ca1c2a0445f30a9d0dc75458e2e this commit breaks a lot of gesture things, like `gesturesEnabled` becomes not configurable and disabled for `modal` mode, and `Animated.spring` to `Animated.timing` is totally wrong, `transitionConfig` is broken too witch is just fixed by #99 that is such a tiny but importance change but wait for nearly a month to be merged, but reverted in seconds by you guys, so ironic.

then #886 fixed `gesturesEnabled` configuration, but breaks by #984 again, then we get this PR to **fix** this again, don't you guys think it's really silly? 

@ericvicenti @grabbou @satya164 @skevy

nihgwu:
@maxkomarychev You should update the docs too, `gesturesEnabled` is only available for `StackNavigator`, and only for `card` mode now (regression from  https://github.com/react-community/react-navigation/commit/11cab8eab62b6ca1c2a0445f30a9d0dc75458e2e), so you couldn't simply read it from options

alvaromb:
Just wanted to say that tomorrow I'll have some spare time to help with anything regarding this issue. Ping me if I can give a hand with anything. I haven't dived into this issue because I discovered it when leaving the office, so I have to admit I don't have all the context to solve it yet.

maxkomarychev:
Guys, I have another rather fundamental question: why does the original implementation enables this feature by default only for card mode on iOS. Why don't enable this on android by default as well? Any thoughts?

src/views/CardStack.js
```diff
@@ -321,10 +321,14 @@ class CardStack extends Component {
         });
       },
     });
-    
-    const gesturesEnabled = mode === 'card' && Platform.OS === 'ios';
+
+    const { options } = this._getScreenDetails(scene);
+    const gesturesEnabled = mode === 'card' && (
```
nihgwu: What about to add a `invariant` to warn about the mode case?

docs/api/navigators/StackNavigator.md
```diff
@@ -137,7 +137,7 @@ Color for material ripple (Android >= 5.0 only)
 
 #### `gesturesEnabled`
 
-Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android
+Whether you can use gestures to dismiss this screen. Defaults to true on iOS, false on Android. This value when set is only respected in `mode` == `card`.
```
nihgwu: this is `navigationOptions` for screen, should be removed, then add it to `StackNavigatorConfig`

maxkomarychev: @nihgwu Not sure I am following you here. Are you saying that `gesturesEnabled` should be configured via `StackNavigatorConfig`? Having it in `navigationOptions` allows you to override this setting per screen. To me it's pretty much the same as being able to show/hide the navbar itself.

nihgwu: My fault, but I think an extra `StackNavigatorConfig` would be much better, on Android, gestures are disabled by default, I have to add `gesturesEnabled` to all the screens in the stack to make them have the same behavior on both iOS and Android, I think that's insane, so I propose we could get the default value from `StackNavigatorConfig`

ericvicenti: @nihgwu, This should already work, to set it in the StackNavigatorConfig as you mentioned:

```
const App = StackNavigator({
  ...
}, {
  navigationOptions: {
    gesturesEnabled: false,
  }
});
```

The default value for every screen navigationOptions comes from the navigator.

nihgwu: that's great, I used to set it in this way:
```
    navigationOptions: {
      cardStack: {
        gesturesEnabled: true,
      },
    },
```

ericvicenti:
Ok, lets get this in for now. Eventually we can bring back the modal gestures

maxkomarychev:
thanks guys!

docs/api/navigators/StackNavigator.md
alvaromb: @ericvicenti in beta 8 neither your approach or @nihgwu's one are not working, the gestures are not disabled in both cases. Does this PR solves this issue?

grabbou:
You should try `master` to verify if it's still the case.

alvaromb:
I can confirm this works with latest `master`

```
navigationOptions: {
  gesturesEnabled: false
}
```

Not this:

```
navigationOptions: {
  cardStack: {
    gesturesEnabled: true
  }
}
```

Thanks for your work!

onlydave:
End of the day kind of stab at getting gestures working for modals again. Is this the direction your looking at going in? I'm not up to speed on why it was regressed originally.

@ericvicenti @grabbou @satya164 @skevy 

https://github.com/react-community/react-navigation/compare/master...onlydave:master

Ignore the `TabViewPager = TabViewPagerScroll;` We need nested tabs and that's the quickest way we could get them working on android.

xiaoyongchen:
@alvaromb  Just want that effect

PR #1155, opened 2017-04-21
Force the back button to be recognized as such by screen readers.

cannona:
The back button is not recognized as a button otherwise, so users may not realize they can tap it.


grabbou:
Not an expert with accessibility, but shouldn't we specify `label` as well, to indicate what readers should read?

grabbou:
Okay, I need to run `prettier` on this diff manually and copy changes. It's hard to satisfy `eslint` from here.

grabbou:
@ericvicenti this is now green 🥇

Issue #485, opened 2017-02-26
cannot read property routes of undefined - navigation state seems to be getting lost 


chadi-kazan:
using drawer navigation with redux integration, the state is for some reason not getting properly passed down to my components - if i properly understand the error. the following is the navigation object i get when _updateNavigationProps is called in withCachedChildNavigation
```javascript
{
    dispatch:(action)
    goBack:goBack(key)
    navigate:navigate( routeName, params, action)
    setParams:setParams(params)
    state:undefined
```

i did my best to follow the ReduxExample app
here are my navigator
```javascript
    const AppNavigator = DrawerNavigator({
        home: {
            screen: Home
        },
        settings: {
            screen: Settings
        },
        login: {
            screen: LoginScreen
        }
    }
```
my navigation reducer

```javascript
const INITIAL_STATE = {
	index: 1,
    routes: [
        { key: 'InitA', routeName: 'home' },
        { key: 'InitB', routeName: 'settings' },
        { key: 'InitC', routeName: 'login' }
    ]
};
export default function navigation(state = INITIAL_STATE, action) {
    switch (action.type) {
            case 'login':
                return AppNavigator.router.getStateForAction(NavigationActions.back(), state);
            case 'home':
            case 'settings':
                return AppNavigator.router.getStateForAction(action, state);
        }
    return state;
}
```

and my app container

```javascript
@connect(state => ({nav: state.navigation, navigation: state.navigation, data: state.data}))
export default class AppContainer extends Component {
	render(): ReactElement {
		const { dispatch, nav } = this.props;
		const navigation = addNavigationHelpers({dispatch, state: nav});
		return <AppNavigator navigation={navigation} />;
	}
}
```

the render function of AppContainer never gets called by the way, the error occurs before.
sorry to post here, but i am unable to find any other forums to get help on this

grabbou:
Please publish entire example. The code looks good at first sight, might be easier to check when we have a runnable code.

If you know `exponent`, try posting it there.

chadi-kazan:
[here](https://github.com/chadi-kazan/bookmarker)'s the github repo for the example

stief510:
I am having a similar problem #427

chadi-kazan:
ok so i replaced DrawerNavigator with TabNavigator, using the exact same routes, reducers and so on, i didn't change anything else in my routing/redux configuration and TanNavigator works!
[...]

PR #1172, opened 2017-04-22
Improve the Redux docs

cooperka:
Things done in this PR:

1. Update the ReduxExample app to be more robust, using `getStateForAction` and checking for `null` states
2. Document the use of `getStateForAction` with redux
3. Link to the ReduxExample app in the docs
4. Clean up the Contributors guide, and make sure there's only one source of truth for how to run the examples

Among other things, this should help clarify some of the confusions expressed in https://github.com/react-community/react-navigation/issues/485.


grabbou:
Can we get this rebased on top of master (remove `LinkingExample`)?

cooperka:
@grabbou done.

examples/ReduxExample/src/reducers/index.js
```diff
@@ -4,27 +4,31 @@ import { NavigationActions } from 'react-navigation';
...
-    { key: 'InitA', routeName: 'Main' },
-    { key: 'InitB', routeName: 'Login' },
-  ],
-};
-
-const initialAuthState = { isLoggedIn: false };
+const firstAction = AppNavigator.router.getActionForPathAndParams('Main');
+const tempNavState = AppNavigator.router.getStateForAction(firstAction);
+const secondAction = AppNavigator.router.getActionForPathAndParams('Login');
+const initialNavState = AppNavigator.router.getStateForAction(secondAction, tempNavState);
```
grabbou: I get the point, tho I believe this will be more complex for new-commers than the plain object.

cooperka: Unfortunately, I agree it's more complicated, but it does also prevent silly user errors. Given the number of people who couldn't figure out how to use it and got errors like in #485 it seems like this is a better approach. I'd love to write it a simpler way if possible -- any ideas?

cooperka: I could simply leave the initial state there as comments, to clarify what this logic all turns into. Or I could add a helper method that generates a state based on an array of actions --
 e.g. `AppNavigator.router.getInitialStateForPath('Main')` which could also take an array `AppNavigator.router.getInitialStateForPath(['Main', 'Login'])`. Seems like a very common pattern that could warrant a helper method like this.

cooperka: Though perhaps they could go in a future PR so we get this in sooner? Imo this is better than the current state.

grabbou: All righty, I'll leave that up to you since I don't use `Redux` with `react-navigation` actively.

cooperka: Ok sounds good, I'll see what I can do.

PR #1173, opened 2017-04-22
Improve CircleCI infrastructure

cooperka:
Things done in this PR:

1. Just specify node version `6`, and let CircleCI default to the most recent version
2. Add very basic Jest tests to NavigationPlayground
3. Fix the existing Jest tests on ReduxExample
4. Run the example app tests during the CI
    - Tests are run in a `post` script after the main tests have passed, so this shouldn't slow that part down
    - Because of the extra tests, the total CI length does increase by a few minutes; if this becomes problematic, it could be solved by using more parallel builds, or a service like Travis which runs all tests concurrently by default
5. Remove LinkingExample entirely, because it's now part of NavigationPlayground


grabbou:
Can we get this branch rebased and conflicts resolved? Also, let's not include `LinkingExample` in the test pipeline, it was meant to be removed.

Actually I was planning to do it soon, but I think you could try including it in this PR.

cooperka:
@grabbou LinkingExample is not being tested. I'll remove it entirely!

cooperka:
Rebased again after changes to `circle.yml`. Should be good to merge.

grabbou:
Looks good to me. @ericvicenti that looks good to me, what you think?

Issue #1060, opened 2017-04-15
Official documentation does not apply to version fetched from NPM


przemkow:
could you update package version on NPM? 
in official docs you can find code like this with already implemented flat options: 
```
class ProfileScreen extends React.Component {
  static navigationOptions = ({ navigation, screenProps }) => ({
    title: navigation.state.params.name + "'s Profile!",
    headerRight: <Button color={screenProps.tintColor} {...} />,
  });
```
while in your app you still need to write code like bellow because version fetched from NPM does not include flat options feature.
```
class ProfileScreen extends React.Component {
  static navigationOptions = {
    title: navigation.state.params.name + "'s Profile!",
    header: (navigation) => ({
      left: <Button color={screenProps.tintColor} {...} />,
    })
  };
```

henjing:
Indeed it is

DiogoAbu:
With [this commit](https://github.com/react-community/react-navigation/commit/93976d358e27580e568ef016f820dc0253423aa7) they changed the docs.
You can change your package json to fetch directly from git (7165efc is the latest commit):
`"react-navigation": "git+https://github.com/react-community/react-navigation.git#7165efc",`

przemkow:
@DiogoAbu that's exactly what i did, i even left the comment under this commit. 
The problem was that it took me some time to get why official dosc isn't working with my app so it might discourage new users from using the lib :)

grabbou:
We should probably version the website. Not sure if it's possible now, but since we are using similar engine to `react-native`, it should be fairly doable.

Going to put it on a `todo` list.

Alternatively, we should consider deploying website on release rather than on master commit.

akrigline:
Chiming in here, this really threw me off. Documentation being wrong is a big hurdle for people trying this lib out to jump. That commit changelog was super helpful @DiogoAbu.

So is the website's docs more Up to date than the stuff that's in NPM?

grabbou:
Yes, currently website follows master API, which we are likely to change as
it's confusing.

do-n-khanh:
as a new react-native user, this problem almost discourages me to continue to learn react. Luckily, someone on StackOverFlow sent me the link to this page. Please update package version on npm

grabbou:
Apologies for the inconvenience. This release is special in a sense that we
have shipped few breaking changes hence painful website differences.

I'll try to solve this problem as soon as possible.

paradigmist:
Any updates?

[...]

PR #1174, opened 2017-04-23
Add documentation versioning

samijaber:
### Screenshot
screen recording 2017-04-22 at 09 38 pm

### Overview

- improves `build-docs` script to include the versioning of documentation, with `edge` being the cutting-edge, yet-unreleased code documentation.
- adds a `VersionSelector` dropdown menu to allow users to choose the documentation version.
- provide the documentation JSON as a navigation parameter in the website.
- retroactively adds documentation for `1.0.0-beta.7` and `1.0.0-beta.8`. I downloaded the zipped release for `1.0.0-beta.7` and used the `docs-dist.json` that exists there. As for `1.0.0-beta.8`, I checked out this commit e94c9c1061151a1478b47640710bcafdbe738848 and took the documentation from there.
- fixes https://github.com/react-community/react-navigation/issues/1060

Now, when `npm run build-docs` (or `build` or `prepublish`) are executed, the `compile-docs` script will do three things:

- First, the current `edge.json` is renamed to the current version (based on whatever version is in `package.json`). If a file for the current version already exists, it will overwrite it.
- Second, a new `edge.json` is built (this step is what's already there)
- Last, the `docs/index.js` file is rebuilt, in case the 1st step resulted in a new file being created

### To-do

- [ ] improve styling of `VersionSelector` (I wanted to make sure this PR is created first so that I can get feedback on the script while I tinker with the styling).


samijaber:
cc @grabbou : you mentioned trying to solve this in #1060. Take a look at this PR and let me know if it already does the trick! 🙂

grabbou:
Hey, thanks for taking a stub on this! Overall, looking good, I am leaving few comments here and there.

website/src/AppFrame.js
```diff
@@ -36,26 +43,105 @@ const NavigationLinks = ({navigation, className, reverse}) => {
...
+  }
+
+  toggleMenuShown() {
+    this.setState({ menuShown: !this.state.menuShown });
+  }
+
+  handleVersionClick(version) {
+    const { setParams } = this.props.childNavigation;
+    return () => {
+      setParams({ version });
```
grabbou: We shouldn't return a closure here, see comment below where we call onPress.

website/src/AppFrame.js
```diff
@@ -36,26 +43,105 @@ const NavigationLinks = ({navigation, className, reverse}) => {
...
+      isUndef(state.params.version)
+    );
+
+    const currentVersion = docVersionIsUndefined(state) ? 'edge' : state.params.version;
+    const versionsList = Object.keys(docs).map((version) => (
+      <li
+        key={version}
+        onClick={this.handleVersionClick(version)}
+        style={{
+          color: (currentVersion === version) ? '#6b52ae' : '#555'
```
grabbou: nit: `( )` seem unnecessary?

website/src/AppFrame.js
```diff
@@ -36,26 +43,105 @@ const NavigationLinks = ({navigation, className, reverse}) => {
...
+    ));
+
+    return (
+      <div className='versionselector'>
+        <span className="versionbutton" onClick={this.toggleMenuShown}>
+          version: {currentVersion} ▼
+        </span>
+        <ul
+          style={{
+            //display: menuShown ? 'inline' : 'none',
```
grabbou: nit

website/src/AppFrame.js
```diff
@@ -36,26 +43,105 @@ const NavigationLinks = ({navigation, className, reverse}) => {
...
 class AppFrame extends React.Component {
-  state = {isMobileMenuOpen: false};
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isMobileMenuOpen: false,
+    };
+
+    this.setVersion = this.setVersion.bind(this);
```
grabbou:
```
setVersion = (version) => {

}
```
and we don't have to `bind`?

website/src/AppFrame.js
```diff
@@ -36,26 +43,105 @@ const NavigationLinks = ({navigation, className, reverse}) => {
...
+  
   componentWillReceiveProps(props) {
     if (this.props.navigation.state !== props.navigation.state) {
       this.setState({isMobileMenuOpen: false});
       window.scrollTo(0,0);
     }
   }
+  
+  setVersion(version) {
+    return () => this.setState({ version });
```
grabbou: Same as previously, no need to return function here. Prefer `onPress={() => setVersion(version)}` instead

website/src/MDPage.js
```diff
@@ -29,9 +27,20 @@ const getHeadingForLevel = (level) => {
   }
 };
 
+const isDefined = (obj) => obj !== undefined;
+
+const getDocsMD = (docPath, navigation) => {
+  const { state } = navigation;
+  const docVersion = (isDefined(state) && isDefined(state.params))
```
grabbou: AFAIK `navigation.state` is always defined. We can only check for `state.params && state.params.version`

website/src/MDPage.js
```diff
@@ -29,9 +27,20 @@ const getHeadingForLevel = (level) => {
...
+    ? state.params.version
+    : 'edge';
```
grabbou: That logic is already defined in two places - here and in version selector. We could either use `initialRouteParams` for cases when params are not set?

ericvicenti:
Could we possibly change the design to integrate the dropdown into the "docs" link, rather than put this in the center of the header? When the docs link is inactive, it can mention the version, like "DOCS - 1.0.0-beta.8", and when it is active, it has the dropdown functionality.

Also I notice this does not integrate with the path, so deep links may be broken if you attempt to link to an old version of the docs.

Be sure to test this on mobile, and post a screenshot of that behavior.

grabbou:
@samijaber will you have time soon to work on this? If not, let me know and I'll be happy to take it. We need working switch for versions soon as we keep development more and more active!

samijaber:
@grabbou I'm unfortunately somewhat busy over the next few days. Feel free to take over and make the changes you want! Otherwise, I might be able to take a look in a week or so.

kelset:
Hey @grabbou @samijaber is this PR worth to keep open? No updates since April and the issue mentioned #1060 is closed atm.

brentvatne:
wow, thank you so much for doing this, sincere apologies that nobody was able to land it. we're in the process of moving over to using docusaurus for our website/docs, so this PR is no longer relevant, you can follow along at http://github.com/react-navigation/website

PR #1187, opened 2017-04-24
Add support for custom transitionConfig

### LINKED ISSUE #1191: Please, can you add android transitioner like ios?
I read the source code and found that ios can use two transition animations by set mode to 'card' or 'modal',but android can't.
```
    // Use the default Android animation no matter if the screen is a modal.
    // Android doesn't have full-screen modals like iOS does, it has dialogs.
    if (prevTransitionProps && (transitionProps.index < prevTransitionProps.index)) {
      // Navigating back to the previous screen
      return FadeOutToBottomAndroid;
    }
    return FadeInFromBottomAndroid;
  } else {
    // iOS and other platforms
    if (isModal) {
      return ModalSlideFromBottomIOS;
    } else {
      return SlideFromRightIOS;
    }
  }
```
So, can you add android transition animations like ios by set mode to 'card' or 'modal',please?

nihgwu:
`transitionConfig` is broken by #984, this fix it

Fixes #1191


grabbou:
Can we make sure flow tests are passing?

grabbou:
One more thing: Can we make sure this is documented in StackNavigator options, like here: https://github.com/react-community/react-navigation/pull/468/files#diff-d8149c156a733c684c5a21d33a7732daR90 ?

docs/api/navigators/StackNavigator.md
```diff
@@ -85,7 +85,7 @@ Visual options:
   - `screen` - Each screen has a header attached to it and the header fades in and out together with the screen. This is a common pattern on Android.
   - `none` - No header will be rendered.
 - `cardStyle` - Use this prop to override or extend the default style for an individual card in stack.
-- `transitionConfig` - Function to return an object that override default screen transitions.
+- `transitionConfig` - Function to return an object that overrides a default screen transitions.
```
nihgwu: a transitions?

nihgwu:
@grabbou 👍

MiLeung:
Where are the docs for transitionConfig?  All I can find is that it is a "Function to return an object that overrides default screen transitions."

What are the properties of the object being returned?

nihgwu:
``` js
  // custom Modal transition animation
  transitionConfig: () => ({
    transitionSpec: {
      duration: 250,
      easing: Easing.out(Easing.poly(4)),
      timing: Animated.timing,
    },
    screenInterpolator: sceneProps => {
      const { layout, position, scene } = sceneProps
      const { index } = scene

      const height = layout.initHeight
      const translateY = position.interpolate({
        outputRange: [height, 0, 0],
      })

        inputRange: [index - 1, index - 0.99, index],
        outputRange: [0, 1, 1],
      })

      return { opacity, transform: [{ translateY }] }
    },
  }),
```

Amerr:
@nihgwu  can you explain what is  ***screenInterpolator*** and its purpose when its invoked ?

nihgwu:
@Amerr http://www.reactnativediary.com/2016/12/20/navigation-experimental-custom-transition-1.html

voidstarfire:
Where do we get these imports from like "Easing" ?

skoob13:
@voidstarfire do you mean `import { Easing } from 'react-native'`?

kalraneeraj24550:
tansitionConfig working well, if i use stackNavigator, but it is not working when i work with StackRouter, anybody plz help

brentvatne:
please stop spamming @kalraneeraj24550

Issue #1034, opened 2017-04-13
Migrate `website` and `examples` to `prop-types`


grabbou:
We already did that for `src` in https://github.com/react-community/react-navigation/pull/1033. Currently, it uses `React.PropTypes`.

Chinchila01:
I did this in a new branch, how do I push it to git to create a new pull request?
Thanks

kelset:
Quite sure this is fixed. I'll reopen if needed.

PR #1190, opened 2017-04-24
Migrate 'website' and 'examples' to 'prop-types'

azisuazusa:
Implements [#1034](https://github.com/react-community/react-navigation/issues/1034)

Migrate website and examples to  prop-types


grabbou:
Thanks for the PR. However, this is unfortunately not what I was planning to do initially. My idea was to replace `propTypes` entirely with `Flow` and optionally, use `babel-plugin-flow-prop-types` to generate them automatically.

Would you be open to doing such rewrite?

PR #1199, opened 2017-04-24
Add issue template

grabbou:
Adding an issue template we got over in Haul. Hopefully, it helps people to group issues in a better way and provide more info on what is not working.

There are issues that are really hard to understand and info about navigation version is missing.

CC: @satya164


AlanFoster:
It might be good suggesting to add screenshots of issues too. There's been a bunch of cryptic issues on react-navigation such as 'text not appearing in expected place', which have been incredibly hard to understand without further asking for a screenshot.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
+<!---
+BEFORE YOU SUBMIT please search open/closed issues since someone might have asked the same thing before!
+-->
+
+### Current Behavior
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
```
AlanFoster: Should we ask developers to label their own PR in this case?

```
### Type of issue

<!---  Bug | Feature Request | Help --->
```

That extra data might help active developers of react-navigation to see if this is a production issue or just a clarification on documentation etc.

AlanFoster: In the future a bot could rip out this information to automatically add github labels

grabbou:
> It might be good suggesting to add screenshots of issues too. There's been a bunch of cryptic issues on react-navigation such as 'text not appearing in expected place', which have been incredibly hard to understand without further asking for a screenshot.

Good idea, I mention screenshots!

.github/ISSUE_TEMPLATE.md
grabbou: I like the idea, but would prefer to keep the template itself minimal. People usually have that tendency to either skip or not fill in the template. The `current` and `expected` behaviour are formulated in a way that should encourage people to use it. 

Time will show 💃

grabbou:
@ericvicenti updated the wording as requested 💯

PR #1203, opened 2017-04-24
Update docs on headerLeft/Right

### LINKED ISSUE #1124: [Docs] Inaccurate docs for `navigationOptions`
Docs for `navigationOptions` say

```
**headerRight** 
String or React Element to display on the right side of the header

**headerLeft** 
String or React Element to display on the left side of the header
```

Apparently this is not true - it fails on strings.

react-native: 0.43.3
react-navigation: https://github.com/react-community/react-navigation.git#4c55887

grabbou:
Fixes #1124 - regression after flat options.

CC: @ericvicenti

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -114,11 +114,11 @@ Title string used by the back button when `headerBackTitle` doesn't fit on the s
 
 #### `headerRight`
 
-String or React Element to display on the right side of the header
+React Element to display on the right side of the header
 
 #### `headerLeft`
 
-String or React Element to display on the left side of the header
+React Element to display on the left side of the header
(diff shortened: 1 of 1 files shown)


ericvicenti:
Maybe we should just actually support strings? That way we can bring the options to parity with Airbnb Native navigation

grabbou:
Not sure what would be the use case for a string in `headerLeft` though. Maybe Airbnb navigation does something different with them? e.g. takes it and makes a button with that label?

I'd be open to discussing it as a feature request - that PR is just to remove confusion :)

ericvicenti:
Ok, merging now to reduce immediate confusion.

grabbou:
Created #1215 to follow up.

PR #1208, opened 2017-04-24
Fix transitioner.md route

jcsmesquita:
Documentation link to transitioner is wrong. This PR fixes it.


grabbou:
Are you sure? -> https://reactnavigation.org/docs/views/transitioner. 

Using `docs/api/views/transitioner` shows `Custom Router API` page which seems to be always rendered when there's no matching route.

jcsmesquita:
If you want it pointed to an external url then change it to that. The relative path it's currently set to doesn't exist.

PR #1220, opened 2017-04-25
Make `header` configurable

### LINKED ISSUE #53: Support for hiding navigation header when `headerMode: float`

### LINKED ISSUE #100: add ability to specify headerComponent per screen

### LINKED ISSUE #570: Use a custom component to render the Header

grabbou:
Fixes #570
Fixes #100 
Fixes #53

I think it's best to name it `header` (previously, it used to be a nested option getter). 

Unfortunately, that means we have to get rid of `invariant`s that we left back in the day to help people migrate from nested to flat options. To me, it makes sense to get rid of that since we are still in beta and breaking changes can happen. Since we are landing all breaking changes into one release and we will be explaining them in a release notes, it might be worth the better API.


satya164:
LGTM

grabbou:
@satya164 tests are passing now

satya164:
I'll let @ericvicenti  review and merge

lprhodes:
@grabbou I'm unsure how this fixed #53

grabbou:
You can now set `header: null` on any component regardless `headerMode` and it will work. When `float`, it will have a weird animation, but this is out of scope for this PR.

lprhodes:
It looks like you've kept the floatingHeader too though?

grabbou:
Yes
```js
if (headerMode === 'float') {
    floatingHeader = this._renderHeader(this.props.scene, headerMode);
 }
```
in this context, `this.props.scene` corresponds to the active scene that is currently displayed on screen (coming from `Transitioner`).

lprhodes:
Sorry I was comparing to the last beta not master. 

Ideally we need a way for the header to transition between the following without the need for nesting:

- iOS-style navbar aka float
- cover with/without navbar
- modal (push a new stack if you wish to show a nav bar)

This needs knowledge of the previous component within the CardStack so that it can deal with the various scenarios or a rethink on how the header is treated - I'm happy to work on it but I don't want to spend lots of time on one direction if you're looking to go another...

grabbou:
@lprhodes all these use-cases should be satisfied with this PR - previously the `float` mode was problematic. Now, you can hide/show header on per screen basis.

The only thing that could be improved is the way it animates (e.g. should it fade out / slide to top?). That requires more thinking, so I believe it's better to start with this implementation and discuss possible improvements later.

docs/api/navigators/StackNavigator.md
```diff
@@ -96,9 +96,9 @@ Visual options:
 
 Generic title that can be used as a fallback for `headerTitle` and `tabBarLabel`
 
-#### `headerVisible`
+#### `header`
 
-True or false to show or hide the header. Only works when `headerMode` is `screen`. Default value is `true`.
+React Element or a function that given `HeaderProps` returns a React Element, to display as a header. Setting to `null` hides header.
```
ericvicenti: Wouldn't this be a React Class?

src/TypeDefinition.js
```diff
@@ -207,14 +207,14 @@ export type NavigationUriAction = {
...
   headerMode?: HeaderMode,
-  headerComponent?: ReactClass<HeaderProps<*>>,
   cardStyle?: Style,
   transitionConfig?: () => TransitionConfig,
   onTransitionStart?: () => void,
   onTransitionEnd?: () => void,
 };
 
 export type NavigationStackScreenOptions = NavigationScreenOptions & {
+  header?: React.Element<*> | ((HeaderProps) => React.Element<*>),
```
ericvicenti: So, if its an element it would not accept any HeaderProps?

Instead of `((HeaderProps) => React.Element<*>)`, does it make sense to do `ReactClass<HeaderProps>`?

Also I think we should have a `?` here after the `header?:`, to signify that null is valid.

src/routers/createConfigGetter.js
```diff
@@ -98,7 +97,5 @@ export default (
     );
     outputConfig = applyConfig(routeScreenConfig, outputConfig, configOptions);
 
-    validateScreenOptions(outputConfig, route);
```
ericvicenti: any reason to kill this? Seems like it provides useful feedback for people who aren't configuring things properly

docs/api/navigators/StackNavigator.md
ericvicenti: What happens when `headerMode` is not screen?

grabbou: Yeah, it wouldn't be. Aligning with the rest of the API where we don't use `React Class`.

grabbou: When headerMode is not screen (float), header will go away as soon as `this.props.scene` in  `CardStack` changes (on animation end)

src/routers/createConfigGetter.js
grabbou: Yeah, I support that. Sadly, that forbids use of `header` which we want to take advantage from here.

src/TypeDefinition.js
grabbou: > Also I think we should have a ? here after the header?:, to signify that null is valid.

Agree

grabbou: > So, if its an element it would not accept any HeaderProps?

Yeah, it can be either: `header: <View />` or `header: (props) => <View />`.

docs/api/navigators/StackNavigator.md
satya164: The issue with ReactClass is, React will unmount and re-mount the component every time when an inline function is given, because `navigationOptions` will return a new function every time.

aranda-adapptor:
How will we port existing code using customised left/title/right invariants to this new approach? Currently I am getting `Objects are not valid as a React child (found: object with keys {title, right})` with the following code:

```
  static navigationOptions = {
    header: ({ state, navigate }) => ({
      title: <Text>{getScreenTitle(state)}</Text>,
      right: (
          <TouchableOpacity
            onPress={() => navigate("Map")}
            style={{ flex: 1, justifyContent: "center", alignItems: "center" }}
          >
      )
    })
  };
```

grabbou:
Use `headerTitle` and `headerRight`. `header` is now a React Element that
can be displayed in place of a header.

PS. You shouldn't be probably running into this unless you are using master
- we are wrapping up changelog and list of braking changes.

aranda-adapptor:
Yep thanks, I was trying to upgrade react-native to 0.43.4 but that seems to have an incompatibility with react-navigation beta.7. Also I figured I might as well match the new API here sooner rather than later.

aranda-adapptor:
Hmm so it looks like `headerTitle` can be a String or React Element. If I want it to be a function (so I can alter it based on the navigation `state`), I need to replace the whole Header with a custom function returning a React Element?

grabbou:
@aranda-adapptor to make it based on `navigation` state, you can define your entire `navigationOptions` as a function, that is:
```js
   title: `...`,
});
```

aranda-adapptor:
@grabbou thanks, that works nicely!

Jacse:
How would I go about hiding a header in a `StackNavigator` for a specific screen now?
Docs say:

> Setting to null hides header

I have tried:
`````
static navigationOptions = () => ({
    header: null
})
`````
-> `Cannot convert undefined or null to object`

and
````
static navigationOptions = () => ({
    header: () => null
})
````
-> `'header' cannot be defined as a function in navigation options for 'WhateverScreen' screen.`

Am I reading the docs completely wrong?

satya164:
cc @grabbou

lprhodes:
@Jacse I've been using the first option you set without any issues on iOS. Perhaps it's platform related - what are you trying it on? Are you definitely using code from the master branch?

Jacse:
@lprhodes You're right. Upgrading from `1.0.0-beta.8` to `1.0.0-beta.9` resolved the issue - thanks.

clay-to-n:
Should this fix the animation bug (when transitioning from a `header: float` scene to a scene with `header: null`, the header visibly moves)?

The bug was referenced in comments of #53 but in my testing of `^1.0.0-beta.9` it still seems present. Not sure if I'm using the API incorrectly, or if that bug remains open.

lprhodes:
No it doesn't fix the animation bug - I've been working on a fix for that with https://github.com/react-community/react-navigation/pull/1308

lprhodes:
I've spotted an issue with #1308 however in that on occasion, for a brief moment you see a flash of the next screen already animated in. I'll be spending some time figuring it out tomorrow.

gregblass:
Looks like its possible now to pass in a component as a header? Just wondering how to do it exactly? I've got:

`const LandingHeader = () => {
  return (
    <View style={styles.header}>
      Derp
    </View>
  )
}`

then:
`export const MainNavigator = StackNavigator({
  Index: {
    screen: Landing,
    header: <LandingHeader />
  }})`

Is that the idea? (that doesn't work but am I on the right track?)

koenpunt:
@gregblass `header` should be nested under `navigationOptions` and it should be a function. Something like this should work:

```js
export const MainNavigator = StackNavigator({
  Index: { 
    screen: Landing,
    navigationOptions: {
      header: props => <LandingHeader {...props} />
    } 
  }
})

gregblass:
The above syntax doesn't work. Does anyone have a working example of the syntax you'd use?

gregblass:
Sorry I'm new to ES6 and just can't get this working for the life of me. Anyone? Beuler? :(

docs/api/navigators/StackNavigator.md
gregblass: Any way I could see an example of the syntax you'd use here to pass a component in as the header?

gregblass:
That works @koenpunt, thanks!

PR #1234, opened 2017-04-25
Switch pager when swipe and animations are disabled. Fixes #662

### LINKED ISSUE #662: Two nested tab navigators give me a blank scene
I have implemented two nested tab navigator, and when I run the app I see both navigators but a blank scene. If I scroll horizontal, I see the right scene.

I leave the code:

**App.js** (first tab-navigator)
```
const App = TabNavigator({
  Battles: {
    screen: Battles,
  },
  Tournaments: {
    screen: Tournaments,
  },
}, {
  tabBarPosition: 'bottom',
  tabBarComponent: TabBar,
  swipeEnabled: false,
  swipeEnabled: false,
  tabBarOptions: {
    showIcon: true,
    showLabel: false,
  },
});
```

**Battles.js** (second tab-navigator)
```
const Battles = TabNavigator({
  LastBattles: {
    screen: LastBattles,
  },
  PopularBattles: {
    screen: PopularBattles,
  },
  HistoricBattles: {
    screen: HistoricBattles,
  },
}, {
  tabBarOptions: {
    indicatorStyle: {
      backgroundColor: colors.white,
    },
    style: {
      backgroundColor: colors.orange,
    },
  },
});

Battles.navigationOptions = {
  tabBar: {
    label: 'Batallas',
    icon: (
      <Image source={require('../../assets/icons/battles.png')} />
    ),
  },
};
```

Here is a screenshot:
captura de pantalla 2017-03-13 a las 12 01 30 a m

I've checked in the componentDidMount from Battles component and it's being executed.

Any ideas? Thank you!

satya164:
Fixes #662 

Android doesn't support nested scrollviews/viewpagers. Since that what we use for TabView, nested tab navigators are broken on Android.

This PR changes the implementation to not use `ViewPagerAndroid` when swipe and animations are disabled to allow nesting tab navigators. When 2 tab navigators are nested, and either the parent/child navigator has swipe and animation disabled, it will work fine.

In future we can change the check to see if only swipe is disabled after I manage to port the library to use native animations.

Tested both on iOS and Android.

cc @grabbou

### DIFF
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -2,7 +2,7 @@
 
 import React, { PureComponent } from 'react';
 import { StyleSheet } from 'react-native';
-import { TabViewAnimated } from 'react-native-tab-view';
+import { TabViewAnimated, TabViewPagerPan } from 'react-native-tab-view';
 import SceneView from '../SceneView';
 import withCachedChildNavigation from '../../withCachedChildNavigation';
 
@@ -127,6 +127,8 @@ class TabView extends PureComponent<void, Props, void> {
     );
   };
 
+  _renderPager = (props: *) => <TabViewPagerPan {...props} />;
+
   render() {
     const {
       router,
@@ -140,6 +142,7 @@ class TabView extends PureComponent<void, Props, void> {
 
     let renderHeader;
     let renderFooter;
+    let renderPager;
 
     const { state } = this.props.navigation;
     const options = router.getScreenOptions(
@@ -159,17 +162,22 @@ class TabView extends PureComponent<void, Props, void> {
       }
     }
 
+    if (animationEnabled === false && swipeEnabled === false) {
+      renderPager = this._renderPager;
+    }
+
     const props = {
-      style: styles.container,
-      navigationState: this.props.navigation.state,
       lazy,
-      renderHeader,
-      renderFooter,
       animationEnabled,
       swipeEnabled,
+      renderPager,
+      renderHeader,
+      renderFooter,
       renderScene: this._renderScene,
       onRequestChangeTab: this._handlePageChanged,
+      navigationState: this.props.navigation.state,
       screenProps: this.props.screenProps,
+      style: styles.container,
     };
 
     return <TabViewAnimated {...props} />;


grabbou:
Looks good to me. Thanks for tackling that issue!

onlydave:
@satya164 Noticing ScrollViews now need `flex:1` or bottom tabs gets pushed off screen.

Should be able to see it here https://github.com/onlydave/react-navigation/tree/nested

satya164:
@onlydave can you create a new issue for that?

src/views/TabView/TabView.js
```diff
@@ -127,6 +127,8 @@ class TabView extends PureComponent<void, Props, void> {
     );
   };
 
+  _renderPager = (props: *) => <TabViewPagerPan {...props} />;
```
wschurman: This looks like it's rendering the pan shadow (from TabViewPagerPan) component on the normal iOS TabView even when swiping is disabled. Should this use the same platform-specific logic from https://github.com/react-community/react-navigation/commit/0cd3eaa97afe9806f7abe31785ad13f75d786542#diff-6f52f52094e1001dd754981e92d9f933L55 to determine which component to render?

That, or should the condition below on L165 be `animationEnabled === false && swipeEnabled === true` so we only show the pager when the view can be swiped?

Patched both of these solutions locally and both seem to fix the issue; not sure which one is correct.

satya164: @wschurman the overflow will be fixed by #1271

xiaozhouwu:
good, it solves my problem.

src/views/TabView/TabView.js
hama: on android , i found there is still a bug, when i use ViewPagerAndroid with react-navigation . 
ViewPagerAndroid will not swipe normally , i found it is the problem of TabViewPagerPan , it is still use Animated.View transform:

`  style={[
          styles.sheet,
          width
            ? { width: routes.length * width, transform: [{ translateX }] }
            : null,
        ]}
        {...this._panResponder.panHandlers}
      >
        {Children.map(children, (child, i) => (
          <View
            key={navigationState.routes[i].key}
            testID={navigationState.routes[i].testID}
            style={
              width
                ? { width }
                : i === navigationState.index ? StyleSheet.absoluteFill : null
            }
          >
            {i === navigationState.index || width ? child : null}
          </View>
        ))}
      </Animated.View>`

PR #1243, opened 2017-04-26
Modal gestureEnabled

### LINKED ISSUE #902: gesturesEnabled not working in `modal` mode
Since upgrading to RN 0.43.0, using gestures for dismissing a screen presented using the StackNavigator is no longer working.  I was leaving this default before (i.e., true on iOS, false on Android), but it stopped working on iOS after the RN upgrade.  Now, the default iOS setting and/or specifying it deliberately in the navigator options (by specifying cardStack: { gesturesEnabled: true }) has no effect.

Any suggestions?

onlydave:
Fixes #902

Adding swipe down gesture back to modals.

Tested within our app on iOS and android.

Animations on android probably need some work but does the job. Card animation animates downwards.

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -81,8 +81,8 @@ const RESPOND_THRESHOLD = 12;
 /**
  * The distance of touch start from the edge of the screen where the gesture will be recognized
  */
-const GESTURE_RESPONSE_DISTANCE_SIDE = 35;
-const GESTURE_RESPONSE_DISTANCE_TOP = 135;
+const GESTURE_RESPONSE_DISTANCE_HORIZONTAL = 35;
+const GESTURE_RESPONSE_DISTANCE_VERTICAL = 135;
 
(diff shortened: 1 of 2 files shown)


src/views/CardStack.js
```diff
@@ -81,7 +81,8 @@ const RESPOND_THRESHOLD = 12;
 /**
  * The distance of touch start from the edge of the screen where the gesture will be recognized
  */
-const GESTURE_RESPONSE_DISTANCE = 35;
+const GESTURE_RESPONSE_DISTANCE_SIDE = 35;
```
satya164: nit: can we name it as `GESTURE_RESPONSE_DISTANCE_HORIZONTAL` and `GESTURE_RESPONSE_DISTANCE_VERTICAL`?

onlydave: Thought top and side might be a bit more intuitive. Vertical and horizontal do fit with the code that's already there though. Change made

PR #1253, opened 2017-04-26
make test pass

nihgwu:
test fails since #1231
`383:25  error    'options' is assigned a value but never used  no-unused-vars`

@grabbou


grabbou:
Ah, looks like I've sent a PR from a point in master that didn't have `prettier` formatting set up. Thanks!

PR #1258, opened 2017-04-26
[fix] Matched tab bar bottom title number of lines

alvaromb:
## Motivation

Match iOS HIG for UITabBar more closely.

This is how a `TabBarIOS` looks when the title is very long:

screen shot 2017-04-26 at 18 16 29

With this change, it mimics more the default behaviour, but lets the user define a `labelNumberOfLines` to configure if the tab bar needs to take more than one line.

screen shot 2017-04-26 at 18 08 41

**Test plan (required)**

Tests & lint pass for the changed file. Doesn't add more flow errors.

**Code formatting**

Prettier formatting.


alvaromb:
AFAIK, the test fails due to a previous issue, not related with this PR.

alvaromb:
@satya164 can I help you with something else with this PR?

Thanks :)

satya164:
IMO we should just set `numberOfLines` to 1 instead of a prop.

brentvatne:
thanks for this! I agree with @satya164's comment, and I pushed a commit to fix this: https://github.com/react-navigation/react-navigation/commit/3c2b977eca9baa9aa77b583964ea0976b29dfef9

PR #1271, opened 2017-04-27
Hide overflow on individual pages in tab view

satya164:
cc @grabbou

### DIFF
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import React, { PureComponent } from 'react';
-import { StyleSheet } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { TabViewAnimated, TabViewPagerPan } from 'react-native-tab-view';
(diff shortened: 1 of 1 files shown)


santomegonzalo:
Hi, when this PR will be implemented? is kind of weird that shadow...

aimak:
Is there any news on this ? When will it be deployed ?

aimak:
for the record, it has been deployed as of https://github.com/react-community/react-navigation/releases/tag/v1.0.0-beta.10

eflath:
I am using beta 10 but this issue is not fixed

eflath:
My setup is as follows:
```
Host StackNavigator
  TabNavigator
    StackNavigator A
    StackNavigator B
    StackNavigator C
    StackNavigator D
```
the TabNavigator is set to lazy=true. When I first launch, the shadow is not present. The shadow appears only on screens whose nearest neighbor to the right has been loaded. If i switch to StackNavigator B (therefore loading that component) then there is a shadow only on StackNavigator A. There is never a shadow on StackNavigator D.

PR #1282, opened 2017-04-27
WIP: make title-to-back transition look more native on iOS

koenpunt:
While discussing the working of #1083, I noticed that the transition from back button to title was not matching the native iOS implementation.

This is a draft version, where the transition is almost seamless, but the implementation has some rough edges:

https://www.dropbox.com/s/dm1c0t9phcgy8iz/title-to-back.mov?dl=0

Also still have to test if this doesn't introduce issues on Android, and also the RTL implementation is not tested yet.

And to make it truly look like the iOS native transition, the title should move, but the back icon should only fade in and out. For comparison the "App Store" app; https://www.dropbox.com/s/k11exzy8c370app/nav-transition.mov?dl=0
But it might be that that would be too custom for the generic header component.

- [x] iOS
- [ ] Android
- [ ] RTL
- [ ] ...

Let me know what you think!

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -228,6 +228,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
             // todo: determine if we really need to splat all this.props
             ...this.props,
             ...props,
+            widths: this.state.widths,
           }),
         ]}
       >
--- a/src/views/HeaderStyleInterpolator.js
+++ b/src/views/HeaderStyleInterpolator.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { I18nManager } from 'react-native';
+import { I18nManager, Dimensions } from 'react-native';
 
 import type { NavigationSceneRendererProps } from '../TypeDefinition';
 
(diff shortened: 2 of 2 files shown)


grabbou:
This is exactly what I was talking about with @ericvicenti earlier today. Exciting stuff. I'll give it a better review later today.

As far as iOS animations -> isn't `<` staying on the left side and it's only back button title that goes to title?

koenpunt:
> As far as iOS animations -> isn't < staying on the left side and it's only back button title that goes to title?

Yes, but like I mentioned this would require some more customisation to the back button. But maybe we can introduce a `HeaderBackButtonIOS` for that purpose.

grabbou:
Yeah, I think it makes sense to do it (so that we don't have to test Android compatibility). Anyway, we can do it in next PR.

Ping me when this is ready for review!

koenpunt:
> Ping me when this is ready for review!

Kinda is now, although I'm not feeling happy about passing the `widths` state like it is done now. But maybe that's the only way..

koenpunt:
I see @skevy added a MaskedView component (https://github.com/facebook/react-native/commit/8ea6cea39a3db6171dd74838a6eea4631cf42bba), so when that lands this can be used to implement the chevron mask. 

Also I had some thoughts about the current implementation, and I think it would work better if instead of applying the animated style to the full button component, it should be applied and handled by each individual component within the button. 

@grabbou @skevy what are your thoughts on this? I can spend some time on this, but I want to make sure that it's not going to be for nothing. Because I can imagine that the MaskedView was added for a reason and that it is already being worked on behind the scenes?

brentvatne:
thanks for this @koenpunt! it's been a while and I'm not sure if you're still interested in this but I would personally *love* to see this cleaned up. we can be much more responsive now that @ericvicenti and I are working on this nearly full time. please open a new PR if this still interests you and we can discuss and work through it there. also feel free to pick up where @skevy left off with MaskedViewIOS!

PR #1293, opened 2017-04-28
accept style prop on Header component

koenpunt:
To allow composition of the Header component when using a custom header
component, styling is probably one of the options you want to have.

For example, I need to dynamicly change the height of the header based
on scroll position, and without this change I have to inherit from the
Header component and copy the render function, instead of composing the
Header component in my own component.


satya164:
Thanks for the PR!

PR #1297, opened 2017-04-28
Adding reset key documentation

scbrady:
This adds the changes in #789 to the Navigation-Actions documentation.


ThePaulMcBride:
This is insanely useful! I've spent 4 days trying to figure this out. Sooner this is in the docs, the better! 👍

dantman:
Same here. There was essentially a regression in the situation where you use the StackNavigator { DrawerNavigator { StackNavigator } } pattern and this was the fix. I only found out that `key: null` was available or even the fix after running react in debug mode and stepping through lines of the navigation calls, reading the source code, and searching for bugs while intending to report the regression as a bug and finding this not obviously related PR.

MrHubble:
@ThePaulMcBride I've spent a fair amount of time trying to figure this out as well. I'm trying to replace a screen (rather than add it to the stack). Can you share any sample code? I've yet to find an example of where to get the `key` from:

>`key` - *string or null* - optional - If set, the navigator with the given key will reset. If null, the root navigator will reset.

I understand it can be `null`, but how do we get the `key` if we don't want to use `null`? 

Also, do you know how we can get the existing stack of actions?

>`actions` - *array* - required - Array of Navigation Actions that will replace the navigation state

ThePaulMcBride:
@MrHubble The key is the key of the router which is currently on the stack that you want to reset. If you pass in no key at all, it will reset the current stack, if you pass in null, it will reset to the root navigator. If you pass in any other key, it will look for a matching key in your navigators. I hope that helps.

Issue #135, opened 2017-02-01
Add optional key to navigate action, allowing idempotent pushes


Benjamin-Dobell:
Due to the fact react native generally doesn't block the native run loop, navigating to a new screen causes a state change resulting in an async render. Whilst this occurs the screen for the previous state can still be interacted with, so quickly tapping a button twice can easily cause duplicate route pushes resulting in a navigation stack that doesn't necessarily make any sense in the context of the app.

Also, because react navigation does animations in JS rather than using native transitions (iOS), user interaction is incorrectly enabled whilst rendering the new state. So even post (virtual DOM) render, during the transition the previous screen still has interaction enabled, allowing double taps as described above.

The ideal solution would be to disable interaction in response to a navigation action, ~~however as the communication between JS-land and native-world is async, this is virtually impossible.~~

_**EDIT:** Come to think of it, it's not at all impossible, it's just a bit of a pain to implement._

As such, the most logical solution is to ensure all state changes are idempotent. In the case of redux, this means that _all_ actions should be idempotent (not just those related to navigation transitions).

Whilst the responsibility of idempotency of non-navigation actions is clearly up to app developers, react-navigation should probably ensure its own actions (and helper functions) help enforce idempotency.

Benjamin-Dobell:
In my own in-dev redux app (admittedly not thoroughly tested yet) I've implemented the following to try enforce idempotency:

Action creation helper functions:

```
// Although we can technically return a NavigationBackAction without a key, taking the navigation (state)
// allows us to prevent duplicate dispatches e.g. back button pressed multiple times before the transition occurs.
export function goBack(navigation: Navigation, key: ?string) : NavigationBackAction {
  return {
    type: 'Back',
    key: key === undefined ? navigation.state.key : key,
  }
}

// Nothing special about this function...
export function navigate(routeName: string, params?: NavigationParams, subrouterAction?: NavigationAction) : NavigationNavigateAction {
  return {
    type: 'Navigate',
    routeName,
    params,
    action: subrouterAction
  }
}
```

Navigation reducer:

```
// TODO: Presumably there's a more efficient way... right?
function paramsEqual(params1: ?NavigationParams, params2: ?NavigationParams) {
  if (params1 == params2) {
    return true
  }

  if (!params1 || !params2 || Object.keys(params1).length != Object.keys(params2).length) {
    return false
  }

  for (let [key, value] of Object.entries(params1)) {
    if (value !== params2[key]) {
      return false
    }
  }

[...]

PR #1303, opened 2017-04-29
Currently a user can double click on a button while an animation is i…

ghuh:
Currently a user can double click on a button while an animation is in progress resulting in the same navigation call getting triggered twice.  This prevents duplicate navigation to the same route.


dantman:
-1 #135 has some other ideas on how to solve this issue, and this PR would break react-navigation in some cases

This PR rejects all actions where `routeName` matches the current `routeName` whether unintentional or intentional. (and seems to do it across multiple levels of navigator)

However there are a wide variety of situations where an app may intentionally desire navigating to another route with the same name, notably the same routeName but different params. eg: You're on `{ routeName: 'Profile', params: { user: 'Alice' } }` and a friend link to "Bob" wants to navigate to `{ routeName: 'Profile', params: { user: 'Bob' } }`. This PR would reject that.

ghuh:
It looks like #135 is a lot of talk without anything happening.  I'd be happy to close this PR if there is another solution incoming.

If not, I'd be happy to add support for checking params.  That is easy enough to do.

Let me know what you guys are planning.

dantman:
Checking for params just sounds like stacking a hack on top of another hack to work around the flawed nature of the hack. You're now doing a shallow/deep comparison of objects (which this lib doesn't do) just to work around something that there are better ways of handling.

If you want to work around this now, you don't even need any changes to react-native. Just set a `this._lock = true;` in your event handler when you make the navigate call, reject any events while `this._lock` is true, and unset that after a few seconds. Or use a debounce function from a library like lodash.

ghuh:
I agree that this is a hack solution.  However, having to do tracking of `this._lock = true` outside of `react-navigation` is also a hack (a worse one in my opinion).  This issue should be handled inside the library.

I'll leave this PR open for now so that others may reference it until a more elegant solution is implemented.

matthamil:
I'm going to mark this PR as "On Hold" for now. There has been discussion on this enhancement at https://github.com/react-community/react-navigation/issues/135, https://github.com/react-community/react-navigation/issues/271, and https://github.com/react-community/react-navigation/pull/1787. I think that this is a much needed part of the library; however, we should be careful to accept any "hack" solutions. A hack is better than no solution, in my opinion, but I'm curious what you think @skevy

ghuh:
Yeah, I've been following some of the other discussions and I 100% agree that this should not be merged.  There are much better solutions coming.

brentvatne:
you can handle this now with idempotent pushes, still need to write up docs here: https://github.com/react-navigation/react-navigation.github.io/issues/42 though!

`this.props.navigation.navigate({routeName: 'yourRoute', key: 'pick-a-key'});` -- it will only navigate to that once (based on key) regardless of how many times you fire it

PR #1308, opened 2017-04-29
Improved animation between card with/without headers

### LINKED ISSUE #1295: Animation when navigating between header and no-header cards (iOS)
I'm creating this issue so that there can be a discussion about how best to make native feeling animations when navigating between different views/cards in the stack when one view has a header and the other does not.

Android is simple as it works well with the existing `headerMode: 'screen'`

iOS requires more finesse:

**no-header to no-header & header to no-header (push, swipe back and pop)**
These are simple as you can just use `headerMode: 'screen'` for the transition.

**no-header to header (push)**
The header should transition inside the card (screen) and then move outside the card (float) once the transition has completed so that it's ready to perform float animations.

**no-header to header (pop)**
The swipe gesture will need to swap the header from being outside (float) to inside (screen) the card immediately.

**no-header to header (swipe back)**
The same as pop (though this should happen when the user starts the pan gesture). It must revert if the action is cancelled.

So essentially we need to be able to position the header inside or outside the card depending on the previous (not parent) card.

In order to achieve this without a major restructure of code, the CardStack must be aware of what the previous (and next when an animation is starting) cards are in the stack.

It must also know when the animation is starting or ending.

It'd be great to get your thoughts on this before I work any further on it.

Thanks!

lprhodes:
Intended to fix #1295


edo1493:
Are there any news on this? Thanks.

FullstackJack:
Please merge this. We need this fix. The header changes on the previous screen before (on push) and after (on pop) the screen transition so the previous card gets header jitter. That looks totally jarring without this fix!

edo1493:
@lprhodes ?

lprhodes:
Sorry I've just been working on merging beta 10 into the branch I'm using ready for a project launching this week.

I'll submit a cleaner pull request later this week as this is pretty messy now.

edo1493:
@lprhodes what happened to this? Thanks.

lprhodes:
I was hoping to eventually get everything working as it does natively but unfortunately the last animation is proving difficult...that being the movement of the tabbar when navigating from a view with one to a view without. I also ended up setting state in a hard way which needs another look at (i.e. this.state.blah = something) but was the easiest solution at the time to work with everything else.

I've not had much time of late for open source, sorry - this fork is working with decent animations in a couple of production apps though https://github.com/lprhodes/react-navigation/

brentvatne:
I believe that @skevy did some work based on this recently, hopefully he can upstream it! I'm going close this as it's out of date and it sounds like @lprhodes doesn't have time to update it. This is a desirable feature though and we would like to see an updated PR for it. Thanks for your work on this @lprhodes, sorry nobody was around to land it.

brentvatne:
Also made a feature request so people can upvote if it's important to you: https://react-navigation.canny.io/feature-requests/p/improved-animation-between-card-withwithout-headers

marcinkornek:
https://github.com/react-navigation/react-navigation/issues/2732#issuecomment-341717998 this is a good workaround for this problem - it worked for me. It also works with iPhone X but you need to use some helper (eg. `react-native-iphone-x-helper`) to check the height of status bar

```
import { Platform, StatusBar } from 'react-native'
import { ifIphoneX } from 'react-native-iphone-x-helper'

const getStatusBarHeight = () => {
  if (Platform.OS === 'ios') {
    return ifIphoneX(44, 20)
  }

  return StatusBar.currentHeight
}
```

Issue #1304, opened 2017-04-29
Get initial route from nested navigators for custom navigation store


svicalifornia:
### Current Behavior

I have a deeply nested navigator structure:

    (Modal) Stack Navigator
      Drawer Navigator (which React Navigation has implemented as two nested Tab Navigators)
        Stack Navigator
          Screen A
          …
        Stack Navigator
          Screen B
          …
      Modal X

I'm supplying my own navigation to the root navigator, with a pruned initial state that doesn't include any routes for the inactive "tabs" of the DrawerNavigator:

      // Root navigator is a ModalNavigator
      {
        index: 0,
        routes: [
          { key: 'DrawerNavigator', routeName: 'DrawerNavigator', index: 0, routes: [
            { key: 'DrawerClose', routeName: 'DrawerClose', index: 0, routes: [
              { key: 'StackNavigator', routeName: 'StackNavigator', index: 0, routes: [
                { key: 'Splash', routeName: 'Splash' },
              ]},
            ]},
          ]},
        ],
      };

When I try to navigate from `Welcome` to a modal screen (in the root), `getStateForAction` first looks in the children of `Welcome` (there aren't any), then the other tabs of the `StackNavigator` (finding nothing there either), and then moves up to the TabNavigator created for the drawer. If tries to go to the second tab, and that tab is not in my state routes, so the lookup fails.

I realize that I'm not supplying a full initial state, but Navigation has handled my simpler pruned version until now. Even hard-coding this much state is prone to breakage as the navigators will likely be restructured over time, and hard-coding a full and complete state for every single node in the navigator tree seems ridiculously redundant.

### Expected Behavior

1. Would be great if Navigation would check whether the tab route exists in the current state, and if not, creates a default route for that tab based on `initialRouteName`, before attempting to traverse down that tab.

2. I would also love to have a function that I call on my RootNavigator (or pass a navigator to) and get back a deeply nested state object representing the initial state based on `initialRouteName`s throughout the nested navigator config. Then I could use that instead of hard-coding the initial state, have even less duplicated code, easily adapt to changes in the navigator configs, and totally avoid the issue described here resulting from an incomplete hard-coded initial state.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  1.0.0-beta9
| react-native     |  0.43.4
| node             | v7.9.0
| npm or yarn      |  4.2.0

svicalifornia:
[...]

Issue #937, opened 2017-04-06
Nested redux navigators stop working


cooperka:
### Issue description

In the following example, I've nested the original StackNav inside a new StackNav, and connected just the top-level one to my redux state. It does maintain its own state, but it no longer seems to be possible to control any of its children. The app loads as usual, but none of the buttons work anymore.

Here's a very simple diff to reproduce this (call it "StackInStack A"), based on the ReduxExample app: https://github.com/cribspot/react-navigation/commit/3ef0a223bdb921518e8a92aa2b204bd518b20a4a ~~(Old, outdated one: https://github.com/cribspot/react-navigation/commit/8afcfd3315e1a276fc18b7a48096229787c8af7c)~~

Here's a second diff ("StackInStack B") doing the same thing but in a slightly different way; this time it doesn't even render, giving `Expect nav state to have routes and index, {"routeName":"Root","action":{"type":"Navigation/NAVIGATE","routeName":"Login"},"key":"Init"}`: https://github.com/cribspot/react-navigation/commit/9332bf1eab3b9d691b7ad7517fae168be6dd1c9e

Similar error with a DrawerNav nested in a StackNav ("DrawerInStack A", gives `Expect nav state to have routes and index, {"routeName":"Root","key":"Init"}`): https://github.com/cribspot/react-navigation/commit/af0f571a59df96fbb5ab3ec8865ceff1a6c6a308 (though, of course, it works if we navigate to a route other than the drawer itself: https://github.com/cribspot/react-navigation/commit/5260b0962aa01df2d15502883d5900809770d1c3, "DrawerInStack B").

In order to control the children, they must be given their own, separate reducer like so: https://github.com/react-community/react-navigation/compare/8afcfd3...110fce9d. This means for every Navigator in your app, you need to create a reducer and duplicate logic to control anything inside of that navigator.

Personally, I would expect the default behavior to be that any navigator passes its props (such as `navigation` and `onNavigationStateChange`) down to any nested navigators and that `router.getStateForAction` would return a hierarchy of routes so you only need one reducer. This could, of course, be explicitly prevented by manually setting the props to `null` in a particular child.

Related issue here: https://github.com/react-community/react-navigation/issues/1041

### Initial state output

StackInStack A (it renders, but buttons aren't clickable) has an initial state of:

```js
{
  "index": 0,
  "routes": [
    {
      "routeName": "Root",
      "key": "Init"
    }
  ]
}
```

[...]

PR #1311, opened 2017-04-29
Improved code example to handle nested navigators

svicalifornia:
Updated the code example on the Redux Integration docs page to use a simpler, less fragile way of initializing state, which also works with nested navigators, where the previous one failed.

See #1304.

### DIFF
--- a/docs/guides/Redux-Integration.md
+++ b/docs/guides/Redux-Integration.md
@@ -2,16 +2,16 @@
 
 To handle your app's navigation state in redux, you can pass your own `navigation` prop to a navigator. Your navigation prop must provide the current state, as well as access to a dispatcher to handle navigation options.
 
-With redux, your app's state is defined by a reducer. Each navigation router effectively has a reducer, called `getStateForAction`. The following is a minimal example of how you might use navigators within a redux application:
+With redux, your app's state is defined by a reducer. Each navigation router effectively has a reducer, called `getStateForAction`. You can dispatch `NavigationActions.init` to initialize your router state.
+
+The following is a minimal example of how you might use navigators within a redux application:
 
 ```
-import { addNavigationHelpers } from 'react-navigation';
+import { addNavigationHelpers, NavigationActions } from 'react-navigation';
 
 const AppNavigator = StackNavigator(AppRouteConfigs);
 
-const initialState = AppNavigator.router.getStateForAction(AppNavigator.router.getActionForPathAndParams('Login'));
-
-const navReducer = (state = initialState, action) => {
+const navReducer = (state, action) => {
   const nextState = AppNavigator.router.getStateForAction(action, state);
 
   // Simply return the original `state` if `nextState` is null or undefined.
@@ -24,12 +24,20 @@ const appReducer = combineReducers({
 });
 
 class App extends React.Component {
+  componentWillMount() {
+    this.navigation.dispatch(NavigationActions.init());
+  }
+
+  get navigation() {
+    return addNavigationHelpers({
+      dispatch: this.props.dispatch,
+      state: this.props.nav,
+    })
+  }
+
   render() {
     return (
-      <AppNavigator navigation={addNavigationHelpers({
-        dispatch: this.props.dispatch,
-        state: this.props.nav,
-      })} />
+      <AppNavigator navigation={this.navigation} />
     );
   }
 }


docs/guides/Redux-Integration.md
```diff
@@ -2,16 +2,16 @@
...
+
+The following is a minimal example of how you might use navigators within a redux application:
 
 ```
-import { addNavigationHelpers } from 'react-navigation';
+import { addNavigationHelpers, NavigationActions } from 'react-navigation';
 
 const AppNavigator = StackNavigator(AppRouteConfigs);
 
-const initialState = AppNavigator.router.getStateForAction(AppNavigator.router.getActionForPathAndParams('Login'));
```
cooperka: I commented on your original issue. Could we actually just change this to `const initialState = AppNavigator.router.getStateForAction(NavigationActions.navigate({ routeName: 'Login' }));`? This one line may be the only change necessary (I hope!)

https://github.com/react-community/react-navigation/issues/937#issuecomment-297166585

svicalifornia: No, I tried that on my app, and it failed to navigate from a null initial state to any other route name. Then I discovered the Init dispatch action on another docs page. Init works.

svicalifornia:
@cooperka Sure thing. It'll be a day or so before I can get to that.

docs/guides/Redux-Integration.md
svicalifornia: To be clear, the failure with navigate was somewhere in RNav, trying to access the previous route but failing to access the nonexistent index.

I also tried to set an initial route of `{index: -1, routes: []}`, but that didn't work either.

It would be nice if navigate handled this initial null/empty case better, but it's not really intended for that. Init explicitly is intended for that.

svicalifornia: I will try the solution proposed in #937 again, just to be sure.

cooperka: My main concern with this PR is that it's just a bit more complex to set up for first-time users. It's not a very typical redux pattern. I have it working in a medium-large app with drawer+tabs+stacks all nested and it works great for me with just the changes in #937 in my reducer.

cooperka:
Here's a patch for the example app (I actually needed to do this for something else I'm trying to debug, may as well share):

```diff
diff --git a/examples/ReduxExample/src/reducers/index.js b/examples/ReduxExample/src/reducers/index.js
index 6eeb3b6..9796f63 100644
--- a/examples/ReduxExample/src/reducers/index.js
+++ b/examples/ReduxExample/src/reducers/index.js
@@ -4,10 +4,8 @@ import { NavigationActions } from 'react-navigation';
 import { AppNavigator } from '../navigators/AppNavigator';
 
 // Start with two routes: The Main screen, with the Login screen on top.
-const firstAction = AppNavigator.router.getActionForPathAndParams('Main');
-const tempNavState = AppNavigator.router.getStateForAction(firstAction);
-const secondAction = AppNavigator.router.getActionForPathAndParams('Login');
-const initialNavState = AppNavigator.router.getStateForAction(secondAction, tempNavState);
+const tempState = AppNavigator.router.getStateForAction(NavigationActions.navigate({ routeName: 'Main' }));
+const initialNavState = AppNavigator.router.getStateForAction(NavigationActions.navigate({ routeName: 'Login' }), tempState);
 
 function nav(state = initialNavState, action) {
   let nextState;
```

JimmyDaddy:
This code not work for me, I got `Cannot read property 'undefined' of undefined`

lucalaissue:
Hey guys ! can you merge this pull request ? this is urgent ! The current exemple is just wrong, if you want to help people merge this PR !

Thanks

seantempesta:
This really needs to be solved.  I wasted at least 10 hours on this manually tweaking the nav data structure before I stumbled on @cityvoice's solution (issue a reset first on the main navigator).

https://github.com/react-community/react-navigation/issues/1304

docs/guides/Redux-Integration.md
FredericRuaudel: I had an issue with nested navigator and the docs' proposed solution for Redux integration, and I fixed it by simply replacing what you proposed @cooperka. It would be great to integrate just that to begin with. ;)

brentvatne:
hello! I'm not clear on what the right thing to do here is because I do not use redux with react-navigation, but I do know that the redux documentation has been overhauled recently and it has also moved over to https://github.com/react-navigation/react-navigation.github.io! please re-open this PR with the appropriate changes over the new docs there and I will be sure it gets timely review

Issue #998, opened 2017-04-12
Support non-route/non-touchable items in drawer (subheaders, dividers)


dantman:
The drawer currently only permits touchable route items and doesn't have a way to add non-touchable/non-route items. This is true even when building your own `contentComponent`, `DrawerView.Items` uses `navigation.state.routes` directly so you can't use `DrawerView.Items` multiple times with subsets of the routes.

![screenshot_20170411-203334](https://cloud.githubusercontent.com/assets/53399/24940225/de2e49ce-1ef6-11e7-97be-e6eb86641bed.png)

ericvicenti:
True, this is a limitation of `DrawerView.Items`. If you want to add the feature, feel free to send in a pull request.

But we have no current plans to add this feature because you can already customize the `contentComponent`, and what you render inside of there isn't really a concern of the navigation library.

cooperka:
@dantman check out #1317 :smile_cat:

dantman:
@cooperka That looks like too much of a hack.

I already created a fix to allow for `customComponent`s to be implemented with any sort of section headers, dividers, and more. I use it not just for section headers and dividers, but also for that pattern you see in some Google apps where there's a header inside the drawer with a large button with a dropdown arrow that makes a separate set of drawer items for signing in/out.

See #1039

cooperka:
Replied in the PR. Thanks for the feedback and link!

PR #1317, opened 2017-04-30
Add "drawerOnPress" handler + demo "account picker"

cooperka:
### Motivation

So we can do things like this:

sign in demo

And this:

refresh button demo

### Changes

1. Allow callbacks for drawer items via `drawerOnPress` without needing to write your own custom `contentComponent`
2. Allow rendering custom drawer items via `contentOptions.itemComponent`
3. Factored out and exported `DrawerNavigationItem` for modularity and reuse
4. Updated the demo app to show off some of the new features, including the account picker shown above

No breaking changes, as far as I know.

### Usage

See the changes to the demo app for an implementation example.

### Test Plan

1. Run the NavigationPlayground app
2. Tap the Drawer Example and try it out

Please let me know what you think; the API makes sense to me but I'm open to changing it.

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -163,6 +163,11 @@ String, React Element or a function that given `{ focused: boolean, tintColor: s
 
 React Element or a function, that given `{ focused: boolean, tintColor: string }` returns a React.Element, to display in drawer sidebar
 
+#### `drawerOnPress`
+
+Callback to handle tap events. Typical behavior would be to call `navigation.navigate('DrawerClose')` after performing your other logic.
+If explicitly set to `null`, there will be no touch feedback for this drawer item; this can be used e.g. for header items / section separators.
+
 ### Navigator Props
 
 The navigator component created by `DrawerNavigator(...)` takes the following props:
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -131,6 +131,7 @@ const styles = StyleSheet.create({
 ### `contentOptions` for `DrawerItems`
 
 - `items` - an array of routes to use in place of `navigation.state.routes`
+- `itemComponent` - a component to use in place of `DrawerNavigatorItem` for each item
 - `activeTintColor` - label and icon color of the active label
 - `activeBackgroundColor` - background color of the active label
 - `inactiveTintColor` - label and icon color of the inactive label
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -130,6 +130,7 @@ const styles = StyleSheet.create({
 
 ### `contentOptions` for `DrawerItems`
 
+- `items` - an array of routes to use in place of `navigation.state.routes`
 - `activeTintColor` - label and icon color of the active label
 - `activeBackgroundColor` - background color of the active label
 - `inactiveTintColor` - label and icon color of the inactive label
--- a/examples/NavigationPlayground/js/Drawer.js
+++ b/examples/NavigationPlayground/js/Drawer.js
@@ -145,7 +145,7 @@ const DrawerExample = DrawerNavigator(drawerRoutes, {
   contentOptions: {
     activeTintColor: '#e91e63',
     items: drawerItems,
-    drawerItemComponent: CustomDrawerItem,
+    itemComponent: CustomDrawerItem,
   },
 });
 
--- a/examples/NavigationPlayground/js/Drawer.js
+++ b/examples/NavigationPlayground/js/Drawer.js
@@ -4,14 +4,15 @@
 
 import React from 'react';
 import {
+  View,
   Button,
   Platform,
   ScrollView,
   StyleSheet,
-  Text,
 } from 'react-native';
 import {
   DrawerNavigator,
+  DrawerItems,
   DrawerItem,
 } from 'react-navigation';
 import MaterialIcons from 'react-native-vector-icons/MaterialIcons';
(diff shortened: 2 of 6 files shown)


dantman:
While there may be some merits to this change on its own (being able to override the onPress for the drawer item of a route that behaves differently than a normal item) it doesn't allow you to insert actual subheaders with proper subheader styles (font, color, and a few other changes) or insert dividers (no text or icon, a completely separate line component). So it does not actually resolve #998.

cooperka:
@dantman thanks for the feedback and the link to your earlier PR (I looked through the open PRs before but the title must not have stood out to me). I think you're right: this `onPress` logic is useful on its own, but your PR is much better for implementing custom section headers. My goal was uniformity, not customizability.

We should wait to merge this until your PR is merged, then I'll rebase.

I definitely agree it's hacky to create "fake scenes" just to add section headers like was done here, so I'll also update the example app to use your logic e.g. for the "Mailboxes" header shown in the gif.

cooperka:
Rebased on #1039 and changed a few things, including the acceptance of `itemComponent`; description is updated. @dantman I'd love to hear your thoughts again if you care to give them. It's less hacky now.

dantman:
I like the direction itemComponent is going. However rather than it being used to handle non-routes, I'd much prefer `itemComponent` being used primarily to replace the styling/layout of route items.

eg: You might use itemComponent to replace the normal rendering of drawer items with a real Material Design library's (Drawer)Item component (it could adhere to MD perfectly, integrate into the library's automatic theme handling, or handle the material dark theme) or a component that renders it like a native iOS list item. This provides proper responsibility isolation and black-boxing, react-navigation becomes responsible exclusively for all internal route/navigation logic and all the layout/design can be controlled by the app.

Then `DrawerView` can be responsible for the basic drawer container/opening; `customComponent` can be responsible for the drawer header, sections, and other static design elements; `DrawerViewItems` can be responsible for the logic of handling a list of routes (all the route related bits of logic); and `itemComponent` can be responsible for only the visual layout and design of a drawer item.

To that end I think DrawerItem should not be responsible for a bunch of the logic it inherited from DrawerViewItems:
* `focused` should be passed as a prop instead of left to the item
* An `onPress` should be resolved and passed instead of passing `drawerOnPress`
* `icon` and `label` should be passed as a prop instead of extracted from `renderItem` and `getLabel`

cooperka:
I generally agree, especially with your bullets at the end -- and that's actually the approach I tried to take at first -- but there are a few issues I'm running into:

- I want to render non-route drawer items, but have them look the same as the default drawer items
- I don't want to write duplicate logic in a `contentComponent` to handle things like determining whether the current item is focused, etc.
- The current default `renderIcon`, `getLabel`, and `getScreenOptions` all expect the item to be a real route, and throw an error if it's not
- Therefore some logic needs to be modified at the level of `DrawerNavigatorItems` or higher to allow for non-route items if that's the one handling the logic as you suggest (and I agree with your suggestion, it's just not particularly simple to do)

I'll continue working on it and see if I can find a more elegant way of doing this.

dantman:
Even so, it's too much of a hack. I think we're better off ditching the idea of making DrawerViewItems handle non-routes. route handling is supposed to be what we're relying on react-navigation for, not a specific visual design. Non route things can be handled by the `contentComponent`.

> I want to render non-route drawer items, but have them look the same as the default drawer items

Actually, if the default `DrawerNavigatorItem ` is exported, with what I described you should still be able to do that. Instead of passing non-routes to `DrawerViewItem` you'd just use `DrawerNavigatorItem` directly in your `contentComponent`.

```js
<View>
  <DrawerNavigatorItem
    {...props}
    onPress={...}
    icon={...}
    label={} />
  <DrawerViewItems {...props} />
</View>
```

Then all your items look the same, `DrawerViewItems` is your black-box tool for rendering any react-navigation route items you have, and you have complete control over the layout of everything.

cooperka:
If I'm understanding correctly, that works perfectly fine except for this point:

> - I don't want to write duplicate logic in a `contentComponent` to handle things like determining whether the current item is focused, etc.

and I'll add another:

- I want to be able to put non-routes in between real routes (e.g. section divider lines like in your screenshot on your PR)

After [implementing](https://github.com/cribspot/react-navigation/commit/8572ef074651592fb8b612bcacc9b058c2ebb965) your suggestions (currently on a WIP branch), [DrawerNavigatorItems.js](https://github.com/cribspot/react-navigation/blob/8572ef074651592fb8b612bcacc9b058c2ebb965/src/views/Drawer/DrawerNavigatorItems.js) has 93 lines of code that would all essentially need to be copied and tweaked in order to render non-route items using a custom `contentComponent`. I completely agree with your point about react-navigation handling **routes** and not custom render logic, but I also don't want to make everyone with custom drawer items re-invent the wheel.

Hopefully I've explained the problem more clearly. It's also possible I'm misunderstanding what you're suggesting?

dantman:
* "Is this focused" logic for routes goes in `DrawerNavigationItems`. Non-routes are not part of react-navigation's item focus so having to decide on your own if something is 'focused' is unavoidable.
* Putting non-routes in between routes is not done by trying to pass non-routes to `DrawerNavigationItems` but instead by using multiple instances of `DrawerNavigationItems` (think of it as your tool to render a list of routes). You take `navigation.state.routes` and split it up into multiple arrays.

eg: here's a crude way of having the first route item, a non-route, then the rest.
```js
<DrawerNavigationItems {...props} items={[props.navigation.state.routes[0]]} />
<DrawerItem {...props} ... />
<DrawerNavigationItems {...props} items={props.navigation.state.routes.slice(1)} />
```

> After implementing your suggestions (currently on a WIP branch),

This is a little different than I imagined:
* `DrawerNavigatorItem` should be navigation agnostic, ie: there should be no `defaultOnPress`. `DrawerNavigationItems` should pass the default `onPress` (that closes the drawer and calls navigate) to it when there isn't an override given for that route.
* `activeTintColor`, `inactiveTintColor`, `activeBackgroundColor`, and `inactiveBackgroundColor` should still be part of `DrawerNavigatorItem`. The item is still responsible for how it looks when it's active and inactive, it's just not responsible for knowing "how" it is focused it's simply told whether it's focused or not (the `focused` boolean prop). Then your non-route code gets to do the same, pass the props you get through and use `focused` to tell the item if it's focused or not.

It's not directly part of this commit, but remember that when you're moving around props you shouldn't remove flow types from `type Props` just because they are now passed through to another component instead of used directly. You still want the props to be type checked (flow probably fails anyways when you don't). If you want to improve the types / be DRY, then when passing props through you can use `&` (or flow's new `...` syntax) to combine the flow `Props` with the `Props` for the component you're passing to.

cooperka:
Splitting the routes like that makes sense. Thanks for the detailed example, I think that will work 👍

cooperka:
Done.

cooperka:
*Note to reviewers: I'll be traveling for the next couple weeks. Hopefully this can be merged without issue, but I'll try to respond to comments when possible (or on May 23 when I return). Branch is rebased and up to date, and owners have permission to push directly to this branch:*

```console
git remote add cooperka https://github.com/cribspot/react-navigation.git
git fetch cooperka
git checkout -b drawer-on-press
```

cooperka:
This is now quite outdated; closing in favor of https://github.com/react-community/react-navigation/pull/2183.

PR #1320, opened 2017-05-01
Don’t use `Init#` keys for reset

dantman:
This Init{index} pattern guarantees that the same keys will always be re-used when doing a RESET.
This behaviour doesn’t match how the rest of the router works and causes bugs.

Namely if you are on `{ routeName: ‘Dashboard’, params: { user: ‘A’ } }` and use RESET to `{ routeName: ‘Dashboard’, params: { user: ‘B’ } }` when the user switches, it will act like a `setParams` action instead of a `RESET` action and the Dashboard will not be re-mounted.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -248,19 +248,19 @@ export default (
         return {
           ...state,
           routes: resetAction.actions.map(
-            (childAction: NavigationNavigateAction, index: number) => {
+            (childAction: NavigationNavigateAction) => {
               const router = childRouters[childAction.routeName];
               if (router) {
                 return {
                   ...childAction,
                   ...router.getStateForAction(childAction),
                   routeName: childAction.routeName,
-                  key: `Init${index}`,
+                  key: _getUuid(),
                 };
               }
               const route = {
                 ...childAction,
-                key: `Init${index}`,
+                key: _getUuid(),
               };
               delete route.type;
               return route;


ericvicenti:
Yeah, this change makes sense. Landing now

rroehrig:
Hey, this change breaks our snapshot tests because of the random Init key.
Any suggestions for a fix?

dantman:
> Hey, this change breaks our snapshot tests because of the random Init key.
> Any suggestions for a fix?

Other parts of react-navigation have uuid keys, those snapshot test would eventually break anyways.

Snapshot test should have some way to mark keys that are non-deterministic.

dbertella:
I'm also interest in the discussion, does anybody find a solution to to successfully run snapshot tests on navigators?

dbertella:
I ended up, writing my custom serializer to parse the json and remove the generated id.
Now every snapshot test have `MY-CUSTOM-KEY` as a key.
Let me know what you think about that.

```
// navigationSerializer.js

const regex = /^Init-id-/;

const eachRecursive = (obj) => {
  const temp = obj;
  for (const k in temp) {
    if (typeof temp[k] === "object" && temp[k] !== null) eachRecursive(temp[k]);
    else if (k === "key" && regex.test(temp[k])) {
      temp[k] = "MY-CUSTOM-KEY";
    }
  }
};

export default {
  print: (val: Object) => eachRecursive(val),
  test: (val: Object) => val && Object.prototype.hasOwnProperty.call(val, "key"),
};


// navigator.test.js
import serializer from '../navigationSerializer';
expect.addSnapshotSerializer(serializer);
```

matt-oakes:
As a simpler solution, you can add this line to your Jest startup file:

```
Date.now = jest.fn(() => 1480000000000)
```

That's using an internal implementation detail of the `_getUuid` method to work though. I think it would be best if at the very least this method was exported in a way in which it could be mocked independently of the rest of the library.

dbertella:
I tryed it before writing the serializer but I don't know why I still get the different uid, logging the test I saw that the date used as a key was before the date logged inside the test, so I assumed it wasn't possible to mock.

matt-oakes:
@dbertella I only got it to work when I added it to a file which was in the [`setupFiles` `package.json` property](https://facebook.github.io/jest/docs/en/configuration.html#setupfiles-array). I presume that the line which calls `Date.now()` in the library runs too early if you mock it in the test file itself. This does mean that `Date.now` is mocked for _all_ of the tests, but that's not a problem for me.

dbertella:
Yeah I thought so, I didn't want to mock Date.now for every test but thanks for pointing it out, it's indeed a good solution. I don't know yet if the serializer it's a good option

matt-oakes:
@dbertella I couldn't actually get your solution to work. I'm not sure if it was because I was setting it up incorrectly, or wheather it was because I was testing root navigators which don't have a `key` property.

dbertella:
I think you have to try it on nested navigator yes. The only thing is doing, looping into the serialized json looking for key === key with values that match the regex `/^Init-id-/` and change the content for it. Some keys I saw where mY component name, so it will not affect those.

dusi:
@matt-oakes we managed to get around this issue by using `require` instead of `import` in our tests. This way the `const` will be set after you're able to mock the date properly.

```js
// ComponentUsingStackNavigator-snapshot-test.js
const ComponentUsingStackNavigator = require("../ComponentUsingStackNavigator").default;

Date.now = jest.fn(() => 42);
```

PR #1322, opened 2017-05-01
DrawerLockMode configuration option with top-level navigator update capability

jeffreymendez1993:
This pr adds drawerLockMode to determine whether or not to allow the drawer to be open by swipe events. This uses screenProps={{lockMode:this.state.lockMode}} that allows it to update accordingly with a top-level navigator. You can also specify the lockMode with   navigationOptions:{
    lockMode: 'unlocked | locked-closed | locked-open'
  }, through normal configuration of your navigator but will not respond to updates.

Use the onNavigationStateChange prop and add some logic this works perfect for my nested StackNavigator inside the DrawerNavigator.

  navigateCheck = (navigation, route) => {
    // enable/disable drawer.
    let drawerEnable = route.routes[0].routes.find( ( ele ) => {
     return ele.index !== 0;
    });
    if ( drawerEnable.index && drawerEnable.index >= 1) {
     this.setState({lockMode: 'locked-closed'});
   } else {
     this.setState({lockMode: 'unlocked'});
   }
  }

<DrawerStack ref={nav => { this.navigator = nav; }} screenProps={{lockMode:this.state.lockMode}} onNavigationStateChange={this.navigateCheck}/>

works for iOS and android.

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -81,7 +81,7 @@ The route configs object is a mapping from route name to a route config, which t
 
 
 ### DrawerNavigatorConfig
-
+- `drawerLockMode` - Specifies the [lock mode](https://facebook.github.io/react-native/docs/drawerlayoutandroid.html#drawerlockmode) of the drawer.
 - `drawerWidth` - Width of the drawer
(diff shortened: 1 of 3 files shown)


jeffreymendez1993:
Should I make these changes into one single commit?

PR #1324, opened 2017-05-01
Pass tintColor to tabBarLabel within navigationOptions

lprhodes:
This expands on #1156 in that it now passes tintColor to tabBarLabel functions as per the documentation.


satya164:
Thanks for the PR!

PR #1333, opened 2017-05-01
Change from BackAndroid to BackHandler

### LINKED ISSUE #1330: BackAndroid is deprecated [react-native 0.44.0]
In react native 0.44.0, BackAndroid is deprecated. Got a warn :/

Thanks !

zackify:
closes #1330 

If someone else could double check this that would be awesome, but looks like the completely same syntax, just a different name!

### DIFF
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import invariant from 'fbjs/lib/invariant';
-import { BackAndroid, Linking } from './PlatformHelpers';
+import { BackHandler, Linking } from './PlatformHelpers';
 import NavigationActions from './NavigationActions';
 import addNavigationHelpers from './addNavigationHelpers';
 
@@ -19,7 +19,7 @@ type NavigationContainerProps = {
   onNavigationStateChange?: (
     NavigationState,
     NavigationState,
-    NavigationAction,
+    NavigationAction
   ) => void,
 };
 
@@ -37,11 +37,11 @@ type State = {
  */
 export default function createNavigationContainer<T: *>(
   Component: ReactClass<NavigationNavigatorProps<T>>,
-  containerOptions?: {},
+  containerOptions?: {}
 ) {
   invariant(
     typeof containerOptions === 'undefined',
-    'containerOptions.URIPrefix has been removed. Pass the uriPrefix prop to the navigator instead',
+    'containerOptions.URIPrefix has been removed. Pass the uriPrefix prop to the navigator instead'
   );
 
   class NavigationContainer extends React.Component<void, Props<T>, State> {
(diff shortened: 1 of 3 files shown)


mica16:
Got a warning too, the replacement would be cool :) 
Thanks

src/createNavigationContainer.js
```diff
@@ -19,7 +19,7 @@ type NavigationContainerProps = {
   onNavigationStateChange?: (
     NavigationState,
     NavigationState,
-    NavigationAction,
+    NavigationAction
```
dominicrj23: Shouldn't we avoid making unnecessary changes in the code structure? 
Applies to a few other places too.

nihgwu:
so you are going to drop the support for RN < 0.44?

zackify:
Sorry that was due to prettier. I will fix it

jakecraige:
Any update from the core team on this one? This one is frustrating to have with RN 0.44 since it throws a `console.warn` which comes up every time you load the app in dev.

weby:
As a temporary workaround, you can add this line to your App.js to prevent yellow box messages.

`console.ignoredYellowBox = ['Warning: BackAndroid']`

ugiacoman:
@weby Thanks!

src/createNavigationContainer.js
cooperka: @zackify These trailing commas are intentional; they help give better diffs when adding/removing/reorganizing things in a list of items. Please keep them in :)

zackify:
I know I totally agree. In fact I have trailing commas turned on in my atom
plugin, don't know how it happened. I'll try to fix the commit tomorrow.

m1ckm:
Ok if it's a need and I can fix tell me what to do but I wasn't impressed with all the up shit shoot and gay stuff. That was being poorly translated.

src/createNavigationContainer.js
zackify: Yes... somehow prettier had trailing commas off on my work computer!! My bad.

zackify:
Okay, I fixed the commas!

satya164:
Travis is failing because `BackHandler` is undefined. You also need to update the RN version probably?

dominicrj23:
I don't think we should merge this until we bump react-native

SNNikitin:
How about news?

mettin:
@dominicrj23 this is currently blocking me from updating my project to .44. 
I think its save to assume that people who are upping their react-navigation won't by using an older RN version?

satya164:
> this is currently blocking me

It's a deprecation warning. It shouldn't block you. You can just suppress the warning.

skevy:
Sorry Zach! I didn't see this one and already reviewed #1520.

Thanks for your work here. Hope to see more contributions from you! :)

PR #1341, opened 2017-05-02
[Minor] Update Stacks examples for better clarity

cooperka:
It was previously a bit unclear what was going on because of the wording used. Now all screens are called "screens", tabs are called "tabs", "Open" means push a screen on the stack, and "Go to tab" means change the tab without pushing on top. It's a small change that hopefully helps clear up some confusion and make it more intuitive.

Also adds the header back to the StacksOverTabs example so the scene titles are visible.

---

*Note to reviewers: I'll be traveling for the next couple weeks. Hopefully this can be merged without issue, but I'll try to respond to comments when possible (or on May 23 when I return). Branch is rebased and up to date, and owners have permission to push directly to this branch:*

```console
git remote add cooperka https://github.com/cribspot/react-navigation.git
git fetch cooperka
git checkout -b better-stacks-example
```

### DIFF
--- a/examples/NavigationPlayground/js/StacksInTabs.js
+++ b/examples/NavigationPlayground/js/StacksInTabs.js
@@ -50,9 +50,6 @@ const MyProfileScreen = ({ navigation }) => (
     navigation={navigation}
   />
 );
-MyProfileScreen.navigationOptions = ({ navigation }) => ({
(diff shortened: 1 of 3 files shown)


ericvicenti:
Looks great, thanks @cooperka, and enjoy your travels!

Issue #1318, opened 2017-04-30
Can't navigate from header


IgorVanian:
1.0.0-beta.9
No error messages, no warnings...
Clicking the button fires goToNewEvent but doesn't navigate to the route.

```
...
events: {
	screen: StackNavigator({
		events: { screen: HomeScreen },
		newEvent: { screen: NewEventScreen }
	})
}
...
```
```
class NewIcon extends Component {
	goToNewEvent = () => {
       		this.props.navigation.navigate({
        		routeName: 'newEvent'
    		});
    	};
  	render() {
	    return (
	      	<TouchableOpacity onPress={() => this.goToNewEvent()}>
			<Icon name='new-message' color='black' size={25} />
		</TouchableOpacity>
	    );
  	}
}

class HomeScreen extends Component {
	static navigationOptions = ({navigation}) => ({
		tabBarIcon: ({focused}) => (focused ? HomeIcon : HomeIconWhite),
		headerRight: <NewIcon navigation={navigation} />,
		title: <Text style={{color: 'black'}}>Events</Text>,
		headerStyle: {
			backgroundColor: 'white'
		}
	})

	constructor(props) {
	    super(props);
	}

	render() {
		return (
			<View style={{flex: 1}}>
			</View>
		);
	}
}
```

koenpunt:
You should call `navigate('newEvent')` instead. The format with `{ routeName: 'newEvent' }` is used for the `NavigationActions.navigate` method.

IgorVanian:
Generates this error when I try to navigate: 

"Invalid data message - all must be length: 36"

![image](https://cloud.githubusercontent.com/assets/16898357/25578397/94917552-2e6e-11e7-9322-93058063b399.png)

```
import React, { Component } from 'react';
import { 
	View, 
	Text 
} from 'react-native';

class NewEventScreen extends Component {


	constructor(props) {
	    super(props);
	}

	render() {
		return (
			<View style={{flex: 1}}>
			</View>
		);
	}
}

export default NewEventScreen;
```

IgorVanian:
Also, this error logs with the previous one:

`Converting circular structure to JSON`

despreston:
getting the same message

despreston:
[...]

PR #1343, opened 2017-05-02
clarify title option

despreston:
If you attempt to use something other than a string, it will throw an error in some situations. #1318 provides more details about the error. If it is expecting a certain type, the docs should clarify that.


satya164:
Thanks for the PR!

PR #1348, opened 2017-05-03
Prevent navigating twice when tapping too fast

Doko-Demo-Doa:
**Changes**
The problem is somehow caused by tapping too fast to navigate (intentionally or not). I have solved my problem by making minor changes to `addNavigationHelpers.js` . By setting the debounce threshold to 500ms, the problem is now gone for navigating.

Tested on both platforms with these devices and emulators:

- iOS emulator (iPhone 5s, 6, 6s).
- Android emulator (Genymotion Xperia API 16).
- iOS real device: iPad Air 2 (16GB, iOS 9).
- Android real device: Samsung Galaxy J1, Samsung Galaxy S5.

**Usage**

Use `navigate` method normally or `dispatch(navigateAction)`.

**Test plan (required)**

- Add a button in one screen to navigate to another screen.

- Try using with `this.props.navigation.dispatch` or `this.props.navigation.navigate("Screen2")` like follow (assuming you are not using redux):

With `dispatch` method:
```
const navigateAction = NavigationActions.navigate({
  routeName: "Screen2"
});

this.props.navigation.dispatch(navigateAction);
```
With `navigate` method:
```
this.props.navigation.navigate("Screen2");
```

### DIFF
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -42,7 +42,6 @@ export default function<S: *>(navigation: NavigationProp<S, NavigationAction>) {
           },
           500,
         );
-
         return true;
       }
       return false;


nihgwu:
check further discussion in this thread #135

skyworld:
I this the best way to solve this problem is Setting Timeout on press event

```

import React, { Component } from 'react';
import { StyleSheet, Platform, StatusBar, Text, View,TouchableHighlight } from 'react-native';


/**
 * 自己封装的touchable事件
 * 
 */
export default class Touchable extends Component {

	constructor(props) {
		super(props);

		var self = this;
		self.state = {
			disabled:false,
		}

		this._onPress=function(){
			var self = this;
			self.props.onPress();
			if(self.props.delayPress > 0){
				self.setState({disabled:true});
				setTimeout(function(){
					self.setState({disabled:false})
				},self.props.delayPress);
			}
		}
    }

    static propTypes = {


    };

    static defaultProps = {
        onPress:function(){},
        delayPress:0,
        underlayColor:'rgba(0,0,0,0.2)'
    };

    render() {
        return (
            <TouchableHighlight 
            	style={this.props.style}
            	onPress={this._onPress.bind(this)} 
            	underlayColor={this.props.underlayColor}
            	disabled={this.state.disabled}>
            		{this.props.children}
        	</TouchableHighlight>
        );               
    }
}
```

usage
```
<Touchable  
                  delayPress={400} {/* delay 400ms for next click response */}
                  onPress={this.showThirdShare}>
                  <View style={styles.contextItem}>
                    <Text style={styles.text}>推荐给好友</Text>
                    <Text style={styles.nextIcon}>&#xe612;</Text>
                  </View>
                </Touchable>
```

Doko-Demo-Doa:
@skyworld Well that's a solution, I guess, but...

- As @nihgwu said, 500ms or 400ms is too arbitrary.
- You probably do not want to set timeout manually per click event.

Anyway I still believe 500ms is arbitrary but usable in most cases.

jeffreymendez1993:
I solved this issue by keeping track of the current route that if its the same route after the function call to block the action.

MovingGifts:
@jeffreymendez1993 I am facing the same issue. Do you mind sharing the code and where to place it to block the second action for the same route?

wwwillchen:
@MovingGifts - if you're using Redux, you might find this helpful. In my navigation reducer I have a simple 'throttling' mechanism.:

```
// outside of reducer
const recentlyVisitedRoutes = new Set();

// inside reducer
  if (action.type === 'Navigation/NAVIGATE' || action.type === 'Navigation/BACK') {
    if (recentlyVisitedRoutes.has(action.routeName)) {
      return state;
    }
    recentlyVisitedRoutes.add(action.routeName);
    setTimeout(() => { recentlyVisitedRoutes.delete(action.routeName); }, 150);
  }

```

MovingGifts:
@wwwillchen Works great! Thank you for jumping in :)

thebigredgeek:
Is there a solution for idempotentency yet?  This is a really annoying issue :(

dijonkitchen:
This is related to #1588

nightspirit:
Here is our solution. Navigate/Back action essentially trigger the screen transition. If you blcok the subsequent Navigate/Back action until the animation ended. This should solve the problem.

You can use [InteractionManager.runAfterInteractions](https://facebook.github.io/react-native/docs/interactionmanager.html) to determine whether the animation is ended.

This solution only works for those using Redux to manage the navigation state. For integrating Redux, please check out the [documentation](https://reactnavigation.org/docs/guides/redux).

In the RootContainer where you feed the navigator navigation props

```javascript
_dispatch = (...arg) => {
  const type = arg[0].type // get action type
  if (type !== 'Navigation/NAVIGATE' && type !== 'Navigation/BACK') {
   // Non Navigate/Back action should dispatch right away
    this.props.dispatch.apply(null, arg)
  } else if (!this.state.blocking) {
    this.setState({ blocking: true }) // set blocking flag to true
    this.props.dispatch.apply(null, arg)
    InteractionManager.runAfterInteractions(() => {
      this.setState({ navigating: false }) // release after animation ended
    })
  }
}

// We are using Drawer Navigator with nested Stack/Tab navigator
// You should be able to change to any other Navigator without problem
render () {
  return (
      <AppDrawer
        navigation={addNavigationHelpers({
          dispatch: this._dispatch,
          state: this.props.navigation
        })}
      />
    )
}
```

While in screen transition, all the incoming Navigate/Back action (unwelcome action) will be dropped. This works really well for us on both iOS & Android.

Noitidart:
@jeffreymendez1993 may you please share your code on how you did that?

jeanregisser:
I adapted @nightspirit so it works for apps not using Redux to manage the navigation state.

```jsx
function fixNavigator(Navigator: *) {
  return (...args: any[]) => {
    class FixedNavigator extends Navigator(...args) {
      _blockDispatch: boolean;

      constructor(...constructorArgs: any[]) {
        super(...constructorArgs);

        const originalDispatch = this.dispatch;

        this.dispatch = (...dispatchArgs) => {
          // console.log('Dispatch', ...dispatchArgs);

          const { type, routeName } = dispatchArgs[0]; // get action
          if (
            (type !== 'Navigation/NAVIGATE' && type !== 'Navigation/BACK') ||
            (type === 'Navigation/NAVIGATE' &&
              (routeName === 'DrawerClose' ||
                routeName === 'DrawerOpen' ||
                routeName === 'DrawerToggle'))
          ) {
            // Non Navigate/Back action or Drawer action should dispatch right away
            originalDispatch(...dispatchArgs);
          } else if (!this._blockDispatch) {
            this._blockDispatch = true;
            originalDispatch(...dispatchArgs);
            InteractionManager.runAfterInteractions(() => {
              this._blockDispatch = false; // release after animation ended
            });
          } else {
            console.warn(
              'Navigation dispatch blocked as an interaction is in progress!',
              ...dispatchArgs,
            );
          }
        };
      }
    }

    return FixedNavigator;
  };
}

const RootNavigator = fixNavigator(StackNavigator)({
  Home: {
    screen: HomeScreen,
  },
  Details: {
    screen: DetailsScreen,
  },
});
```

dann1609:
I solved forward navigation using this conditional in my reducer

```
if(state.routes[state.routes.length-1].routeName!=action.routeName){
    return nextState || state;
  }else{
    return state;
  }
```

brentvatne:
you can handle this now with idempotent pushes:

`this.props.navigation.navigate({routeName: 'someRoute', key: 'pick-a-key'});`

see: https://github.com/react-navigation/react-navigation/pull/3393

we need to write docs about this: https://github.com/react-navigation/react-navigation.github.io/issues/42 and also we need to think about the ergonomics around it and how we can, when possible, handle it automatically for users. thanks for this PR and the thoughtful discussions around here everyone! let's move discussions over to https://github.com/react-navigation/rfcs/issues/16 for that

dann1609:
@brentvatne what about NavigationActions.navigate ??
Your solution do not work with it.

PR #1360, opened 2017-05-03
Changed View to Animated.View for TabBarBottom

zakster12:
Changing the View to Animated.View for TabBarBottom connected to this discussion:
https://github.com/react-community/react-navigation/issues/1270


zakster12:
@satya164 Formatting is done. Thank you again!

PR #1378, opened 2017-05-04
Dynamic Modals 

jeffreymendez1993:
* added the capability to adjust the mode per screen route through screenProps and a top level-navigator.

### DIFF
--- a/src/navigators/StackNavigator.js
+++ b/src/navigators/StackNavigator.js
@@ -52,7 +52,7 @@ export default (
     <CardStackTransitioner
       {...props}
       headerMode={headerMode}
-      mode={mode}
+      mode={props.screenProps.mode || mode}
       cardStyle={cardStyle}
       transitionConfig={transitionConfig}
       onTransitionStart={onTransitionStart}
--- a/src/navigators/StackNavigator.js
+++ b/src/navigators/StackNavigator.js
@@ -52,7 +52,7 @@ export default (
     <CardStackTransitioner
       {...props}
       headerMode={headerMode}
-      mode={props.screenProps.mode || mode}
+      mode={props.screenProps && props.screenProps.mode || mode}
       cardStyle={cardStyle}
       transitionConfig={transitionConfig}
       onTransitionStart={onTransitionStart}


ericvicenti:
Hmm, doesn't this cause broken-looking animations? Could you re-open once you've done the following?

- Test thoroughly and describe how you tested it
- Add an example to the playground app that utilizes this new option
- Take screenshots/videos of the behavior on both platforms
- Write documentation

jeffreymendez1993:
@ericvicenti  It does not. It works perfectly fine for me on iOS and android. Sure i'll do it this weekend when I get spare time.

jameskhamil:
@jeffreymendez1993, I know that is a daunting list of requirements from @ericvicenti, but are you still interested in doing it?

herkyl:
@jeffreymendez1993 could you write a small example on how to use it, I would love to test it out

JulianKingman:
Would it make more sense to make `mode` a navigationOptions property?

jeffreymendez1993:
Will try to get to this this weekend sorry about that! and yes @jameskhamil

jeffreymendez1993:
Actually it kinda brings you down a bad rabbit hole if you have gestures enabled and you are trying to pop a scene but the current scene is a modal type. Say you navigate to a Modal -> NormalScene -> popBack to the Modal -> now try to pop a scene as a gesture. You actually have to manually track mode and it gets annoying. This should be added to the navigationConfig as an option.

acomito:
I need this! Excited for the PR.

JulianKingman:
@jeffreymendez1993 if mode were allowed in navigationOptions, you could pass it to _getTransitionConfig here: https://github.com/react-community/react-navigation/blob/5e075e1c31d5e6192f2532a815b1737fa27ed65b/src/views/CardStack/CardStack.js#L429
I think you can get the options from _getScreenDetails (here: https://github.com/react-community/react-navigation/blob/5e075e1c31d5e6192f2532a815b1737fa27ed65b/src/views/CardStack/CardStack.js#L145)

Otherwise, maybe you could keep it in the state somehow?

slicejunk:
for people coming from ex-navigation, 
a list of all the features that do no work any more would save a lot of time.

JulianKingman:
I started a separate PR

Issue #1261, opened 2017-04-26
Nested drawers issue


lorenzsell:
I have a drawer nested in a stack nav.  that drawer has a stack nav with another drawer nested inside it.  Before I upgraded versions, I had no trouble opening the main drawer when I was on the main screen with navigate('DrawerOpen') and opening the secondary drawer when I was on the nested stack nav.  After upgrading, navigate('DrawerOpen') always opens the sub drawer and I can't figure out how to open the top level drawer.

lorenzsell:
@cooperka - sorry to ping you directly on this.  i know you worked on some drawer stuff before.  i've literally spent days trying to debug this and still struggling to make it work properly.  do you have any thoughts on what the problem might be?

cooperka:
Hmm, nothing in particular stands out to me. It could either be a new bug, or an old bug you were accidentally relying on that got fixed ;)

To figure out which commit broke your code, you could try using [`git bisect`](http://webchick.net/node/99) which is a great tool for quickly discovering where an issue started. It might be easiest to use the react-navigation example app and tweak it in order to reproduce the bug. If you're able to figure out which commit caused it to stop working, please post it here and we can try to help further!

lorenzsell:
@cooperka - I'm still struggling to figure out what's causing this.  I've copied my nav set up into a bare bones app.  Would you mind taking a look to see if anything strikes you as wrong?  As far as I can tell I've followed all the examples perfectly: https://github.com/lorenzsell/react-nav-test

cooperka:
I appreciate the clean code, it looks great! Nothing I can tell that's wrong with it though :( have you tried `git bisect` yet? It's usually helpful to know which change broke your code, then you know what to look for.

You might also try using `getStateForAction(NavigationActions.navigate({ routeName: 'Login' }))` as suggested here: https://github.com/react-community/react-navigation/pull/1311. I'm not sure if it's related.

lorenzsell:
ok.  i'm trying to figure this out with git bisect.  I'm somewhat new to git and npm.  maybe you can point me in the right direction.  if i understand correctly, in order to do this, i have to fork react-navigation and clone it locally.  then i have to link my project to the locally cloned version by npm installing from the local directory.  then i git bisect the local react-navigation version and refresh my simulator to see where the problem occurs.

I'm having trouble npm installing the local version.  I get:

`Error: Couldn't find preset "react-native" relative to directory "/Users/lorenzsell/DEV/react-navigation"`

[...]

PR #1379, opened 2017-05-04
fix to nested drawer issue

lorenzsell:
This is a minor bug fix to address a nested drawer issue referenced here: #1261 caused by changes from https://github.com/react-community/react-navigation/commit/2dc869fedd27d304c2f0ddd606761320b8bec923.

The issue was caused when you had nested drawer navigators and tried to open the parent navigator.  Instead, the child navigator would open.  This fix checks to see if action.routeName is equal to 'DrawerOpen' or 'DrawerClose' before handling child routers.

This is my first pull request ever, so I apologize if something is not quite right.


src/routers/StackRouter.js
```diff
@@ -186,7 +186,11 @@ export default (
       }
 
       // Handle navigation to other child routers that are not yet pushed
-      if (action.type === NavigationActions.NAVIGATE) {
+      if (
+        action.type === NavigationActions.NAVIGATE &&
+        action.routeName !== 'DrawerOpen' &&
```
satya164: I don't want to hard-code route names in the router. Especially since these names are random names that drawer navigator chose to use and they don't have any special meaning.

satya164:
cc @grabbou

src/routers/StackRouter.js
lorenzsell: Can you recommend a different way to do this?  My understanding is that this code manages navigation to child routers that are not yet pushed but this approach seems to be creating a bug where existing child drawer navigators are being executed instead of the parent drawer navigator.  This was not a problem before this piece of code was introduced in https://github.com/react-community/react-navigation/commit/2dc869fedd27d304c2f0ddd606761320b8bec923.  Am I missing something or is there a better way to do this?

satya164: @lorenzsell I don't have the context for the commit, so I am not sure, maybe @grabbou or @ericvicenti can shed some light.

The ideal fix will be in drawer navigator's code if it involves these route names.

ericvicenti:
I agree with @satya164. This may be the desired behavior for the drawer, so lets override this behavior in the drawer navigator rather than customizing the stack router.

brentvatne:
closing because inactive and also https://github.com/react-navigation/react-navigation/issues/1261#issuecomment-363915547

PR #1384, opened 2017-05-05
Fixed error when navigation.dispatch a RESET action on StackNavigator (#691)

ralkan:
Fixed navigation.dispatch a RESET action from propagating to child navigators by explicitly calling with a stateName param (#691)

When keeping the state of your navigation in Redux and you have more than one navigation, on dispatching a RESET action, all reducers will receive it and try to reset the stack which leads to errors like:

```
There is no route defined for key Home. Must be one of:
'FacebookAuthentication','EmailAuthentication'
```

Now you can explicitly give your StackNavigator a name and make sure that the RESET action is only being handled by that specific navigator's reducer. All the other reducers will just return the current state as is.

Ex:

```
const App = StackNavigator({
  Login: {
    screen: Login
  },
  Authenticated: {
    screen: Authenticated
  }
}, {
  stateName: 'MainAppNav', // If you don't give it a name it will just use the value of initialRouteName
  initialRouteName: 'Login'
})
```

And then navigation.dispatch a reset action like so:

```
const resetAction = NavigationActions.reset({
   index: 0,
   stateName: 'MainAppNav',
   actions: [
     NavigationActions.navigate({ routeName: 'Authenticated' })
   ]
});
this.props.navigation.dispatch(resetAction);
```

### DIFF
--- a/src/navigators/StackNavigator.js
+++ b/src/navigators/StackNavigator.js
@@ -23,6 +23,7 @@ export default (
   stackConfig: StackNavigatorConfig = {},
 ) => {
   const {
+    stateName,
     initialRouteName,
     initialRouteParams,
     paths,
@@ -35,6 +36,7 @@ export default (
     navigationOptions,
   } = stackConfig;
   const stackRouterConfig = {
+    stateName,
     initialRouteName,
     initialRouteParams,
     paths,
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -55,6 +55,9 @@ export default (
 
   const initialRouteName = stackConfig.initialRouteName || routeNames[0];
 
+  // If stateName is not explicitly set, just use the initialRouteName as default
+  const stateName = stackConfig.stateName || initialRouteName;
+
   const initialChildRouter = childRouters[initialRouteName];
   const paths = stackConfig.paths || {};
 
@@ -135,6 +138,7 @@ export default (
         };
         // eslint-disable-next-line no-param-reassign
         state = {
+          stateName,
           index: 0,
           routes: [route],
         };
@@ -243,6 +247,11 @@ export default (
       }
 
       if (action.type === NavigationActions.RESET) {
+        // Prevent the RESET action from propagating on child components if
+        // the stateName param is set and just return state as is
+        if (action.stateName && action.stateName !== state.stateName) {
+          return state;
+        }
         const resetAction: NavigationResetAction = action;
 
         return {


lucasleelz:
@ralkan Can you modify the test case? Make the test pass, I am very looking forward to this feature.

ralkan:
@lucasleelz Yes I'll try to see if I can get the test pass today when I go home.

ralkan:
@lucasleelz Done! 👍

lucasleelz:
@ralkan thank you, but there is a little bit of conflict ~

ralkan:
@lucasleelz Yeah, when I tried to rebase the branch with the current master and resolved the conflict the tests suddenly didn't pass due to jslint errors. So, I rolled back before pushing.

itswaze:
Where should I put that `const resetAction = NavigationActions.reset({` code?

0x6e6562:
@ralkan I can confirm that 3fe8a4f does fix an issue I'm experiencing with not being able to access the root navigator object with parallel nested routers.

Hopefully this can inspire an update to the maintained API, since there seems to be quite a bit of interest in this topic.

Many thanks for the patch :-)

ralkan:
@0x6e6562 I'm happy to see that more people can benefit from this patch! You're very welcome ;)

0x6e6562:
For anybody following this, #789 is apparently an alternative fix for the same issue. However it did not solve the issue for me, whereas this patch does. #789 appears to suggest that if you set a null key, then the action will be forwarded to the root navigator:

```js
const resetAction = NavigationActions.reset({
  index: 0,
  key: null,
  actions: [
   NavigationActions.navigate({ routeName: 'Auth' })
  ]
});
```

However, this did not work for me in practice.

kcfgl:
@0x6e6562 I don't "believe" it solved it for me either.  But with all the various issues and comments surrounding this pattern of navigation, my head is swimming about what I've tried and not tried.  

For now,  I'm running parallel navigation, one is for Pre/Login screens and then a 2nd one, a TabNavigator, with various tabs (and stacks inside the tabs) when authenticated. 

I need to be able to reset a route within a specific tab to go back to index 0 in the tab and delete the previous routes. But when I tried to do the reset above, it worked, but it also replaced all the other initial screens of the other tabs. And I'm still not sure the parallel option is best either, but at least that works.

Wish there was a straightforward solution to deal with unauthenticated and authenticated flows. And easily reset. Because the rest of react-navigation is excellent.

0x6e6562:
@kcfgl I assume you are saying that #789 did not fix your issue.

It might be the case with me that I've misunderstood how to use #789 or whether something else is borked in my app.

However, I think I'm coming from a similar place as you - I've spent a lot of time trying all kinds of different things and maybe I've introduced some operator error into my app.

I also wish there were a more straight forward solution to this common problem. I'm in the process of migrating from expo/ex-navigation to react-navigation. My ex-navigation stack manages the auth/non-auth state in the `render()` method of a custom component, which I gather is an anti-pattern. So I was hoping that react-navigation might have incorporated this into its design, but given the amount of confusion around this subject, that might not be the case (yet).

kcfgl:
@0x6e6562 Yes, sorry for the confusion, https://github.com/react-community/react-navigation/pull/789 did not solve it for me. But like you, I'm wondering if I'm just doing something wrong.  I'm going to give the https://github.com/react-community/react-navigation/commit/3fe8a4f5c2772d14b06250d124e5e15214b7e945 patch a try today. 

But overall, I just wish there was a clear design from react-navigation because in theory, it IS going to be the best navigation out there.

0x6e6562:
@ralkan Is there a way to reset to the initial route of the parent navigator?

In this flow diagram, if I navigate from `Screen C` back to the nested `Auth` stack (flow 1), I will go through the `Auth` stack progressing say from `Screen A` (the initial screen) to `Screen B`. I can then switch to the parallel `Main` stack back to `Screen C` (flow 2). But if I try to repeat flow 1, the state of the `Auth` stack is `Screen B`, whereas I would like to switch to the initial screen by default.

flow

Is this something that you have come across?

0x6e6562:
Feels quite hacky, but resetting both the `Master` stack and the `Auth` stack at the same time achieves the effect of defaulting the flow back to the initial screen:

```js
const resetAuthStack = NavigationActions.reset({
   index: 0,
   stateName: 'AuthStack',
   actions: [
     NavigationActions.navigate({ routeName: 'A' })
   ]
});

const resetMasterStack = NavigationActions.reset({
   index: 0,
   stateName: 'MasterStack',
   actions: [
     NavigationActions.navigate({ routeName: 'Auth' })
   ]
});

this.props.navigation.dispatch(resetAuthStack);
this.props.navigation.dispatch(resetMasterStack);
```
IIRC there is some way of setting a nested target, but I'd need to go back through a bunch of different discussion threads to find out.

augustoabreu:
@0x6e6562 @kcfgl I am also having problems with auth/non-auth flow. Actually i have two main problems: the reset action and hide/show item on TabBar. Besides that, i am trying to reset all state of a tab screen and i'm hope that this PR allow me to do that.

XeniousSoftware:
+1 for this pull request, it works perfectly for navigating between nested navigators.

ralkan:
@kcfgl It should be possible to do what you want. You should give a unique stateName to all of your navigators, and dispatch a reset action just on that specific navigator.

@0x6e6562 @augustoabreu  Yes, that hacky way should be the way for now. (Which actually is not as hacky as it seems) Maybe I could make it so that instead of passing just one stateName as a string, you could be able to pass an array of stateNames to dispatch one reset action and reset 2 or more screen at the same time.

Or I could change this fix so that when you dispatch an action on a parent navigator that it bubbles down to its child navigators. But I'm not sure if that can be easily done right now.

@XeniousSoftware I'm happy to see that this fix worked for you!

0x6e6562:
@ralkan The bubble down approach feels like it might be the most re-useable in a more complex set up - how large would this patch be (considering that it would be good to get this back into the main tree some day)?

Alternatively it would be interesting to see how passing an array of state names would work in practice - in theory it seems like a good generalization.

suhshinhwang:
Do we have an update on this PR?

diz00:
@ralkan Thank you so much for this fix. I've spent 2 days trying to find a way to fix the bug with 2 nested StackNavigators.

This needs to be merged and documented.

dashracer:
Really need this PR

silvainSayduck:
I agree that this pull request is needed (or a similar fix)!

Also, in my app I am sharing the same screen component in multiple stackNavigators, and it feels a bit stupid to need a different name in each case (`nav1Landing: {screen: Landing}` and `nav2Landing: {screen: Landing}` so that `Navigation.navigate('nav1Landing')` doesn't also navigate to the landing screen in nav2 in the background.
For this, I just added the same fix in StackRouter.js for the NAVIGATE action:

```js
/* react-navigation/src/routers/StackRouter.js#L165-174 */
// Handle explicit push navigation action
      if (
        action.type === NavigationActions.NAVIGATE &&
        childRouters[action.routeName] !== undefined
      ) {
        // Prevent the NAVIGATE action from propagating on child components if
        // the stateName param is set and just return state as is
        if (action.stateName && action.stateName !== state.stateName) {
          return state;
        }
        ...
```
This seems to work as intended (but I can't say I have thoroughly tested).
Maybe I'm just really confused on how to use react-navigation, but being able to distinguish between navigators when passing actions seems completely a must, given how many navigators even a simple app can end up using.

Silvaire:
I agree that this pull request is needed (or a similar fix)!

Also, in my app I am sharing the same screen component in multiple stackNavigators, and it feels a bit stupid to need a different name in each case (`nav1Landing: {screen: Landing}` and `nav2Landing: {screen: Landing}` so that `Navigation.navigate('nav1Landing')` doesn't also navigate to the landing screen in nav2 in the background.
For this, I just added the same fix in StackRouter.js for the `NAVIGATE` action:

```js
/* react-navigation/src/routers/StackRouter.js#L165-174 */

      // Handle explicit push navigation action
      if (
        action.type === NavigationActions.NAVIGATE &&
        childRouters[action.routeName] !== undefined
      ) {
        // Prevent the NAVIGATE action from propagating on child components if
        // the stateName param is set and just return state as is
        if (action.stateName && action.stateName !== state.stateName) {
          return state;
        }
        ...
```

This seems to work as intended (but I can't say I have thoroughly tested).

Maybe I'm just really confused on how to use react-navigation, but being able to distinguish between navigators when passing actions seems completely a must, given how many navigators even a simple app can end up using.

maitham1:
Really need this PR

lucasleelz:
Really need this PR！
Every time I build, I have to use my own private service。

GitHubTracey:
I have several nested Navs ( Tab > Stack > Stack etc ) where I would love to be able to reset various levels based on actions from the user. Being able to provide a state name would be fantastic! I'd love to see this PR merged in!

spencercarli:
Hey all - I think this PR looks good! I need to test it but one thing that will be necessary is some documentation on how to use this.

spencercarli:
Also, there are some linting errors.

sergeylaptev:
Any progress on this?

JCMais:
It's saddening that multiple PRs that would solve multiple people problems are being neglected by this project owners.

tiagoalves:
I might be doing something wrong, but this PR doesn't work for me, as is. When I dispatch the `NavigationActions.reset()` action, the new parameter `stateName` is discarded.

I checked the code, and the `NavigationActions.reset()` method doesn't even consider the `stateName` parameter ([here](https://github.com/ralkan/react-navigation/blob/fix-navigation-reset-error/src/NavigationActions.js#L57)). So... how's everyone else being able to use the fix in this PR?

If I create the action manually, like shown below, it works:

```
    dispatch({
      type: NavigationActions.RESET,
      index: 1,
      stateName: 'SignedInNav',
      actions: [
        NavigationActions.navigate({ routeName: "main" }),
        NavigationActions.navigate({ routeName: "newUser" }),
      ]
    })
```

browniefed:
Looks like just some linting issues. Would be great to get this landed.

ohtangza:
Lots of devs seems still suffering from this issue...

rongierlach:
Facing this issue currently, would love to see a merge!

sraka1:
Is this not a duplicate of https://github.com/react-navigation/react-navigation/pull/789?

PR #1327, opened 2017-05-01
Update react-native-tab-view

satya164:
Release notes - https://github.com/react-native-community/react-native-tab-view/releases/tag/v0.0.62
Fixes #1248

PR #1392, opened 2017-05-05
Update dependencies

stoffern:
Update dependencies to latest:
`react-native-drawer-layout-polyfill`

Solves: 
https://github.com/react-community/react-navigation/issues/763


package.json
```diff
@@ -76,8 +76,8 @@
     "hoist-non-react-statics": "^1.2.0",
     "path-to-regexp": "^1.7.0",
     "prop-types": "^15.5.8",
-    "react-native-drawer-layout-polyfill": "1.2.0",
-    "react-native-tab-view": "^0.0.61"
+    "react-native-drawer-layout-polyfill": "1.3.0",
+    "react-native-tab-view": "^0.0.62"
```
onlydave: 0.0.62 does fix the endless loop issue on iOS but breaks nested tabs again on Android

stoffern:
@onlydave reverted

onlydave:
@satya164 Might be able to shed some light on it. 

I've been poking around for the last hour and it seems to work if you disable animations and swipe on all the tabs. Parent and nested. Or if you set `renderPager = (props: *) => <TabViewPagerPan {...props} />;` in TabView.js. Not familiar enough with react-native-tab-view to know what's going on properly.

satya164:
@onlydave I'm not sure what exactly you mean, but it sounds like this https://github.com/react-native-community/react-native-tab-view/issues/225

In that case, it's a very old issue with viewpager and it's just that you're running into it now (probably something with the react upgrade caused it to resurface).

onlydave:
@satya164 Sorry, should have updated my comment after I finished up. 

All I changed in the end was react-native-tab-view to 0.0.62. To see if it would fix the issue with endless looping when moving between tabs quickly. React and react native are at 16.0.0-alpha.6 and 0.44.0. Didn't change them.

Mayabe more closely related to #1234 

I'll try #1327 and see how that goes

satya164:
@onlydave make sure to open an issue with a gif/video if you repro it

stoffern:
Updated now @satya164, my bad

satya164:
you need to rebase. there are conflicts

stoffern:
@satya164 dont know how you rebase conflics.
Please note that this bug renders the `react-navigation` drawer quite useless on Android

skevy:
Done in https://github.com/react-community/react-navigation/commit/f3a958dca1180af87611a35fb64dbd8dd57da898

Thanks for your work on this.

PR #1397, opened 2017-05-05
Add testID to HeaderBackButton

stevoland:
We are using Appium for end2end testing and need to be able to access the back button. An appropriate `testID` is the way to do this.

**Test plan (required)**

Here you can see the back button identified in Appium:

appium2


ericvicenti:
Excellent, thank you!

PR #1404, opened 2017-05-06
Display Transitioner title correctly

ThePaulMcBride:
Fix a bug where the title, "Transitioner", was only visible if you moused near it.


ericvicenti:
This was a useful link. Maybe we can fix the actual display bug instead?

It should be easy. Add a className to the a tag here: https://github.com/react-community/react-navigation/blob/master/website/src/MDPage.js#L65

Then filter by the className here: https://github.com/react-community/react-navigation/blob/master/website/src/App.css#L23,L29

Please test the fix before submitting the PR!

PR #1415, opened 2017-05-08
Fix link to Transitioner docs

keithito:
This changes the docs to correct a broken link to the Transitioner page.

**Test plan (required)**
This is a documentation-only change, there are no tests.


keithito:
Looks like the old link works fine on the documentation site. I'll close this.

Issue #1041, opened 2017-04-14
Nested navigation with redux


MyTotoro:
Hi, I've gotten redux integrated and have the parent level navigator (stack) pretty much figured out. The issue I'm trying to fix is now to integrate/use the nested navigator (drawer)

```
MyStackNavigator
-Splash
-Login
-MainDrawerNavigator

MainDrawerNavigator
-Home
-Settings
-etcscreen
```

I've used reactotron and my state tree looks like this:
![image](https://cloud.githubusercontent.com/assets/3621921/25032896/2f8dfd0e-20a6-11e7-8749-acd034972654.png)

My question is how to handle

1. programmatically opening the drawer
2. navigating to another screen (settings for example)
3. reset navigation stack completely and go back up to Splash screen in parent navigator

I've tried increasing the index to 1 so that DrawerOpen is selected but that didn't seem to accomplish anything:

```
const newState = Object.assign({}, state)
newState.routes[0].index = 1
//return MainNavigation.router.getStateForAction(navigateAction, state)
return newState
```

I looked around on the issues and found that one user suggested using:

```
this.props.dispatch({
  type: 'Navigate',
  routeName: 'Home',
  action: {
    type: 'Navigate',
    routeName: 'Main',
  },
```

I've tried integrating this and didn't get anywhere

samridhgupta:
+1

DongShi:
+1

lorenzsell:
+1

byk04712:
+1

asmurianwar:
+1

zacholas:
+1

dinhha:
+1

X-Tender:
Did someone have an Example with Redux + Nested Navigations?

RUIFERNANDE5:
+1
I need to be able to reset a structure like the following:

```
Auth:
  - Login
  - Register
Main
  - Dashboard
```
I can only navigate from Login to Main/Dashboard with dispaching navigate action.
How can I achieve this with reset?

cooperka:
Related: https://github.com/react-community/react-navigation/issues/937. There don't seem to be docs for any of these use cases :(

State also seems to be really weird when nesting routers. I wonder if there's just a bug.

[...]

PR #1419, opened 2017-05-09
Fix #1312 #1041 #937

### LINKED ISSUE #1312: setParams on root of nested Stack doesn't work
If I have nested StackNavigators (Stack within Stack) and I use setParams from the root of the inner Stack, the params get set on the root of the outer Stack - as they both have the key "Init" and the outer matches first

Why don't Stack roots get assigned unique keys?

helloyou2012:
- Fix #1312 #1041 #937 (Init to Init-UUID)

### DIFF
--- a/src/routers/__tests__/StackRouter-test.js
+++ b/src/routers/__tests__/StackRouter-test.js
@@ -325,7 +325,7 @@ describe('StackRouter', () => {
     const initState = TestRouter.getStateForAction(NavigationActions.init());
     expect(initState).toEqual({
       index: 0,
-      routes: [{ key: 'Init', routeName: 'foo' }],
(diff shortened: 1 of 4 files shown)


src/routers/__tests__/StackRouter-test.js
```diff
@@ -602,15 +602,15 @@ describe('StackRouter', () => {
...
+        key: 'Init-id-0-17',
       },
       state,
     );
     expect(state2 && state2.index).toEqual(0);
     /* $FlowFixMe */
     expect(state2 && state2.routes[0].routes[0].routes).toEqual([
       {
-        key: 'Init',
+        key: 'Init-id-0-17',
```
ericvicenti: This looks pretty tedious to maintain.. because one test may disrupt the key from other tests

PR #1431, opened 2017-05-09
allow uriPrefix to be a regular expression

koenpunt:
when accepting both a protocol url and a normal url, it's hard to define a static prefix.

So for example, in the application I'm currently working on we use iOS applinks for deeplinking, but also want to use a protocol url (`myapp://...`) for in-app linking. Doing this using a static string prefix I also have to use do parsing of the path in `getActionForPathAndParams`, but when using a regex this extra parsing can be omitted.

**NB.** using a regex does already work, but it's nice to have it typed too.

### DIFF
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -15,7 +15,7 @@ import type {
 } from './TypeDefinition';
 
 type NavigationContainerProps = {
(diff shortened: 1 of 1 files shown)


ericvicenti:
Nice catch, thanks!

TIL `split` can take a regex

PR #1371, opened 2017-05-04
changed BackAndroid to BackHandler

Rewieer:
BackAndroid is deprecated since 0.44.0 for BackHandler.

PR #1432, opened 2017-05-09
Migrate from BackAndroid to BackHandler

r3nya:
Hello!

`BackAndroid` [is deprecated now](https://facebook.github.io/react-native/docs/backandroid.html) and because of this I updated code.

Please check my PR.


nihgwu:
#1333 #1371 All with no backward compatibility

r3nya:
@nihgwu oh … I think that I need to close my PR. :)

PR #1483, opened 2017-05-12
	Fix custom back button props & visibility

Misioka:
I have to have option to add custom styled back button.
So I want to provide a React element styled button.
I need navigation props to be passed to component, that I can call this.props.navigatios.goBack()

I also don't want to handle visibility of button on my own.
If I'm in root it should return null, but that functionality is implemented below custom renderer.

For testing you should build custom back button.
e.g.
`  <TouchableOpacity onPress={() => { his.props.navigation.goBack(null) }} style={styles.row}>
        <View style={styles.backCircle}>
<Icon
            name='ios-arrow-dropleft-circle-outline'
            size={20}
            color="#112233"
          />
        </View>
        <Text>{'back'}</Text>
      </TouchableOpacity>`

Then you can use it e.g.

`headerLeft: props => <BackButtonContainer {...props} />`

Test when it is showing, test on press.

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -136,6 +136,7 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
         tintColor={options.headerTintColor}
         title={backButtonTitle}
         truncatedTitle={truncatedBackButtonTitle}
+        titleStyle={options.headerBackTitleStyle}
         width={width}
       />
     );
@@ -283,14 +284,15 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
       position,
       screenProps,
       progress,
+      style,
       ...rest
     } = this.props;
 
     const { options } = this.props.getScreenDetails(scene, screenProps);
-    const style = options.headerStyle;
+    const headerStyle = options.headerStyle;
 
     return (
-      <Animated.View {...rest} style={[styles.container, style]}>
+      <Animated.View {...rest} style={[styles.container, headerStyle, style]}>
         <View style={styles.appBar}>
           {appBar}
         </View>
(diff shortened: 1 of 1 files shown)


pvinis:
no, this was changed just now.
the way you changed it, it will not allow for a left button on the 0-index screen.
this pr should be not accepted.
actually, i will push a new pr today, that adds a `headerBack`. if you use `headerLeft`, its an element, does whatever you want. if you use `headerBack`, its an element that is still a back button. if you use `headerBackTitle` and `headerBackTitleStyle` is a normal os back button, and you only control the text and the style of it.
look forward for it later today.

Misioka:
It's not enaught to change style and text
My client wants 
image
So i need to do it like that

pvinis:
yes. thats what im writing now. you will do
```
headerBack: (
<View>
  <View style={styles.backCircle}>
    <Icon name='ios-arrow-dropleft-circle-outline' size={20} color="#112233" />
  </View>
  <Text>{'back'}</Text>
</View>
)
```

Misioka:
Can you please handle somehow 0-index hide? 
You definately know best solution. Thanks a lot.

pvinis:
yes. the back button will only appear on non-zero indexes. like a back button should :D

pvinis:
#1492 <- and here is it

alenia:
HeaderLeft is not necessarily the back button, so this change will remove the ability for custom headerLeft on pages where the screen index is zero.

alenia:
https://github.com/react-community/react-navigation/pull/1538 here's another PR that might help

vonovak:
#2712 is a similar PR

vonovak:
A PR which allows to customize the back button behaviour was recently merged: #2712

Since the mentioned PR covers the use cases mentioned here, I will close this PR as being redundant. With that being said, thank you for your efforts, @Misioka !

PR #1492, opened 2017-05-12
allow back button customization. more than just title change and tint.

pvinis:
This PR allows the user to use any react element as a back button.
We can already use something like
```
const Navigator = StackNavigator(
  {
    Splash: { screen: SplashScreen },
    Main: { screen: MainScreen },
  },
  {
    initialRouteName: 'Splash',
    navigationOptions: {
      headerMode: 'screen',
      headerStyle: {
        backgroundColor: Colors.navBarBackground,
        borderBottomWidth: 1 / PixelRatio.get(),
        borderBottomColor: Colors.separator,
      },
      headerTitleStyle: {
        color: Colors.textBlue,
        fontFamily: Fonts.type.base,
      },
```
Now we can also add something like
```
    headerBack: <Image source={Images.iconBack} />,
```
or like
```
    headerBack: (
      <View style={{
        alignItems: 'center',
        flexDirection: 'row',
      }}>
        <Image source={Images.iconBack} />
        <Text style={{paddingRight: 10, }} numberOfLines={1}>BACK</Text>
      </View>
    ),
```
This will be wrapped and made into a button that goes back.
The difference with `headerLeft` is basically that `headerLeft` can be anything, and added in any screen. `headerBack` is always a button that goes back, that contains any custom element the user gives, and it will only appear on non-zero index screens.

Without a custom `headerBack` (regular back button, automatically rendered, as before):
img_0052
(`PavlosTv` is the title of the previous screen)

With a custom `headerBack` (custom element, wrapped as a `goBack` button):
img_0051

With a custom `headerBack` (custom element, wrapped as a `goBack` button):
img_0053

### DIFF
--- a/src/views/Header.js
+++ b/src/views/Header.js
@@ -7,6 +7,7 @@ import React from 'react';
 import { Animated, Platform, StyleSheet, View } from 'react-native';
 
 import HeaderTitle from './HeaderTitle';
+import HeaderBackButtonWrapper from './HeaderBackButtonWrapper';
 import HeaderBackButton from './HeaderBackButton';
 import HeaderStyleInterpolator from './HeaderStyleInterpolator';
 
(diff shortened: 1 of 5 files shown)


koenpunt:
Isn't the same achievable with something like:

```js
navigationOptions: ({ navigation }) => {
  const left = navigation.state.index === 0 ? null : <MyBackButton/>;
  return { headerLeft: left };
}
```

matthamil:
This feature already exists for react-navigation, as @koenpunt has mentioned. If this is a new feature, how is this different from the existing API as seen here https://github.com/react-community/react-navigation/blob/ff99d6ec0a2d7a62adfd8d1e6a3c906fcedf294a/src/views/Header.js#L151

https://reactnavigation.org/docs/navigators/stack#headerLeft

koenpunt:
There might be a caveat here, because I think the index isn't passed to the `navigationOptions`. To workaround that fact you could set `headerLeft: null` on the root navigator to explicitly disable the back button.

pvinis:
I will check, but I also think the index is not in there. And it feels cleaner this way anyway. we have `headerLeft` for `put this on the left always`, which could be hamburger button, or search or something, and we now have `headerBack`, which is basically customizing the regular `back` button. Dont you guys agree?

aksonov:
@koenpunt Is it bug that index is not passed to navigationOptions? Is it going to be fixed?

pvinis:
god.. this repo needs more active people that work on issues and prs.

src/views/Header.js
```diff
@@ -119,25 +120,35 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
...
       return null;
     }
-    const backButtonTitle = this._getBackButtonTitleString(props.scene);
-    const truncatedBackButtonTitle = this._getTruncatedBackButtonTitle(
-      props.scene
-    );
     const width = this.state.widths[props.scene.key]
       ? (this.props.layout.initWidth - this.state.widths[props.scene.key]) / 2
       : undefined;
+    var backButtonElement = options.headerBack;
```
matthamil: I think it would be better to extract the logic that determines `backButtonElement` into a function that returns the value of what `backButtonElement` should be rather than imperatively determining what its value is. See `this._getBackButtonTitleString` as an example.

pvinis:
should be better now

nonameolsson:
This is a great PR! Could it be fixed and merged?

pvinis:
i will try to fix it in the weekend, even though im planning to move away from react-navigation soon.

pvinis:
closing in favor of #2869.
i fixed everything, added a test/example, and it sits on top of master, which is one commit ahead of beta.16, so it should be pretty easy to merge.
thanks again.

PR #1500, opened 2017-05-14
stop using Animated.spring

nihgwu:
following #1036

`Animated.spring` with `bounciness = 0` is not the right way to mimic a decay animation, the animation doesn't finish immediately when the `Animated.Value` gets to the `toValue`, it will still bounce in a short time

I didn't measure the bouncing duration, but I can tell it's there, in my app, I get two screens with different `barStyle` of `StatusBar`, it's visually sensible that I have to wait for about 100ms until the `StatusBar` changes the barStyle. Although in #1036 I've skipped the `Transitioner` animation after swiped back, but the bouncing time is still there, after change `Animated.spring` to `Animated.timing`, the `StatusBar` changes the barStyle immediately after transition animation

I also changed the animation config in `src/views/TransitionConfigs.js` because when you click the `goBack` button, the bouncing time is also there with the same reason

This PR also resolved the bouncing issue when swiping back with a extreme speed introduce in #1036 

@ericvicenti maybe you would be interested in this 😄

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -10,7 +10,6 @@ import {
   Platform,
   View,
   I18nManager,
-  Dimensions,
   Easing,
 } from 'react-native';
 
@@ -37,8 +36,6 @@ import type {
 import TransitionConfigs from './TransitionConfigs';
 
 const emptyFunction = () => {};
-
-const windowSize = Dimensions.get('window');
 
 type Props = {
   screenProps?: {},
@@ -67,7 +64,9 @@ type Props = {
 };
 
 /**
- * The duration of the card animation in milliseconds.
+ * The max duration of the card animation in milliseconds after released gesture.
+ * The actual duration should be always less then that because the rest distance 
+ * is always less then the full distance of the layout.
  */
 const ANIMATION_DURATION = 250;
 
@@ -234,8 +233,10 @@ class CardStack extends Component {
     if (headerMode === 'float') {
       floatingHeader = this._renderHeader(this.props.scene, headerMode);
     }
-    const { navigation, position, scene, mode, scenes } = this.props;
+    const { navigation, position, layout, scene, scenes, mode } = this.props;
     const { index } = navigation.state;
+    const isVertical = mode === 'modal';
+
     const responder = PanResponder.create({
       onPanResponderTerminate: () => {
         this._isResponding = false;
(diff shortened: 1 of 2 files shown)


src/views/CardStack.js
```diff
@@ -65,7 +69,7 @@ type Props = {
 /**
  * The duration of the card animation in milliseconds.
  */
-const ANIMATION_DURATION = 200;
+const ANIMATION_DURATION = 250;
```
ericvicenti: can you describe how you came to this number? Ideally it would be via measurement, but you can also say 'approximation'

src/views/CardStack.js
```diff
@@ -36,6 +38,8 @@ import TransitionConfigs from './TransitionConfigs';
 
 const emptyFunction = () => {};
 
+const windowSize = Dimensions.get('window');
```
ericvicenti: I think we may be able to use `props.layout` instead of `Dimensions` for this. Generally Dimensions is a hack because the screen size may theoretically change over time (and windowSize will get stale). Or theoretically the card stack may not take up the full screen and this would be wrong

src/views/CardStack.js
```diff
@@ -199,30 +195,28 @@ class CardStack extends Component {
     animatedSubscribeValue(props.position);
   }
 
-  _reset(resetToIndex: number, velocity: number): void {
-    Animated.spring(this.props.position, {
+  _reset(resetToIndex: number, duration: number): void {
+    Animated.timing(this.props.position, {
       toValue: resetToIndex,
-      duration: ANIMATION_DURATION,
+      duration,
```
ericvicenti: Ah, I like this technique of controlling duration instead of the 'velocity'. Its a lot more clear what is happening

src/views/CardStack.js
nihgwu: It's a approximate value from `src/views/TransitionConfigs.js`, I assume the time for swipe a full screen is equal the `goBack` duration, although that value is configurable via `TransitionConfig`, but actually if we swipe back, the distance is less then the full distance, so the duration should be always less then 250ms, and I think it a proper value, unless we can configure it or read from `transitionConfig`
So I'd prefer to update the comment right now

ericvicenti: Sure, feel free to just update the comment and mention the conflict with TransitionConfigs

src/views/CardStack.js
nihgwu: updated, and the test failure is irrelevant with this PR, seems it was introduced by the recent change on the `Init` key

src/views/CardStack.js
nihgwu: The core reason for this PR is that `Animated.spring` with `bounciness = 0`  will introduce a bouncy gap _after_(visually) transition animation

ericvicenti:
Ok, looking good!

PR #1509, opened 2017-05-14
Update/Fix Prettier + ESLint Config

skevy:
Just a few updates to code style.

- Changes Prettier trailing comma setting to "es5" so we can write Node scripts
- Adds eslintignore
- Removes eslint config in website dir, so it will use the repo-wide config; runs Prettier on website
- Runs Prettier on other misc files


ericvicenti:
nice! Will we add a test for prettier to help us avoid regressing it?

skevy:
@ericvicenti what kind of test?

skevy:
To be clear -- this wasn't a regression.

skevy:
Do you want to just run prettier in CI or something? Eslint should catch it actually. If somebody doesn't run eslint --fix before they commit (or have it built into their editor), Eslint will fail with an error (and we're already running eslint in CI)

ericvicenti:
Well it would be nice if we tested that every source file in the repo === prettier(the source file). That way we would have good signal when people inevitably forget to run it on their PRs

ericvicenti:
Ah, ok. I didn't realize that ESLint had that level of verification

skevy:
> Ah, ok. I didn't realize that ESLint had that level of verification

Yep -- we use https://github.com/prettier/eslint-config-prettier.

ericvicenti:
Ok, so why were tests passing before you made all these fixes? I would think that they would fail lint

skevy:
Because I updated Prettier.

You just reminded me though -- Eslint isn't being run on the `website` folder in CI. I will add that here.

PR #1510, opened 2017-05-14
Fix Flow and Tests

skevy:
- Fixes NavigationActions.js flow typing (we need to use string literals -- flow can't infer these)
- Updates eslint on StackRouter/TabRouter tests
- Use expect.objectContaining so that we don't rely on keys

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -1,11 +1,15 @@
-const namespacedAction = (action: string) => `Navigation/${action}`;
+/**
+ * @flow
+ */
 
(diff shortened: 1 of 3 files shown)


ericvicenti:
Looks great, thanks for fixing this! Are there any other tests that need to be updated?

skevy:
@ericvicenti I don't think so? I was just trying to get it into a passing state. I'm happy to come back around at a different time to fix other tests

ericvicenti:
Thanks for getting it green! I'm thinking we should switch to objectContaining for other tests.

PR #1515, opened 2017-05-14
StackRouter.getStateForAction: replace instead of push when only child changes

faceyspacey:
This is a fix for this issue: https://github.com/react-community/react-navigation/issues/1508

It's more of a preliminary proposal, as I'm not yet sure all the far-reaching effects. What needs to be tested is a DrawerRouter and another StackRouter as a the child. This assumes the TabRouter is a child and works nicely. I'll hopefully be testing other routers soon and pushing any changes. If anyone with more experience with `react-navigation` could take a look at this, I'd greatly appreciate it.


faceyspacey:
this is a key one--probably the truest *bug fix* of all my PRs. ...i made an additional change from the first commit. the fix even applies without nested child routers.

ericvicenti:
This code has changed quite a bit, and now that we support idempotent navigation with key, I suspect that the best fix is within getActionForPathAndParams. Let me know if you can still help with this, @faceyspacey, and I can guide you on the fix!

PR #1516, opened 2017-05-14
provide config options for `openPath` and `closePath`

faceyspacey:
This PR allows you to replace the `DrawerOpen` and `DrawerClose` default paths. I think it's very important because if you have a top level DrawerNavigator, and you're using universal deep-linking basically no path you use is going to match up with links from your website because they will all start with `DrawerOpen/Close`.


brentvatne:
looks like these exist now! https://github.com/react-navigation/react-navigation/blob/95565487ec1cd209b8bcee6c962a67f85e497bb8/src/navigators/DrawerNavigator.js#L41-L43

PR #1520, opened 2017-05-15
Fix backAndroid warning

GregoryNative:
Pull request created for fixing this warning:
"Warning: BackAndroid is deprecated.  Please use BackHandler instead."


src/PlatformHelpers.native.js
```diff
@@ -1,5 +1,13 @@
...
+import {
+  BackAndroid as OldBackAndroid,
+  BackHandler as NewBackAndroid,
+  Linking,
+} from 'react-native';
 
-export { BackAndroid, Linking };
+const fullVersion = require('react-native/package.json').version.split('.')[1];
+
+let BackHandler = fullVersion < 44 ? OldBackAndroid : NewBackAndroid;
```
chirag04: just look for BackHandler and use it if available otherwise use BackAndroid. no need to check for RN version

GregoryNative: yeap, you are right.

src/PlatformHelpers.native.js
```diff
@@ -1,5 +1,11 @@
 /* @flow */
 
-import { BackAndroid, Linking } from 'react-native';
+import {
+  BackAndroid as OldBackAndroid,
+  BackHandler as NewBackAndroid,
```
skevy: Import this just as BackHandler.

src/PlatformHelpers.native.js
```diff
@@ -1,5 +1,11 @@
 /* @flow */
 
-import { BackAndroid, Linking } from 'react-native';
+import {
+  BackAndroid as OldBackAndroid,
```
skevy: Call this `DeprecatedBackAndroid`.

PR #1523, opened 2017-05-15
Expo Playground Demo

DanielMSchmidt:
# Motivation

To give potential users of our library a better understanding of the look and feel on real devices having the current playground always up to date on expo can provide a lot of benefit. As the expo version uses the normal RN playground as dependency there is no "magic" that might break in the future.

# Test plan

You can cd into the expo playground example directory, `npm install` and try the expo commands. (You need to install the expo CLI before that though)


circle.yml
```diff
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
```
skevy: Looks like you have a merge conflict here.

website/src/App.js
```diff
@@ -43,13 +43,24 @@ type DocPageConfig = {
 };
 
 const createDocPage = (
+<<<<<<< HEAD
```
skevy: Merge conflict...

website/src/AppFrame.js
```diff
@@ -51,9 +55,13 @@ class AppFrame extends React.Component {
       state: navigation.state.routes[navigation.state.index],
     });
     const { state } = navigation;
+<<<<<<< HEAD
```
skevy: Merge conflict...

circle.yml
skevy: Looks like you missed it here @DanielMSchmidt.

skevy:
One more thing @DanielMSchmidt -- you need to run eslint --fix on the files you changed, so that they're prettiered.

DanielMSchmidt:
Sure, will do so later

DanielMSchmidt:
@skevy Ran prettier, can we add running the format command in the readme, as pre-commit hook, and in the CI maybe? Could do a PR if you like 👍

skevy:
@DanielMSchmidt yah -- we should definitely do this.

We shouldn't run Prettier FOR you, but we should run eslint pre-commit and recommend that you run it.

We can use Husky + Lint-Staged for this -- maybe you can put together a PR?

MrHubble:
@DanielMSchmidt @skevy when I scan the QR code on my phone I get "no experience found": 

file_000

PR #1525, opened 2017-05-15
Exclude __tests__ from published files

migueloller:
### Motivation
When installing `react-navigation` there are various Flow errors that happen as a result of test files within the `src` folder. This PR removes those files from the build output.

### Test Plan
No new code added.

### Code Formatting
No new code added.


migueloller:
Thank you! 😄

PR #1566, opened 2017-05-17
feature($navKey): add router-aware actions + reducers via navKey-tagged actions

faceyspacey:
## Motivation
The motivation behind this PR is to make it so when using Redux your actions are tagged to a specific router, and only the corresponding reducer (via `router.getStateForAction`) responds to the action. This is done by tagging all actions dispatched through `navigation.dispatch` with a `navKey` and by creating your Navigators with a `NavigatorOption` of the same `navKey` value, eg:

```js
export const MyStack = StackNavigator(routeConfigs, { 
   navKey: 'myStack'
})
```

Sample action dispatched:
```js
{
   type: 'Navigation/NAVIGATE,
   routeName: 'Foo',
   navKey: 'myStack'
}
```

*reducer*:
```js
import MyStack from '../components/MyStack'

const myStackReducer = (state, action) => {
  return MyStack.router.getStateForAction(action, state) || state
}

combineReducers({ myStack: myStackReducer })
```

The primary mechanism to achieve this is in `addNavigationHelpers`:

```js
// ...
export default function<S: *>(navigation: NavigationProp<S, NavigationAction>) {
  const oldDispatch = navigation.dispatch

  navigation.dispatch = (action: NavigationAction): NavigationAction => {
    action.navKey = typeof action.navKey !== 'undefined'
      ? action.navKey
      : navigation.navKey

    return oldDispatch(action)
  }

  return {
    ...navigation,
    goBack: (key?: ?string): boolean =>
      //...
```

Then in `getStateForAction` for all routers you have (simplified):

```js
      if (action.navKey && action.navKey !== navKey) {
        return state
      }
     // ...
```

After that, it's just up to the navigators to have a config option to specify the navKey, which is passed down to the router it creates. 

##Test plan (required)
I know there is a desired form for these PRs, but there are ton of issues and PRs on this repo, and I'm not sure if the maintainers are interested in some of my PRs. I'm more than willing to take these PRs all the way home. I'd just like to gauge interest and discuss things first, as this is a large codebase. And by the way this package is amazing, well-architected, and nailing it--I realize it's a long row to tow in order to essentially become feature-complete. So I made a few issues, and I figured rather than just requesting stuff, I'd essentially put in some work and *showcase* what I had in mind. After discussing things--perhaps learning about considerations that are hard to determine when new to the codebase--I'm more than willing to take direction and do what you need.

## Final Notes
As for this feature, basically it seems complete redux-integration + deep-linking isn't fully handled by react-navigation. Like, there is quite a bit of work to do to get it correct. For one, using  `getActionForPathAndParams` in combination with `getPathAndParamsForState` + deep-links produces inaccurate state with nested routers. I have a PR for that:

https://github.com/react-community/react-navigation/pull/1515
> note: at the end of the day, that issue with the deep-links is that `getStateForAction` doesn't respond well to actions with nested actions that links produce when transformed into their corresponding action. The way the system currently excels is when dispatching flatter actions, which is why even though the above PR doesn't directly handle links, it's very related to it (since links converted to actions for nested routers result in nested actions). Here's the associated Issue: https://github.com/react-community/react-navigation/issues/1508

Another such issue is the one this PR handles. It seems making Navigation powered by state was what inspired the idea of this package, but when it came time to build it, it was built based on its own internal state system, and didn't fully take into consideration things like multiple disconnected router "families" and the like. Anyway, these are all problems very important to me that I've been working on holistically coming up for solutions for you guys. Please search your PRs and Issues for my username and you'll see all my stuff is geared towards this one agenda of **redux-integration + deep-linking + multiple router families**. My goal is to give React Navigation first-class support for this stuff. I bring this up because it's all very related, and for effective decisions to be made they should be made as a group.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -10,15 +10,10 @@ export type HeaderMode = 'float' | 'screen' | 'none';
 
 export type HeaderProps = NavigationSceneRendererProps & {
   mode: HeaderMode,
-  router: NavigationRouter<
-    NavigationState,
-    NavigationAction,
-    NavigationStackScreenOptions
-  >,
-  getScreenDetails: NavigationScene => NavigationScreenDetails<
-    NavigationStackScreenOptions
-  >,
-  style: Style,
+  router: NavigationRouter<NavigationState, NavigationAction, NavigationStackScreenOptions>,
+  getScreenDetails: (
+    NavigationScene,
+  ) => NavigationScreenDetails<NavigationStackScreenOptions>,
 };
 
 /**
@@ -73,7 +68,7 @@ export type NavigationStateRoute = NavigationLeafRoute & {
 
 export type NavigationScreenOptionsGetter<Options, Action> = (
   navigation: NavigationScreenProp<NavigationRoute, Action>,
-  screenProps?: {}
+  screenProps?: {},
 ) => Options;
 
 export type NavigationRouter<State, Action, Options> = {
@@ -90,12 +85,10 @@ export type NavigationRouter<State, Action, Options> = {
    */
   getActionForPathAndParams: (
     path: string,
-    params?: NavigationParams
+    params?: NavigationParams,
   ) => ?Action,
 
-  getPathAndParamsForState: (
-    state: State
-  ) => {
+  getPathAndParamsForState: (state: State) => {
     path: string,
     params?: NavigationParams,
   },
@@ -119,7 +112,7 @@ export type NavigationScreenOption<T> =
   | T
   | ((
     navigation: NavigationScreenProp<NavigationRoute, NavigationAction>,
-    config: T
+    config: T,
   ) => T);
 
 export type Style =
@@ -147,13 +140,11 @@ export type NavigationScreenConfigProps = {
 
 export type NavigationScreenConfig<Options> =
   | Options
-  | (NavigationScreenConfigProps &
-    (({
-      navigationOptions: NavigationScreenProp<
-        NavigationRoute,
-        NavigationAction
-      >,
-    }) => Options));
+  | (NavigationScreenConfigProps & ((
+    {
+      navigationOptions: NavigationScreenProp<NavigationRoute, NavigationAction>,
+    },
+  ) => Options));
 
 export type NavigationComponent =
   | NavigationScreenComponent<*, *>
@@ -175,6 +166,7 @@ export type NavigationParams = {
 export type NavigationNavigateAction = {
   type: 'Navigation/NAVIGATE',
   routeName: string,
+  navKey?: ?string,
   params?: NavigationParams,
 
   // The action to run inside the sub-router
@@ -183,11 +175,13 @@ export type NavigationNavigateAction = {
 
 export type NavigationBackAction = {
   type: 'Navigation/BACK',
+  navKey: string,
   key?: ?string,
 };
 
 export type NavigationSetParamsAction = {
   type: 'Navigation/SET_PARAMS',
+  navKey?: ?string,
 
   // The key of the route where the params should be set
   key: string,
@@ -198,12 +192,14 @@ export type NavigationSetParamsAction = {
 
 export type NavigationInitAction = {
   type: 'Navigation/INIT',
+  navKey?: ?string,
   params?: NavigationParams,
 };
 
 export type NavigationResetAction = {
   type: 'Navigation/RESET',
   index: number,
+  navKey?: ?string,
   key?: ?string,
   actions: Array<NavigationNavigateAction>,
 };
@@ -211,6 +207,7 @@ export type NavigationResetAction = {
 export type NavigationUriAction = {
   type: 'Navigation/URI',
   uri: string,
+  navKey?: ?string,
 };
 
 export type NavigationStackViewConfig = {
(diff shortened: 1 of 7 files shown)


faceyspacey:
The primary remaining consideration is directly targeting nested routers. Currently there are issues with deep-linking where if you take a path, convert it to an action via router.getActionForPathAndParams the resulting action (which will have nested actions) isn't handled the same way a flat action is handled, which attempts to let child routers handle it in getStateForAction. So perhaps simply fixing that bug (which I have a have another PR which attempts to at least solve part of it: #1515), will solve the problem of dispatching actions within one router family that targets another router family. But if not, this PR may want to be amended to allow for child routers (which don't have a corresponding reducer of their own), to still have their own navKey, which can be directly targeted, even though their state is calculated in a parent reducer.

brentvatne:
curious what you think about this @ericvicenti

ericvicenti:
Hey @faceyspacey, thanks for putting so much thought into this- I think you’re on to something here, and we have been moving in this direction a bit, with the new push/pop actions that are able to target the router by key.

One concern with this specific proposal is how navigation state for child routers get a key defined by the parent router. This becomes handy if, for example, you want to push two of the same child routers on to a stack. If the key was defined statically like this, they would conflict. 

So although this may be 9 months late, I’d love to work with you on improving the redux and deep linking use cases. And I definitely enjoyed reading your comments! Would you be interested in reviving this effort?

faceyspacey:
Hey guys, super overloaded right now. But while I have your attention here's what I want to propose, as it's something I'm going to be doing, but with your buy-in could be better: an independent StackNavigator. No nesting. No other navigators. Made with redux in mind, and ultimately since it's just a StackNavigator, it probably can just receive everything as props, regardless of whether you're using Redux. 

When React Navigation was just about to come out a year ago, that's what I thought you were going to release. For me, at least, that's the pain point I envisioned as the primary goal. 

On another note: I love the innovation in the nested routes concept. I totally zeroed on in one of your initial posts about React Navigation regarding the difficulty in reconciling history stacks. I've spent a lot of time thinking about how browser history stacks can in fact be reconciled across multiple navigators--and I've come up with several algorithms that work, as long as each entry is tagged with a `NavKey`. In addition, in Redux-First Router, I released a package that makes it so you use React Navigation with all the navigators *NOT* nested, and instead used "flatly." 

I came to a conclusion as a result of all this: the goal of React Navigation seemed to be anything and everything to everyone, and essentially bit off more than it can chew. Built-in nesting + state management added to the overload. Essentially instead of the focus being the best Navigator driven by external state, it became the best complete navigation solution for those that don't want to manage their own state. I think there's a huge opportunity in going back to the basics here. I for one don't need state-driven drawers or tab bars. For intermediate+ developers it's very easy to do that. It's totally possible to combine them with the StackNavigator solution Im proposing just by using `navKeys`. I've done it. Here's all the info about what I built for integrating Redux First Router with React Navigation last spring:

https://github.com/faceyspacey/redux-first-router/blob/master/docs/react-native.md#first-class-react-navigation-support

Unfortunately I put that initiative on pause. And my conclusion was to circle back and produce an independent StackNavigator. 

I know you guys are super busy guys--one last thing: So I mentioned reconciling multiple navigators from one history stack array. Obviously this is less of a problem in Native, but for the browser I've managed to put together a solution that properly remembers your entire browser history and keeps it in sync with the secret array of entries the browser is keeping. And it works in all modern browsers, and has solid fallbacks, etc. It took going the mile to put multiple workarounds to use to make possible, but I've felt it essentially BS that we've been settling for browsers not giving us this capability. It's meant mobile web apps can't have Navigators that stay in sync with the URL, and as a result we see far less navigators on web, certainly not ones with RESET + GO. And as far as RESET + GO goes, they are the least frequent action, but happen at the most crucial points in time (e.g. after making a purchase and you want to reset the stack, etc). So anyway I put my mind to making the next release of Redux-First Router (which is being renamed to Rudy) to solving this problem. It's now officially as if the browser, popstate, etc, doesn't exist from the point of view of a Redux application, and you get full history tracking. WHICH MEANS: in combination with navKeys, we can properly reconcile the histories array to do things like reset one of multiple navigators. You have to make some choices, but there's multiple options to do it. I'll give one example: 

```js
const historyEntries = [
  { path: '/step-1', navKey: 'mainStack' },
  { path: '/step-2', navKey: 'mainStack' },
  { path: '/settings', navKey: 'drawerStack' },
  { path: '/settings/sound', navKey: 'drawerStack' },
  { path: '/step-3', navKey: 'mainStack' }
]
```

Now it comes time to reset the stack after step-3. One algorithm eliminates all entries no in the `mainStack`:

```js
const historyEntries = [
  { path: '/step-1', navKey: 'mainStack' },
]
```

Another, pushes `/step-1` to the front while keeping entries from other navKeys just as they are:

```js
const historyEntries = [
  { path: '/settings', navKey: 'drawerStack' },
   { path: '/step-1', navKey: 'mainStack' },
]
```

It's been a while but I believe I came up with 3-5 different algos. I wrote 3 of them and tested it with React Navigation and they worked swimmingly.

In short, for a large # of use-cases, it is possible to reset independent navigators, while reconciling the overarching history stack (which in the browser we have no choice but to allow to continue to drive the app, eg. back/next browser buttons). Picking an algo can be as simple as choosing a key/val in your Redux action.

So this has been my focused. I noticed you wanted to make your tools available for web. The reducers/routers obviously already are. I see a clear far more narrow path forward for this stuff if it's *just the StackNavigator*. And with it far easier to make cross-compatible between Native + the browser. And more importantly, because there are finally entries-management tools for the browser, actually applicable in the browser. For, a StackNavigator in the browser that gets un-synced from the address bar is ultimately useless and unsafe bet to hedge your app on. I think many have made the decision just to make apps where *order* isn't important--i.e. you can link to/from any page.

I'll be releasing Rudy in a week or 2. I think once you see it, the idea of a cross-compatible *standalone* StackNavigator will be something we can collaborate on. As far as RNav, I'm just not going to be able to do it--the approach of combining all these navigators encumbers things too much and isn't needed for advanced developers willing to put it all together in order to maintain flexibility/control (reads: less potential for having to resolve to workarounds).

My plan is/was to rip out StackNavigator from here or wix's. Whatever was the most advanced and most ready for the given strategy, come time to execute. I'm estimating 6-8 weeks away before I can devote time to Navigators again. But I know you guys are *The Navigator Guys*. I've studied plenty of this code. It's been a while, but I know I got the chops to perform the surgery. Perhaps your blessing is all that would be required. Wouldn't want to do it and you guys feel taken advantage of. Even tho our code is out there as free open source, I feel it's always best to talk to the original authors about serious forks. If we did it together in some capacity, it could be a win-win for us all. **The React world needs a powerful flawless standalone state-driven Native navigator by now. :)**

PR #1574, opened 2017-05-18
Fix flow type for NavigationParams

### LINKED ISSUE #252: Values for NavigationParams are defined as strings (is it on purpose?)
TypeDefinition.js defines `NavigationParams` as a map where each value (parameter) must be a string.

`export type NavigationParams = {
  [key: string]: string,
};`

Yet, in the same file there's a documented example, where the value is an int. 

```
/**
   * Params passed to this route when navigating to it,
   * e.g. `{ car_id: 123 }` in a route that displays a car.
*/
```

To me it looks like NavigationParams type definition is wrong, but I couldn't be sure, since you may have thought about URL routing, where each parameter must be string.

yspychala:
Fix #252 

In _TypeDefinition.js_, `NavigationParams` should use a `mixed` type instead of `string`, since you may have a param with an integer, or an array for example.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -169,7 +169,7 @@ export type NavigationNavigator<T, State, Action, Options> = ReactClass<T> & {
 };
 
 export type NavigationParams = {
-  [key: string]: string,
+  [key: string]: mixed,
 };
 
 export type NavigationNavigateAction = {


mikelambert:
Also fixes https://github.com/react-community/react-navigation/issues/1640

Would love to see this merged in!

PR #1577, opened 2017-05-18
import View and Button from react-native

ldehai:
When add new Sceen, the sample use 'View' and 'Button', but the code didn't notice user to import them.

 It will be crash when running. Especially for the novice, it will cost time to find the problem.

### DIFF
--- a/docs/guides/Guide-Intro.md
+++ b/docs/guides/Guide-Intro.md
@@ -116,6 +116,19 @@ const SimpleApp = StackNavigator({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


koenpunt:
If you follow the guide from top to bottom, a set of imports is already documented. 
So I think this change would be redundant.

ldehai:
Yes, you're right. init app has imported 'View', but no 'Button'. like this:
```
import {
  AppRegistry,
  StyleSheet,
  Text,
  View
} from 'react-native';
```

PR #1582, opened 2017-05-18
Revert "Tweak iOS animation easing, gesture tweaks"

nihgwu:
https://github.com/react-community/react-navigation/commit/296becd5d1491f2c00bae0657a6bfaa199ed2f8f#commitcomment-22192854

**PLEASE DO NOT MAKE ANY BREAKING CHANGE ON THE MASTER BRANCH DIRECTLY**

### DIFF
--- a/src/views/CardStack.js
+++ b/src/views/CardStack.js
@@ -72,24 +72,24 @@ type Props = {
  * The actual duration should be always less then that because the rest distance 
  * is always less then the full distance of the layout.
  */
-const ANIMATION_DURATION = 500;
+const ANIMATION_DURATION = 250;
 
 /**
  * The gesture distance threshold to trigger the back behavior. For instance,
- * `1/2` means that moving greater than 1/2 of the width of the screen will
+ * `1 / 3` means that moving greater than 1 / 3 of the width of the screen will
  * trigger a back action
  */
-const POSITION_THRESHOLD = 1 / 2;
+const POSITION_THRESHOLD = 1 / 3;
 
 /**
  * The threshold (in pixels) to start the gesture action.
  */
-const RESPOND_THRESHOLD = 20;
+const RESPOND_THRESHOLD = 12;
 
 /**
  * The distance of touch start from the edge of the screen where the gesture will be recognized
  */
-const GESTURE_RESPONSE_DISTANCE_HORIZONTAL = 25;
+const GESTURE_RESPONSE_DISTANCE_HORIZONTAL = 35;
 const GESTURE_RESPONSE_DISTANCE_VERTICAL = 135;
 
 const animatedSubscribeValue = (animatedValue: Animated.Value) => {
@@ -202,7 +202,7 @@ class CardStack extends Component {
     Animated.timing(this.props.position, {
       toValue: resetToIndex,
       duration,
-      easing: Easing.linear(),
+      easing: Easing.inOut(Easing.ease),
       useNativeDriver: this.props.position.__isNative,
     }).start();
   }
@@ -218,7 +218,7 @@ class CardStack extends Component {
     Animated.timing(position, {
       toValue,
       duration,
-      easing: Easing.linear(),
+      easing: Easing.inOut(Easing.ease),
       useNativeDriver: position.__isNative,
     }).start(() => {
       this._immediateIndex = null;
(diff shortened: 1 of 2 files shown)


skevy:
Hi there!

Sorry if this affected you.

We need to be able to push things to the master branch to be able iterate quickly.

Most of the time, things will go through PR review in public. In this case, it was a small change and was discussed privately among some members of the core team. The intention was that further review of this change would go through review (aka, UX review, not code review) before we launched the next beta to the public.

***The master branch should not be treated as stable.***

This change _was_ tested, and this transition duration is done this way to match the default iOS push transition more closely.

The reason that it's 500ms is because the actual duration of the default iOS push transition is 500ms. Even though it seems shorter than that in practice, I took the time and measured this in After Effects, along with the bezier curve.

Generally, unless you're paying _extremely_ close attention, in practice you may not even notice the above change in your app. But the idea is that it makes the animation smoother and more responsive.

This change is a positive one, and there will be more animation / gesture changes coming in the future to align react-navigation more closely with platform default.

Thanks for your concern!

skevy:
Also, re: the stability of the master branch.

As you have demonstrated with this PR -- reverting is easy.

When we committed the gesture change, we understood that it was breaking, and that it could cause issues.

As I look at the PRs / issues you've personally filed towards correcting the issues introduced there, you've done great work!

Basically, the way that we need to run this project, at this point in it's life, is to move fast and attempt to fix some of the long standing issues with features + user experience. If some of those things happen without extensive enough review and cause some kind of regression, we can take the approach that you've taken previously, which is to accept PRs that fix these issues! This is a good thing. I'd rather push forward, and revert if necessary, then have many things sitting in review for long periods of time and never be merged (we actually have this problem right now, in some cases, with some of the open pull requests).

When we reach 1.0 + stability -- we can run this project a bit more traditionally, with more rigorous testing requirements + code review. The community deserves that, of course. But at the moment, moving quickly is paramount.

skevy:
One thing that we'd appreciate, however, is constructive PRs.

There's no need to get frustrated about a breaking change that wasn't released. These things happen.

As you mentioned in your commit comment -- you have not tested this change. So you don't actually know if something was broken by it or not. 

Instead of reverting as a knee-jerk reaction -- let's make sure we provide test cases, etc for why a revert should happen.

Admittedly -- I could have put more details in that commit as to what it does. I will follow up with some code comments + more changes to iOS animations that explain my reasoning (in a PR) and tag you, if you're curious.

nihgwu:
Moving fast shouldn't be the reason not committing via PR, OK?
You can still merge it as fast as your wish, but at lest someone can review it or test it. But now the only thing we can do is to admit it and use it!

On 18 May 2017, at 22:53, Adam Miskiewicz <notifications@github.com<mailto:notifications@github.com>> wrote:


One thing that we'd appreciate, however, is constructive PRs.

There's no need to get frustrated about a breaking change that wasn't released. These things happen.

As you mentioned in your commit comment -- you have not tested this change. So you don't actually know if something was broken by it or not.

Instead of reverting as a knee-jerk reaction -- let's make sure we provide test cases, etc for why a revert should happen.

Admittedly -- I could have put more details in that commit as to what it does. I will follow up with some code comments + more changes to iOS animations that explain my reasoning (in a PR) and tag you, if you're curious.

nihgwu:
I'm always using the master branch of this package in my app, this kind of change make me lose the tracking on what's happening on it

Do you still remember the issues you've introduced by your swift moving changes? Animated.timing with speed?

If you made the change via PR, it's really easy to be noticed and revised

On 18 May 2017, at 22:53, Adam Miskiewicz <notifications@github.com<mailto:notifications@github.com>> wrote:


One thing that we'd appreciate, however, is constructive PRs.

There's no need to get frustrated about a breaking change that wasn't released. These things happen.

As you mentioned in your commit comment -- you have not tested this change. So you don't actually know if something was broken by it or not.

Instead of reverting as a knee-jerk reaction -- let's make sure we provide test cases, etc for why a revert should happen.

Admittedly -- I could have put more details in that commit as to what it does. I will follow up with some code comments + more changes to iOS animations that explain my reasoning (in a PR) and tag you, if you're curious.

nihgwu:
You called this as a knee-jerk reaction, but I take it as a protest

On 18 May 2017, at 22:53, Adam Miskiewicz <notifications@github.com<mailto:notifications@github.com>> wrote:


One thing that we'd appreciate, however, is constructive PRs.

There's no need to get frustrated about a breaking change that wasn't released. These things happen.

As you mentioned in your commit comment -- you have not tested this change. So you don't actually know if something was broken by it or not.

Instead of reverting as a knee-jerk reaction -- let's make sure we provide test cases, etc for why a revert should happen.

Admittedly -- I could have put more details in that commit as to what it does. I will follow up with some code comments + more changes to iOS animations that explain my reasoning (in a PR) and tag you, if you're curious.

skevy:
Ok.

Again, I'm sorry if this affected you in a negative manner. Part of using the master branch of a repo in your app means that you should a) be locking to a specific commit and b) you should be watching the commits in the repo.

PRs or not -- we can make no guarantees at this point in the project about the stability of the master branch.

My point about constructive PRs still stands. I'm happy to chat about things like this in a constructive manner. PRs with bold text and all caps are not how I like to have conversations.

skevy:
I'm going to lock this thread.

If you'd like to chat about this further, please feel free to reach out to me on Twitter (@skevy) or Facebook messenger (@skevy on Facebook).

PR #1590, opened 2017-05-18
Add screen option to allow interactivity during transition

chromakode:
In some cases it is desirable for screens to be interactive sooner than the hardcoded `MIN_POSITION_OFFSET` 99% progress into the transition animation. This patch adds optional per-screen configuration of how soon in the transition the screen will become interactive.

In the case of Keybase Chat, when the back button is pressed on a conversation, we currently have the perception of early taps being dropped while the current conversation is sliding away. This change makes the conversation list feel more responsive to fast taps.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -235,6 +235,7 @@ export type NavigationStackScreenOptions = NavigationScreenOptions & {
   headerRight?: React.Element<*>,
   headerStyle?: Style,
   gesturesEnabled?: boolean,
+  transitionInteractivityThreshold?: number,
 };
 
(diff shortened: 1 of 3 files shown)


brentvatne:
cc @ericvicenti - curious what you think about this

brentvatne:
@chromakode - any chance you could rebase this? I think it's a nice addition

ericvicenti:
This looks good. But I wonder if we should fix this for everybody instead of adding yet another option. What value do you use in Keybase chat, @chromakode?

brentvatne:
closing this due to inactivity but if someone would like to pick it up and own it, please reopen and we'll review!

PR #1592, opened 2017-05-19
Fix queued transition being dropped if interrupted

chromakode:
During rapid updates to the card stack, `_queuedTransition` can be replaced before the animation has a chance to be performed. This can lead to a state in which navigation was performed but the card positions are not updated, leaving you stuck on the previous route. This patch makes `_queuedTransition` an array which executes in order, so that no queued transitions are dropped.

An alternate approach might be to apply the end state of the queued transition immediately and skip it. I didn't try this route because I was concerned this could cause abrupt jumps in transitions.

### DIFF
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -50,11 +50,11 @@ class Transitioner extends React.Component<*, Props, State> {
   _transitionProps: NavigationTransitionProps;
   _isMounted: boolean;
   _isTransitionRunning: boolean;
-  _queuedTransition: ?{
+  _queuedTransition: Array<{
     nextProps: Props,
     nextScenes: Array<NavigationScene>,
     indexHasChanged: boolean,
-  };
+  }>;
 
   props: Props;
   state: State;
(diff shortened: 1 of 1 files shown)


skevy:
Do you have a video of this happening? I'm not sure I've ever seen this behavior.

chromakode:
@skevy here's an example. Note that we have a shim to use the CardStack component with our own router state, which might be involved in the repro.

The first time in the gif, clicking the conversation list item opens a conversation. The second time it gets stuck, registering the click (and resulting nav) but the CardStack stays on the previous route forever. You have to get the timing just right, clicking while the animation is still transitioning.

stuck

ghuh:
This fixes https://github.com/react-community/react-navigation/issues/1721

ofairfoul:
@skevy I have also experienced this behaviour pretty often when navigating quickly often resulting in a blank screen or a transparent screen which is pushed over the current one, the only way to recover from is to swipe back from the left. I think this only happens with nested stack navigator within a tab navigator.

brentvatne:
this is no longer an issue on 1.0, there were various changes to transitioner (including the `isTransitioning` state) which resolved this. thank you for the PR, sorry nobody ever got around to reviewing it! I'm new around here but plan to handle PR reviews very quickly, just check the activity for the last month :)

Issue #1597, opened 2017-05-19
Warning: BackAndroid is deprecated.  Please use BackHandler instead.


Lizhooh:
Warning: BackAndroid is deprecated.  Please use BackHandler instead.

| software         | version
| ---------------- | -------
| react-navigation  |  1.00--beta.9
| react-native   |   0.44 
| node             |  7.10
| npm or yarn      | 4.12

kelset:
This is not strictly an error related to the react-navigation, you just need to change the physical android handler to:

```
import { BackHandler } from 'react-native';

  // code

  componentDidMount() {
    BackHandler.addEventListener('backPress');
  }

  // some more code

  componentWillUnmount() {
    BackHandler.removeEventListener('backPress');
  }

```

acomito:
@kelset would you put this in your top-level component? I'm not using BackHandler myself so it must be used somewhere in react-navigation? I just upgraded to 44 and see this now.

chrischen-melb:
same here, after I started using StackNavigator...
also not sure where shall I apply @kelset 's solution, tried put it in navigation related component, warning stays

kelset:
@acomito yes I've added those lines to the top-level component.

I am running the same React Native version (0.44) but I'm pointing react-navigation to `master`, maybe it's a fix not yet in a release?

acomito:
@kelset I'm also using 44 (but via expo 17) and I believe I'm pointing to master. But I am still seeing the error.

I feel like this has to do something with react-navigation and maybe `navigation.goBack()` or BackAndroid being used somewhere in the library?


```
{
  "dependencies": {
    "antd-mobile": "^1.1.3",
    "apollo-client": "^1.1.1",
    "expo": "^17.0.0",
    "graphql-tag": "^2.0.0",
    "lodash": "^4.17.4",
    "meteor-apollo-accounts": "^2.0.1",
    "moment": "^2.18.1",
    "react": "16.0.0-alpha.6",
    "react-apollo": "^1.2.0",
    "react-native": "https://github.com/expo/react-native/archive/sdk-17.0.0.tar.gz",
    "react-native-aws3": "0.0.8",
    "react-native-elements": "^0.11.2",
    "react-navigation": "git+https://github.com/react-community/react-navigation.git",
    "react-redux": "^5.0.4",
    "redux": "^3.6.0",
    "redux-devtools-extension": "^2.13.0",
    "redux-form": "^6.6.3",
    "redux-persist": "^4.6.0",
[...]

PR #1605, opened 2017-05-19
BackAndroid is deprecated.

abossard:
PR for #1597 
Checked it, seems to work fine.


skevy:
This was fixed in https://github.com/react-community/react-navigation/pull/1520, and I just merged that.

skevy:
Thanks though!

PR #1616, opened 2017-05-20
[WIP] Use Lerna

skevy:
This is attempt #2 in getting Lerna into react-navigation.

The reasons have been discussed before, but to recap:

1. We want to split out the "views" of react-navigation into "react-navigation-views". This will let us do that. We may want more packages in the future, as well. I really, really, really, want to avoid react-navigation from getting super monolithic. Setting the repo up this way is only a small part of that battle, of course, but if the infra is there, it makes it easier for us to make that call in the future.

2. We have examples + a website in this repo, and prior to this PR, we had a ton of dirty hacks in webpack configs, rn-cli.config.js files, etc. to get them both to work.

3. Simplifies and makes consistent scripts, linting, etc.

So this PR migrates us to Lerna, as well as a few other things:

- Moves both examples to "create-react-native-app", removes crazy configs.
- Moves react-navigation inside `packages/` directory
- Removes ejected create-react-app code, migrates to CRA 1.0.
- Reorganizes npm scripts so that they work appropriately with the new repo
- Adds Husky + lint-staged to lint (catching Prettier things) on pre-commit.

I want to address a concern that @satya164 and @grabbou have raised to me in the past:

> Recently at callstack we needed to use a new unpublished commit in jest and were unable to install it from GIT because it uses lerna. I think this will be a problem too, unless there's a solution I'm not aware of.

I think the way to resolve this is to fork the repo, make the change, publish `react-navigation` to `@skevy/react-navigation`, and then use babel-plugin-module-resolver:

```json
{
  "plugins": [
    ["module-resolver", {
      "alias": {
        "react-navigation": "@skevy/react-navigation"
      }
    }]
  ]
}
```

I know this isn't quite as easy as forking and installing from Git, but it's not _too_ bad. Hopefully, people won't have to do this very often (especially as we get more stable) and also, @satya164 and I discussed moving to semantic-release, so we could release more often. Even if we don't do that, we can commit to a weekly release schedule (while in beta), so that people get fixes quickly.


skevy:
cc @ericvicenti

ericvicenti:
Excellent, thank you for doing this!

Looks like tests are failing though.

Also, perhaps to deal with the 'unpublished commit' issue, perhaps we could configure our CI to upload nighly releases to npm `@next`

lerna-debug.log
```diff
@@ -0,0 +1,33 @@
+0 silly input [ 'eslint', '.' ]
```
satya164: Oops!

examples/ReduxExample/rn-cli.config.js
```diff
@@ -0,0 +1,11 @@
+const path = require('path');
+
+module.exports = {
+  getProvidesModuleNodeModules: () => [
+    // `${path.resolve(__dirname)}/node_modules/react-native`,
+    'react-native'
+  ],
+  getAliases: () => ({
+    'react-native': `${path.resolve(__dirname)}/node_modules/react-native`,
```
satya164: nit: Maybe `path.resolve(__dirname, '/node_modules/react-native')` looks nicer?

examples/ReduxExample/rn-cli.config.js
```diff
@@ -0,0 +1,11 @@
+const path = require('path');
+
+module.exports = {
+  getProvidesModuleNodeModules: () => [
+    // `${path.resolve(__dirname)}/node_modules/react-native`,
```
satya164: nit: Unused comment

examples/ReduxExample/README.md
```diff
@@ -1,5 +1,220 @@
-# Redux example
```
satya164: Probably we should get rid of this generated README and use a simpler README saying how to run the project,

satya164:
It'll be nice to group card stack related views in there own folder :)

satya164:
It's amazing considering how many hacks we had to do earlier :D

ericvicenti:
Also, this is a purely cosmetic complaint, but I'd find it easier to navigate the repo if all of the modules were located at the root as opposed to being stuffed inside `/packages`. Plus I think it may be appropriate for examples to be in the same namespace as the main modules. Depending on your definition of 'example', we may even consider the navigators to be example navigators, because you may want to define your own navigators if you don't intend to use the default views/routers.

skevy:
> Looks like tests are failing though.

@ericvicenti that's why it's "WIP" -- I'll finish it this week. :)

> Also, perhaps to deal with the 'unpublished commit' issue, perhaps we could configure our CI to upload nighly releases to npm @next

I think this is fine -- we could do a "nightly" dist-tag.

> Also, this is a purely cosmetic complaint, but I'd find it easier to navigate the repo if all of the modules were located at the root as opposed to being stuffed inside /packages

Haha you mentioned this to me in person when we were talking about create-react-workspace. I personally disagree -- I think it's more straightforward that the "packages" are in "packages", and thus you know that anything in that folder is published to npm. I also like that we wouldn't have to maintain a list of projects in lerna.json. It's also basically what every other mono-repo does. 

If you really feel super strongly though, I can change it.

skevy:
> The requirement for trying from GIT seems reasonable. Will it be possible to point the alias to the directory in the repo instead? It seems so.

Actually, this is a really good point.

```json
["module-resolver", {
  "alias": {
    "react-navigation": "./node_modules/react-navigation/packages/react-navigation"  
  }
}]
```

That should work fine.

ericvicenti:
>Haha you mentioned this to me in person when we were talking about create-react-workspace. I personally disagree -- I think it's more straightforward that the "packages" are in "packages", and thus you know that anything in that folder is published to npm. I also like that we wouldn't have to maintain a list of projects in lerna.json. It's also basically what every other mono-repo does.
>
>If you really feel super strongly though, I can change it.

So in your mind, should the website go into `examples/`? I guess that makes sense. But who's to say that apps can't or shouldn't be published to npm? 😉 

You have a good point about the `lerna.json` but it sounds like we could potentially fix that in lerna itself.

skevy:
@ericvicenti I don't think the website should go into examples, because even though it's kind of an example, it's also the website :p.

Basically, I'm proposing we leave it alone right now (aka, as I have it in this PR), and then when Yarn workspaces is released (which we would switch to), we can re-eval.

brentvatne:
😢

PR #428, opened 2017-02-21
Added complex example

si-harps:
A more complex example of redux integration with nested StackNavigator / TabNavigator structure. Creating pull request as agreed with @grabbou here https://github.com/react-community/react-navigation/issues/179#issuecomment-278584750.

PR #1625, opened 2017-05-21
Complex example

si-harps:
Follow up from #428 - reopening with updates based on previous issues


examples/ComplexExample/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "private": true,
+  "scripts": {
+    "start": "node node_modules/react-native/local-cli/cli.js start",
+    "test": "jest"
+  },
+  "dependencies": {
+    "react": "15.4.2",
+    "react-native": "0.40.0",
+    "react-native-vector-icons": "^4.0.0",
+    "react-navigation": "git+https://github.com/react-community/react-navigation.git",
```
matthamil: Is there any way the `ReactNavigation` dependency could be version locked? This is a great example of Redux integration with React Navigation; however, with the breaking changes that have happened with the API recently, I worry that this could cause issues in the future if more changes were to occur.

si-harps:
HI @matthamil - thanks for the suggestion. I'm wondering whether it might be better to remove the dependency from the example package altogether so that it can be developed alongside react-navigation? I'm happy to manage any breaking changes as and when. Please see the latest commit with this change.

matthamil:
@si-harps I think that this would be a better idea than filling the core repo with examples. I think the lack of real-world examples is a big problem with the current state of the documentation.

jnbt:
@matthamil I also like the idea of an example repository and I'm more than willing to contribute there. I think every attempt to cleanup the issues in the core repo could help.

matthamil:
@grabbou @satya164 @skevy Thoughts on moving the examples to a separate repo?

matthamil:
Another idea would be to put examples like this into a *RealWorld* example repo like https://github.com/gothinkster/realworld

satya164:
>  Thoughts on moving the examples to a separate repo?

We don't want to move examples to separate repo. It makes it harder to update and test them.

si-harps:
@matthamil whilst the idea of a separate repo sounds interesting, my suggestion was to keep the examples where they are and use the core react-navigation dependency in the example...

```
"devDependencies": {
    "babel-jest": "18.0.0",
    "babel-preset-react-native": "1.9.1",
    "jest": "18.1.0",
    "react-test-renderer": "15.4.2",
    "react-navigation": "file:../.."
  }
```

This will make it easier to develop the example alongside the core project without having to version lock the dependency. What do you think?

skevy:
Sorry this has sat here for a bit.

>>> We don't want to move examples to separate repo. It makes it harder to update and test them.

I spoke to @satya164 about this, and I _do_ think we should make a new repo with examples.

The only example I want in this repo is the NavigationPlayground, which later on we can use for E2E/instrumentation tests. 

I created an org: https://github.com/react-navigation

I don't really have time to review this whole example now @si-harps, but I really appreciate you making it.

For now, please throw it on your own Github, and I'm happy to link to it in the docs/README (could you send a PR for that)?

Then hopefully, in the near future, we can take the ReduxExample out of this repo, throw it in the react-nav org, and then transfer your Complex example to that org as well.

Thanks again, and sorry for taking so long to make this decision.

Issue #1647, opened 2017-05-23
Initial route params not passed to tab navigator


zash80:
Having a StackNavigator with 'initialRouteName' and 'initalRouteParams' where the initial route routes to a TabNavigator with tabs relying on the initial route parameter, this initial route parameter is not passed down to the tabs.

If the StackNavigator is re-configured to pass parameter via navigate action to the screen containing the TabNavigator all is working fine, means parameter is passed down to the tabs.

### Current Behavior
The parameters defined in StackNavigator 'initalRouteParams' are not passed down to a nested TabNavigator.

### Expected Behavior
The parameters defined in StackNavigator 'initalRouteParams' should be passed down to a nested TabNavigator.

### Code

```javascript
const SampleStack = StackNavigator({
  Screen1: {
    screen: SimpleScreen
  },
  Screen2: {
    screen: TabScreen
  }
}, {
  initialRouteName: 'Screen2',
  initialRouteParams: { name: 'value' }
});
```
```javascript
const TabScreen = TabNavigator({
  Tab1: {
    screen: SimpleTab
  },
  Tab2: {
    screen: TabRelyingOnParameters
  }
}, {
  tabBarPosition: 'bottom',
  tabBarComponent: MyTabBar
});
```
```javascript
TabRelyingOnParameters.navigationOptions =
  ({ navigation, navigation: { state: { params: { name } } } }) => ({
    header: (
      <Header>
        <Left>
          <Button onPress={() => navigation.goBack(null)} transparent>
            <Icon name='arrow-back' />
          </Button>
        </Left>
        <Body>
          <Title>{name}</Title>
        </Body>
        <Right />
      </Header>
    )
  }
);
```
### Environment

| software         | version
| ---------------- | -------
| react-navigation |  1.0.0-beta.10
| react-native     |  0.44.0

MrHubble:
What about adding the `initialRouteParams` as a config option to the TabRouter as per the StackRouter?

StackRouter:
![screen shot 2017-06-13 at 2 13 26 pm](https://user-images.githubusercontent.com/2251631/27065760-9db56e68-5042-11e7-82da-2435e53a48b9.png)

TabRouter:
![screen shot 2017-06-13 at 2 14 36 pm](https://user-images.githubusercontent.com/2251631/27065777-add19c36-5042-11e7-97f9-a27417218f80.png)

and then we could have `initialRouteParams` specific to TabNavigator:
````
const MyApp = TabNavigator({
  Home: {
[...]

PR #1648, opened 2017-05-23
#1647: pass initial state params down to routes on tab navigators ini…

zash80:
#1647: pass initial state params down to routes on tab navigators initial navigation action


skevy:
See my comment on #1647

brentvatne:
agree with @skevy that this should be named `initialRouteParams` rather than `params`, can you update and rebase? thanks!

PR #1676, opened 2017-05-24
Expose CardStackTransitioner as export (closes #1326)

### LINKED ISSUE #1326: Expose CardStackTransitioner as export
I prefer to use the transitioners and other components directly over using the helper functions since I want more fine tuned control over them.

Previously, I was using the CardStack component by itself, but within the last couple of betas a lot of its functionality has been moved to CardStackTransitioner and doesn't have much useful behavior by default. Currently the only normal way to get an instance of the component is by creating an entire StackNavigator.

I keep my router declarations separate, so the main issue I have against just using a StackNavigator instance is that I can't use an existing router instance.

rockmacaca:
(from #1326)
I prefer to use the transitioners and other components directly over using the helper functions since I want more fine tuned control over them.

Previously, I was using the CardStack component by itself, but within the last few betas a lot of its functionality has been moved to CardStackTransitioner and it now doesn't have much useful behavior by default. Currently the only normal way to get an instance of the component is by creating an entire StackNavigator.

I keep my router declarations separate, so the main issue I have against just using a StackNavigator instance is that I can't use an existing router instance.

### DIFF
--- a/src/react-navigation.js
+++ b/src/react-navigation.js
@@ -45,6 +45,9 @@ module.exports = {
   get Transitioner() {
     return require('./views/Transitioner').default;
   },
+  get CardStackTransitioner() {
+    return require('./views/CardStackTransitioner').default;
+  },
   get CardStack() {
     return require('./views/CardStack').default;
   },


skevy:
Sure! Thanks!

sagark1510:
How to use it? Can you add  an example?

rockmacaca:
@sagark1510 Basically exactly the same as you would use `StackNavigator()`. I'd recommend looking at the function's source to see what it is doing. That second `StackNavigatorConfig` argument contains all the props passed to the `CardStackTransitioner`. You'll need an existing router instance that you can set as a static member of your component (or pass to `createNavigator `).

The only advantages to using it this way is being able to use an existing router instance and greater freedom in how props are assigned.

PR #1687, opened 2017-05-25
Correct sizing of header back button image on iOS

wschurman:
**Summary**

This fixes the slight blurring of the back icon in stack navigator on iOS due to a mismatch between absolute pixel sizes of the icon (and the `@2x`, `@3x`, etc) and the stylesheet width and height.

The alternative here would be to regenerate all the iOS icons at the correct resolution (20x12 @ 1x), but I believe @grabbou has the master file (7a20389e04a41de65e145a456c7dcacb787aad24).

**Test plan**

Before:
simulator screen shot may 25 2017 10 30 37 am

After:
simulator screen shot may 25 2017 10 30 48 am

### DIFF
--- a/src/views/HeaderBackButton.js
+++ b/src/views/HeaderBackButton.js
@@ -118,8 +118,8 @@ const styles = StyleSheet.create({
   },
   icon: Platform.OS === 'ios'
     ? {
-        height: 20,
-        width: 12,
+        height: 21,
+        width: 13,
         marginLeft: 10,
         marginRight: 22,
         marginVertical: 12,


wschurman:
For reference, the `@1x` size of the asset is 21x13. https://github.com/react-community/react-navigation/blob/master/src/views/assets/back-icon%401x.ios.png

Noor0:
is there a way to set the height of the header of a stack navigator?
StackNavigator({...routes},{
   headerStyle:{
     height: 20
   }
});

PR #1694, opened 2017-05-26
Remove non-configurable properties from contentOptions in DrawerNavigator doc

Agontuk:
As far as I understand, `items`, `activeItemKey ` & `onItemPress` can not be configured from DrawerNavigatorConfig. They are useful when defining custom `contentComponent`.

We can add a separate section about which props are available for custom `contentComponent`. That way it'll be less confusing I think.

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -130,13 +130,10 @@ const styles = StyleSheet.create({
 
 ### `contentOptions` for `DrawerItems`
 
-- `items` - the array of routes, can be modified or overridden
-- `activeItemKey` - key identifying the active route
 - `activeTintColor` - label and icon color of the active label
 - `activeBackgroundColor` - background color of the active label
 - `inactiveTintColor` - label and icon color of the inactive label
 - `inactiveBackgroundColor` - background color of the inactive label
-- `onItemPress(route)` - function to be invoked when an item is pressed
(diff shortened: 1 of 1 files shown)


ajostergaard:
I changed my `src/views/Drawer/DrawerSidebar.js` to allow `items` to be a function which gets `screenProps` but am constantly flummoxed by how to push a sensible pull request so gave up. I guess I'm saying that I'd prefer this feature to be fixed rather than removed from docs.

```
--- a/src/views/Drawer/DrawerSidebar.js
+++ b/src/views/Drawer/DrawerSidebar.js
@@ -88,7 +88,7 @@ class DrawerSidebar extends PureComponent<void, Props, void> {
         <ContentComponent
           {...this.props.contentOptions}
-          items={state.routes}
+          items={this.props.contentOptions.items && this.props.contentOptions.items(this.props.screenProps) || state.routes}
```

brentvatne:
@ajostergaard - that makes sense, it's unfortunate you had to struggle to find a way around that. overall the drawer component is not well loved and we need to revisit it -- we'll likely pull it out into an external repo where people who use it heavily can iterate on it independently of react-navigation. I'm going to close this PR for now as we have moved over to a new website repository at https://github.com/react-navigation/react-navigation.github.io but I appreciate you opening it @Agontuk -- if you could submit a new PR that indicates that these are passed into the content view it would be helpful!

PR #1698, opened 2017-05-27
Removed dependency on fbjs

gabrielbull:
fbjs is a big part of the dependencies of this project, as can be seen here:

```diff
├─┬ react-navigation@1.0.0-beta.10 invalid
│ ├── clamp@1.0.1 extraneous
+│ ├─┬ fbjs@0.8.12
+│ │ ├── core-js@1.2.7
+│ │ ├─┬ isomorphic-fetch@2.2.1
+│ │ │ ├── node-fetch@1.7.0 deduped
+│ │ │ └── whatwg-fetch@1.1.1 deduped
+│ │ ├── loose-envify@1.3.1 deduped
+│ │ ├── object-assign@4.1.1 deduped
+│ │ ├── promise@7.1.1 deduped
+│ │ ├── setimmediate@1.0.5
+│ │ └── ua-parser-js@0.7.12
│ ├── hoist-non-react-statics@1.2.0
│ ├── path-to-regexp@1.7.0 extraneous
│ ├── prop-types@15.5.10 deduped
│ ├── react-native-drawer-layout-polyfill@1.3.1 extraneous
│ └── react-native-tab-view@0.0.65 extraneous
```

It is required for only two small functions that can be easily incorporated into the source code of this project.

It also require `isomorphic-fetch` which, if inadvertently imported in a react-native project, will break the project.

img_0011

fbjs is also not recommended for use outside of Facebook's internal projects, as mentioned in it's README:

 > __Note:__ If you are consuming the code here and you are not also a Facebook project, be prepared for a bad time. APIs may appear or disappear and we may not follow semver strictly, though we will do our best to. This library is being published with our use cases in mind and is not necessarily meant to be consumed by the broader public. In order for us to move fast and ship projects like React and Relay, we've made the decision to not support everybody. We probably won't take your feature requests unless they align with our needs. There will be overlap in functionality here and in other open source projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
   },
   "dependencies": {
     "clamp": "^1.0.1",
-    "fbjs": "^0.8.12",
     "hoist-non-react-statics": "^1.2.0",
     "path-to-regexp": "^1.7.0",
     "prop-types": "^15.5.10",
--- a/src/StateUtils.js
+++ b/src/StateUtils.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import invariant from 'fbjs/lib/invariant';
+import invariant from './utils/invariant';
 
 import type { NavigationRoute, NavigationState } from './TypeDefinition';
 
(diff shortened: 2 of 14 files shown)


jnbt:
@gabrielbull I'm agreeing 101%

PR #1713, opened 2017-05-29
Upgrade React Native Tab View to v0.0.66

satya164:
This adds 2 changes:

https://github.com/react-native-community/react-native-tab-view/commit/13243a2c6e048c73b089de63da58b7320fdb8bf0
https://github.com/react-native-community/react-native-tab-view/commit/27429c9fcc7b50598732c38062b2ead270f93618

I'm so stupid that this didn't occur to me earlier. This should now make it possible to absolutely position the tabs properly.

cc @skevy

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "path-to-regexp": "^1.7.0",
     "prop-types": "^15.5.10",
     "react-native-drawer-layout-polyfill": "^1.3.0",
-    "react-native-tab-view": "^0.0.65"
+    "react-native-tab-view": "^0.0.66"
   },
   "jest": {
     "notify": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -4074,9 +4074,9 @@ react-native-drawer-layout@1.3.0:
   dependencies:
     react-native-dismiss-keyboard "1.0.0"
 
-react-native-tab-view@^0.0.65:
-  version "0.0.65"
-  resolved "https://registry.yarnpkg.com/react-native-tab-view/-/react-native-tab-view-0.0.65.tgz#b685ea3081ff7c96486cd997361026c407302c59"
+react-native-tab-view@^0.0.66:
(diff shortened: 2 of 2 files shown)


nihgwu:
https://github.com/facebook/react-native/pull/13105 does no help on this?

skevy:
cc @satya164 ^^

satya164:
@nihgwu I haven't tested it, but looks like it might fix the bugs with zIndex. Though it's not in stable version yet, so the changes in react-native-tab-view help current stable versions without any downside.

michaelduminy:
When I use this my absolutely positioned top tabs are visible again, but not clickable.

satya164:
@michaelduminy you probably need to specify `zIndex: 1` manually for the tab bar since that's the only thing which was removed.

michaelduminy:
@satya164 thanks for responding :)

My style passed to the tabBar has always had `zIndex` set, it only stopped working in the latest version of this library. Now when I set any `zIndex` value (on an absolutely positioned top tab bar) I get this warning:
```
The specified child already has a parent. You must call removeView() on the child's parent first.
```

On RN 0.44.2 this works with beta.9 but not in beta.11.

I played around with the rendering of things and found that if I remove all contents of the tab scene being rendered, returning null from `render()` with `zIndex` set on the tab bar, then the tab bar renders and is clickable. With hotloading I then replace the null with the original contents and that makes everything work as expected. I think this means there's probably an order of rendering that is the problem. Not sure how to solve this so I've reverted to beta.9 for now.

satya164:
Looks like a bug in RN. You should file a bug report in the RN repo.

michaelduminy:
@satya164 everything is fixed if this change is reverted:
https://github.com/react-native-community/react-native-tab-view/commit/48741e7ea77d22e0b010a94d7985d186cf565539

Issue #1626, opened 2017-05-21
Update react-native-drawer-layout-polyfill to 1.3.1


Jacse:
`react-native-drawer-layout-polyfill@1.3.1` fixes a bug which failed to clear interactions meaning that functions invoked through `InteractionManager.runAfterInteractions` were never called.

For this reason it is currently not possible to do the "render inexpensive component until navigation is done"-trick in a `DrawerNavigator`.

See [react-native-drawer-layout#16](https://github.com/react-native-community/react-native-drawer-layout/issues/16)

Jacse:
Cc @skeva @satya164

satya164:
Can you send a PR for that?

Jacse:
Moved to #1716

PR #1716, opened 2017-05-29
Update drawer layout polyfill to 1.3.1

Jacse:
`react-native-drawer-layout-polyfill@1.3.1` fixes a bug which failed to clear interactions meaning that functions invoked through `InteractionManager.runAfterInteractions` were never called.

For this reason it is currently not possible to do the "render inexpensive component until navigation is done"-trick in a `DrawerNavigator`.

See [react-native-drawer-layout#16](https://github.com/react-native-community/react-native-drawer-layout/issues/16)

Moved from issue #1626

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "hoist-non-react-statics": "^1.2.0",
     "path-to-regexp": "^1.7.0",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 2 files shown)


satya164:
This also needs to update the `yarn.lock`, so you'll need to run `yarn add react-native-drawer-layout-polyfill` locally. Sorry :)

PR #1742, opened 2017-05-31
Fix documention header styles

ThePaulMcBride:
Any heading in the documentation that were also links would be set to `visibility: hidden` in css.

An example of this is here: https://reactnavigation.org/docs/views/

This is a quick fix which will display these titles.

There is a larger underlying issue though. When a title is also a link, it isn't assigned an id. This is caused by functionality which generates the id string. For example, the code in this section returns an object (react-node) if the title is a link

https://github.com/react-community/react-navigation/blob/master/website/src/MDPage.js#L57-L65

I'm happy to take a look at this if you let me know how you would like it dealt with. In the meantime, this pull request will display any hidden titles.


skevy:
Thanks!

PR #1753, opened 2017-06-01
Fix broken redux example, catch errors in reducer

braco:


docs/guides/Redux-Integration.md
```diff
@@ -12,15 +12,23 @@ const AppNavigator = StackNavigator(AppRouteConfigs);
...
 
 const navReducer = (state = initialState, action) => {
-  const nextState = AppNavigator.router.getStateForAction(action, state);
+
+  let nextState;
+
+  try {
+    nextState = AppNavigator.router.getStateForAction(action, state);
+  }
+  catch (e) {
```
AlanFoster: A reducer shouldn't throw exceptions; What is the error here?

matthamil: I am also curious about the motivation behind this change.

PR #1756, opened 2017-06-01
export action constants and creators

braco:
It would be helpful to export action constants and creators from NavigationActions.js. One such example is a singleton service that seems to [keep showing up in issues](https://github.com/react-community/react-navigation/issues/1439#issuecomment-303661539)

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -11,6 +11,8 @@ const RESET = 'Navigation/RESET';
 const SET_PARAMS = 'Navigation/SET_PARAMS';
 const URI = 'Navigation/URI';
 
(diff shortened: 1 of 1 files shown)


src/NavigationActions.js
```diff
@@ -23,6 +25,8 @@ const reset = createAction(RESET);
 const setParams = createAction(SET_PARAMS);
 const uri = createAction(URI);
 
+export const creators = { back, init, navigate, reset, setParams, uri };
```
matthamil: Would this be better to be named as `actionCreators`? Your comment in the `export default` clarifies that these are action creators. Curious on your thoughts just so we remove any ambiguity in the name.

brentvatne:
not clear to me why we need to export these separately, you can already use them from the default export

PR #1764, opened 2017-06-03
Fix issue #618 - TabBarBottom should hide when Keyboard is activated

mmalfertheiner:
This small fix makes the TabBarBottom disappear when the Keyboard is shown.
I explained the work around, which I currently use in my production app, in issue #618 and now I made a pull request, because it looks like more people have this problem.


danieloprado:
Any updates?

kelset:
Hey @mmalfertheiner can you update this PR to remove the conflicts? I think it's a good feature to add.

SupriyaKalghatgi:
@mmalfertheiner Any updates on this?

I need this fix in my project

vonovak:
thank you @mmalfertheiner

richeterre:
Hm, for me this doesn't fix the issue when using `TabBarBottom` on Android – instead, it now breaks on iOS, too 😕

vonovak:
@richeterre can put together a minimal repro that shows your problem? Thanks.

richeterre:
@vonovak, I tried (https://snack.expo.io/BJaxu9hR-) but couldn't get Expo Snack to install the latest `master` commit. Any idea how to do that?

Anyway, in my app I have a vanilla `GiftedChat` (`react-native-gifted-chat`) component rendered inside a `TabNavigator`'s tab. When tapping the chat input, the appearing keyboard should cover everything below the chat input field. This works on iOS with `beta.19`, but on Android the tab bar is showing between the keyboard and input field.

With `master`, it now breaks on both platforms: The appearing keyboard pushes up the chat, but once its animation ends, the chat content jumps down all the way behind the keyboard.

vonovak:
@richeterre does the following (ie. wrapping with KeyboardAvoidingView) work for your app on beta 19? https://snack.expo.io/B1udZbp0b

richeterre:
@vonovak, wrapping the chat with a `KeyboardAvoidingView` (no matter its `behavior`) doesn't fix my layout on Android with beta 19, and in fact slightly breaks it on iOS as well (chat input stays behind keyboard). I'm using `android:windowSoftInputMode="adjustResize"`.

Will give it a try with the latest `master` as well!

404sand808s:
I have the exact same issue as @richeterre and I believe it's due to the TabBar disappearing right as the keyboard finishes animating upwards. The TextInput (which is wrapped in a KeyboardAvoidingView) which is showing nicely on the upwards animation then snaps back down right beneath the keyboard, and playing with keyboardVerticalOffset (which is set to 64) results in bugs one way or another if changed away from 64. I've attached a gif to show the issue as well.

ezgif-3-c831fd5401

vonovak:
@404sand808s the PR uses `keyboardDidShow` event (see the code) could you try replacing that with `keyboardWillShow` and tell us if that fixes your issue?

vonovak:
@404sand808s @richeterre can you please provide reproductions of your issue with beta.19? (Or provide a PR that fixes your issues?) I'm going to revert this PR because I came to the conclusion that using `KeyboardAvoidingView` should handle the potential problems including the ones you have, but without a repro we won't be able to debug those.

richeterre:
It's Launch Week for me, but I'll try to put something together as soon as these crazy days are over 😄  What I can say is that beta.19 only breaks on Android when used with `GiftedChat` – on iOS the tab bar is never an issue for me. Glad to hear this is reverted for now, I felt it introduced more problems than it fixed!

bilalsyed001:
beta.19. Im using android:windowSoftInputMode="adjustResize" this PR fixes issue for Android & iOS. KeyboardAvoidingView doesnt work in my case

PR #1786, opened 2017-06-06
Improve iOS slide transitions

stefanroex:
It that the default transitions didn't match the native iOS transitions. I've made a few tweaks to make the transitions more similar to native transitions:

- The shadow is a more subtile.
- The underlying card moves 30% of the screen width instead of 10 pixels.
- The underlying card gets darker instead of lighter when transitioning to the new card.

I've attached some gifs to illustrate the differences. Sadly, it's barely noticeable in the gifs, yet it makes a big difference when you try it on an actual device.

### Before:
before

### After:
after


nihgwu:
add `containerStyle` is not a good idea, you can set it in `cardStyle`

stefanroex:
@nihgwu You won't get the same results when you're only changing the `cardStyle`. 

All `Card` components are rendered within a `CardStack`. When we transition between cards the opacity of individual `Card` components is animated as is defined in `CardStackStyleInterpolator.js`. The underlying container, the `CardStack` component, is visible when a card is opaque. It makes sense have control over the styling of visible elements. It will be impossible to mimic the iOS transitions without altering this.

Changing the `cardStyle` only affects individual cards. This doesn't have any effect on the underlying content when the transition changes the opacity.

I'm not sure if the `TransitionConfig` should have a `containerStyle` property. On one hand it's styling, not a transition. On the other hand the container will only be visible when you change the opacity of a card, which I only see happening during a transition. It's in some way _part_ of the transition, e.g. the Android transition doesn't require a dark background.

I'm open for suggestions!

nihgwu:
@stefanroex sorry for my quick reply, I've got you motivation now, but I think you can move the `containerStyle` to `StackNavigatorConfig`

stefanroex:
@nihgwu I thought about moving it to `StackNavigatorConfig` as well, but the background is dependent on which transition is used. The containerStyle is different for iOS and Android transitions.

nihgwu:
@stefanroex but I think `containerStyle` is still useful for other use-case, for you scenario, I think you can simply pass the `backgroundColor` in `screenInterpolator`

stefanroex:
@nihgwu Hmmm. I'm not sure, but I think `screenInterpolator` only sets the style of a single card, right? How would that work?

mickaelw:
@stefanroex Hello, I saw yesterday that your pull request was approved :) I update my version react navigation and I saw your code but for iOS I don't see the new animation.. I have to do something special?

stefanroex:
@mickaelw The PR isn't reviewed or merged yet, so updating react-navigation won't have any effect. You could reference my PR repository from you're package.json via `"react-navigation": "git+ssh://git@github.com/stefanroex/react-navigation.git#master"`. This won't give you any future updates though.

skevy:
This is great.

Thanks.

koenpunt:
Nice! Noticed the difference some time ago, but didn't have the time to create a PR. One thing that still can be improved is the shadow, which should fade out the more the screen moves out of the screen. Although I think that's more work to implement, since the shadow can't be natively animated.

skevy:
@koenpunt you're right about that! Maybe we'll get that fixed some time in the future :)

PR #1811, opened 2017-06-09
allow empty path to be matched

koenpunt:
previously when opening a deeplink without a path, it would result in an array of 2 empty strings, which is are falsy values, so the full url was passed on.

by checking if the value is actually `undefined` instead of falsy we have no more false positives


koenpunt:
Any word on this? ✨

skevy:
Thanks!

PR #1817, opened 2017-06-09
DrawerNav: don't push screen to stack if it's already focused

Agontuk:

Right now if same screen is opened from drawer multiple times, it'll push the screen to the stack every time. So navigating back will render the same screen over & over. This commit fixes that.

The issue was explained in a comment in here [https://github.com/react-community/react-navigation/issues/954#issuecomment-302872661](https://github.com/react-community/react-navigation/issues/954#issuecomment-302872661)

### DIFF
--- a/src/views/Drawer/DrawerSidebar.js
+++ b/src/views/Drawer/DrawerSidebar.js
@@ -75,9 +75,11 @@ class DrawerSidebar extends PureComponent<void, Props, void> {
     return null;
   };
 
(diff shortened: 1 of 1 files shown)


aaronbuchanan:
thanks for working on this!

nonameolsson:
Wow, thanks! I have been struggling with this problem!

nonameolsson:
What is it that has to be done before merging this one?

skevy:
Thanks!

PR #1824, opened 2017-06-11
fix movedDistance

nihgwu:
fix the wrong movedDistance value, sorry for my foolish mistake made in #1500 , ping @ericvicenti


ericvicenti:
This make sense to me, but can you please document: What exact behavior does this fix, and how did you test the fix?

cc @skevy who can probably land this.

nihgwu:
that's a minor fix that the `movedDistance` should be `dx` not `moveX`, `moveX` is the gesture release position

PR #1827, opened 2017-06-11
add doc for transitionConfig

### LINKED ISSUE #1490: Provide an example of using a custon transitionConfig in StackNavigator
Currently, docs say the following:
    `transitionConfig` - Function to return an object that overrides default screen transitions

However, no example is provided what that function should look like.

nihgwu:
close #1490


matthamil:
I think a cleaned up version of timrc's answer in https://github.com/react-community/react-navigation/issues/1490#issuecomment-304801223 on the same issue would be a great addition to this in the documentation.

nihgwu:
It is a demo for all the `StackNavigator` options, and I think it's enough comparing with the other demos in the docs

matthamil:
This example will work for now. I think it would be great to eventually have more examples in the documentation. Thank you for taking the time to submit this PR!

spencercarli:
Will you move this example down into the "Examples" section at the end of this file? Below the existing content in "Examples". A good title would be

"Modal StackNavigator with Custom Screen Transitions"

Thanks!

PR #1841, opened 2017-06-12
pass screenProps to TabBarComponent

koenpunt:
To support for example filtering the tabs (like @satya164 suggested here: https://github.com/react-community/react-navigation/issues/717#issuecomment-287392565)  based on a property passed to a parent navigator, I need to have to access to the screenProps in the tab bar.


satya164:
Thank you

Issue #1436, opened 2017-05-10
Can't reset from inside of nested navigato


msdark:
### Current Behavior
Current definition of my navigators
```javascript
const AppNavigator = StackNavigator({
  Home: { screen: SigninScreen },
  Tabs: { screen: TabNavigator }, // A Tab Navigator
});

const TabScreen = TabNavigator({
  Saldos: { screen: SaldosScreen },
  Operaciones: { screen: OperacionesNavigator }, // a Stack Navigator
  Retenciones: { screen: RetencionesScreen },
  Ejecutivo: { screen: EjecutivoScreen },
});

const OpNavigator = StackNavigator(
  {
    Operaciones: { screen: OperacionesScreen },
    OperacionesDetails: { screen: OperacionesDetails },
  })
```
The app start with the `Home` screen, after user enter the login details the app redirects to the `TabNavigator` and the user is only able to navigate through that set of screens. (back button is disabled).

I add a logout button
```javascript 
const logout = NavigationActions.reset({
  index: 0,
  actions: [NavigationActions.navigate({ routeName: 'Home' })],
});

const LogoutButton = ({ dispatch }) => (
  <TouchableOpacity
    onPress={() => dispatch(logout)}
    style={{
      flexDirection: 'row',
      justifyContent: 'flex-start',
      alignItems: 'flex-start',
    }}
  >
    <Icon
      name="log-out"
      style={{ color: 'white', marginRight: 10, marginTop: 5 }}
    />
  </TouchableOpacity>
);

LogoutButton.propTypes = {
  dispatch: PropTypes.func.isRequired,
};

export default connect()(LogoutButton);
```

But onPress the button this errors raise
> There is no route defined for key Home.
> Must be one of:
> 'Operaciones, OperacionesDetails'

That errors shows only tow possible routes. That routes belongs to the `OpNavigator` but the current active screen is `Saldos` that belongs to `TabNavigator`.

What is the correct way to setup the navigators and perform the stack reset?.

The desired stack is.
|- Home (always the first screen to be loaded)
|- LoadingScreen (used in the transition from Home to Tabs)
|-- Tabs
|---- Operaciones (is a tab that holds two other screens as a Stack Navigator)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  git+https://github.com/react-community/react-navigation.git
| react-native     |  https://github.com/exponent/react-native/archive/sdk-16.0.0.tar.gz
| node             | v7.7.2
[...]

Issue #335, opened 2017-02-14
How to navigate from nested StackNavigator to parent's router screen?


goncalomarques:
I have several nested navigators with the following structure:

```
StackNavigator
    Home
    Login
    Authenticated -> CustomTabNavigator
        ScreenOne -> StackNavigator
            ScreenOneDetails
            ScreenOneSettings
        ScreenTwo
```

How can I navigate from `ScreenOneDetails` to `Home`?

Already tried to dispatch it as an action:
```
const resetAction = NavigationActions.reset({
    index: 0,
    actions: [
        NavigationActions.navigate({ routeName: 'Home' })
    ]
});
this.props.navigation.dispatch(resetAction);
this.props.navigation.goBack(null);
```

But all I get is an error - `There is no route defined for key splash. Must be one of: 'ScreenOneDetails','ScreenOneSettings'`

I know that this makes sense, as the `navigation` prop that is passed to `ScreenOneDetails` is the one created by the `ScreenOne` StackNavigator. However, is there any way to access the `navigation` prop created by parent navigators? And how would I go the other way around, e.g: from `Login` to `ScreenOneSettings`?

nico1510:
First off, I'm having the same questions but here's an idea what you could try:
> And how would I go the other way around, e.g: from `Login` to `ScreenOneSettings`?

The [docs ](https://github.com/react-community/react-navigation/blob/master/docs/guides/Screen-Navigation-Prop.md#navigate---link-to-other-screens) include an example I think. Specifically: 
> - `action` - (advanced) The sub-action to run in the child router, if the screen is a navigator.

So you would call something like:

`
navigate('ScreenOne', {}, NavigationActions.navigate({
          'ScreenOneSettings'
        }));
`
As to your first problem:

> However, is there any way to access the `navigation` prop created by parent navigators?
have you tried passing down the navigation prop from the root `StackNavigator` as a prop down to your child `StackNavigator` ? Technically I think this is how is should work by reading [this section of the docs](https://github.com/react-community/react-navigation/blob/master/docs/guides/Custom-Navigators.md) but in practice I got some errors when trying this.

goncalomarques:
Thanks @nico1510 

I had already tried that before creating this issue, but today I managed to make it work. I still have some issues related to the CustomTabNavigator navigationOptions, as I'm passing a TabsNavigator and want to keep the top navbar dynamically linked to the tabs pages

zoontek:
[...]

PR #1888, opened 2017-06-15
Fix wrong getStateForAction for reset action

wsxyeah:
I ran into the issue which is discussed in #335, #1436.

The `reset ` doesn't work in a nested stack navigator. I dig into the code, and found this in `StackRouter`:

`if (action.type !== NavigationActions.RESET || action.key !== null) `

I found the condition would be `true` if the `action.key` is `undefined`, and according to the documentation, we don't pass `key` to the `reset` action, which means `reset` action will always run into this branch, but there is another one below: 
`if (action.type === NavigationActions.RESET) { \\ handleSomething() }`
I was thinking this might be the right one.

I haven't read the code thorough, I might be wrong but it actually solve the `reset` issue in my project.

So I hope this PR could solve this issue permanently, and please test this carefully, I don't want to affect other functionalities since I'm not sure about this.


src/routers/StackRouter.js
```diff
@@ -139,7 +139,7 @@ export default (
       }
 
       // Check if a child scene wants to handle the action as long as it is not a reset to the root stack
-      if (action.type !== NavigationActions.RESET || action.key !== null) {
+      if (action.type !== NavigationActions.RESET || !!action.key) {
```
matthamil: I think it would be best to be explicit here and not use `!!` because this will coerce values like `0` into `false` where `0` might be a valid key in some instances (either in the current version or in future versions).

skevy: Agreed with this -- let's explicitly check for `undefined` in addition to `null`.

bdbergeron:
Very curious about this fix being merged. It seems to solve the nested navigator reset issue for me if I use `action.key != null`. That will check against `undefined` and `null` values.

ericvicenti:
This has already been fixed by another commit

PR #1913, opened 2017-06-17
Reduce re-renders on header back button

AlanFoster:
First PR related to [Improve performance for React Navigation](https://github.com/react-community/react-navigation/issues/1466)

Although the HeaderBackButton is a `PureComponent` - we are binding a new function on each re-render which still causes a rerender of HeaderBackButton


satya164:
Thank you for the PR

PR #1922, opened 2017-06-18
Use animatable text component for the header title

fdnhkj:
# Motivation
This PR replaces `Text` component by its animatable version `Animated.Text`.
Having an animatable component for the header title allows features like the change of header title opacity based on the scroll animation.

auto-hiding-header

### DIFF
--- a/src/views/HeaderTitle.js
+++ b/src/views/HeaderTitle.js
@@ -2,7 +2,7 @@
 
 import React from 'react';
 
-import { Platform, StyleSheet, Text } from 'react-native';
+import { Platform, StyleSheet, Animated } from 'react-native';
 
 import type { Style } from '../TypeDefinition';
 
@@ -11,14 +11,13 @@ type Props = {
   style?: Style,
 };
 
-const HeaderTitle = ({ style, ...rest }: Props) => (
-  <Text
+const HeaderTitle = ({ style, ...rest }: Props) =>
+  <Animated.Text
     numberOfLines={1}
     {...rest}
     style={[styles.title, style]}
     accessibilityTraits="header"
-  />
-);
+  />;
 
 const styles = StyleSheet.create({
   title: {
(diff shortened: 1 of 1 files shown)


matthamil:
Is this a necessary change? Can this be achieved by using the library's [headerTitle](https://reactnavigation.org/docs/navigators/stack#headerTitle) navigationOption?

fdnhkj:
To me, it is.
If I would have to implement that feature with `headerTitle` navigationOption, I would keep a copy of react-navigation's `HeaderTitle` component within my app and adapt it with this change.

Instead, I think it is better to do it directly in react-navigation's `HeaderTitle` component.
It is like the Animated present in [`Header`](https://github.com/react-community/react-navigation/blob/master/src/views/Header.js#L290) component.

fdnhkj:
@matthamil Please let me know if you have any concern about this change.

skevy:
@fdnhkj this seems fine to me. Can you fix the linting errors? You need to run HeaderTitle through Prettier.

fdnhkj:
@skevy Done :)

fabriziogiordano:
@fdnhkj would you mind pasting an example on how to use this new feature?

fdnhkj:
@fabriziogiordano I didn't have time to test it but this should give an idea.
```
// Within a screen component
static navigationOptions = ({ navigation }) => {
    return {
      title: 'MyScreen',
      headerTitleStyle: {
        opacity: navigation.state.params === undefined
          ? 1
          : navigation.state.params.headerTitleOpacity
      }
    };

componentWillMount() {
    this.animatedValue = Animated.Value(1);
      headerTitleOpacity: this.animatedValue
    });
  }

componentDidMount() {
    Animated.startTiming(this.animatedValue, {
        toValue: 0.5,
        duration: 500
    }).start();
}
```

PR #1911, opened 2017-06-16
Use PropTypes from prop-types #3

szhigunov:
Remove warning from React about PropTypes, use prop-types instead

PR #1928, opened 2017-06-18
Use PropTypes from prop-types ( examples )

szhigunov:
Remove warning from React about PropTypes, use prop-types instead.
Contains #1909 #1910 #1911


matthamil:
@szhigunov Thank you for fixing these warnings. Can you update this branch with `master` so I can merge it?

PR #1968, opened 2017-06-22
Update examples

szhigunov:
Use `prop-types` instead of `React.PropTypes`


szhigunov:
I modified previous pull request #1928 to keep commit history more clear.

PR #1972, opened 2017-06-23
Added typescript definitions

gabrielbull:
For us using Typescript, it would be easier to have the definition in the repo instead of adding a separate dependency. It would also allow us to maintain the Typescript definition in here instead of on DefinitelyTyped.

This definition is imported directly from DefinitelyTyped and has been created by @huhuanming, @mhcgrq, @andrewlinfoot, @agentilela, @habibialkaff, @fangpenlin, @jeremistadler and @AleksandarDev.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,662 @@
+// Type definitions for react-navigation 1.0
+// Project: https://github.com/react-community/react-navigation
+// Definitions by: Huhuanming <https://github.com/huhuanming>
+//                 mhcgrq <https://github.com/mhcgrq>
+//                 fangpenlin <https://github.com/fangpenlin>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+import * as React from 'react'
+import {
+  Animated,
+  ViewStyle,
+  TextStyle
+} from 'react-native'
+
+export type HeaderMode = 'float' | 'screen' | 'none'
+
+export function addNavigationHelpers(params: any): any
+
+type SubViewProps = NavigationSceneRendererProps & {
+  onNavigateBack?: () => void,
+}
+
+type SubViewRenderer = (subViewProps: SubViewProps) => React.ReactElement<any> | undefined
+
+export type HeaderProps = NavigationSceneRendererProps & {
+  mode: HeaderMode,
+  onNavigateBack?: () => void,
+  renderLeftComponent: SubViewRenderer,
+  renderRightComponent: SubViewRenderer,
+  renderTitleComponent: SubViewRenderer,
+  tintColor?: string,
+  router: NavigationRouter,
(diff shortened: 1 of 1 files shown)


huhuanming:
cool!
And should resolve the conflicts at first

AleksandarDev:
We should
- [ ] move `react-navigation-tests.tsx` too
- [ ] add ts tests to CI 
- [ ] ensure index.d.ts is published with npm package
- [ ] mark @types/react-navigation as obsolete (create issue on dt)
- [ ] update index.d.ts on dt to warn fellow developers not to update that package
- [ ] mention this (breaking) change in changelog/release notes in react-navigation's next release
- [ ] eventually remove package from DefinitelyTyped

See how Redux was handled: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9247

huhuanming:
Look here.

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-navigation

nonameolsson:
@huhanming Is that made by someone in this thread?

huhuanming:
0 - 0

I added typings to `DefinitelyTyped/DefinitelyTyped`, but no one maintain this thread.

nonameolsson:
@huhuanming Great! Sorry if it's a dumb question, but how would I do to use it with VS Code?

huhuanming:
VSCode will be autoload the typings in JavaScript Project.
But in TypeScript Project, you should add `@types/react-navigation` into your `package.json`, and install it.

nonameolsson:
What's the status on this? Could it be fixed so that we can also get #1975

huhuanming:
I am waiting for it too, butno updates.
Maybe they think that we can use @types/react-navigation to do same work and really doesn't need adding it to this repo.

AleksandarDev:
@gabrielbull You should update the PR and remove react-native from DT repo. 

Here are the instructions on removing the package from DT... https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

gabrielbull:
This PR should be merged and the todo list moved to another issue so we can continue working on it.

matthamil:
Hi, there was recently a PR merged that updated some of the Flow types for the package. Can you check to see if any changes from that PR might impact the types in this one?

brentvatne:
this is out of date so I'm going to close, but I would love to see typescript definitions added to the project. we recently added a flow interface (https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js) and published to flow-typed, and it'd be great to do the same for definitelytyped. if you're still interested, you will definitely get an expedient review next time as we have a couple of dedicated people working on this project now. thanks for your efforts!

brentvatne:
(managing the definition here and syncing out to definitelytyped is a good call, it's what we do for flow/flow-typed)

PR #1975, opened 2017-06-23
Feature/with navigation typescript definition

gabrielbull:
Dependent on PR #1972

Added withNavigation type definition to typescript definitions.


brentvatne:
this is out of date so I'm going to close, but I would love to see typescript definitions added to the project. we recently added a flow interface (https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js) and published to flow-typed, and it'd be great to do the same for definitelytyped. if you're still interested, you will definitely get an expedient review next time as we have a couple of dedicated people working on this project now. thanks for your efforts!

Issue #1889, opened 2017-06-15
Path starting with / leads to unexpected errors


slorber:
### Current Behavior

If you have a router with a path starting with "/",  it will lead to hard to debug errors, particularly inside `getActionForPathAndParams` 

The problem is that we use `const parts = path.split('/');` so if the path starts with a "/", then parts[0] will always be `""`.

For the record, current implementation is:

```
 getActionForPathAndParams(path: string, params: ?NavigationParams) {
      return (
        order
          .map((tabId: string) => {
            const parts = path.split('/');
            const pathToTest = paths[tabId];
            if (parts[0] === pathToTest) {
              const tabRouter = tabRouters[tabId];
              const action: NavigationNavigateAction = NavigationActions.navigate(
                {
                  routeName: tabId,
                }
              );
              if (tabRouter && tabRouter.getActionForPathAndParams) {
                action.action = tabRouter.getActionForPathAndParams(
                  parts.slice(1).join('/'),
                  params
                );
              } else if (params) {
                action.params = params;
              }
              return action;
            }
            return null;
          })
          .find((action: *) => !!action) ||
        order
          .map((tabId: string) => {
            const tabRouter = tabRouters[tabId];
            return (
              tabRouter && tabRouter.getActionForPathAndParams(path, params)
            );
          })
          .find((action: *) => !!action) ||
        null
      );
    },
```

### Expected Behavior

1) throw error on routers declared with path starting with a `/` for better DX

OR

2) accept paths starting with "/" and make `getActionForPathAndParams`work by removing leading "/" to get first path part.

kelset:
Pinging OP @slorber since this issue has not been active for a while, and it's related to an old version of the lib.

Please let me know if you want this to remain open, because if I get no answer in the next 7 days I will close it.

kelset:
Hi there @slorber ,

In an effort to cleanup this project and prioritize a bit, since this issue had no follow up since my last comment I'm going to close it.

If you are still having the issue with the latest version (especially if it's a bug report) please check the current open issues, and if you can't find a duplicate, open a new one that uses the new Issue Template to provide some more details to help us solve it.

PR #2033, opened 2017-06-30
Throw error if route config path has leading slash symbol

ksvitkovsky:
Fix for issue #1889 

Currently it is not obvious for developer that something has gone wrong if route config paths have leading slashes. With this fix, error is being thrown on validation of route configs.

Added new test to cover this case.

### DIFF
--- a/docs/api/routers/StackRouter.md
+++ b/docs/api/routers/StackRouter.md
@@ -41,7 +41,7 @@ const MyApp = StackRouter({ // This is the RouteConfig:
 
 Each item in the config may have the following:
 
-- `path` - Specify the path (should not start with the "/" character) and params to be parsed for item in the stack
+- `path` - Specify the path and params to be parsed for an item in the stack. The path value should not start with the `/` character.
 - `screen` - Specify the screen component or child navigator
 - `getScreen` - Set a lazy getter for a screen component (but not navigators)
 
--- a/docs/api/routers/StackRouter.md
+++ b/docs/api/routers/StackRouter.md
@@ -41,7 +41,7 @@ const MyApp = StackRouter({ // This is the RouteConfig:
 
 Each item in the config may have the following:
 
-- `path` - Specify the path and params to be parsed for item in the stack
+- `path` - Specify the path (should not start with the "/" character) and params to be parsed for item in the stack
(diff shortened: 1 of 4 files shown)


src/routers/validateRouteConfigMap.js
```diff
@@ -55,6 +55,14 @@ function validateRouteConfigMap(routeConfigs: NavigationRouteConfigMap) {
           '}'
       );
     }
+
+    if (routeConfig.path) {
+      invariant(
+        routeConfig.path[0] !== '/',
+        `Invalid path '${routeConfig.path}' for route '${routeName}'. ` +
+          `Path should not start with the slash symbol`
```
matthamil: I think it would be best to be explicit with the error message and say `Path should not start with the "/" character.` "Slash" could be slightly ambiguous since the slash could be a forward or back slash, yet only one throws an error.

ksvitkovsky:
I thought about proposed implementation too, but decided to go with warning, because:
- it does not add complexity to path handling logic
- although it is nice to allow both ways, our focus is to provide universal solution, so setting some rules might not hurt
- paths without slash just seem cleaner :-)

docs/api/routers/StackRouter.md
```diff
@@ -41,7 +41,7 @@ const MyApp = StackRouter({ // This is the RouteConfig:
 
 Each item in the config may have the following:
 
-- `path` - Specify the path and params to be parsed for item in the stack
+- `path` - Specify the path (should not start with the "/" character) and params to be parsed for item in the stack
```
matthamil: I think for the sake of clarity, this should be reworded to:

```
`path` - Specify the path and params to be parsed for an item in the stack. The path value should not start with the `/` character.
```

docs/api/routers/TabRouter.md
```diff
@@ -37,7 +37,7 @@ const MyApp = TabRouter({ // This is the RouteConfig:
 
 Each item in the config may have the following:
 
-- `path` - Specify the path for each tab
+- `path` - Specify the path (should not start with the "/" character) for each tab
```
matthamil: See the other review comment about rewording.

PR #2098, opened 2017-07-08
Fix Transitioner link in Views.md doc

matthamil:
This fixes an issue on https://reactnavigation.org/docs/views/ where the Transitioner link does not appear unless a user mouses over it.


spencercarli:
Seems this was done at some point already

PR #2099, opened 2017-07-08
Fix broken documentation links

ixje:


spencercarli:
The ids use capital letters so to make sure it continues to work I'm going to leave it as is. Thank you though!

ixje:
It looks like something changed because now the documentation actually points to the right sections, where previously it redirected you to https://reactnavigation.org/docs/routers/api for some reason. Either way, I'm fine with it being closed. As long as the docs work as they're supposed to do :)

Issue #1584, opened 2017-05-18
Cannot read property 'setNativeProps' of undefined


veljkomatic:
I have
```js
 TabNavigator({
     user: {
        screen: StackNavigator ({
           userSetting: { screen: UserSetting }
       })
     },
     shop: {
         screen: StackNavigator ({
             shopScreen: { screen: Shop}
         }, {
           navigationOptions: {
            tabBarVisible: false,
          }
        })
     },
     support: {
        screen: StackNavigator ({
           supportScreen: { screen: Support}
       })
     }
```

So I set tabBarVisible false on shop child element...but I cant navigate on support it crashes withCannot read property 'setNativeProps' of undefined

EslamNasser:
I'm having the same problem with a similar setup.
Anybody got a solution to this?

josephrstraub:
+1 

Works on Android but not iOS

EslamNasser:
@josephrstraub how did you get it to work on android?

romanenko:
I have the same issue with the similar setup: it occurs on switching between tabs, when one of the tab has `tabBarVisible: false`. And I'm also having this error only in iOS.

vqc:
It seems like an error _is not_ thrown when the screen you are navigating too also has the tab bar hidden.

hyubs:
👍  Encountered this issue too. Commenting out the `tabBarVisible: false` fixes the issue. Using iOS simulator and react-navigation@1.0.0-beta.9.

(though I still need to figure out a workaround to hide the tab bar)

valetarton:
I also encountered this issue and resolved it with taking away the tabBarVisible: false
Though I might add this issue came up when I both set route config dynamically AND statically

HeKnOw:
Can confirm issue is happening as described, and removing tabBarVisible "resolves" the issue

zinoviev:
+1 If it's intended behavior will be nice to have noted in docs

doomsower:
Have same problems with master branch
For me unstable workaround is to add setTimeout in componentDidUpdate in TabViewPagerPan in `node_modules/react-navigation/node_modules/react-native-tab-view` to
```js
  componentDidUpdate(prevProps: Props<T>) {
    if (prevProps.navigationState.index !== this.props.navigationState.index) {
      setTimeout(this._transitionTo, 0, this.props.navigationState.index);
    }
  }
```
[...]

Issue #1398, opened 2017-05-05
Set tabBarVisible to be false in tabNavigator cause navigating crash


Marswang92:
### Current Behavior
I've got a screen which has a `navigationOption` as follows:
```
  static navigationOptions = {
    tabBarVisible: false,
    tabBarIcon: ({ tintColor }) => (
      <Icon
        name={ 'camera-iris' }
        color={ tintColor }
        size={60}
        style={{
          marginBottom: 30,
        }}
      />
    )
  };
```
Once I call `this.props.navigation.goBack()` in my screen, the App crash and give following error message. If I change `tabBarVisible` back to true, everything works properly. Anybody can help me with it?

![screen shot 2017-05-05 at 9 55 15 am](https://cloud.githubusercontent.com/assets/14301745/25755743/fc0fae1c-3178-11e7-8268-6e80003818b0.png)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  1.0.0-beta.9
| react-native     |  react-native
| react             | 16.0.0-alpha.6

Marswang92:
Actually just solved by setting `animationEnabled` to be false in `tabBarConfiguration`, but is this behavior expected?

matthamil:
Does this still crash when you remove `tabBarIcon`? Taking a guess that it could be a conflict between rendering the tab bar icon and also declaring that no tab bar should be present.

acomito:
Is this for android?

Marswang92:
@matthamil I haven't tried that yet but yep I think that can be the cause. Will test that out later today.

@acomito It just happened to me on iOS but I'll test it on Android once I've got time.

hbar-digital:
I am experiencing the same issue even when removing `tabBarIcon`. This started occurring after upgrading from `1.0.0-beta.6` to `1.0.0-beta.9`.

matthamil:
@hbar-digital Can you confirm that this does or does not happen when you remove `tabBarIcon` when you flag `tabBarVisible` as `false`?

hbar-digital:
I am getting this error regardless of whether `tabBarIcon` is set. If `tabBarVisible` is set as false, I am unable to navigate (`goBack` or `navigate`).

hbar-digital:
Looks like the issue is here:

https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabView.js#L164

[...]

PR #2101, opened 2017-07-09
Fix back navigation when tabBarVisible is false ...

gusgard:
Workaround that fix back navigation when `tabBarVisible` is `false` and `tabBarPosition` is `bottom`. The issue is in `react-native-tab-view` because doesn't support hide the children elements.

Related issues: #1398 and #1584

Render `renderTabBar` with `display: 'none'` instead of nothing.


src/views/TabView/TabView.js
```diff
@@ -110,54 +110,56 @@ class TabView extends PureComponent<void, Props, void> {
...
       return null;
     }
+
+    const { state } = navigation;
+    const options = router.getScreenOptions(
+      this.props.childNavigationProps[state.routes[state.index].key],
+      screenProps || {}
+    );
+
+    const tabBarVisible = options.tabBarVisible == null
```
matthamil: Can you change the `==` to an equality using `===`?  This ternary is hard to read.

gusgard: Updated 👍

gusgard:
@matthamil the test fails because of `warning  Unexpected use of weak type "any"  flowtype/no-weak-types` maybe it should be better to disable the eslint rule? or is used when a new version is released?

src/views/TabView/TabView.js
```diff
@@ -110,54 +110,56 @@ class TabView extends PureComponent<void, Props, void> {
...
     return (
       <TabBarComponent
         {...props}
         {...tabBarOptions}
-        screenProps={this.props.screenProps}
         navigation={this.props.navigation}
         getLabel={this._getLabel}
         renderIcon={this._renderIcon}
         animationEnabled={animationEnabled}
+        style={!tabBarVisible && styles.hidden }
```
cooperka: This is unclear syntax for many people, please use the more conventional `style={!tabBarVisible ? styles.hidden : undefined}`

src/views/TabView/TabView.js
```diff
@@ -110,54 +110,56 @@ class TabView extends PureComponent<void, Props, void> {
...
       return null;
     }
+
+    const { state } = navigation;
+    const options = router.getScreenOptions(
+      this.props.childNavigationProps[state.routes[state.index].key],
+      screenProps || {}
+    );
+
+    const tabBarVisible = options.tabBarVisible === null
```
cooperka: I just saw @matthamil's comment [above](https://github.com/react-community/react-navigation/pull/2101#discussion_r126574026) recommending this originally. To be clear, this code was simply moved from lower down and needs to check for `undefined` also, otherwise the tabs all disappear if `tabBarVisible` is not explicitly set.

gusgard: Thanks for clarifying, I tested with beta.12 and works fine. Could you tell me with what flow fails?

src/views/TabView/TabView.js
gusgard: Sure

src/views/TabView/TabView.js
```diff
@@ -110,22 +110,37 @@ class TabView extends PureComponent<void, Props, void> {
...
     return (
       <TabBarComponent
         {...props}
         {...tabBarOptions}
-        screenProps={this.props.screenProps}
         navigation={this.props.navigation}
         getLabel={this._getLabel}
         renderIcon={this._renderIcon}
         animationEnabled={animationEnabled}
+        style={tabBarVisible ? undefined : styles.hidden}
```
cooperka: This clobbers the `style` prop in `tabBarOptions`. Fix:

```diff
diff --git a/src/views/TabView/TabView.js b/src/views/TabView/TabView.js
index 7cf6653..bae866a 100644
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -13,12 +13,15 @@ import type {
   NavigationState,
   NavigationRouter,
   NavigationTabScreenOptions,
+  Style,
 } from '../../TypeDefinition';
 
 export type TabViewConfig = {
   tabBarComponent?: ReactClass<*>,
   tabBarPosition?: 'top' | 'bottom',
-  tabBarOptions?: {},
+  tabBarOptions?: {
+    style?: Style,
+  },
   swipeEnabled?: boolean,
   animationEnabled?: boolean,
   lazy?: boolean,
@@ -34,7 +37,9 @@ export type TabScene = {
 type Props = {
   tabBarComponent?: ReactClass<*>,
   tabBarPosition?: 'top' | 'bottom',
-  tabBarOptions?: {},
+  tabBarOptions?: {
+    style?: Style,
+  },
   swipeEnabled?: boolean,
   animationEnabled?: boolean,
   lazy?: boolean,
@@ -149,7 +154,7 @@ class TabView extends PureComponent<void, Props, void> {
         getOnPress={this._getOnPress}
         renderIcon={this._renderIcon}
         animationEnabled={animationEnabled}
-        style={!tabBarVisible ? styles.hidden : undefined}
+        style={[!tabBarVisible ? styles.hidden : undefined, tabBarOptions && tabBarOptions.style]}
       />
     )
   };
```

gusgard: Thanks for the review, already added

gusgard:
@cooperka I will test all the PR during the weekend.

gusgard:
@matthamil could you check it?

etaiklein:
+1

etaiklein:
I tried this patch as is and got errors (display: 'none' isn't a valid property).
I think this fix is the right idea, but needs more attention.

matthamil:
Assigning myself so I can take a look at this in the next couple of days.

gusgard:
@etaiklein what version of react-native are you using? Also, I will check it again with the last merge from master.

@matthamil great roadmap!

dchersey:
Confirming that this will fix navigation FROM a tab whose tab bar is hidden (original issue in #1584).
If so, I'll wait for the merge... thanks!

gusgard:
@dchersey I tested with the last version of master a few minutes ago and works fine. Please test it with this version `https://github.com/gusgard/react-navigation#1f411bc1f45cebae0c6d7961bbe8f70420495eac`

dchersey:
Thank you @gusgard, :thumbsup: that version does the trick!

gusgard:
@matthamil I merged with 1.0.0-beta.16 and works correctly. Could you take a look?

lucasbento:
@gusgard: hey, thanks for this PR, I'm using it with `react-navigation: "v1.0.0-beta.19"` and tab hides fine but doesn't show the images of the icon of each tab when it appears back again, is there something that I can do?

gusgard:
@lucasbento could you share the code? I never tested with images but with components work fine.

src/views/TabView/TabView.js
```diff
@@ -93,11 +93,30 @@ class TabView extends React.PureComponent {
...
     }
 
+    const { state } = navigation;
+    const options = router.getScreenOptions(
+      this.props.childNavigationProps[state.routes[state.index].key],
+      screenProps || {}
+    );
+
+    const tabBarVisible =
+      options.tabBarVisible === null || options.tabBarVisible === undefined
```
brentvatne: `options.tabBarVisible === false ? false : true`

gusgard: Ok, is there a reason why to not use `const tabBarVisible = options.tabBarVisible !== false;`

brentvatne: that's fine too, the current form is just a bit verbose

gusgard: Updated!

satya164:
Can you file a bug report for this in https://github.com/react-native-community/react-native-tab-view with a repro? This is weird. Also, does this issue still exist? There have been a lot of changes to react-native-tab-view since the PR was opened. Just need to ensure that this is still an issue.

gusgard:
Yes, still happen I will try to have a working example for this weekend.

ronal2do:
Any update?

AbhishekCode:
Version 1.5.9 and issue still exist. Even workaround which was working on older version is not working now.

gusgard:
@ronal2do @AbhishekCode What version of RN are you using it?

AbhishekCode:
I am using 0.51.0.  
By the way i have found a workaround for now https://github.com/react-navigation/react-navigation/issues/1584#issuecomment-377938796

ronal2do:
"react": "16.0.0",
    "react-native": "0.49.3",
@gusgard But, If possible to pass `tabBarStyles` can be solve my problem, I have several internal levels of tabbar hide/show, only 1 has this problem

AbhishekCode:
@ronal2do You can create your own tabbarcomponent, then inside that you can check this.props.navigation which contains whole navigator props including styles. Then you can overwrite style depending on your tab.

brentvatne:
this belongs on https://github.com/react-navigation/react-navigation-tabs now :)

PR #2121, opened 2017-07-11
Don't forward back actions to inactive children of TabRouters

monteithpj:
**Motivation**

If we have a TabRouter with child routers and we receive a navigation action, we first delegate to the active child, then try to handle it ourselves, then iteratively try to delegate to inactive children. This behavior makes sense if we have a Navigate action, but is very strange for a Back action (Why push forward, to inactive children, when the action is inherently going backwards). This change simply returns the current state if a TabRouter gets a Back action that isn't handled by the active child and isn't handled by the TabRouter itself.

**Test plan**

Ensured that no other unit tests are affected by this change, added a unit test to check that the behavior is working


brentvatne:
@ericvicenti - motivation here makes a lot of sense to me, could you take a look?

PR #2143, opened 2017-07-14
Add prerelease versions of react 16 to package.json.

### LINKED ISSUE #2133: react-navigation cannot be used with react-native
### Current Behavior
The current `react` peerDependency of `react-navigation` is incompatible with the `react` peerDependency of `react-native`.

`react-native`: `react@16.0.0-alpha.12`
`react-navigation`: `react@*`

As per https://github.com/npm/node-semver#prerelease-tags, the `*` does not include prerelease tags, thus:

```
> semver.satisfies('16.0.0-alpha.12', '*')
false
> semver.satisfies('16.0.0-alpha.12', '^16.0.0-0')
true
```

### Expected Behavior
I would expect them to be compatible.

A way to achieve this might be by adding the following peerDependency in `react-navigation`:

```
react@* || react@^16.0.0-0
```

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  `1.0.0-beta.11`
| react-native     |  `0.44.0`
| node             |  `4.0.0`
| npm      | `2.14.2`

It may only be an issue in old versions of node or npm; I realise we're outdated here. Nevertheless other frameworks have more flexible peerDependencies:

```
react-redux@5.0.5 wants react@^0.14.0 || ^15.0.0-0 || ^16.0.0-0
```

jslvtr:
This fixes #2133 by adding prerelease version of React to package.json.

**Test plan (required)**

No code changes made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "react-test-renderer": "^15.4.2"
   },
   "peerDependencies": {
-    "react": "*",
+    "react": "* || ^16.0.0-0",
     "react-native": "*"
   },
   "dependencies": {


spencercarli:
This is no longer necessary - thank you for the contribution though!

PR #2180, opened 2017-07-18
Fix custom `items` prop for DrawerItems

cooperka:
## Motivation

The `items` prop was added in [this PR](https://github.com/react-community/react-navigation/pull/1039), but unfortunately it never actually worked due to being immediately overridden by `items={state.routes}`.

This fixes the prop by allowing everything specified in [`contentOptions`](https://reactnavigation.org/docs/navigators/drawer#contentOptions-for-DrawerItems) to override the default behavior.

## Test Plan

Try it out on the example app, e.g. in `Drawer.js`:

```js
    initialRouteName: 'Drafts',
    contentOptions: {
      items: [{
        key: 'Drafts',
        routeName: 'DraftsScreen',
      }],
    },
```

### DIFF
--- a/src/views/Drawer/DrawerSidebar.js
+++ b/src/views/Drawer/DrawerSidebar.js
@@ -98,7 +98,6 @@ class DrawerSidebar extends PureComponent<void, Props, void> {
     return (
       <View style={[styles.container, this.props.style]}>
         <ContentComponent
-          {...this.props.contentOptions}
           navigation={this.props.navigation}
           items={state.routes}
           activeItemKey={
@@ -109,6 +108,7 @@ class DrawerSidebar extends PureComponent<void, Props, void> {
           renderIcon={this._renderIcon}
           onItemPress={this._onItemPress}
           router={this.props.router}
+          {...this.props.contentOptions}
         />
       </View>
     );


dantman:
The point of #1039 was not to allow items to be overridden in contentOptions, that doesn't make sense. The PR also does not permit arbitrary objects to be passed to items.

The purpose of the PR was to allow a custom drawer component to provide a filtered list of `navigation.state.routes` items to `DrawerView.Items` instead of the whole list.

dantman:
😰 I think when I wrote the documentation for `DrawerView.Items` I mistook the `contentOptions` documentation for documentation about the `DrawerView.Items` props. `items`, `activeItemKey`, and `onItemPress` are not intended to be overridden in `contentOptions`. They are intended to be used in a `customComponent`.


Like this modified version of the ["Providing a custom `contentComponent`"](https://reactnavigation.org/docs/navigators/drawer#Providing-a-custom-contentComponent) example in the documentation. That hides one item and renders a Home item separate from the rest of the drawer items.
```js
import { DrawerItems } from 'react-navigation';

const CustomDrawerContentComponent = (props) => (
  <View style={style.container}>
    <DrawerItems
      {...props}
      items={items.filter((item) => item.routeName === 'Home' && item.routeName !== 'HiddenRoute')} />
    <Text>Section</Text>
    <DrawerItems
      {...props}
      items={items.filter((item) => item.routeName !== 'Home' && item.routeName !== 'HiddenRoute')} />
  </View>
);

  container: {
    flex: 1,
  },
});
```

cooperka:
@dantman thank you for your clarifications.

This change is currently needed for https://github.com/react-community/react-navigation/pull/2183, and unfortunately I don't have time right now to refactor. If you can provide the same feature as proposed in https://github.com/react-community/react-navigation/pull/2183 without overriding `items` in this way, I would be happy to use that instead, but for now I consider this good enough to be quite useable.

dantman:
@cooperka This is not at all necessary for that change. In fact it's counterproductive as `items` should not be abused to add non-navigation items, #2183 already makes it possible to do that without using `items`.

Using `itemComponent` to override drawer items with a different style of drawer item doesn't require customization of `items`.

And to add non-navigation items that is what `contentComponent` is for. And you've already separated `DrawerNavigatorItem` out and made it a generic UI component that doesn't require any navigation specific props – which means you can now use it outside of `DrawerNavigatorItems` to make non-navigation items that look just like the default navigation items.

For example, your example of adding a refresh item and a non-item section header could easily be done without modifying `items` just by doing:
```js
import { DrawerItems, DrawerItem } from 'react-navigation';

// ...
  initialRouteName: 'Drafts',
  contentComponent: (props) => (
    <View style={style.container}>
      <DrawerItem
        icon={<MaterialIcons name='refresh' size={24} style={{ color: tintColor }} />}
        focused={false}
        label={props.navigation.state.params.isRefreshing ? 'Refreshing...' : 'Refresh'}
        onPress={doRefresh} />
      <Text>Section</Text>
      <DrawerItems {...props} />
    </View>
  ),
```

In fact doing it this way is much easier to understand, uses less code (than creating a custom drawer item whose sole purpose is to use a hack to fake in a non-interactive section header), and is much more flexible (complete control over where everything goes and no limits on how you style things).

cooperka:
Please correct me if I'm misunderstanding your point, but `DrawerNavigatorItems` calculates 7 different props to pass to `DrawerNavigatorItem`. To me it seems easier to use those calculations rather than calculate them all yourself, e.g.

```js
      const focused = activeItemKey === route.key;
      const tintColor = focused ? activeTintColor : inactiveTintColor;
      const scene = { route, focused, index, tintColor };
      const icon = renderIcon(scene);
      const label = getLabel(scene);
      ...
      key={route.key}
      onPress={() => onItemPress({ route, focused })}
      {...drawerItemProps}
```

Sure, it's not **that much** code, but I prefer to let the library handle its own rendering and not duplicate things.

The user always has the option to do it your way instead, I just personally find this way easier.

dantman:
@cooperka None of those calculations are useful when you're trying to add a new item that is not a navigation item.

non-navigation items don't get implicitly focused because they aren't navigation items that can be active:
```js
const focused = activeItemKey === route.key;
```

active/inactive tint color calculation doesn't work when the focus calculation doesn't work in the first place:
```js
const tintColor = focused ? activeTintColor : inactiveTintColor;
```

this isn't actually a calculated prop:
```js
const scene = { route, focused, index, tintColor };
```

`renderIcon` doesn't actually render or calculate an icon, it just asks the `drawerIcon` function the user provided to render an icon for it; there's no point doing this for a non-navigation item when the user can just provide that icon directly to the item component:
```js
const icon = renderIcon(scene);
```

non-navigation items always define their own labels, the extra steps of going through `getLabel` and `navigationOptions` to get that custom label are just wasted:
```js
const label = getLabel(scene);
```

key is not necessary when you are not using an array:
```js
key={route.key}
```

non-navigation inherently have their own custom onPress handling so the default onItemPress is useless:
```js
onPress={() => onItemPress({ route, focused })}
```

So none of the calculations done by `DrawerItems` are relevant when you pass something that is not a react-navigation drawer route to it. All of them either just waste your time or force you to override them with something that actually works.

cooperka:
Have a look at the [Drawer.js example](https://github.com/react-community/react-navigation/pull/2183/files#diff-c9d0de489f8782c6f37355ab854f5d04) in my PR. I'm using the default `DrawerNavigatorItems` with a unitary `items` array to render each unmodified drawer item. You're right that `items` isn't needed for custom items, but that's not what it's being used for.

cooperka:
Your code example [above](https://github.com/react-community/react-navigation/pull/2180#issuecomment-327309784) is great for adding some extra items to the top of the drawer, and in that case it works great (though using a custom header component would have worked just as well with even less code). However, I wanted to provide a more generic solution that allows inserting arbitrary custom drawer items **in between** the default ones.

dantman:
@cooperka My solution is more generic. I didn't include an example for inserting items in between / dividing the routes up into sections because I've already given examples for that.

https://github.com/react-community/react-navigation/pull/2180#issuecomment-323000634
https://github.com/react-community/react-navigation/issues/795#issuecomment-322999805

If you want something big/flexible/complex and real-world, heres a bit of code from an app I'm working on:
https://gist.github.com/dantman/caefc8778303dd1609659860bd4729ab

This is a pretty complex real-world use where I:
- Automatically divide up drawer items into multiple sections with dividers in between
- Implicitly give some of those sections section headers
- Take one of those items/sections (the Teams) and insert some custom drawer items that are based on data I fetch elsewhere from an API (ie: dynamic drawer items that do not require routes or a fixed list of items that is known to the app)
- Give the drawer a ScrollView
- Give the drawer a DrawerHeader (it's a custom component that looks like the [Material Design navigation drawer](https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-content)'s header)
- Gives the drawer a slide down account layer (see Google apps on Android where pressing the button in the drawer header opens a separate account drawer)
- This account drawer contains a custom logout item and company items which are more items based on data fetched from an API like the team items. The logout item is a custom item just because that is how I structured logout in my app (login/logout are top level screens that are at a higher level than the drawer), but it's equally possible for the account drawer to have login/logout/account items that are actual drawer routes.

This example looks pretty big, but all that is because this example itself is complex (99% of that largeness is related to the app, not the react-navigation code).

cooperka:
I guess I remain unconvinced that forcing people to do it your way is the right solution here. This PR leaves the option up to the developer.

I'll leave it up to the react-navigation maintainers to decide whether to merge. Note that if the decision is made to close this, then we should open a PR to fix the incorrect docs (as referenced in the description).

cooperka:
@ericvicenti agreed since we're moving https://github.com/react-navigation/react-navigation/pull/2183 to a different repo like you suggested.

Unfortunately, I've been working on non-RN projects recently so a project like this won't have much overlap with my work. I may find time to contribute later, but probably not for several months.

dantman:
@cooperka If you're doing drawer navigator in another library and doing items with a basic layout, I've got a Material Design library and I'd be happy to add a material version of the drawer item in it.

Its probably a similar timeframe as you, the client that has the RN app I work the most on wants me to help build something else for their web systems before I get back to the app.

satya164:
We have some material design compliant drawer items here which you can reuse https://callstack.github.io/react-native-paper/drawer-item.html

dantman:
@satya164 Oh, there's yet another MD library? Mine is https://github.com/material-native/material-native/ I was trying to solve the problem that there used to be 2 different Material Design implementations, both unmaintained, and both implementing an incomplete set of components that don't cover everything the other library does.

PR #2185, opened 2017-07-19
Fix handle back button in android

le0pard:
In new react native you need to handle different event.


le0pard:
Closed, already fixed

PR #2223, opened 2017-07-24
[Closes #2158] Support flux standard actions

### LINKED ISSUE #2158: Compatibility with libraries like redux-actions (Flux Standard Actions)
### Current Behavior
The action payload is like this:
{ type: 'Navigation/NAVIGATE', routeName: 'DrawerOpen' }

### Expected Behavior
There is a standard for the payload, many people already use.
{ type: 'Navigation/NAVIGATE', payload: { routeName: 'DrawerOpen' } }
See https://github.com/acdlite/flux-standard-action

Would be great if this lib could become compatible! Because with this change existing redux support libs would work!

Things like createAction https://redux-actions.js.org/docs/api/createAction.html#createactiontype-payloadcreator
```
// actions.js

import { createAction } from "redux-actions";
import { NavigationActions } from "react-navigation";

export const openMenu = createAction(NavigationActions.NAVIGATE, () => { return { routeName: "DrawerOpen" } });
```

mindmelting:
This PR closes issue #2158 and adds support for [Flux Standard Actions](https://github.com/acdlite/flux-standard-action) (commonly used in redux)

I decided not to refactor the existing createAction method, as this would most likely introduce breaking changes for developers who are expecting actions to be created as-is.

This simply checks if an action is FSA compliant, and unpacks the payload object, to be compliant with the existing expected action contents. Given this check, there should be no additional impact on existing implementations.

Unit tests for StackRouter and TabRouter have been added.

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -55,7 +55,6 @@ export default (
       passedAction: NavigationAction | { action: NavigationAction },
       inputState?: ?NavigationState
     ): ?NavigationState {
-      // eslint-disable-next-line no-param-reassign
       let action = NavigationActions.mapDeprecatedActionAndWarn(passedAction);
 
       // Support flux standard actions by expanding payload


brentvatne:
@mindmelting - thanks for this! at the moment i don't see a compelling argument for changing this behavior internally in react-navigation, but if you would like to make a case for it, please open a rfc at https://github.com/react-navigation/rfcs and i'd be happy to discuss further. :)

PR #2224, opened 2017-07-24
Add flex:1 to show MainScreenNavigator in Guide-Nested.md

BenoitClaveau:
NavigatorWrappingScreen is not displayed without flex.

### DIFF
--- a/docs/guides/Guide-Nested.md
+++ b/docs/guides/Guide-Nested.md
@@ -88,7 +88,7 @@ In this case, the NavigatorWrappingScreen is not a navigator, but it renders a n
 class NavigatorWrappingScreen extends React.Component {
   render() {
     return (
-      <View>
+      <View style={{flex: 1}}>
         <SomeComponent/>
         <MainScreenNavigator/>
       </View>


docs/guides/Guide-Nested.md
```diff
@@ -88,7 +88,7 @@ In this case, the NavigatorWrappingScreen is not a navigator, but it renders a n
 class NavigatorWrappingScreen extends React.Component {
   render() {
     return (
-      <View>
+      <View style={{flex: 1}}>
```
matthamil: I think this is an unnecessary change and only makes the example more verbose.

spencercarli:
Thanks for this PR @BenoitClaveau! As @matthamil said - this isn't really necessary or relevant to that example therefore I'm going to go ahead and close this but please don't let that stop you from submitting future PRs!

docs/guides/Guide-Nested.md
svcondor: Without flex: 1 then the screen is blank.
With flex:1 it works as described.
 react-navigation@v1.0.0-beta.12

svcondor:
This IS relevant. @BenoitClaveau  is correct, the example does not work without flex.
The following snack shows the issue
https://snack.expo.io/rJxeBf4sW
uncomment line 27 to fix the issue

matthamil:
@svcondor The error is due to a misunderstanding of React Native components. Ultimately, we do not want to make the user think that they have to render `View` elements with `flex: 1` for the library to work. The code comment above the code snippet in question notes that the component is _not_ a navigator. It is a React component. This is outside of the scope of navigation, and ultimately a decision that the end user should make. I think that if you would want to see this change, I would add a note to the comment above the code block saying that you will need to add some kind of style to set a height on the rendered component, whether that be with flex or whatever solution the user deems appropriate.

svcondor:
@matthamil I agree that the tutorial is about navigation not style and that `flex: 1` is not necessarily the solution.   I just don't think a "Getting Started" example should render as blank.

matthamil:
I'd accept a PR if you wanted to add a comment above the code snippet explaining that the example will render blank on a device without `flex: 1` :)

svcondor:
PR #2639 submitted

PR #2183, opened 2017-07-19
Factor out DrawerItem for easy "account picker" demo

cooperka:
## Motivation

To make it easy to do things like this:

<div>

</div>

And this:

<div>

</div>

## Changes

1. Factor out `DrawerNavigatorItem` into its own component
2. Add an `itemComponent` prop to [`contentOptions`](https://reactnavigation.org/docs/navigators/drawer#contentOptions-for-DrawerItems) (default is `DrawerNavigatorItem`)
3. Allow the `onPress` handler to be `null` to remove touch feedback
4. Update the NavigationPlayground Drawer example to show off the account picker

No breaking changes, as far as I know.

## Usage

See the changes to the demo app for an implementation example.

## Test Plan

You'll probably want to rebase locally on https://github.com/react-community/react-navigation/pull/2170 so that the example app works (verified no conflicts, easy rebase). It also depends on the fix in https://github.com/react-community/react-navigation/pull/2180.

1. Run the NavigationPlayground app
2. Tap the Drawer Example and try it out

Please let me know what you think; the API makes sense to me but I'm open to changing it.

## Notes

Originally submitted as https://github.com/react-community/react-navigation/pull/1317, but due to other PRs that have been merged, the logic has changed significantly enough to warrant opening a separate PR.

PR #2294, opened 2017-08-02
Fix drawer side menu contentOptions props load order

rpopovici:
Fix drawer side menu contentOptions props load order and move onItemPress option to navigationOptions to be accessible per screen


dantman:
Partial duplicate of #2180, also #2183 is the way to go for overriding icon press.

rpopovici:
Why https://github.com/react-community/react-navigation/pull/2183 is not merged already?
and https://github.com/react-community/react-navigation/pull/2180 moves this.props.contentOptions to the bottom which might override other stuff than items

dantman:
@rpopovici I have another commit (#1313) that hasn't been merged. My guess is that react-navigation just isn't bothering to review any enhancements to the library till they get the v1 fundamental pieces finished.

I just added 2 comments to #2180 about why `items`, `activeItemKey`, and `onItemPress` are not meant to be overridden from `contentOptions`.

PR #2334, opened 2017-08-08
idempotent pushes

rpopovici:
The problem: idempotent navigation push actions currently are not supported because:
1. dispatch function is totally broken. Internally is using setState(async call), it's leaking side effects and it's not able to queue actions.
2. each push receives an unique random id by default and there is no way to control this

The solutions:
<del>1. Try to use timeout or debounce. These solutions work very well but they introduce time based constraints and most of the time you don't want this.
2. Fix dispatch asynchronicity by adding a completion handler . This might work in most cases but will not be able to handle situations when you have decoupled components and not able to use calbacks
3. Fix  dispatch by queuing the actions dispatched and serving those when setState is finished</del>
4. Add user defined navigation keys and default to routeName if none is provided
5. Make pushes idempotent only for two consecutive screens with the same key. 
e.g:
Invalid: Home -> Profile_with_samekey  -> Profile_with_samekey
Invalid: Home -> Profile_with_samekey -> Settings -> Profile_with_samekey
Valid: Home -> Profile_with_key1 -> Profile_with_key2

In this way to we fix our debounce problem but we don't have to enforce stricter rules for navigation and this will works also for BACK action

<del>In this PR https://github.com/react-community/react-navigation/pull/2400 I tried to fix all these issues by introducing consecutive idempotent pushes and fixed the asynchronicity problem of dispatch by adding a callback but also I added an action queue
So you should be able to write both:
dispatch(action1)
dispatch(action2)
or 
dispatch(action1, (state, prev, action) => {
dispatch(action2)
})
Both cases will be handled properly </del>

I did some testing on our project, but other than that not so much. Anyone can feel free to do more thorough testing if he wishes.


src/routers/StackRouter.js
```diff
@@ -93,7 +87,7 @@ export default (
       passedAction: NavigationStackAction,
       state: ?NavigationState
     ) {
-      const action = NavigationActions.mapDeprecatedActionAndWarn(passedAction);
+      const action = NavigationActions.mapDeprecatedActionAndWarn(passedAction);
```
dantman: Trailing line added here.

rpopovici:
@dantman you can pass additional key parameter to your push call to differentiate between consecutive routes. Uniqueness is not automatically handled anymore, but user defined as it should be IMO

dantman:
@rpopovici It should be the other way around, allowing the user to specify the key is nice to allow the user to control the key in the situations where reasonable (the subset of apps that have unique limits on navigation or specific cases where you want to control the screen later from a nested navigator using the key), but it should not be required. When the developer makes a call to `dispatch` it should be presumed that they want the action dispatched, not that they made a mistake by not handling double press and requiring them to pass a custom `key` that they previously did not require.

rpopovici:
@dantman you are right, it's not the users business to take care of key assignment and it should happen automatically. My solution is a compromise from the ideal case by defaulting the key to the route's name if not provided already and asking to provide an unique key if you want to push same route consecutively. Maybe I should add a console warning in case you are trying to do this and let the user know what is expected of him.
Also, IMO the case where you want to push same route consecutively multiple times is rare, it's an exception. I don't really have an use case in mind right now. Fixing the usual cases has higher priority than treating a corner case. I think KISS should apply to this project as well..

rpopovici:
for reset action to work properly PR https://github.com/react-community/react-navigation/pull/2292 should be consider as well

dantman:
Oh right, besides being a breaking change the fixed keys will likely be a regression that re-introduces the bug fixed by #1320.

rpopovici:
@dantman working on v2. Wait for it

rpopovici:
Reworked my initial PR #2334
    async dispatch with callback and insured synchronicity through action queuing
    requestAnimationFrame before dispatching in order to fix some transitioner weird behavoir where some unmounted screens get remounted for a split second while transitioning
    idempotent pushes for routes with identical name
    optional unique push key. overrides route name

dannycochran:
Any chance to update and merge this? I know there's a rewrite for react-navigation in progress, but this remains a pretty big pain point for a lot of users on the current version.

dantman:
@ericvicenti Instead of idempotent pushes using next route keys like this me and @rpopovici seemed to have settled on using current route keys like how `goBack` works. 

This is described in this discussion: https://github.com/react-navigation/react-navigation/pull/2578#issuecomment-330835051

ericvicenti:
This approach seems like a good way to handle idempotent pushes. I'm gonna close this out and do this in a new PR. The only things I plan on doing differently:

- always properly generate a UUID when key is missing, so users can ignore key and still push multiple of the same routeName
- tests

PR #2343, opened 2017-08-09
idempotent pushes v2

rpopovici:
- async dispatch with callback and insured synchronicity through locking and action queuing 
- idempotent pushes for routes with identical name and identical initial params if provided
- optional unique push key. overrides route name and params matching

For a detailed description check my previous PR https://github.com/react-community/react-navigation/pull/2334


nihgwu:
IMO optional unique is the right approach 👍

PR #2348, opened 2017-08-10
Pin react-native-tab-view with ~ instead of ^

hawkrives:

> Explain the **motivation** for making this change. What existing problem does the pull request solve?

Because react-native-tab-view insists on sticking at 0.0.*, 
the caret version pinning won't install "minor" updates.

Semver rules are weird with 0. major versions.

Is this something you'd be interested in?

> Prefer **small pull requests**. These are much easier to review and more likely to get merged. Make sure the PR does only one thing, otherwise split it.

**Test plan (required)**

> Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

I'm updating a dependency, so let's see what the CI says.

**Code formatting**

> Look around. Match the style of the rest of the codebase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "path-to-regexp": "^1.7.0",
     "prop-types": "^15.5.10",
     "react-native-drawer-layout-polyfill": "^1.3.2",
-    "react-native-tab-view": "^0.0.67"
+    "react-native-tab-view": "~0.0.67"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",


hawkrives:
Thanks!

Palisand:
@hawkrives @spencercarli After this change and a fresh install, **react-native-tab-view@0.0.69** (rather than 67) is installed, which results in the following warning:
```
Warning: Failed prop type: The prop `onIndexChange` is marked as required in `TabViewAnimated`, but its value is `undefined`.
```
This is probably responsible for #2510.

hawkrives:
Agreed. It is difficult to use semver with a library that doesn't follow semver, it seems. 

I suppose this should be reverted if tab-view has made breaking changes.

matthamil:
@spencercarli It looks like there are big changes between the react-native-tab-view versions, particularly with the warning listed above in @Palisand 's comment. Can you confirm that this is working fine for you or throwing the same error?

MovingGifts:
Same warning as @Palisand

spencercarli:
Yep this is my bad. #2527 should fix this

PR #2354, opened 2017-08-10
Fix ActiveScreen in custom navigation

JPeer264:
The way as it was before was good, but when there are another Navigations, such as TabNavigation, inside this TabNavigation. The entire nested navigation won't get rendered.

I was sitting here for hours in the source code, but this finally made the move.


alexkuc:
Thank you! You are a life saver! I had the same issue and haven't thought of ActiveScreen being the issue.

tobyf93:
OMG thank you @JPeer264.  I too was stuck on this issue for hours!

PR #2400, opened 2017-08-16
queued transitions

rpopovici:
<del>Currently, dispatch is broken because it's using setState internally and it's not exposing any way to treat asynchronicity. This PR is trying to address this issues and also it's doing the same thing for the Transitioner.</del>(due to redux integration this is handled by https://github.com/react-community/react-navigation/pull/1313) Transitions are not being properly queued when received in componentWillReceiveProps  which can cause sometime "shadow" rendering for some screens. Basically, some stalled screens will be remounted for a split second before they go away.
This PR is a continuation to my other PR about idempotent pushes https://github.com/react-community/react-navigation/pull/2334

<del>So you should be able to write both:
dispatch(action1)
dispatch(action2)
or
dispatch(action1, (state, prev, action) => {
dispatch(action2)
})
Both cases will be handled properly</del>

### DIFF
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -53,12 +53,7 @@ class Transitioner extends React.Component<*, Props, State> {
   _prevTransitionProps: ?NavigationTransitionProps;
   _transitionProps: NavigationTransitionProps;
   _isMounted: boolean;
-  _isTransitionRunning: boolean;
-  _queuedTransition: ?{
-    nextProps: Props,
-    nextScenes: Array<NavigationScene>,
-    indexHasChanged: boolean,
-  };
+  _queuedTransitions: Array<Props>;
 
   props: Props;
   state: State;
@@ -86,8 +81,7 @@ class Transitioner extends React.Component<*, Props, State> {
     this._prevTransitionProps = null;
     this._transitionProps = buildTransitionProps(props, this.state);
     this._isMounted = false;
-    this._isTransitionRunning = false;
-    this._queuedTransition = null;
+    this._queuedTransitions = [];
   }
 
   componentWillMount(): void {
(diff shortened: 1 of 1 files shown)


dantman:
You're changing transitioner behaviour, do you have an example to demonstrate how the change in transitioner behaviour works in all cases for your test plan?

From my understanding of what you've done with transition end is that the transitioner will now always wait for the previous transition to finish and then do the next.

So given this:
```js
navigator.navigate('IntermediateScreen');
navigator.navigate('DeeperScreen');
```

The transitioner will always visually show the app transitioning to `IntermediateScreen` and then when its finished displaying on the screen will then transition to `DeeperScreen`.

I would think that a developer writing that instead of writing something that waits for a transitionEnd event would want to transition directly to `DeeperScreen`. And I expect the double transition without any user interaction would be confusing and annoying to users (you double the standard transition time and make them wait).

And given `setState` followed by `navigate`, if setState were to have a transition such as closing an action menu, then the transitioner would wait for the animation of the menu closing to complete before navigating to the next screen instead of going directly to the next screen.

dantman:
Side note, this is a breaking change. `dispatch` now returns `undefined` instead of returning a boolean indicating whether the action was valid (ie: an invalid action will now be silently ignored instead of being a handleable error).

rpopovici:
@dantman my intention was not to deliver batched navigation transitions. This is not working any way with the current implementation. I was trying to fix the existing implementation which was queuing one transition already btw, but this was happening without properly waiting for animation(async) and setState(async) to complete. 

Batched navigation transitions are possible right now only with single navigate action + child action
or via reset with multiple actions as parameter

Your example:
```
navigator.navigate('IntermediateScreen');
navigator.navigate('DeeperScreen');
```
should not work as expected any way since you are not properly dispatching stuff. The internal state will be messed up anyway. What happens, for example, if you are trying to call navigation.setParams() in IntermediateScreen componentDidMount? Can you guarantee that the second call to navigator.navigate('DeeperScreen'); will not going to mess up the prev state?

Anyway, the reason I was toying with this code is because with the current implementation  
stalled screens sometimes are being remounted for a split second when reset action is called.

rpopovici:
@dantman  Nice catch about dispatch returning undefined. I have to fix that

dantman:
@rpopovici That is impossible.

That breaking change is the reason that my change works the exact way that it does. I'm working on a long reply to you in that PR that might help explain the complexities and interactions going on here.

rpopovici:
It can return a promise or add a callback..

dantman:
Ok, though it'll still be a breaking API change that will require a major semver increment.

rpopovici:
IMO dispatch signature is wrong any way since it's exposing a sync interface but internally it's async

dantman:
@rpopovici Actually, now that I remember, you can't add a callback and change the return value to promise won't guarantee that the response from `this.props.navigation.{dispatch,navigation,...}` will always be a promise.

#1206 is the bug about multiple dispatches in the same tick, there was a *****lot***** of discussion about possible solutions that lead up to PR #1313.

One of those limiting factors that came up during discussion is that `dispatch` isn't a react-native interface. It's based off the redux dispatch method and react-navigation explicitly promotes replacing the internal dispatcher with the Redux dispatcher when using redux.
https://reactnavigation.org/docs/guides/redux

rpopovici:
@dantman thanks for the explanation. I am not using redux :) and could not foresee this use case

ericvicenti:
At the moment I think we need to focus on making the transitioner simpler, not more complicated. Once it is a bit easier to understand, I'd love to support advanced features like this. If anybody wants to help rewrite Transitioner, ping me and I'll help you get started- its a pretty isolated project.

PR #2415, opened 2017-08-18
prevent error `RawText "" must be wrapped in an explicit <Text> compo…

rohozhnikoff:
prevent error `RawText "" must be wrapped in an explicit <Text> component`, when title for some reason is empty string

this react-native error has no adequate stack, and i waste 4 hours for juggling pieces, until i get the idea to trace error by myself and finded this bug


rohozhnikoff:
for javascript, empty string is Falsy and interpreter return it (last statement) from expression
but JSX send it to View, like a children, which cause this nasty bug

src/views/Header/HeaderBackButton.js
```diff
@@ -97,7 +97,7 @@ class HeaderBackButton extends React.PureComponent<DefaultProps, Props, State> {
             source={asset}
           />
           {Platform.OS === 'ios' &&
-            title &&
+            typeof title === 'string' &&
```
matthamil:
```js
typeof '' === 'string' // true
```

This will still render the `<Text>` when `title === ''`.

rohozhnikoff: i think this behavior is correct, and more perfomant on blinking

matthamil: Sorry for the delay on this-- looking back at this again, would it not be better to check `typeof backButtonTitle` since that is what is being rendered?

wbercx:
Thanks for this. I was getting pretty frustrated trying to find out what was going on here.

Making the error go away is one thing, but it would make sense to me if the back button just said `Back` if your previous screen had an empty title. That way you don't end up with just the icon.

Thoughts?

kelset:
Thanks @rohozhnikoff ! And thanks @vonovak for reviewing!

rohozhnikoff:
tnx guys )

react-native decline my pull-request with general improves of this bugs
they move this error to Native side, and i dont know how to fix it ((

PR #2441, opened 2017-08-22
Accessibility labels can be added to TabNavigator

momoDeako:
on both iOS and Android

Added title and accessibilityLabel to TypeDefinition

**Test plan (required)**
Created two tabNavigator and give title and header to one of them(It's called Recent in the example), and title, header and accessibilityLabel to another one(It's called All in the example). 

Result: 
App works just fine, and users are able to see accessibilityLabel using different inspectors. Please see attachment for more details. 

iOS Accessibility Inspector
screen shot 2017-08-21 at 4 06 15 pm

Android uiautomatorviewer
screen shot 2017-08-21 at 4 07 56 pm

Sample Code(Same for both iOS and Android): 
screen shot 2017-08-21 at 4 22 21 pm

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -130,6 +130,14 @@ export type NavigationLeafRoute = {
    * e.g. `{ car_id: 123 }` in a route that displays a car.
    */
   params?: NavigationParams,
+  /**
+   * Generic title that can be used as a fallback for accessibilityLabel.
+   */
+  title?: string,
+  /**
+   * A label use by VoiceOver to identify the element.
+   */
+  accessibilityLabel?: string,
 };
 
 export type NavigationStateRoute = {
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -135,7 +135,7 @@ export type NavigationLeafRoute = {
    */
   title?: string,
   /**
-   * A label use by VoiceOver to identify the element.
+   * A label used by VoiceOver to identify the element.
    */
   accessibilityLabel?: string,
 };
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -64,7 +64,7 @@ export default (
         const routes = order.map((routeName: string) => {
           let accessibilityLabel;
           if (
-            [routeName].navigationOptions &&
+            routeConfigs[routeName].navigationOptions &&
             routeConfigs[routeName].navigationOptions.accessibilityLabel
           ) {
             accessibilityLabel =
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -64,11 +64,10 @@ export default (
         const routes = order.map((routeName: string) => {
           let accessibilityLabel;
           if (
-            routeConfigs[routeName].screen &&
-            routeConfigs[routeName].screen.navigationOptions
+            routeConfigs[routeName].navigationOptions
           ) {
             accessibilityLabel =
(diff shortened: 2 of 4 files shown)


src/TypeDefinition.js
```diff
@@ -130,6 +130,14 @@ export type NavigationLeafRoute = {
    * e.g. `{ car_id: 123 }` in a route that displays a car.
    */
   params?: NavigationParams,
+  /**
+   * Generic title that can be used as a fallback for accessibilityLabel.
+   */
+  title?: string,
+  /**
+   * A label use by VoiceOver to identify the element.
```
matthamil: nit: used

src/routers/TabRouter.js
```diff
@@ -62,6 +62,15 @@ export default (
       let state = inputState;
       if (!state) {
         const routes = order.map((routeName: string) => {
+          let accessibilityLabel;
```
matthamil: Is there any way to change the accessibility label programmatically at run-time? It looks like the accessibilityLabels are set when the initial tab routes are set up.

momoDeako: Hey Matt! Thank you for reviewing this for me! 
So here is my thought process about this part. I noticed iOS and Android implemented this a bit differently. When we create a tab for Android, it create the tab using [TabBarTop](https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabBarTop.js), and TabBarTop uses [TabBar](https://github.com/react-native-community/react-native-tab-view/blob/master/src/TabBar.js) from [react-native-tab-view](https://github.com/react-native-community/react-native-tab-view) library. So that for achieving your suggestion, some changes to react-native-tab-view library is required. 
Also,  for iOS react-navigation did the magic in [TabBarButtom](https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabBarBottom.js). 
What's your suggestion?

momoDeako:
@matthamil  I see what you meant now. accessibilityLabel should be defined when the page is created not when it get rendered. 

Working on it now.

src/views/TabView/TabBarBottom.js
```diff
@@ -153,10 +153,13 @@ export default class TabBarBottom extends PureComponent<
             outputRange: (outputRange: Array<string>),
           });
           const justifyContent = this.props.showIcon ? 'flex-end' : 'center';
+          const accessibilityLabel = route.accessibilityLabel || route.title;
           return (
             <TouchableWithoutFeedback
               key={route.key}
               onPress={() => jumpToIndex(index)}
+              accessible={true}
+              accessibilityLabel={accessibilityLabel}
```
matthamil: Is there ever a situation where the route title would be undefined? This would negate the `accessible={true}` if there is an `undefined` label.

matthamil: Perhaps the `accessible` boolean should only be `true` when a title OR label is detected.

spencercarli:
Going to close this because accessibility label & more was added in #2684

tommeier:
@spencercarli but #2684 is only for tab bar bottom? not for top?

spencercarli:
@tommeier hmm good point. This one is also only for tab bar bottom but I would accept a PR to add it to tab bar top

rcorrie:
@spencercarli Seems like this stuff isn't supported in the latest version? Or am I missing something?

spencercarli:
I'm not sure, I'm no longer actively helping in the project.

Issue #2410, opened 2017-08-17
[Question] Avoid status bar overlap on all screens


James2516:
I want all screens on my app to appear below the status bar on both iOS and Android. Currently I have to use `headerStyle: { marginTop: 24 }` on all the stack navigators.

Is there a way to do this globally? Where is the appropriate top level component to add the `StatusBar` in a Redux app? (eg. which part of https://github.com/react-community/react-navigation/tree/master/examples/ReduxExample)?

Noitidart:
I do the same thing. Note that on iOS marginTop should be 20. On Android you should use `StatusBar.currentHeight` instead of hard coding 24 (it usually is 24 though).

riwu:
Where do u wrap your whole app? https://github.com/react-community/react-navigation/tree/master/examples/ReduxExample Wrapping it at `<Provider>` does not work.

Noitidart:
Ok so I lied, I don't really wrap my whole app. Each screen/card has a `<Header>` component. I just gave that `<Header>` a extra padding at the top so it shows nicely behind the status bar.

ro-savage:
Shouldn't this be defaulted into the app? 

That the header always appears below the statusbar (unless specified otherwise).

As @efkan pointed out, updating [line 24](https://github.com/react-community/react-navigation/blob/b310b153fd76edd27ecbf4d6d2d71ead1d33a530/src/views/Header.js#L37) in header.js fixes this.

spencercarli:
Hi! In an effort to get the hundreds of issues on this repo under control I'm closing issues tagged as questions. Please don't take this personally - it's simply something we need to do to get the issues to a manageable point. If you still have a question I encourage you to re-read the docs, ask on StackOverflow, or ask on the #react-navigation Reactiflux channel. Thank you!

Issue #1168, opened 2017-04-22
statusbar covering the tabs in expo app on android


manjeets12:
I am just started to work with expo and just installed [example expo app](https://expo.io/@community/react-navigation-playground) on my android 5.1. In that my statusbar is covering some parts of tabs and I am not sure how to resolve this

expo/react-navigation example screenshots
![screenshot_2017-04-22-21-58-16 1](https://cloud.githubusercontent.com/assets/17762875/25306275/76c19160-27a7-11e7-9a9c-fd438db21a82.png)
![screenshot_2017-04-22-21-58-37 1](https://cloud.githubusercontent.com/assets/17762875/25306277/81ee0104-27a7-11e7-8604-c950650f7927.png)

and now here is screenshot from expo/ex-navigation example app

![screenshot_2017-04-22-22-03-02 1](https://cloud.githubusercontent.com/assets/17762875/25306292/c5eb6964-27a7-11e7-95ae-867dcb7e1e91.png)
![screenshot_2017-04-22-22-03-09 1](https://cloud.githubusercontent.com/assets/17762875/25306303/e157c990-27a7-11e7-8c32-18d3d704c76a.png)

and see now how good it is looking, what changes should I do to make it look like same

dantman:
Expo implicitly sets the StatusBar to translucent. You'll need to add `paddingTop` to `headerStyle` on Android to accommodate the status bar size.

k15a:
There is [`Constants.statusBarHeight`](https://docs.expo.io/versions/v16.0.0/sdk/constants.html#expoconstantsstatusbarheight) from the `expo` package which will be the height of the statusBar.

grabbou:
This is a duplicate of #12

ro-savage:
Code sample, for anyone looking for it:

```js
const App = StackNavigator({
  Home: { screen: HomeScreen },
  About: { screen: AboutScreen },
  }, {
    navigationOptions: {headerStyle: { marginTop: Expo.Constants.statusBarHeight } }
  })
```

noamnelke:
Small tweak to @ro-savage's code sample, so it doesn't break ios:

```javascript
const App = StackNavigator({
  Home: { screen: HomeScreen },
  About: { screen: AboutScreen },
}, {
  navigationOptions: { headerStyle: { marginTop: (Platform.OS === 'ios') ? 0 : Expo.Constants.statusBarHeight } }
})
```

fc:
[...]

Issue #12, opened 2017-01-27
Play nicely with translucent StatusBar on Android 


brentvatne:
In Expo we use a translucent StatusBar by default on Android, and some normal React Native projects may do this as well because it makes it easier to reason about cross-platform since the iOS StatusBar is always translucent.

We may need to upstream a constant to React Native that we can use to check if the StatusBar is translucent, and then add any necessary paddingTop to the navigation bar when it is.

datwheat:
I just wanna +1 this because I just used react-navigation for the first time (with Exponent) and realized that I needed to handle android statusbar paddingTop unlike when I use ex-navigation where it handles it for me. It would be really nice for new users to get that functionality out of the box 👍🏿

lyahdav:
@datwheat I ran into this same issue with react-navigation and Exponent. Can you provide an example of your code that fixed this? I'm not understanding where to add the paddingTop. I basically followed the instructions [here](https://reactnavigation.org/docs/intro/) to use react-navigation.

kingr:
@lyahdav Here's a few ways you can do it.

**TabNavigator**
Pass a second argument to the a Navigator with a custom config containing styles.
[Docs](https://reactnavigation.org/docs/navigators/tab#TabNavigatorConfig)

`const BasicApp = TabNavigator({
   Home : { screen: Home },
   Manufacturers : { screen: Manufacturers },
   },
   {
    tabBarOptions: {
        style: {
            backgroundColor: 'orange',
            marginTop: 24
        },
    }
});`

**StackNavigator**
Inside a Screen add a header config containing your styles to navigationOptions.
[Docs](https://reactnavigation.org/docs/navigators/stack#Screen-Navigation-Options)

`
static navigationOptions = {
        title: 'something else',
        header: {
            style: {
                backgroundColor: 'orange',
                marginTop: 24
            }
        }
    }
`

**DrawerNavigator**
No example, but it also takes a second argument with a config where you can add styles.
[Docs](https://reactnavigation.org/docs/navigators/drawer#DrawerNavigatorConfig)

Ehesp:
The way I've been going for now is to have the statusBar as always transparent, with a color of transparent:

```
<StatusBar
  backgroundColor={'transparent'}
  translucent
/>
```

In the styles for the navigators, add the following. It allows for a different header color (Android):

```
  tabBarOptions: {
    style: {
      backgroundColor: 'orange', // Main color
      borderTopWidth: 24,
[...]

PR #2446, opened 2017-08-22
Set STATUSBAR_HEIGHT for android.

### LINKED ISSUE #2164: STATUSBAR_HEIGHT for Android
I use Android Emulator (Nexus 5X with API 23) and my component is seemed as swiped to below for 25 px.

If I'm right, [at this line](https://github.com/react-community/react-navigation/blob/b310b153fd76edd27ecbf4d6d2d71ead1d33a530/src/views/Header.js#L37) `25` should be assigned to `STATUSBAR_HEIGHT` value for Android.

gunn:
This is to fix the `Header` component not taking the status bar into account on android:

android_header

Currently you have to patch it manually:

```javascript
if (Platform.OS=="android") {
  headerStyles.paddingTop = Constants.statusBarHeight
  headerStyles.height     = Constants.statusBarHeight + 56
}
```

This fixes #2164 and #2410.

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -4,7 +4,7 @@
 
 import React from 'react';
 
-import { Animated, Platform, StyleSheet, View } from 'react-native';
+import { Animated, Platform, StyleSheet, View, StatusBar } from 'react-native';
 
 import HeaderTitle from './HeaderTitle';
 import HeaderBackButton from './HeaderBackButton';
@@ -34,7 +34,7 @@ type HeaderState = {
 };
 
 const APPBAR_HEIGHT = Platform.OS === 'ios' ? 44 : 56;
-const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
+const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : StatusBar.currentHeight;
 const TITLE_OFFSET = Platform.OS === 'ios' ? 70 : 56;
 
 class Header extends React.PureComponent<void, HeaderProps, HeaderState> {


vonovak:
I'm actually not seeing this. are you using RN or expo?

James2516:
I can confirm this problem exists on Expo (with create-react-native-app)

gunn:
@vonovak Expo via create-react-native-app. The screenshot is actually from the official demo - https://expo.io/@react-navigation/NavigationPlayground

powdahound:
See also: #1168, #12

src/views/Header/Header.js
```diff
@@ -34,7 +34,7 @@ type HeaderState = {
 };
 
 const APPBAR_HEIGHT = Platform.OS === 'ios' ? 44 : 56;
-const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
+const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : StatusBar.currentHeight;
```
dantman: Please add a `|| 0` to the end to ensure that `STATUSBAR_HEIGHT` defaults to `0` instead of `undefined`. RN defines `StatusBar.currentHeight` as "Android only" and `react-navigation` runs on platforms that are not iOS or Android such as the web.

dantman:
I should also note that this will be a breaking change, it'll break all react-native apps that are not built with Expo and do not set a translucent StatusBar.

translucent StatusBar is not a default in react-native, it's Expo specifically that defaults the StatusBar on Android to translucent. So all normal apps previously built with `react-native init` and react-navigation will now have padding added to the top of the header.

moriyoshi:
This appearance is caused by the attribute `android:windowTranslucentStatus` being set in styles.xml as mentioned and I don't either think it's a good idea to add the margin for the status bar to the header's height of the header regardless of how the app is built out.

P.S. the attribute is also the cause of [this](https://forums.expo.io/t/android-theme-causes-window-not-to-resize-when-keyboard-is-open/1582) (by using `KeyboardAvoidingView` you can avoid the issue tentatively on expo but it gets wrong on the native build then because the component is wrongly designed.)  My best guess is the attribute will be removed from the default in any way.

src/views/Header/Header.js
dantman: Also this isn't actually correct.

On Platform.Version < 19 (api levels below 19) the relevant status bar height is always 0 because the ability to make status bars translucent was added in API level 19. (react-native supports API level 18).

dantman: Which itself is also incorrect in this context. Because while Android introduced the ability to make status bars translucent in KitKat (API level 19) using an xml declaration ([the way Expo does it](https://github.com/expo/expo/blob/9a9dfb4103cb0cd5bfff8a97710350321655059f/android/app/src/main/res/values/styles.xml)) they didn't add the ability to make it translucent dynamically (the way `<StatusBar />` and most react-native apps that are translucent do it) until Lollipop (API level 21).

So if you are using Expo translucency doesn't apply `< 19`, if you are using `<StatusBar translucent />` it doesn't apply `< 21`.

outaTiME:
Hi guys, any update for this feature? it will awesome if this patch must be include the same logic for the drawer ...

work as expected in StackNavigator:

screen shot 2017-10-13 at 12 21 49

but doesn't with the DrawerNavigator:

screen shot 2017-10-13 at 12 21 51

vonovak:
here's how ex-navigation handles this, if any expo user wants to take care of this: https://github.com/expo/ex-navigation/blob/3166a2a09e5bb50e30e4aeafed1dd68386a84205/src/ExNavigationBar.js#L30

outaTiME:
Finally i use the same style for StackNavigator and DrawerNavigator, here the snippet code:

´´´js
import Expo from 'expo';
import { StyleSheet, View, Platform } from 'react-native';
import { DrawerNavigator, DrawerItems, StackNavigator } from 'react-navigation';

const UserScreen = StackNavigator(
  {
    Main: {
      screen: MainUserScreen,
    },
  },
  {
    navigationOptions: ({ navigation, screenProps }) => ({
      headerStyle: [{
        backgroundColor: '#4DBCE9',
      }, styles.navigationPadding],
      headerTitleStyle: {
        color: '#fff'
      },
    })
  }
);

const MainScreen = DrawerNavigator({
  User: {
    screen: UserScreen,
  },
}, {
  contentComponent: props => (
    <View style={[{ flex: 1 }, styles.navigationPadding]}>
      <DrawerItems {...props} />
    </View>
  ),
});

  navigationPadding: {
    ...Platform.select({
      ios: {
        // pass
      },
      android: {
        paddingTop: Expo.Constants.statusBarHeight,
        height: Expo.Constants.statusBarHeight + 56,
      },
    })
  }
});
´´´

kelset:
@davepack @brentvatne is this PR related to what you were saying about "improving" status bar?

brentvatne:
i resolved this in https://github.com/react-navigation/react-navigation/releases/tag/v1.0.0-beta.26

PR #2491, opened 2017-08-29
Provide intro for create-react-native-app and react-native init

probeadd:
Sorry for creating new PR again, I accidentally deleted my fork. So, here I fixed everything that needed. Thank you.


probeadd:
@matthamil @nonameolsson Please kindly check my PR. Thank you.

docs/guides/Guide-Intro.md
```diff
@@ -20,6 +20,20 @@ react-native run-android # or:
...
+cd SimpleApp
+
+# Install the latest version of react-navigation from npm
+npm install --save react-navigation
+
+# Run the new app
+npm start
+
+# This will start a development server for you and print a QR code in your terminal.
+
```
spencercarli: Make sure to close this code block with. Would also be helpful to give htem some more context around how to verify it's running - such as

"This will start a development server for you and print a QR code in your terminal. Scan the QR code using the Expo app on your iOS or Android device."

docs/guides/Guide-Intro.md
```diff
@@ -55,6 +69,7 @@ const SimpleApp = StackNavigator({
   Home: { screen: HomeScreen },
 });
 
+// if you are using create-react-native-app you don't need this line
```
spencercarli: If they're not using `AppRegistery` then they have to `export default` the component. We can just add that into the example above. 

However, I feel like it's confusing to comment out a line when going through a getting started guide. Makes me want to split the tutorials up a bit 🤔 

That shouldn't be a blocker for right now though.

matthamil: FWIW React Native does this in their own documentation: https://facebook.github.io/react-native/docs/tutorial.html#content

spencercarli: Ah hadn't realized that! Then the only thing to do here is add that `export default`

spencercarli:
@probeadd looping back around with you regarding the requested changes - would love to merge this. Thanks! 😄

PR #2492, opened 2017-08-29
Feature/toggle

GantMan:
Every drawer menu needs a toggle. Especially when navigating to a drawer 'DrawerOpen' twice can cause bugs/issues.  **To solve this currently** everyone needs to know about drawer state and structure, which causes code to look like so:

```js
    headerLeft: <Text onPress={() => {
      if (navigation.state.index === 0) {
        navigation.navigate('DrawerOpen')
      } else {
        navigation.navigate('DrawerClose')
      }
    }}>
      Toggle Drawer
    </Text>
```

This is not readable or future-proof.  Instead toggle should be a first class citizen of drawers.
```js
    headerLeft: <Text onPress={() => navigation.navigate('DrawerToggle')}/>
        Toggle Drawer
    </Text>
```

That is what this PR includes.  Also updates docs.

### DIFF
--- a/src/views/Drawer/DrawerView.js
+++ b/src/views/Drawer/DrawerView.js
@@ -63,6 +63,12 @@ export default class DrawerView<T: *> extends PureComponent<void, Props, void> {
       const { routes, index } = nextProps.navigation.state;
       if (routes[index].routeName === 'DrawerOpen') {
         this._drawer.openDrawer();
+      } else if (routes[index].routeName === 'DrawerToggle') {
+        if (this._drawer.state.drawerShown) {
+          this.props.navigation.navigate('DrawerClose');
+        } else {
+          this.props.navigation.navigate('DrawerOpen');
+        }
       } else {
         this._drawer.closeDrawer();
       }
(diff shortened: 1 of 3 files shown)


src/views/Drawer/DrawerView.js
```diff
@@ -63,6 +63,12 @@ export default class DrawerView<T: *> extends PureComponent<void, Props, void> {
       const { routes, index } = nextProps.navigation.state;
       if (routes[index].routeName === 'DrawerOpen') {
         this._drawer.openDrawer();
+      } else if (routes[index].routeName === 'DrawerToggle') {
+        if (this._drawer.state.drawerShown) {
+          this.props.navigation.navigate('DrawerClose');
```
spencercarli: To keep it consistent with DrawerOpen/DrawerClose how about using `this._drawer.closeDrawer();` and `this._drawer.openDrawer();` instead?

davepack: Sorry, missed this before approving

GantMan: @spencercarli - using `this._drawer.closeDrawer();` here won't update the state, or fire off the events. 

 If we switch this out it will animate into place, not fire any events, and then suddenly hide due to not being in that new state.

This way solves it.

PR #2498, opened 2017-08-30
getComponentForState made more generic

Rabursky:
It's just one line...
routeName (and what component is being rendered) should depend directly and only on the navigation state.


matthamil:
Can you explain the motivation behind this change?

Rabursky:
@matthamil 
`order` const is being calculated only once when the router is created.
Router should render components based on its current state and not on preconfigured one from the startup.
In my project I am using TabBar as child navigator of another navigator. Parent navigator dynamically change order and availability of tabs. It doesn't work because of this one line, I had to monkey-patch the router.

Hope it explains my motivation a bit better...

Rabursky:
@matthamil Did previous explanation help?

satya164:
It makes sense to me that we should use the passed state to determine the component. However other methods are using `order` too. So we would need to change it.

Issue #1466, opened 2017-05-11
Improve performance for React Navigation


satya164:
I have heard several people say that the perf on low-end Android devices is bad. And it's not quite as fast as ex-navigation used to be in my mid-end Android device either.

Some easy to fix issues will be:
1. Identify unnecessary re-renders and optimize them
2. Cache stuff like navigation options

aqnaruto:
ex-navigation is much better than this one .

AlanFoster:
@satya164 As a new comer to react-native, how can someone contribute to this issue?

> 1. Identify unnecessary re-renders and optimize them

What would be the best approach to investigate this further? As far as I can tell for react-native the Perf tools still work, but would we need to use systrace etc for this particular issue?

Out of curiosity; Are there any tools for showing react-native redraws visually, such as Chrome's functionality or the React Developer tools?

aqnaruto:
ok,it not not quite as fast as ex-navigation used to be, and it is still not surport goback-gesture in android

AlanFoster:
@satya164 I've been looking into this more, viewing react native react videos etc. I'm more aware of systrace now. 

Unfortunately it's not well documented how to compile the react native ios simulator yourself to enable the Javascript flame graph profiler.  Do you have any context on this @satya164? I'd love to take a deeper look into this

AlanFoster:
### Identify unnecessary re-renders and optimize them

react-navigation has some components that bind a new function on each render, which doesn't satisfy `PureComponent`. I have put a PR up to fix one of these issues https://github.com/react-community/react-navigation/pull/1913

The most expensive renders seem to be related to the Card/Container components, I believe this is due to the inline styles that are being used to calculate the position:

![image](https://user-images.githubusercontent.com/1271782/27265052-0d7f5460-5486-11e7-8d8a-54f4d3da3753.png)

### Cache stuff like navigation options

Caching navigation options would be good; Particularly in the redux examples which create new goBack/navigate/setParam functions on each state change. I was able to partially cheat with a small shim locally - but I'm not sure what the best library level API Is for this. Perhaps a higher order component, something similar to recompose?

For the mean-time I had just gone with a small shim based on the default redux example:

```
const navigationCacher = function () {
  let _helpers = undefined;
  const cachedFunctions = {
    goBack(...args) {
      return _helpers.goBack.apply(undefined, args);
    },

    navigate(...args) {
      return _helpers.navigate.apply(undefined, args);
    },

    setParams(...args) {
[...]

Issue #360, opened 2017-02-16
Disable logging?


sankhadeeproy007:
I just followed the example from the docs and noticed some lag when rendering heavy components on Android. On enabling debugger, it was somewhat faster as the Navigation actions were logged on chrome. Question is, how do I disable logging from the router? I could not find anything regarding that in the docs (or maybe I didn't know where to look for it).

agrcrobles:
are you using a redux middleware for logging?

ElliottJRo:
I think @sankhadeeproy007 is talking about the dispatch logging. If thats the case, I don't think there is a way to disable it atm. You can disable it yourself by modifying local version in `/src/createNavigationContainer.js`

satya164:
I think we should use a custom function for logging so we can enable it for development and not for library consumers. PR is welcome.

ericvicenti:
This behavior lives in [`createNavigationContainer`](https://github.com/react-community/react-navigation/blob/a7b124305393a8e392621a32e49fa96a3a14c6c9/src/createNavigationContainer.js). It also logs URI requests

I think we should add something like `onNavStateChange` to `containerOptions`, which is currently documented in the [wrong](https://github.com/react-community/react-navigation/issues/403) [place](https://reactnavigation.org/docs/navigators/custom).

The default behavior of `onNavStateChange` would literally just be the logging function, so if you want to disable logging you could simply override it with an empty function.

ARMATAV:
+1 Please add an option for disabling the logging

tomykho:
+ 1

joemckie:
+1, related to [this issue](https://github.com/evgenyrodionov/redux-logger/issues/32). I notice a definite lag in some screens due to logging.

csotiriou:
+1 for providing a way to disable logging.

React native's log is extremely slow. We are noticing 7 seconds (having debugging disabled!) when we have a screen with 30 items inside, due to the navigator displaying the contents of the state, which has a Realmjs object with 30 items inside it.

geekvijay:
+1 for Disabling Logging. it takes one at least a second to navigate to another screen and resulting in bad UX.

Is anyone figured out any patch for this ?

adi69:
`RCTSetLogThreshold(RCTLogLevelError);` will disable all the logging except in case of Error or Fatal Exception.

tomykho:
+1

iboxgithub:
@ericvicenti  Don't you think the logging should be disabled by default?

[...]

PR #2501, opened 2017-08-31
Use react-native-config to read the REACT_NAV_LOGGING environment variable when running in React Native

levibuzolic:
## Issue

Navigation state changes aren't logged when running on native.

## Cause

The conditional for enabling logging was using `process.env.REACT_NAV_LOGGING` which can't be set on native platforms. `process.env` will only ever contain `NODE_ENV`.

## Fix

The community accepted method of using environment variables in React Native is to use [`react-native-config`](https://github.com/luggit/react-native-config). This pull request adds `react-native-config` as a dependency and uses it only on native platforms, web will still use `process.env`.

I've also updated the docs to give some indication that this environment variable exists and needs to be set in order to enable logging.

### DIFF
--- a/docs/api/navigators/Navigators.md
+++ b/docs/api/navigators/Navigators.md
@@ -22,7 +22,7 @@ To learn how to create screens, read about:
 
 ### Calling Navigate on Top Level Component
 
-In case you want to use Navigator from the same level you declare it you can use react's [`ref`](https://facebook.github.io/react/docs/refs-and-the-dom.html#the-ref-callback-attribute) option:  
+In case you want to use Navigator from the same level you declare it you can use react's [`ref`](https://facebook.github.io/react/docs/refs-and-the-dom.html#the-ref-callback-attribute) option:
 ```js
 const AppNavigator = StackNavigator(SomeAppRouteConfigs);
 
@@ -38,7 +38,7 @@ class App extends React.Component {
   }
 }
 ```
-Please notice that this solution should only be used on the top level navigator.  
+Please notice that this solution should only be used on the top level navigator.
 
 ## Navigation Containers
 
@@ -48,11 +48,11 @@ When rendering one of the included navigators, the navigation prop is optional. 
 
 For the purpose of convenience, the built-in navigators have this ability because behind the scenes they use `createNavigationContainer`. Usually, navigators require a navigation prop in order to function.
 
-Top-level navigators accept the following props:  
+Top-level navigators accept the following props:
 
 ### `onNavigationStateChange(prevState, newState, action)`
 
-Function that gets called every time navigation state managed by the navigator changes. It receives the previous state, the new state of the navigation and the action that issued state change. By default it prints state changes to the console.
+Function that gets called every time navigation state managed by the navigator changes. It receives the previous state, the new state of the navigation and the action that issued state change. By default it prints state changes to the console if the `REACT_NAV_LOGGING` environment variable is set.
 
 ### `uriPrefix`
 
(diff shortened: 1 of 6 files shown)


matthamil:
I'm not sure this is a dependency that we want to add to react-navigation. This lib aims to be a JS-only solution to navigation for React Native. react-native-config introduces native dependencies and installation instructions that increase the barrier to entry. This native module dependency also breaks any examples that we have working with [Expo](https://expo.io/).

You can pass an `onNavigationStateChange` prop to that navigator instead of modifying the source directly. This should resolve your problem.

levibuzolic:
Thanks for the response @matthamil, I can completely understand the reasons for not adding a native dependency like this.

Unfortunately passing an `onNavigationStateChange` prop isn't particularly feasible with nested navigators as you'd need to pass that prop to every navigator -- not impossible, but certainly not ideal.

Given the docs say that logging should be the default action, I think it should be enabled by default and be an opt-out (in development). Would we be able to swap `process.env.REACT_NAV_LOGGING` for `process.env.NODE_ENV === 'development'` so that logging is enabled by default.

I recall seeing some issues around some people complaining that this slows down native devices, so I could possibly simplify/reduce the data that is logged to some simple strings rather than the entire state and action objects.

Thoughts?

brentvatne:
@levibuzolic - I think you might be misunderstanding the environment that this JS is executed in: it runs inside JavaScript Core on your phone. So, there is no `process.env.NODE_ENV` because there is no node. Environment variables don't make much sense in React Native because the execution environment is the user's phone -- think of the React Native environment more like a web browser tab than a node process.

As for logging being default, in my opinion, I don't think it's a good idea -- imagine that every library did this, your logs would be a waterfall of spam.

> Unfortunately passing an onNavigationStateChange prop isn't particularly feasible with nested navigators as you'd need to pass that prop to every navigator -- not impossible, but certainly not ideal.

Make a wrapper for the navigators that you use and pass through the prop automatically, then use the wrapper everywhere :) You could alternatively just expose some function on react-navigation to set the logging state and perhaps the maintainers would be more likely to accept such a PR. Maybe something like this:

```
import { configureLogs } from 'react-navigation';

configureLogs({
  transitions: true,
  logFunction: () => {},
});
```

levibuzolic:
> @brentvatne: I think you might be misunderstanding the environment that this JS is executed in: it runs inside JavaScript Core on your phone. So, there is no process.env.NODE_ENV because there is no node. Environment variables don't make much sense in React Native because the execution environment is the user's phone -- think of the React Native environment more like a web browser tab than a node process.

`NODE_ENV` does exist actually, it's the only ENV that React Native sets, this allows `invariant` to work correctly in development.

Looking over the issues about logging -- #360, #1466 and 1919bff719908a885610e5005a9a594e473ce7fd there seems to be significant demand for it to be off by default. Which I can completely understand.

I think @brentvatne's suggestion for a global state/configuration is worthwhile, I'll take a look at implementing something along those lines.

levibuzolic:
Thinking a little more, how do you feel about having a global `onNavigationStateChange` rather than something specific to configuring logging?

I feel like it's likely that users of libraries like Redux might appreciate being able to globally set `onNavigationStateChange` so they can also handle nested navigators.

brentvatne:
sorry my bad, I forgot it was polyfilled here: https://github.com/facebook/react-native/blob/3a031cc93a87834c3b63f96f12a8b3cdb94d3eb6/Libraries/Core/InitializeCore.js#L85-L90

not to be confused however with the environment being anything like node.

PR #2520, opened 2017-09-03
Fix animation (#1493)

94cstyles:
Fix animation display exception problem。#1493

- before the change
   before the change
- after the change
   after the change

While using the [react-navigation-addons](https://github.com/satya164/react-navigation-addons/pull/19) to enhance goBack，support routeName and step。

example


matthamil:
Hi! Thanks for the PR! Can you explain the motivation behind this change?

94cstyles:
Sorry, forgot to fill in the instructions。Hope to be able to quickly merge into the project,Because this PR to solve the problem of animation, This question plagued a lot of people。

reyalpsirc:
I just copied your `CardStackStyleInterpolator.js` in my project and used it as the `transitionConfig.screenInterpolator` and it works. However, the header still has the problem when using the go back to a previous route. Does `HeaderStyleInterpolator.js` also needs changes?

94cstyles:
Yes, `HeaderStyleInterpolator.js` needs to be changed. Because my setting is `headerMode: 'screen'`, so no title animation found abnormal.

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -49,17 +72,21 @@ function forHorizontal(
...
+
+  const inputRange = [index - 1, index, index + 1];
   const outputRange = I18nManager.isRTL
     ? ([-width, 0, width * 0.3]: Array<number>)
     : ([width, 0, width * -0.3]: Array<number>);
 
-  // Add [index - 1, index - 0.99] to the interpolated opacity for screen transition.
-  // This makes the screen's shadow to disappear smoothly.
-  const opacity = position.interpolate({
+  let opacity = {
```
matthamil: To prevent reassignment or mutation of opacity, could the logic that determine's `opacity`'s value be its own function?

matthamil: Also I noticed that you removed the comments, was there motivation behind this?

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -68,17 +95,33 @@ function forHorizontal(
...
+  };
+
+  if (isBack) {
+    if (index === lastIndex) {
+      translateX.inputRange[0] = activeIndex;
+      return {
+        transform: [{ translateX: position.interpolate(translateX) }],
+      };
+    } else if (index === activeIndex) {
+      translateX.inputRange[2] = lastIndex;
```
matthamil: Can you refactor this to use fewer magic numbers?

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -20,6 +23,26 @@ import type { NavigationSceneRendererProps } from '../../TypeDefinition';
  */
 
 /**
+ * Format the sequence of the scenes
+ * @param scenes
+ * @param index
+ * @returns {NavigationScene}
+ */
+function formatScenes(scenes: Array<NavigationScene>, index: number) {
```
matthamil: nit: JSDoc says this returns `NavigationScene`, it looks like you are returning `Array<NavigationScene>`. I would also add the appropriate Flow return type.

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -20,6 +23,26 @@ import type { NavigationSceneRendererProps } from '../../TypeDefinition';
...
+  let scene = scenes[index];
+  if (!scene.isActive && index === scene.index) {
+    scenes = [].concat(scenes);
+    const activeIndex = scenes.findIndex(
+      (item: NavigationScene) => item.isActive
+    );
+    const activeScene = scenes[activeIndex];
+    scenes.splice(activeIndex, 1);
```
matthamil: Can this be refactored to use non-mutative `Array.prototype` methods?

94cstyles: Before that I had created a new array。 my initial intention was to modify the queue order of the new array ， so I used `splice` here without any clash. Why must we use non-mutative medthod.

src/views/CardStack/CardStackStyleInterpolator.js
94cstyles: The intention of this function is to correct the order of scenes, because in reset, there will be the same index of the scene, or the emergence of scene order reversed

src/views/CardStack/CardStackStyleInterpolator.js
94cstyles: scenes: [A , B , C , D]
D -> A , 
B and C Does not need to show, so modified `opacity = {opacity: 0}`

94cstyles: I just cancel the merge。

src/views/CardStack/CardStackStyleInterpolator.js
94cstyles: According to the original animation effect, there is no way to reduce the magic numbers.

94cstyles:
Rewrite the implementation of CardStack and Header animations
[Test Project](https://github.com/94cstyles/react-navigation-test)

src/views/CardStack/CardStackStyleInterpolator.js
matthamil: Can you add the appropriate flow return type to it? :)

tundak:
Good to see, when this will be part of reactnavigation ?

94cstyles:
@spencercarli Please merge the PR and make it part of `react-navigation`.

reyalpsirc:
@94cstyles While that doesn't happen, we just use your fork :)

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -53,30 +55,30 @@ function forHorizontal(
   }
 
   const index = scene.index;
-  const inputRange = [index - 1, index, index + 1];
-
   const width = layout.initWidth;
```
vonovak: nit: could you move this constant down to where it is actually being used? There are a couple more places in the code where you create a const and use it much later. Can you please move these assignments closer to where they are put into use? Other than that this looks good to me, so we'll merge. Thanks :)

94cstyles: I can not understand your intention, you said the need to modify the place, the original document is so written. Is this modification necessary?

vonovak: hi, what I have in mind is this: imagine you need some object on line 1000 of your code. You only need it in that one place. Where are you gonna create that object? - probably on line 999 because creating it on line 200 will be confusing for the reader. To give an example (read the comments):

```js
  const index = scene.index; // You think: okay, we now have a variable called index
  const interpolate = animatedInterpolate(props); // okay, here we call some function

  if (!interpolate) return { opacity: 0 }; // oh, if the function returned some falsy value, we return early. Why did we create the index variable?

    inputRange: ([interpolate.first, index, interpolate.last]: Array<number>),
    outputRange: ([0, 1, 1]: Array<number>),
  });
```
Do you see what I mean? It's just to improve readability, it will not affect the results. Can you do the changes? Thanks.

hparyani49:
@94cstyles i want to use your code for goBackTo(i.e goBack('routeName')) and to solve that animation issue also.

Can you help me how to use ur code until it is merged in official react-navigation repo

94cstyles:
@hparyani49 [example](https://github.com/94cstyles/react-navigation-test)
use my fork [react-navigation](https://github.com/94cstyles/react-navigation) and [react-navigation-addons](https://github.com/94cstyles/react-navigation-addons)

hparyani49:
Thank You @94cstyles for the example.

src/utils/animatedInterpolate.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * get the first and last animated interpolation
+ * When the return value is null, the scene should be set to invisible.
+ * @returns {{first: Number, last:Number} || null}
+ */
+function animatedInterpolate({ scene, scenes }) {
```
vonovak: hi @94cstyles - I'm sorry for delivering the review in chunks but I haven't had time to sit and do a proper review until now. I'd like to ask you again for some minor changes, only in this one file -
1 . please rename this function so it is very clear about what it does. `animatedInterpolate` doesn't convey much information.
2 . please rename the variables used in the body of this function. Right now there are variables `activeIndex`, `scenesActiveIndex`, `scenesCurrentIndex`. Some of these adjectives / names sound very similar (active, current, `activeIndex`, `scenesActiveIndex`) and people who read the code will not understand the difference (eg. `activeIndex` vs. `scenesActiveIndex` vs. `scenesCurrentIndex`). Could you change the names so that the reader doesn't get so easily confused?
3 . please add flow types

the body of the function could also be made nicer, by not mutating the `first` and `last` variables and rewriting the if, for example

```js
function getFirstAndLastSceneIndicesForInterpolationInputRange({ scene, scenes }) {
  const lastIndex = scenes.length - 1;
  const index = scene.index;
  const isBack = !scenes[lastIndex].isActive;
  if (!isBack) {
    return { first: index - 1, last: index + 1 };
  } else {
    // please rename the constants; you can even put this in a separate function if you like
    const activeScene = scenes.find(item => item.isActive);
    const scenesActiveIndex = scenes.findIndex(item => item === activeScene);
    const activeIndex = activeScene.index;
    const scenesCurrentIndex = scenes.findIndex(item => item === scene);

    if (scenesCurrentIndex === lastIndex) {
      return { first: activeIndex, last: index + 1 };
    } else if (index === activeIndex && scenesCurrentIndex === scenesActiveIndex) { // eslint-disable-line
      return { first: index - 1, last: lastIndex };
    } else if (scenesCurrentIndex > scenesActiveIndex || index === activeIndex) { // eslint-disable-line
      return null;
    }
  }
}
```

That's it! Thank you so much much for your work. As a side note - it seems like you've been using force push to keep all your work in a single commit. Please do not do this - it makes the review process harder by not seeing the changes you made after addressing the review requests. We can squash the commits into one before merging to the master. Thank you.

94cstyles: Thank you for your feedback, let me realize that my lack.

src/utils/getFirstAndLastSceneIndicesForInterpolationInputRange.js
```diff
@@ -0,0 +1,38 @@
...
+function getFirstAndLastSceneIndicesForInterpolationInputRange(
+  props: NavigationSceneRendererProps
+): SceneIndicesForInterpolationInputRange | null {
+  const { scene, scenes } = props;
+  const index = scene.index;
+  const lastSceneIndicesInQueue = scenes.length - 1
+  const isBack = !scenes[lastSceneIndicesInQueue].isActive
+
+  if (isBack) {
+    const currentSceneIndicesInQueue = scenes.findIndex((item: NavigationScene) => item === scene);
```
vonovak: hi! Thanks for being so quick! I'm really sorry, but the naming still has room for improvement. My main point was to remove possible confusion between *active* and *current*. Now we have `currentSceneIndicesInQueue` and `activeSceneIndicesInQueue`. The difference between these is still hard to tell. Moreover, you used 'indices', which is a plural and that suggests the variable is some sort of collection of values, but it is only a single number. Also, I'm not quite sure about the word 'queue' since that gives an impression of some FIFO stuff going on, and I thinks it's better to think of `scenes` as of a list/array (js doesn't really make a difference between these two). I'm guessing English is not your first language, so if you feel like these differences are beyond your resolving power I can rename the variables, push the changes to your branch and then merge. It's still going to count as your contribution. This is just an offer, you can give it another shot if you want :)

src/views/CardStack/CardStackStyleInterpolator.js
```diff
@@ -51,33 +53,25 @@ function forHorizontal(
   if (!layout.isMeasured) {
     return forInitial(props);
   }
+  const interpolate = getFirstAndLastSceneIndicesForInterpolationInputRange(props); // eslint-disable-line
```
vonovak: This is kinda my mistake, but um... it would be nicer not to use `eslint-disable-line` so many times... I'm guessing it's because of the long function name? If that's the cause, maybe `getSceneIndicesForInterpolationInputRange` will make eslint happy?

src/utils/getFirstAndLastSceneIndicesForInterpolationInputRange.js
94cstyles: Yes, Chinese is my mother tongue. But here is indeed confusing, because a scene with two index, one is called the key of the index, the other is in the scene stack index. This time I re-named, I do not know whether to meet your naming norms, if still not through, trouble you help me to rename it, thank you.

vonovak:
thanks a lot @94cstyles

kelset:
Awesome work, thanks for the PR! 💪

rmn528:
hello, I download the react navigation project with this changes, but when I tested in my app, the reset action continues poping the screens from the stack, each of them with the back animation, can you help me with an example of how implement your changes?

94cstyles:
@rmn528 [example](https://github.com/94cstyles/react-navigation-test)

rmn528:
thank you @94cstyles, I downloaded your example and checked it and I saw a difference between your example and my project, I'm using redux, I already applied the changes in my router as your example has it and it's still not working, do you thing that redux could be the guilty?

blaues0cke:
@rmn528 I am also using redux and the only thing I had to do to make this work was this:

image

eballeste:
ok i finally got this to work with redux but it's important to know and easily overlooked that you **need** to include @94cstyles branch of [react-native-addons](https://github.com/94cstyles/react-navigation-addons) and use it's enhance method on your StackNavigator... 

`import { enhance } from 'react-navigation-addons';`
`export default MainNavigator = enhance(StackNavigator)({`

eballeste:
`react-native-addons` somehow now causes all of my screens to re-render everytime I try to reset the navigation using `NavigationActions.reset()`.

eballeste:
Warning: Do not use react-native-addons and the suggested fixes for using the goBack method by specifying a routeName. It will cause unintended side effects, specifically the reset function will cause your screens to mount again, triggering all of it's internal life cycle events.

There is a supposed PR for this #3085, so maybe go down that path instead.

94cstyles:
The current PR is only fix animation-related issue. No fix goBack does not support routeName.
`react-native-addons` is just an experimental component. You can not use it.

rmn528:
I have a doubt, I'm ussing the tab navigation from react navigation and I have 3 screens (A, B, C) and I start the tab navigation in the A screen, when I press the tab that carry me to the C screen from the A screen, I notive that the animation passes through B screen, is there a way to avoid the B screen when I go from A to C?

wkoutre:
Referencing this commit: [gorokhovich/react-navigation@30e6d0c](https://github.com/gorokhovich/react-navigation/commit/30e6d0cc73406e7697f5b846a3a7c6a1ddca9862)

For project-related reasons, I'm on `react-navigation@1.0.0-beta.15`.

a) The code in gorokhovich's commit makes sense
b) It works as a plug-and-play solution to this reset animation issue... 3 screens deep or 15 screens deep, the UI pops back only one screen.

woowalker:
it seems not fixed on react-navigation: 1.5.11, am i missing something ?

brentvatne:
@woowalker - please create a new issue that follows the template

PR #2526, opened 2017-09-05
Fix Website Build

spencercarli:
Should fix website build so we can get doc updates live.

## Test Plan

Follow instructions in `website/README.md` to build the site.


src/utils/invariant.js
```diff
@@ -24,7 +24,7 @@
 
 var validateFormat = function(format) {};
 
-if (__DEV__) {
+if (global.__DEV__) {
```
spencercarli: Any negative ramifications for this? It was causing a `ReferenceError` when testing the production web build.

matthamil:
Looks like it built correctly on my machine :tada:, awesome 😄 ! Not sure about the implications of changing `__DEV__` to `global.__DEV__` though.

src/utils/invariant.js
skevy: It's probably fine -- but one other way to solve it would be to polyfill __DEV__ in webpack:

```
new webpack.DefinePlugin({
      __DEV__: isDebug,
    }),
```

spencercarli: Seems that's how it's [already being done](https://github.com/react-community/react-navigation/blob/master/website/config/webpack.config.prod.js#L206). Not sure why it isn't working 🤔

PR #2527, opened 2017-09-05
Fixed issue with tab navigation

manubamba:
dependency `react-native-tab-view` has renamed `onRequestChangeTab` to `onIndexChange` in the version `0.0.69` that was being resolved by `~0.0.67`.
Minor fix for the same.


cooperka:
See https://github.com/react-community/react-navigation/pull/2348, I think we want to keep `~` instead of `^`. Other than that this works for me!

spencercarli:
This looks good to me - for some reason circle is failing. Looking into it

PR #2544, opened 2017-09-06
fixed execution order for child actions in TabRouter

nico1510:
This is a small fix which fixes the execution order of actions which contain a sub action. Before this fix the following scenario did not work: Suppose there is a simple DrawerNavigator.

```javascript
const drawer = DrawerNavigator({
  AStack: { screen: AStack },
  BStack: { screen: BStack },
});
```

Further suppose that the BStack route is active and dispatching the following action: 

```javascript
subAction = NavigationActions.reset({
      index: 0,
      actions: [
        NavigationActions.navigate({
          routeName: 'someOtherRouteInAStack',
        })
      ]
   });
 }
this.props.navigation.navigate('AStack', undefined, subAction);
```

When this is done, one would expect the following execution order:

1. navigate to AStack
2. reset AStack's state
3. navigate to someOtherRouteInAStack

But instead the TabRouter trys to execute the sub action (the reset action) first. This leads to an error since the routeName "someOtherRouteInAStack" in the actions array is not known to the currently active route which is still BStack.

With the fix everything works as expected.

The bug was discovered by me and another user here https://github.com/react-community/react-navigation/pull/1914#issuecomment-325816788 so once it is fixed I can submit a PR which fixes the issue described there.

**Test plan (required)**

A test in StackRouter is failing in the master branch so naturally it is also failing in my branch but it is not related to this fix.


nico1510:
did someone have a look at this ? Should I add a test case for this ?

kelset:
@vonovak can you double check my old review?

exKAZUu:
Maybe we should fix the test `src/routers/__tests__/TabRouter-test.js` to clear out CircleCI's error.

nico1510:
any progress here ?

monteithpj:
This case should be fixed by #875

nico1510:
thanks for letting me know @monteithpj  ;-)

PR #2568, opened 2017-09-11
Expose useNativeAnimations for Drawer navigator

jkimbo:
By enabling `useNativeAnimations` the drawer navigator is significantly quicker.

**Test plan (required)**

Option is opt in so shouldn't affect any current users.

### DIFF
--- a/src/navigators/DrawerNavigator.js
+++ b/src/navigators/DrawerNavigator.js
@@ -32,7 +32,7 @@ const DefaultDrawerConfig = {
     Dimensions.get('window').width - (Platform.OS === 'android' ? 56 : 64),
   contentComponent: DrawerItems,
   drawerPosition: 'left',
-  useNativeAnimations: false,
+  useNativeAnimations: true,
 };
 
 const DrawerNavigator = (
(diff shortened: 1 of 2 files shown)


AndrewIngram:
Videos showing the different:

Before: https://www.youtube.com/watch?v=O3-YMhocBVw
After: https://www.youtube.com/watch?v=dSpF-WkP1Yc

src/navigators/DrawerNavigator.js
```diff
@@ -32,6 +32,7 @@ const DefaultDrawerConfig = {
     Dimensions.get('window').width - (Platform.OS === 'android' ? 56 : 64),
   contentComponent: DrawerItems,
   drawerPosition: 'left',
+  useNativeAnimations: false,
```
kelset: One thing: I think this should be set to true

jkimbo: I defaulted to false because I didn't want to accidentally break someones app. Happy to change it though.

spencercarli: Yeah I don't think it would have any negative ramifications to set this to default.

jkimbo: Ok done!

jkimbo:
@kelset thanks! Any eta on when it will be released?

kelset:
Probably soon ;) We are working to get the roadmap to 1.0 out asap atm ;)

PR #2573, opened 2017-09-13
Update Guide-Intro.md to include missing code for the first example

ocie:
I was going through the first example on https://reactnavigation.org/docs/intro/, but I was getting an error because SimpleApp was not being exported.  Once I added this export, the app worked for me.  I was running this via expo. Not sure if that is important.

### DIFF
--- a/docs/guides/Guide-Intro.md
+++ b/docs/guides/Guide-Intro.md
@@ -55,9 +55,9 @@ const SimpleApp = StackNavigator({
   Home: { screen: HomeScreen },
 });
 
-AppRegistry.registerComponent('SimpleApp', () => SimpleApp);
-
 export default SimpleApp;
+
+AppRegistry.registerComponent('SimpleApp', () => SimpleApp);
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guides/Guide-Intro.md
```diff
@@ -56,6 +56,8 @@ const SimpleApp = StackNavigator({
 });
 
 AppRegistry.registerComponent('SimpleApp', () => SimpleApp);
+
+export default SimpleApp;
```
kelset: can you move it above the AppRegistry?

ocie: It should work. I'll update the pull request.

matthamil:
This is due to using CRNA. I think we're going to go with https://github.com/react-community/react-navigation/pull/2491 since it adds comments for CRNA users and RN CLI users.

ocie:
@matthamil So it sounds like my change would be redundant considering #2491. I'll close this pull request then.

matthamil:
Don't let this discourage you from submitting another PR in the future :)

PR #2581, opened 2017-09-14
Trivial: cleanup warnings in test suite

plasticine:
### Before
screen shot 2017-09-14 at 4 47 41 pm

### After
screen shot 2017-09-14 at 4 47 24 pm

### Test plan (required)

1. Run `jest` without this diff and see noisey warnings
2. Run `jest` with this diff and don’t see noisey warnings

### DIFF
--- a/src/routers/__tests__/StackRouter-test.js
+++ b/src/routers/__tests__/StackRouter-test.js
@@ -869,6 +869,7 @@ describe('StackRouter', () => {
   });
 
   test('Maps old actions (uses "Handles the reset action" test)', () => {
+    global.console.warn = jest.fn();
     const router = StackRouter({
       Foo: {
         screen: () => <div />,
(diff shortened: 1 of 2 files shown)


spencercarli:
@plasticine ah, linter doesn't like the line length. Could you fix those (keep it under 80 characters per line)?

plasticine:
> Could you fix those (keep it under 80 characters per line)?

Ah — yep of course.

spencercarli:
Ah! It's being touchy. Looks like the only way to get it to pass is to use the format

```
expect(console.warn).toBeCalledWith(
  expect.stringContaining(
    "The action type 'Init' has been renamed to 'Navigation/INIT'"
  )
);
```

plasticine:
@spencercarli haha — yeah, sorry I just noticed that. Would you consider a PR for a prettier pre-commit hook?

matthamil:
@plasticine I think that would be a good idea. I've noticed a few PRs now that have not been ran through Prettier before submission. :)

spencercarli:
@plasticine I would love to see that in a PR! Would save a lot of people a lot of time 😄

PR #2619, opened 2017-09-20
Fix Flow types (and update RN version)

Ashoat:
The main reason there are errors on a stock RN app with react-navigation is that the `.flowconfig` had `[ignore]`s that caused the types to drift from stock packages. The first commit in this PR fixes `.flowconfig`, the second and third update package and type library versions, and the remaining commits fix the resultant Flow errors.


Ashoat:
Hey @mrm1ck, I'm not really sure what you're talking about, and your PR doesn't really explain it. What test failures are you addressing? Is the new folder you're creating (`cli/temp/test/reports`) some sort of magic folder for some test suite? What test suite exactly?

Sorry if I'm missing something obvious; I'm not too failure with how `react-navigation` tests.

@skovhus - Good find. Right now there is a `flow/react-navigation.js` file that types `src/react-navigation.js`, but it's not included when `react-navigation` package is installed in another package, and on top of that it just types everything as `any`.

I went ahead and removed that file (as well as `flow-typed/react-native.js`) and things seem fine. However, adding `@flow` to `src/react-navigation.js` revealed another type error: `createNavigator` takes 4 parameters, but in the website code only one is passed to it. It's worth noting that the docs only talk about one parameter too. I'll try to investigate a bit more to figure out what's going on there,

Ashoat:
@grabbou, in 93976d358e27580e568ef016f820dc0253423aa7 ("Introducing flat options") you updated `createNavigator` to take four params instead of one. However, you didn't update the [docs](https://reactnavigation.org/docs/navigators/custom#createNavigator) or the nine `createNavigator` calls in the [website code](https://github.com/react-community/react-navigation/blob/master/website/src/App.js#L61).

The type errors with the website code didn't get uncovered until now, since the Flow configuration on this project has been broken for quite a while. I am trying to fix it.

Can you provide some guidance on what we should do here? Are the new parameters optional? If not, how should we update the website code? Should the docs still mention `createNavigator` at all?

src/react-navigation.js
```diff
@@ -1,5 +1,5 @@
 /*
- * @noflow - get/set properties not yet supported by flow. also `...require(x)` is broken #6560135
```
skovhus: Which version of flow does this break?

Alternative would be not to use `get` here.

Ashoat: Basically we're relying on [this Flow functionality](https://github.com/facebook/flow/commit/6dea70976b4519a2e0fcc69977eed8699ea66243), which was already enabled in our `.flowconfig` with `unsafe.enable_getters_and_setters=true`. This is the same thing React Native does, so I think it's a safe bet.

Ashoat: By the way, using this version of the PR (with typechecking enabled in the library definition) brings up a whole lot more (mostly) legitimate Flow errors on my personal project. I'm working through those errors right now to make sure nothing is wrong with the types in React Navigation.

skovhus: Greatness. Do you see this error https://github.com/react-community/react-navigation/pull/2616 ?

I guess this is due to an old version of Flow that a lot of users are still using (e.g. older React Native projects).

Ashoat:
Okay, so I went ahead and integrated this version of the library into my personal app. I got a whole bunch of new Flow errors, which was a good indication that things were working! I worked through all them, and I identified one typing issue and one unfortunate shortcoming of Flow.

**Typing Issue**

The issue has to do with the `TabNavigator`/`StackNavigator`/etc. component factories. The component they spit out should be possible to use with just a `navigation` prop, according to [the docs](https://reactnavigation.org/docs/guides/redux). However, the `NavigationNavigatorProps` type includes [two other required props](https://github.com/react-community/react-navigation/blob/master/src/TypeDefinition.js#L389-L390). [`createNavigationContainer.js`](https://github.com/react-community/react-navigation/blob/master/src/createNavigationContainer.js) is where the component factory logic is defined.

Things seem to work okay if I just set these two other props (`screenProps` and `navigationOptions`) to optional in `NavigationNavigatorProps`. This is what I did to "fix" the issue I addressed to @grabbou above as well. Not sure if that's what we should be doing.

If things typecheck, I guess we're good. But for all of these cases, it'd be great if somebody could elucidate why these props were originally listed as private.

@grabbou - three questions currently for you, about type errors that have been exposed after fixing up Flow configuration:
1. The question asked [above](https://github.com/react-community/react-navigation/pull/2619#issuecomment-330887870).
2. You added `screenProps` to `NavigationNavigatorProps` in the same commit that introduced the `createNavigator` type issue above: 93976d358e27580e568ef016f820dc0253423aa7 ("Introducing flat options"). Is it really a required prop for all components created from `TabNavigator`/`StackNavigator`/etc.?
3. Same question for `navigationOptions`, which you added in a252b46c500ee2d789ced6ff4f75a400d7931ed0 ("Breaking: Replace `containerOptions` with just `props`").

**Flow Shortcoming**

Unfortunately, it seems impossible to properly type your exact `NavigationState` in your application. I'm using Redux, and there are some cases where I reach into particular parts of the `NavigationState` to get something that I know is there. This relies on a particular configuration of the `NavigationRoute` tree. However, each `NavigationStateRoute`'s children are typed as `Array<NavigationRoute>`, and it doesn't seem to be possible to [refine the type of an Array to a Tuple](https://flow.org/en/docs/types/tuples/#toc-tuples-don-t-match-array-types), nor does it seem possible to declare a particular type for a specific member of an array using object notation.

src/react-navigation.js
Ashoat: I haven't seen that error in the process of building this PR. I'll comment with a thought on your PR.

Ashoat:
I can confirm that the latest version of this PR typechecks both in the root directory, the NavigationPlayground example, and in my personal app. The only possibly sketchy parts are the three questions addressed to @grabbou above.

It's worth noting that if ReduxExample was Flow-typed, we would've noticed the "Typing Issue" in my last comment earlier. As it stands, I only noticed it when integrating with my personal app, which means there's no real to detect when `NavigationNavigatorProps` is mistyped within React Navigation.

matthamil:
I'll review this and get it merged over the weekend 😄

src/TypeDefinition.js
```diff
@@ -451,7 +451,7 @@ export type NavigationSceneRendererProps = NavigationTransitionProps;
 export type NavigationTransitionSpec = {
   duration?: number,
   // An easing function from `Easing`.
-  easing?: (t?: number) => number,
+  easing?: (t: number) => number,
```
matthamil: Two commits for the same change :). Nice catch, again 😄

matthamil:
@Ashoat Can you run `yarn run format` on this PR so CircleCI doesn't fail with lint/prettier checks?

Ashoat:
Went ahead and ran `yarn run format` and committed the results. The first parallel test (`npm run lint && npm run flow-check && npm run jest`) ran fine, but the second parallel test (`node node_modules/jest/bin/jest.js`) is failing due to some React Native issue it seems?

```bash
$ node node_modules/jest/bin/jest.js
 FAIL  ./App.test.js
  ● Test suite failed to run

    TypeError: Cannot read property 'BLOB_URI_SCHEME' of undefined
      
      at Object.<anonymous> (node_modules/react-native/Libraries/Blob/URL.js:21:21)
      at node_modules/react-native/Libraries/Core/InitializeCore.js:174:40
      at getValue (node_modules/react-native/Libraries/Utilities/defineLazyObjectProperty.js:44:10)
      at ContextifyScript.Script.runInContext (vm.js:53:29)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        5.059s
Ran all test suites.
error Command failed with exit code 1.
Exited with code 1
```

GertjanReynaert:
@Ashoat you might want to take a look at this issue https://github.com/facebook/react-native/issues/15810 for the `BLOB_URI_SCHEME` error

Ashoat:
The `BLOB_URI_SCHEME` issue discussed above happens in NavigationPlayground, and is fixed in RN 0.48.4, so I upgraded to that (b3570bf).

In the process of debugging that I discovered that the test config uses `yarn link` with NavigationPlayground. When I tried this on my local copy, more Flow errors popped up. One of them was because the absolute paths Flow compares are different when using `yarn link`, and so one the lines I removed in the first commit of this PR had to be re-added to NavigationPlayground's `.flowconfig` (6bba368). Besides that, there were three legitimate type errors that I corrected with e9a48b1, e790d45, and 0f58da1.

Looks like the tests fully pass now!

kelset:
Thanks for taking such a good care of this PR! Now that tests are passing I think @matthamil will easily give his 💯 - it looks good to me ;D

GertjanReynaert:
@Ashoat Sorry for the late response, but I'm glad you got it fixed.

Really great work, and I'm really excited to see the community effort in this project ignited again!

PR #2629, opened 2017-09-21
Improve StatusBar/Header Behavior 

spencercarli:
## Motivation

While working on adding support for iPhoneX/iOS11 I saw incorrect behavior with the header. Currently, when a device goes to landscape mode the status bar is hidden (as expected) but the header component height stays the same. This can be seen in the following GIF.

current-status-bar-behaviour

## Test Plan

Install `react-navigation` from `spencercarli/react-navigation#improve-status-bar-behavior` into the NavigationPlayground example and spin your device around. You should see the following behavior.

new-status-bar-behaviour


src/views/withOrientation.js
```diff
@@ -0,0 +1,41 @@
...
+
+    componentWillUnmount() {
+      Dimensions.removeEventListener('change', this.handleOrientationChange);
+    }
+
+    handleOrientationChange = ({ window }) => {
+      const isLandscape = isOrientationLandscape(window);
+      this.setState({
+        isLandscape,
+        isPortrait: !isLandscape,
```
matthamil: Do we need `isPortrait` if it's already the opposite of `isLandscape`?

spencercarli:
See #2669

PR #2635, opened 2017-09-23
Fixes issues with change to navigation.navigate(object only)

ShMcK:

`navigation.navigate('DrawerOpen')` and its similar calls no longer works. An error is thrown asking that you replace DrawerOpen calls with an object of `{ routeName: 'DrawerOpen' }`. This fixes issues in: docs, example, DrawerView component.

See the example code in the repo for an example.

### DIFF
--- a/docs/api/navigators/DrawerNavigator.md
+++ b/docs/api/navigators/DrawerNavigator.md
@@ -65,14 +65,14 @@ const MyApp = DrawerNavigator({
 To open and close drawer, navigate to `'DrawerOpen'` and `'DrawerClose'` respectively.
 
 ```js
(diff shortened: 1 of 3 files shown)


ShMcK:
I believe in this case the type checking in the test is behind the current version. 

navigate has moved from accepting a type of string to:

```js
{
     routeName: string,
      params?: NavigationParams,
      action?: NavigationAction
}
```

However, I've also come to realize that **all of the examples** need updating because of this, this PR is only a piece of the problem.

ericvicenti:
I think this is fixed now. Please try the latest versions of the library

PR #2639, opened 2017-09-24
Suggest flex: 1 for embedded navigator

svcondor:
In the documentation for Nesting a Navigator in a Component suggest using flex: 1 if the navigator  renders as blank.

The tutorial might be used by beginners who don't fully understand React render.

This is a modified version of PR #2224 by  https://github.com/benoitclaveau


spencercarli:
Thanks!

PR #2641, opened 2017-09-25
Fix impossible to renderTitle in full width on ios

Freddy03h:
On iOS it's impossible to render a title in Header in full width with existing options.

And it can be annoying for some use case. In mine it's a **search bar on top** :

capture d ecran 2017-09-25 a 11 56 57

So for iOS this code disable left and right only if there is no right and left component to render. The behavior is different from Android because if we only disable left or right the title will not be centered anymore.

capture d ecran 2017-09-25 a 11 58 48


vonovak:
I don't think you want to use `title` for this, but use `header`.

Freddy03h:
Note in this case because it's the header from the stack navigator with default animation.

ezgif com-video-to-gif

vonovak:
so the problem is the animation? Well, if you take a look at the `if` just above the code you have added, you'll see

```
       if (!options.hasLeftComponent) {
         style.left = 0;
       }
        if (!options.hasRightComponent) {
          style.right = 0;
        }
```

the author probably had a reason for limiting this to android. Your approach with an `else` branch seems to be pretty hacky, surely it could be made nicer, imo. I think the first step is finding out why the `if (Platform.OS === 'android') {` is there.

Freddy03h:
On Android the title isn't centered by default, so when there is no component on the left the `left` is set to `0` otherwise there will be a strange blank space.

`left` and `right` are independent on Android because title isn't centered : the title behavior is to take all the space.

In my Android version of the app, the search bar (as title component) don't take all the space because there is a burger icon on the left for the drawer, but on the right it take all the place. See :

screenshot_20170926-134530

This behavior can't happened on iOS because of the centered title, that why there is a specific code about Android here.

But on Android if there is no components at all on the right and on the left, the title take all the space. And I think this behavior is also possible on iOS too !

vonovak:
@Freddy03h I see. I think this makes sense.

spencercarli:
Thanks for this! Just tested it and things look good. The tests were failing so I fixed them - will merge as soon as CircleCI finishes!

Freddy03h:
@spencercarli Thank you !

PR #2661, opened 2017-09-28
Fix goBack

robxyy:


kelset:
What is this PR accomplishing? Can you provide a test case? Is it solving some open issues?

robxyy:
Fix #1213 
AScreen(a-key)->BScreen(b-key)->CScreen(c-key)->AScreen
navigation.goBack(akey)

rpopovici:
@robxyy how are you accessing previous state keys from inside the route, because navigation.state is not exposing any previous routes information?

robxyy:
@rpopovici Save the navigation.state.key and send it to the next screen, and then get the previous status key from navigation.state.params until we want it.

rpopovici:
@robxyy I know that, but it's not convenient. Beside this, goBackFrom was introduced to make goBack idempotent when you press the back button multiple times. You removed that feature now and have nothing to replace it. I believe that goBackTo should be an alternative action not a replacement for goBackFrom

robxyy:
@rpopovici Well, you are right, and I also agree that goBackTo is great. Since goBack is blurry, so I removed this feature and replaced it.

Issue #2658, opened 2017-09-27
1.0.0-beta.13 still has flow errors :(


joeybaker:
### Current Behavior

Thanks to @Ashoat's work, flow support greatly improved in https://github.com/react-community/react-navigation/pull/2619, but it looks like the built version of the package is now throwing new flow errors:

```
node_modules/react-navigation/lib-rn/utils/invariant.js.flow:0
invariant. Duplicate module provider
current provider. See: node_modules/fbjs/lib/invariant.js.flow:0

node_modules/react-navigation/lib-rn/utils/shallowEqual.js.flow:0
shallowEqual. Duplicate module provider
current provider. See: node_modules/fbjs/lib/shallowEqual.js.flow:0

node_modules/react-navigation/lib/utils/invariant.js:0
invariant. Duplicate module provider
current provider. See: node_modules/react-navigation/lib-rn/utils/invariant.js:0

node_modules/react-navigation/lib/utils/invariant.js.flow:0
invariant. Duplicate module provider
current provider. See: node_modules/fbjs/lib/invariant.js.flow:0

node_modules/react-navigation/lib/utils/shallowEqual.js:0
shallowEqual. Duplicate module provider
current provider. See: node_modules/react-navigation/lib-rn/utils/shallowEqual.js:0

node_modules/react-navigation/lib/utils/shallowEqual.js.flow:0
shallowEqual. Duplicate module provider
current provider. See: node_modules/fbjs/lib/shallowEqual.js.flow:0

node_modules/react-navigation/src/utils/invariant.js:0
invariant. Duplicate module provider
current provider. See: node_modules/react-navigation/lib-rn/utils/invariant.js:0

node_modules/react-navigation/src/utils/shallowEqual.js:0
shallowEqual. Duplicate module provider
current provider. See: node_modules/react-navigation/lib-rn/utils/shallowEqual.js:0


Found 8 errors
```

### Expected Behavior
It looks like the root of this problem is to not rely on the `fbjs` package in https://github.com/react-community/react-navigation/pull/1698. I'm pretty unclear why it's desirable to not rely on the package? We could just rely on the package again, that _should_ fix things. 

Another option is to remove `@providesModule` from each of the two files and not export their `.js.flow`files during the build. (This feels less clean to me)

Edit: Yet another option: could we not have `src/utils` not copied into `lib` and `lib-rn`?

### Your Environment
- _Include as many relevant details about the environment you experienced the bug in_

| software         | version
| ---------------- | -------
[...]

PR #2665, opened 2017-09-28
Remove @providesModule declarations in src/utils to avoid Flow conflicts with fbjs package

Ashoat:
See #2658. These `@providesModule` ("Haste") declarations are not necessary since every file imports these modules using the full path. We're removing them because they declare the same modules as the fbjs packages do (the files are actually copied in from there), so Flow gets confused and doesn't know which to use.

### DIFF
--- a/src/utils/invariant.js
+++ b/src/utils/invariant.js
@@ -5,8 +5,6 @@
  * This source code is licensed under the BSD-style license found in the
  * LICENSE file in the root directory of this source tree. An additional grant
  * of patent rights can be found in the PATENTS file in the same directory.
- *
- * @providesModule invariant
  */
 
 'use strict';
--- a/src/utils/shallowEqual.js
+++ b/src/utils/shallowEqual.js
@@ -6,7 +6,6 @@
  * LICENSE file in the root directory of this source tree. An additional grant
  * of patent rights can be found in the PATENTS file in the same directory.
  *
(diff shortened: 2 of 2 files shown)


Ashoat:
I tested that this solves the issue by stripping the `@providesModule` declarations from the following files in a project that's using `react-navigation@1.0.0-beta.13`:

```
node_modules/react-navigation/lib/utils/invariant.js
node_modules/react-navigation/lib/utils/invariant.js.flow
node_modules/react-navigation/lib/utils/shallowEqual.js
node_modules/react-navigation/lib/utils/shallowEqual.js.flow
node_modules/react-navigation/lib-rn/utils/invariant.js
node_modules/react-navigation/lib-rn/utils/invariant.js.flow
node_modules/react-navigation/lib-rn/utils/shallowEqual.js
node_modules/react-navigation/lib-rn/utils/shallowEqual.js.flow
node_modules/react-navigation/src/utils/invariant.js
node_modules/react-navigation/src/utils/shallowEqual.js
```

I also confirmed that the `@providesModule` declarations are missing the generating `lib` and `lib-rn` folders by running `yarn` after making my changes to the files in the `src` folder.

kelset:
Looks good to me, but I'm not sure we are going to do another release just for this fix - what do you think @spencercarli?

spencercarli:
Awesome - thanks!

PR #2667, opened 2017-09-28
Fix "drawerLockMode" typo in DrawerNavigator docs

dgurns:
There's a typo in the docs -- when specifying drawerLockMode in screenProps, it should be "screenProps.drawerLockMode", not "screenProps.lockMode". 

Tested on iOS and Android.


dgurns:
hey, dumb question but this is my first pull request. Now that the PR is approved, @matthamil would you or someone else with write access actually merge the PR?

PR #2669, opened 2017-09-29
Improve StatusBar/Header Behavior

spencercarli:
Replaces #2629.

## Motivation

While working on adding support for iPhoneX/iOS11 I saw incorrect behavior with the header. Currently, when a device goes to landscape mode the status bar is hidden (as expected) but the header component height stays the same. This can be seen in the following GIF.

current-status-bar-behaviour

## Test Plan

Install `react-navigation` from `spencercarli/react-navigation#improve-status-bar-behavior2` into the NavigationPlayground example and spin your device around. You should see the following behavior.

new-status-bar-behaviour

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -9,6 +9,7 @@ import { Animated, Platform, StyleSheet, View } from 'react-native';
 import HeaderTitle from './HeaderTitle';
 import HeaderBackButton from './HeaderBackButton';
 import HeaderStyleInterpolator from './HeaderStyleInterpolator';
+import withOrientation from '../withOrientation';
 
 import type {
   NavigationScene,
@@ -281,14 +282,22 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
       screenProps,
       progress,
       style,
+      isLandscape,
       ...rest
     } = this.props;
 
     const { options } = this.props.getScreenDetails(scene);
     const headerStyle = options.headerStyle;
+    const statusBarHeight = isLandscape ? 0 : STATUSBAR_HEIGHT;
+    const containerStyles = [
+      styles.container,
+      { paddingTop: statusBarHeight, height: APPBAR_HEIGHT + statusBarHeight },
+      headerStyle,
+      style,
+    ];
 
     return (
-      <Animated.View {...rest} style={[styles.container, headerStyle, style]}>
+      <Animated.View {...rest} style={containerStyles}>
         <View style={styles.appBar}>{appBar}</View>
       </Animated.View>
     );
@@ -315,9 +324,7 @@ if (Platform.OS === 'ios') {
 
 const styles = StyleSheet.create({
   container: {
-    paddingTop: STATUSBAR_HEIGHT,
     backgroundColor: Platform.OS === 'ios' ? '#F7F7F7' : '#FFF',
-    height: STATUSBAR_HEIGHT + APPBAR_HEIGHT,
     ...platformContainerStyles,
   },
   appBar: {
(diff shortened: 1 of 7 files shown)


src/views/Header/Header.js
```diff
@@ -281,14 +282,22 @@ class Header extends React.PureComponent<void, HeaderProps, HeaderState> {
       screenProps,
       progress,
       style,
+      isLandscape,
       ...rest
     } = this.props;
 
     const { options } = this.props.getScreenDetails(scene);
     const headerStyle = options.headerStyle;
+    const statusBarHeight = isLandscape ? 0 : STATUSBAR_HEIGHT;
```
matthamil: Maybe we want to rename `statusBarHeight` if there is already a variable named `STATUSBAR_HEIGHT`?  Maybe `landscapeAwareStatusBarHeight`? Might be confusing for others looking at the code base seeing two variables both for status bar height.

spencercarli: Good point. Fixed!

vonovak:
@dantman my guess is the fact that the navbar height on iphone should be different in landscape was previously simply ignored, maybe also because it wasn't _that_ big of a problem, and maybe because someone knew that the old way it was implemented worked on the ipad.

Indeed, this is now a bug. Question is, should simply revert this commit for now, or try to fix this by finding some way of decreasing the height only when the status bar is hidden? I don't know an api which provides this information..?

dantman:
> @dantman my guess is the fact that the navbar height on iphone should be different in landscape was previously simply ignored, maybe also because it wasn't that big of a problem, and maybe because someone knew that the old way it was implemented worked on the ipad.

I do believe this is a permanent change for iOS 11. It's really inconsistent on when I get it before iOS 11, but I do sometimes get a visible status bar and a reported status bar height from iOS in landscape mode.

iPhone X - iOS 11:
simulator screen shot - iphone x - 2017-10-20 at 11 53 14

iPhone 7 Plus - iOS 10:
simulator screen shot - iphone 7 plus - 2017-10-20 at 11 52 56

iPhone 6 Plus - iOS 9:
simulator screen shot - iphone 6s plus - 2017-10-20 at 11 56 21

vonovak:
@dantman sorry, I previously wrote "...navbar height on iphone should be different in landscape..." which is overly simplified. The accurate way to put it is "navbar height on iphone should be different under *some* conditions." Now, what are those conditions? THB, I'm not knowledgeable of native ios development so I'd appreciate if someone could shed some light on this (maybe it has to do with [size classes](https://developer.apple.com/library/content/featuredarticles/ViewControllerPGforiPhoneOS/TheAdaptiveModel.html#//apple_ref/doc/uid/TP40007457-CH19-SW1)?). 

Now, my hypothesis is that the navbar height is smaller when the status bar is hidden, and greater when statusbar is shown (which is not the same as being/not being in landscape mode). Please correct if somebody thinks this is wrong. I'll try to find out in the mean time if that works.

vonovak:
I gave it a try, and it seems it'll work. I also noticed this [PR in rn-core](https://github.com/facebook/react-native/pull/16478), so I'll try to write it in a way that will not require much editing when it is merged.

dantman:
> Now, my hypothesis is that the navbar height is smaller when the status bar is hidden, and greater when statusbar is shown (which is not the same as being/not being in landscape mode). Please correct me if somebody thinks this is wrong. I'll try to find out in the mean time if that works (it should, imo).

Yes, the statusbar height does accurately reflect whether the status bar is shown, baring rare API glitches. The key being that iOS has its own internal patterns for deciding whether it wants to show the status bar in landscape or not.

For example the status bar was hid most of the time when I opened the app and then rotated the simulator to landscape, but was visible most of the time if I opened the app while in landscape mode. There may be a bunch of different conditions based on things like what orientation you open the app in and whether the app was already running.

PR #2674, opened 2017-09-29
fix redirect bug in stacks in tab

sujameslin:
Bug explain:
There's unwanted redirection if action with type `Navigation/SET_PARAMS` is dispatched in stacks in tab navigator. It was critical bug for our app. 

Bug demonstration: https://expo.io/@sujameslin/react-navigation-bug-reproduce
Like you see the bug demo, it was supposed to be showing the first tab, but its showing 2nd tab after load.
Source code: https://github.com/sujameslin/react-navigation-bug-reproduce/blob/master/App.js#L55-L57

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -238,6 +238,10 @@ export default (
       });
       // console.log(`${order.join('-')}: Processed other tabs:`, {lastIndex: state.index, index});
 
+      // keep active tab index if action type is SET_PARAMS
+      index =
+        action.type === NavigationActions.SET_PARAMS ? state.index : index;
+
       if (index !== state.index || routes !== state.routes) {
         return {
           ...state,


src/routers/TabRouter.js
```diff
@@ -241,7 +241,9 @@ export default (
       if (index !== state.index || routes !== state.routes) {
         return {
           ...state,
-          index,
+          // keep active tab index if action type is SET_PARAMS
+          index:
```
matthamil: Can this logic be moved to outside of the returned object?

sujameslin:
@matthamil would you please review again? i updated like you suggested

sujameslin:
@matthamil Hey. 
i updated like you requested and pushed with force mode because i squashed commits for better commit history. 
After force push, it doesn't showing that your change request is not solved. 
I don't know how to fix this but you can see that files are changed like you requested.

Thanks.

zhouanbo:
@sujameslin Thanks! This fixed my issue.

nelxaz:
A little too late to comment but this solved my issue a couple weeks ago! Nice work!

PR #2676, opened 2017-09-29
[iOS 11] Handle landscape tabbar correctly.

spencercarli:
## Motivation

I saw that the behavior of the tabbar has changed in iOS11 - it no longer stacks the icon and label on top of itself in landscape. 

## Test Plan

Install `react-navigation` from `spencercarli/react-navigation#ios-11-tab-bar` into an app with tab bar. I don't think this should affect customizations much (if at all).

I can't get the NavigationPlayground example to detect it's in landscape (limitation of CRNA?) so you'll need to create a new app. You can use [this](https://github.com/react-community/react-navigation/blob/master/examples/NavigationPlayground/js/StacksInTabs.js) as a start point. Should be copy and paste as long as you have `react-native-vector-icons` installed.


spencercarli:
cc @matthamil @kelset could one of you guys review (& merge if all looks good)?

spencercarli:
Good point. Will update to only make these changes if version is >= 11

spencercarli:
Alright, finally got around to fixing this! Here's iOS 10.3

ios10

and here's iOS 11

ios11

PR #2679, opened 2017-09-30
Fixing Intro Guide

### LINKED ISSUE #832: props.navigation undefined
Hello,
I am trying React native for the first time and I want to implement the [Basic Tutorial](https://reactnavigation.org/docs/intro/) into a new react native project but I get an "this.props.navigation" undefined error when opening the main page. 

It seems that this.props doesn't have a navigation method. 

In main app page:
`const App = StackNavigator({
  Signup: { screen: SignupScreen },
});`

In page Render:
`const { navigate } = this.props.navigation;`    
` <Text onPress={() => navigate('Signup')} Sign Up</Text>`

What am I may miss in this case?

kelset:
Fixes #832 - the code in this page was wrong, that `export default` was causing errors.

I've also modified a few phrases here and there to help the comprehension.

### DIFF
--- a/docs/guides/Guide-Intro.md
+++ b/docs/guides/Guide-Intro.md
@@ -6,21 +6,23 @@ Let's use React Navigation to build a simple chat application for Android and iO
 
 First, make sure you're [all set up to use React Native](http://facebook.github.io/react-native/docs/getting-started.html). Next, create a new project and add `react-navigation`:
 
-
 ```sh
 # Create a new React Native App
 react-native init SimpleApp
+cd SimpleApp
 cd SimpleApp
 
 # Install the latest version of react-navigation from npm
 npm install --save react-navigation
 
 # Run the new app
-react-native run-android # or:
+react-native run-android
+# or:
 react-native run-ios
 ```
 
 If you are using `create-react-native-app` instead of `react-native init`, then:
+
 ```sh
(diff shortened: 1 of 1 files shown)


kelset:
I've tested the code locally and with that fix it works fine, but always better to get an extra pair of eyes on it - @spencercarli or @matthamil can you just take a quick look so that we can merge and help introducing new people to `react-navigation`? 😇

docs/guides/Guide-Intro.md
```diff
@@ -66,7 +68,7 @@ class HomeScreen extends React.Component {
   }
 }
 
-export default const SimpleApp = StackNavigator({
+export const SimpleApp = StackNavigator({
```
spencercarli: We had `export default` here so that it would work with `create-react-native-app`. We followed the pattern that [React Native](https://facebook.github.io/react-native/docs/tutorial.html#hello-world) is using. What was the issue that was being caused by `export default`?

kelset: leaving `default` gives me back an `unexpected token` error (even via CRNA).

That said, I'm having troubles making it work with CRNA at all. I'll probably update my PR once I've managed to make it work the CRNA way.

kelset:
Updated my PR to make it work with CRNA & Expo

PR #2684, opened 2017-10-02
Issue-2662: Add support for testID and accessibilityLabel for tab bar…

robertkongsvmx:
# Description

This is to add support for testID and accessibility label for automated UI testing.

See https://github.com/react-community/react-navigation/issues/2662


robertkongsvmx:
@spencercarli Please review and provide feedback

spencercarli:
@robertkongsvmx thanks for submitting! How would I go about testing this?

robertkongsvmx:
@spencercarli You can use appium or Xamarin UI test. Most mobile UI test tools support the accessibilityLabel query to identify UI elements. The testID attribute is currently on iOS, and it seems appium is work on a new driver on android that might support the testID attribute.

For your purpose, you can just verify that the accessibilityLabel/testID is added to the tab item if you don't want to dig into the details of the different UI test framework.

https://developer.xamarin.com/guides/testcloud/uitest/
http://appium.io/

hellogerard:
I know we're not supposed to add "+1" comments, but +1.  I am adding Detox end-to-end testing to a react-navigation app. Detox has the same issue with tabs, and this PR would solve.

hellogerard:
@spencercarli I tested this branch with my app and it worked great. I tapped on the tab successfully from my test and drove the simulator. 

@robertkongsvmx I don't know much about the accessibility props. Question: Do you need to pass both a `testID` _and_ an `accessibilityLabel`?  Can you just pass a single string, and then use that value for both props in the `TabBarBottom`?

simonracz:
@hellogerard accessibilityLabel and testID have different purpose. `accessibilityLabel` is user visible. Think of screen readers for blind people, it will read out accessibilityLabel.

spencercarli:
Thank you!

Issue #2251, opened 2017-07-27
Querystring params not set with the URI action


joeybaker:
### Current Behavior

When using the `URI` action, querystring params aren't set on the active route.

```js  
    import {NavigationActions, StackNavigator} from 'react-navigation'
    
    // very simple component
    const Foo = () => null
    
    const ChildNavigator = StackNavigator({Food: {path: 'foo', screen: Foo}})
    const ParentNavigator = StackNavigator({Child: {path: 'child', screen: ChildNavigator}})
    
    // ... instantiate ....
    
    const path = 'child/foo?bar=baz'
    const action = NavigationActions.uri({ uri: path })
    const nextState = getStateForAction(action)
    
    /*
    {
	index: 0,
	routes: [
	    {
		routeName: 'Child',
		index: 0,
		routes: [
		    // expected: `params: {bar: 'baz'}` to be included
		    {routeName: 'Foo'}
		]
	    }
	]
    }
    */
```

### Expected Behavior

The query string params should be set on the active route.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  1.0.0-beta.11
| react-native     |  0.46.4
| node             |  7.10.0
| npm or yarn      |  0.27.5

PatNeedham:
@joeybaker @spencercarli I figured it out. [Line 363 of `StackRouter.js`](https://github.com/react-community/react-navigation/blob/master/src/routers/StackRouter.js#L363) is not passing the `queryString` variable on to the next `getActionForPathAndParams` call in order to retrieve the nested action!

Changing that line from

    pathMatch.slice(pathMatchKeys.length).join('/')

to

    pathMatch.slice(pathMatchKeys.length).join('/') + '?' + queryString

did the trick for me.

spencercarli:
Just merged the fix - thanks @PatNeedham! Will be out in the next release (hopefully this week).

PR #2695, opened 2017-10-05
Fix Querystring params not set for nested URI (issue #2251)

PatNeedham:
This pull request solves issue #2251 by introducing a `nestedQueryString` variable that is either the original `queryString` or an empty string (when `queryString` is null), that gets passed as part of the path for retrieving the nested action in `StackRouter``s `getActionForPathAndParams`.

**Test plan (required)**
I modified `src/routers/__tests__/StackRouter-test.js` by adding an additional test case, starting on the line `test('Querystring params get passed to nested deep link', () => {`. It is mostly a combination of two existing tests:

1. The one with title `Correctly parses a path with arguments into an action chain` with uri value `'main/p/4/list/10259959195'`
2. The ones titled `Parses paths with a query` and `Parses paths with an empty query value` that make use of uris ending with `?code=test&foo=bar` and `?code=&foo=bar`, respectively.

Video demonstrating the changes on iOS: [https://vimeo.com/236860581](https://vimeo.com/236860581)

For Android I tweaked the command slightly to open the deep link, due to [this StackOverflow question](https://stackoverflow.com/questions/35645414/android-deep-linking-with-multiple-query-parameters) explaining the situation where testing with multiple query params only includes the first one by default. All I had to do was include a backslash character (\\) before the ampersand (&).

Android video demonstration: [https://vimeo.com/236863226](https://vimeo.com/236863226).

On a somewhat related note, the current react-navigation [Deep Linking documentation](https://reactnavigation.org/docs/guides/linking) makes no reference to query params. I'd be more than glad to supplement that as well to cover the topic.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -360,7 +360,7 @@ export default (
       if (childRouters[matchedRouteName]) {
         nestedAction = childRouters[matchedRouteName].getActionForPathAndParams(
           /* $FlowFixMe */
-          pathMatch.slice(pathMatchKeys.length).join('/')
+          pathMatch.slice(pathMatchKeys.length).join('/') + '?' + queryString
         );
       }
 
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -357,10 +357,11 @@ export default (
       // get the action for the path AFTER the matched path for this
       // router
       let nestedAction;
+      let nestedQueryString = queryString ? '?' + queryString : '';
       if (childRouters[matchedRouteName]) {
         nestedAction = childRouters[matchedRouteName].getActionForPathAndParams(
           /* $FlowFixMe */
-          pathMatch.slice(pathMatchKeys.length).join('/') + '?' + queryString
+          pathMatch.slice(pathMatchKeys.length).join('/') + nestedQueryString
         );
       }
 
--- a/src/routers/__tests__/StackRouter-test.js
+++ b/src/routers/__tests__/StackRouter-test.js
@@ -902,4 +902,66 @@ describe('StackRouter', () => {
       )
     );
   });
+
+  test('Querystring params get passed to nested deep link', () => {
+    // uri with two non-empty query param values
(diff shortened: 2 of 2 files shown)


PatNeedham:
Here is the App.js file I'm using: https://gist.github.com/PatNeedham/082613f9c792089b315ea3b5dfe853ed

Ran react-native init NestedQueryParams. Then `npm install --save react-navigation`. Then followed the [deep linking guides](https://reactnavigation.org/docs/guides/linking) making the instructed changes to `ios/SimpleApp/AppDelegate.m` and `android/app/src/main/AndroidManifest.xml`

spencercarli:
This is great - thank you! I appreciate the test and videos, makes reviewing super easy 😄 

Would love an additional PR for improved documentation. Just ping me on it and I'll take a look.

P.S. A+ query values 😆

PatNeedham:
@spencercarli where would be the best place to put the documentation additions? I'm not sure about whether to add a section to the existing Deep linking page (and if so, whether to place it all the way at the bottom, or after `URI Prefix` section & before `iOS`) or create a new page altogether within the `Advanced Guides` category.

joeybaker:
An interesting side-effect of this PR is that all routes now get all params. This might have un-intended effects if you have multiple routes that take the same params. It's a little un-expected because the web wouldn't work this way, but it might be okay native-side if we clearly document that params are "global".

GollyJer:
This is excellent work... thanks!

spencercarli:
Hey sorry I never got back to you on this. I think I accidentally deleted the email notification.

@PatNeedham I think adding it to deep linking is best. After `URI Prefix` and before `iOS`. Docs are kind of a mess (and major pain to add a new pages).

@joeybaker thanks for pointing that out! I agree this would be confusing. Do you (or anyone else) know what the normal behavior is on mobile? I imagine it would be like the web, but mobile is weird 😄

spencercarli:
My initial thoughts would be that params would only be available on the screen that was opened via deep linking and then you would pass the data other screens are interested in via props?

joeybaker:
That’s my thought too.

PR #2698, opened 2017-10-05
Added property inside drawerConfig to backgroundColor

padil:
**DrawerNavigator**

I added a property to **drawerConfig**, which is **drawerBackgroundColor**, because I needed to use the transparent background to make a shadow and I could not.

example:

`const drawerConfig = {
  drawerBackgroundColor: 'transparent'
};`


kelset:
Just one thing, can you kindly submit another PR that adds this prop to the documentation?

padil:
Sure, no problem!

PR #2753, opened 2017-10-14
[WIP] Updating dependencies in package.json

kelset:
Mainly motivated by #1313.

I've tested this version against two test projs, a simple one and a more "realistic/worst case scenario", by using @GantMan drawer project + forked + hardcoded the dependency to the version with updated packages (so in a scenario of old Expo where everything else was kept "old") ([here](https://github.com/kelset/drawer-example)).
Both times the projects kept working fine, so I think we can up the deps 💪

---
EDIT: need to fix all the CicleCI errors, so I've put the [WIP] in front.
These new errors come from the new Flow version, as said by @Ashoat & @Palisand in #2658


kelset:
Small update: I've managed to reduce the flow errors to "only" 221 errors.

Since the update from 52 to 53 has some big updates, I've tried using [the tool they suggest to use](https://github.com/facebook/flow/blob/master/Changelog.md#0530) called [flow-upgrade](https://yarnpkg.com/en/package/flow-upgrade). It detects 170 files but it fails on every single one of them with the error reported [here](https://github.com/facebook/flow/issues/4945).

Ashoat:
Do the examples in the example folder actually work with these changes? If they do, are we concerned about the mainline package.json and .flowconfig being out of whack with the examples’ package.json and .flowconfig?

I was gonna wait until Expo 22 was released and react-native-tab-view was [fixed up](https://github.com/react-native-community/react-native-tab-view/pull/361) before updating the dependencies.

kelset:
Ok I'm dumb I completely overlooked the `/example` folder 😱

Thanks @Ashoat for your feedback, better if I leave this up to you, you have way more experience in tackling Flow configuration 🙏

PR #2759, opened 2017-10-15
Make goBack() behave the same as goBack(null)

tienf4m:
Currently goBack has an optional param, which specifies the route to go back from. If there's no key provided, we'll use the key from the route that goBack is called from (which makes sense).
https://github.com/react-community/react-navigation/blob/f84fe152721f0f6f80f799dbcd9bcbf5f42d61d0/src/addNavigationHelpers.js#L21

The problem is that it doesn't work if we want to go back from a nested navigator to a parent navigator, but goBack(null) works. And this confuses a lot of people, including me. 
I have looked at the code and it seems that if we're in a StackNavigator with only 1 route or a TabNavigator with the first tab active, both goBack() and goBack(null) will result in the child navigator returning the same state, thus the parent navigator will handle the action. But with a non-null key, the navigator will try to find a child route with that key and go back from that, and it won't be able to find that route because it's from a nested navigator:
https://github.com/react-community/react-navigation/blob/f84fe152721f0f6f80f799dbcd9bcbf5f42d61d0/src/routers/StackRouter.js#L277

In the other hand, if the key is null, the parent navigator will just pop:
https://github.com/react-community/react-navigation/blob/f84fe152721f0f6f80f799dbcd9bcbf5f42d61d0/src/routers/StackRouter.js#L285

So if we don't provide a default key when calling goBack(), we'll be able to go back from a nested navigator. In the current implementation, we use the key from current route just to make sure that the app is not going to go back from a wrong route. But I think the change from this PR wouldn't cause any unexpected behaviors since navigators will let the active route handle the action first.

This will fix https://github.com/react-community/react-navigation/issues/1522

### DIFF
--- a/docs/guides/Screen-Navigation-Prop.md
+++ b/docs/guides/Screen-Navigation-Prop.md
@@ -90,7 +90,7 @@ class ProfileScreen extends React.Component {
 
 ## `goBack` - Close the active screen and move back
 
-Optionally provide a key, which specifies the route to go back from. By default, goBack will close the route that it is called from. If the goal is to go back *anywhere*, without specifying what is getting closed, call `.goBack(null);`
+Optionally provide a key, which specifies the route to go back from. By default, goBack will close the currently active route (which most of the time is the route that it is called from).
 
 ```js
 class HomeScreen extends React.Component {
@@ -101,10 +101,6 @@ class HomeScreen extends React.Component {
         <Button
           onPress={() => goBack()}
           title="Go back from this HomeScreen"
-        />
-        <Button
-          onPress={() => goBack(null)}
-          title="Go back anywhere"
         />
         <Button
           onPress={() => goBack('screen-123')}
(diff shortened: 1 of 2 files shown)


kelset:
LGTM  - @spencercarli what do you think?

tienf4m:
I just updated the docs and removed goBack(null) because now goBack() is doing the same thing.

> I've tried it in my prod app and it "broke" the navigation
Can you provide a usecase where it breaks? From my understanding goBack(null) will go back from the currently active route, which is probably what you want. But if there are certain usecases where you want it to goBack from the route is is called from and people are relying on that behavior, I think we could just update the usages of goBack in the docs.

tienf4m:
After going through some other issues with goBack, I think we need to find a way to make goBack(key) work in nested navigators instead of making it behave the same as goBack(null). There are a few ways I can think of but it would increase the complexity of the current code. I'll open another PR when I have a better solution.

PR #2773, opened 2017-10-17
Fixes default drawerWidth to match Material UI patterns.

davepack:
Previously: calculated based on device width regardless of orientation and did not recalculate when orientation changed. If an app was opened in landscape, the drawer would be too wide if the device was rotated to portrait.
Now: calculates based on minimum of device height/width, drawer width remains constant and is guaranteed to always fit screen regardless of orientation.

This is the expected behavior based on observing Google apps (e.g. Gmail). This is also better than recalculating on every orientation change, which would result in variable width drawers and awkward empty space when in landscape in most cases.

Also added support for rotating device in NavigationPlayground example app.

### DIFF
--- a/src/navigators/DrawerNavigator.js
+++ b/src/navigators/DrawerNavigator.js
@@ -23,13 +23,16 @@ export type DrawerNavigatorConfig = {
 } & NavigationTabRouterConfig &
   DrawerViewConfig;
 
+const { height, width } = Dimensions.get('window');
+
+console.log(height, width, Math.min(height, width));
+
 const DefaultDrawerConfig = {
   /*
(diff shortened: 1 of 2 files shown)


src/navigators/DrawerNavigator.js
```diff
@@ -23,13 +23,16 @@ export type DrawerNavigatorConfig = {
 } & NavigationTabRouterConfig &
   DrawerViewConfig;
 
+const { height, width } = Dimensions.get('window');
+
+console.log(height, width, Math.min(height, width));
```
spencercarli: Remove console.log

PR #2790, opened 2017-10-19
Add previous scene to tabBarOnPress - fix #2787

### LINKED ISSUE #2787: Get the previous scene in the tabBarOnPress callback
### Current Behavior
- The callback `tabBarOnPress` [used here](https://github.com/react-community/react-navigation/blob/master/src/views/TabView/TabView.js#L104) takes as arguments `scene: TabScene, jumpToIndex: (index: number) => void` with `scene` as the destination scene.

### Expected Behavior
- In order to implement a custom behavior, I need to know what is the scene we just left. Can we add `prevScene: TabScene` as the previous scene?
- Alternatively: is there a way to retrieve that information? I already inspected the `navigation` object and doesn't store that information.

Thanks

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | #master
| react-native     | "^0.48.4"
| node             | "6.10.3"
| npm or yarn      | "1.1.0"

pietro909:
Goals:

- pass the previous scene to the `tabBarOnPress` callback, together with its current arguments

Why?

- Implementing some custom logic (ie: reset the child navigator of the scene we are leaving) was not possible in other ways

Please see issue #2787

**Test plan (required)**

Being it an argument (which always exists) passed to a callback, the only tests I can deliver are the type checks and the current tests themselves.

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -36,8 +36,13 @@ type Props = {
   jumpToIndex: (index: number) => void,
   getLabel: (scene: TabScene) => ?(React.Node | string),
   getOnPress: (
+    previousScene: NavigationRoute,
     scene: TabScene
-  ) => (scene: TabScene, jumpToIndex: (index: number) => void) => void,
+  ) => ({
+    previousScene: NavigationRoute,
+    scene: TabScene,
+    jumpToIndex: (index: number) => void,
+  }) => void,
   getTestIDProps: (scene: TabScene) => (scene: TabScene) => any,
   renderIcon: (scene: TabScene) => React.Node,
   style?: ViewStyleProp,
@@ -209,6 +214,7 @@ class TabBarBottom extends React.PureComponent<Props, State> {
       isLandscape,
     } = this.props;
     const { routes } = navigation.state;
+    const previousScene = routes[navigation.state.index];
     // Prepend '-1', so there are always at least 2 items in inputRange
     const inputRange = [-1, ...routes.map((x: *, i: number) => i)];
 
@@ -229,7 +235,7 @@ class TabBarBottom extends React.PureComponent<Props, State> {
           {routes.map((route: NavigationRoute, index: number) => {
             const focused = index === navigation.state.index;
             const scene = { route, index, focused };
-            const onPress = getOnPress(scene);
+            const onPress = getOnPress(previousScene, scene);
             const outputRange = inputRange.map(
               (inputIndex: number) =>
                 inputIndex === index
@@ -251,7 +257,9 @@ class TabBarBottom extends React.PureComponent<Props, State> {
                 testID={testID}
                 accessibilityLabel={accessibilityLabel}
                 onPress={() =>
-                  onPress ? onPress(scene, jumpToIndex) : jumpToIndex(index)}
+                  onPress
+                    ? onPress({ previousScene, scene, jumpToIndex })
+                    : jumpToIndex(index)}
               >
                 <Animated.View
                   style={[
--- a/src/views/TabView/TabBarTop.js
+++ b/src/views/TabView/TabBarTop.js
@@ -6,6 +6,8 @@ import { TabBar } from 'react-native-tab-view';
 import TabBarIcon from './TabBarIcon';
 
 import type {
+  NavigationAction,
+  NavigationRoute,
   NavigationScreenProp,
   NavigationState,
   ViewStyleProp,
@@ -26,9 +28,14 @@ type Props = {
   jumpToIndex: (index: number) => void,
   getLabel: (scene: TabScene) => ?(React.Node | string),
   getOnPress: (
+    previousScene: NavigationRoute,
     scene: TabScene
-  ) => (scene: TabScene, jumpToIndex: (index: number) => void) => void,
-  renderIcon: (scene: TabScene) => React.Node,
+  ) => ({
+    previousScene: NavigationRoute,
+    scene: TabScene,
+    jumpToIndex: (index: number) => void,
+  }) => void,
+  renderIcon: (scene: TabScene) => React.Element<*>,
   labelStyle?: TextStyleProp,
   iconStyle?: ViewStyleProp,
 };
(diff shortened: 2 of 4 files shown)


kelset:
Couple things:
1. it would be helpful to modify the docs too to reflect the change ([here](https://reactnavigation.org/docs/navigators/tab#tabBarOnPress))
1. shouldn't jumpToIndex get this "history" param too?

Tbh it should be even something we have more in general, I mean to only add it for this specific case is sort of de-aligning the navigation methods... or is it just me?

pietro909:
Thank you @kelset for the feedback:

1. docs updated in d3b5901
2. I don't think so because `jumpToIndex` is in charge of just go straight to the desired index (if it exists), so as a consumer of _react-navigation_ I wouldn't expect it to take any decision beyond the linear jump itself

> it should be even something we have more in general

I agree, can you help me to come up with a list of the possible entries to be updated? I'm quite new to this library.

Thanks!

kelset:
1. thanks for updating the docs
1. thinking more about it yeah, I agree. Moreover, this is in the context of somebody overriding a "default method", so it makes sense that we provide the dev with all these information.
1.  > I agree, can you help me to come up with a list of the possible entries to be updated? I'm quite new to this library.

I'm quite sure that this method is "unique" at the moment, moreover in the Stack scenario you always know the previous route (= it's linear). And in the Drawer AFAIK there's no method to override the onPress method.

---

By the way, while checking all the files to write the third point I've noticed that you haven't modified `TabBarTop` so that needs to be fixed before merging ;)

spencercarli:
Just noting this for myself or @kelset when we make the next release notes that this will be a breaking API change (though certainly worth it) and we should notate that.

pietro909:
I've updated `TabBarTop` and rebased against `master` branch.

Though I've noticed that I get this when I run `yarn test`:

screen shot 2017-10-25 at 08 56 47

kelset:
Ok, thanks.
This weekend I'll review it.

kelset:
Hey @pietro909 sorry for not giving you feedback earlier, we have been busy with `beta.16` and the issues it spawned - now most of them should be fixed so we can focus again on other PRs.
Can you please fix the merge conflicts?

pietro909:
Thank you @kelset I've fixed conflicts and rebased. Please let me know if there's something else.

I'm looking forward to having this one in the master since we use this feature at Knipster :-)

kelset:
LGTM - @spencercarli or @vonovak can you take a look at it too? I think we can merge.

src/views/TabView/TabBarBottom.js
```diff
@@ -187,7 +193,7 @@ class TabBarBottom extends React.PureComponent<Props> {
           {routes.map((route: NavigationRoute, index: number) => {
             const focused = index === navigation.state.index;
             const scene = { route, index, focused };
-            const onPress = getOnPress(scene);
+            const onPress = getOnPress(previousScene, scene);
```
vonovak: As noted by @spencercarli this is a breaking change. I think we shouldn't introduce breaking changes unless there are very good reasons for doing so. In this case, changing `const onPress = getOnPress(previousScene, scene);` to `const onPress = getOnPress(scene, previousScene);` and changes in a couple more lines will make this a non-breaking change. Imho it's better to do the change here rather to ask users to do it.

It is truth that the order of params in `onPress(previousScene, scene, jumpToIndex)  is slightly more intuitive than `onPress(scene, jumpToIndex, previousScene)`, but I don't think this is a reason for making a breaking change. If there is no other reason for doing the breaking change, I'd say we don't do it. Any thoughts?

vonovak: or perhaps a compromise (still breaking, but less), if you want to have a nicer order of params.

change
```
  getOnPress: (
    previousScene: NavigationRoute,
    scene: TabScene
  ) => (
    previousScene: NavigationRoute,
    scene: TabScene,
    jumpToIndex: (index: number) => void
  ) => void,
```

to

```
  getOnPress: (
    scene: TabScene.
    previousScene: NavigationRoute,
  ) => (
    scene: TabScene,
    previousScene: NavigationRoute,
    jumpToIndex: (index: number) => void
  ) => void,
`

Comment (pietro909) [reply src/views/TabView/TabBarBottom.js]: I see your point, and to be honest I'm not sure what could be a sensible decision: generally, I prefer a meaningful order for function arguments but the breaking change is not a small one.

Given that this is still a _beta_ version, I'd be ready as a developer to accept breaking changes if needed.

File: src/views/TabView/TabBarTop.js [hunk 2]
```diff
@@ -131,6 +137,10 @@ export default class TabBarTop extends React.PureComponent<Props> {
     // TODO: Define full proptypes
     const props: any = this.props;
 
+    // TODO: get navigtion from props
```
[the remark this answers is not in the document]
pietro909: sorry, I forgot it!

src/views/TabView/TabBarBottom.js
spencercarli: I'm back and forth on this API change. I agree with @pietro909 in that this is beta software, therefore breaking changes will happen. This is also a relatively new feature of the library so I'm not sure how many people are actively using this.

On the flip side of that, we want to encourage confidence and minimize breaking changes and @vonovak's first solution would allow that.

Another question: If we do make a breaking change now, what can we do to ensure we don't have another breaking change in the future? Would it be to pass an object with the arguments in there (eliminates the order issue)?

pietro909: I find the idea of an object more scalable as a pattern, like:

    ({
        scene: TabScene,
        previousScene: NavigationRoute,
        jumpToIndex: (index: number)
    }) => void

This one offers two advantages:
1. order doesn't matter anymore
2. can add parameters without breaking the API

On the other hand, it makes less clear what is being passed if the consumer is not using Flow (or Typescript). But I think that this is the reason why we have docs :-)

spencercarli: Exactly - docs are there and we can give added flexibility in the future. @vonovak @kelset any hesitations on that API change? Still breaking but "better" breaking? 😄

kelset: ⛵️

pietro909: Do we have any news? Can I proceed with the object instead of single arguments?

kelset: Let's ping @vonovak ;) Btw I think you still need to remove the comment at line 140 commented by him a while back

vonovak: Yes, I was also thinking about that (using a single object argument), I think its okay to do it that way, so let's go ahead with that.

pietro909:
I've updated both the sources and the docs. Thanks everybody for the insights :-)

kelset:
@vonovak  @spencercarli  I think we are ready to ⛵️

spencercarli:
🎉

PR #2796, opened 2017-10-20
Add tests for navigators

jugglebird:
Add minimal unit tests for the DrawerNavigator, StackNavigator and TabNavigator.

This PR was motivated because I went to add unit tests in my own codebase to a class that inherited from StackNavigator and I wanted to see what StackNavigator's unit tests looked like. Turned out that there weren't any :-)

### DIFF
--- a/src/navigators/__tests__/.eslintrc
+++ b/src/navigators/__tests__/.eslintrc
@@ -1,0 +1,6 @@
+{
+  "extends": "../../../.eslintrc",
+  "env": {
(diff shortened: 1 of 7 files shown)


kelset:
This is awesome! Thanks for submitting :)
I'll leave @GantMan to review it since he's the one in charge of introducing testing - I think he is mostly working on having Detox setup so any Jest PR will surely be appreciated :)

Issue #1772, opened 2017-06-05
headerRight of navigationOptions not able to access "this.refs"


cfs5403:
### Current Behavior
the headerRight of navigationOptions not able to access to this.refs. I get this error "undefined is not an object (evaluating 'this.refs[WEBVIEW_REF]')". The alert method under refreshHandler() able to work but the "this.refs[WEBVIEW_REF].reload()" not work.

below is my code
```

import { WebView } from 'react-native'
import React, { PropTypes, Component } from 'react';
import { Modal, ActivityIndicator, View, Text, Button } from 'react-native';

var WEBVIEW_REF = 'webview';

class WebViewComponent extends Component {
    constructor(props) {
        super(props);
    }

    static navigationOptions = ({navigation}) => {
        const {params = {}} = navigation.state;
        return {
            title: navigation.state.params.title,
            headerRight: <Button
                             title="Refresh"
                             onPress={ () => params.handleRefresh() } />

        };
    };

    componentDidMount() {
        this.props.navigation.setParams({
            handleRefresh: this.refreshHandler
        });
    }

    refreshHandler() {
        alert('hello')
        this.refs[WEBVIEW_REF].reload()
    }

    render() {
        const {state} = this.props.navigation;
        const {url} = state.params;

        return (
            <WebView
                ref={ WEBVIEW_REF }
                style={ { flex: 1 } }
                source={ { uri: url } }
                scalesPageToFit={ true }
                startInLoadingState={ true }
                renderError={ () => this.loadError() } />
            );
    }


    loadError() {
        return (
            <View style={ { flex: 1, alignItems: 'center', justifyContent: 'center', margin: 20 } }>
                <Text style={ { fontSize: 18, fontWeight: 'normal', textAlign: 'center' } }>
                    Not able to load content. Please check your internet connection and try again.
                </Text>
            </View>
        )
    }

}

module.exports = WebViewComponent;

```
### Expected Behavior
Should able to call this.refs on headerRight.

### Your Environment
| software         | version
| ---------------- | -------
| react-navigation |  16.0.0-alpha.6
| react-native     |  0.44.2
| node             | 7.0.0
| npm or yarn      |3.10.8

Exilz:
It's `static` so it's not supposed to work, you can't reach the reference to your class in there.

cfs5403:
@Exilz so there is no way I can access this.refs? Because I need to the right button to refresh my webview. Thanks.

cfs5403:
[...]

Issue #1529, opened 2017-05-16
SetParams headerRight does not render a new button


abettadapur:
### Current Behavior

I cannot change the headerRight content using the setParams method. Here is my code:

```
this.props.navigation.setParams({
    headerRight: <TouchableOpacity>...</TouchableOpacity>
    title: "Selecting"
});
```
The header title changes to "Selecting" as expected, but no new content is rendered on the right side of the header. If I add the same headerRight item to the static navigationOptions object, content is rendered on the right as expected

### Expected Behavior

I should be able to dynamically set the content on the right of the header
### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |  1.0.0-beta9
| react-native     |  43
| node             | 7
| npm or yarn      | 4.*

tulin745:
I met the same problem.

BenBach:
I have the same problem. Any workaround?

tjeffree:
I'm using something like this as a workaround, but there probably is a nicer way:

```javascript
class SomeComponent extends Component {

  static navigationOptions = ({ navigation, screenProps }) => {

    const { state } = navigation
    const { signOutButton } = "params" in state && state.params

    return {
      headerRight: signOutButton && signOutButton()
    }
  }

  componentWillMount() {

    const signOutButton = (
      <Button onPress={() => { this.doSignOut() }} title="Sign Out"></Button>
    )

    this.props.navigation.setParams({ signOutButton: () => signOutButton })
  }
}
```

BenBach:
Thanks works great!

yidc:
This just solve my problem! Thanks!

efkan:
There might be 2 reasons that prevent the changes from being shown:

1. Using PureComponents

An explanation is placed on [VirtualizedList](https://facebook.github.io/react-native/docs/virtualizedlist.html#virtualizedlist) page.

> This is a PureComponent which means that it will not re-render if props remain shallow- equal. Make sure that everything your renderItem function depends on is passed as a prop (e.g. extraData) that is not === after updates, otherwise your UI may not update on changes. This includes the data prop and parent component state.

2. Parameters changes doesn't detect by React-Native because of nested objects
So if the params object is not being changed, React-Native doesn't render the changes.
The same issue valids for [Redux store](http://redux.js.org/docs/recipes/reducers/ImmutableUpdatePatterns.html#updating-nested-objects).

I think that one of or both of these two reasons might cause the rendering problem.

[...]

PR #2806, opened 2017-10-22
Add 'save-pattern' to docs

Jacse:
As evident with the activity on #145 people are unsure how to implement this popular pattern. I took the [recommended solution](https://github.com/react-community/react-navigation/issues/145#issuecomment-304649551), improved it and added it to the docs.

I tried to copy the writing style of the other docs, let me know if I'm off somewhere.

Code has been tested and verified.

Discussed with @kelset.


Jacse:
Further issues created due to confusion on how to implement this: #1989, #1772, #1529

docs/guides/Guide-Headers.md
```diff
@@ -79,4 +79,61 @@ Now, the header can interact with the screen route/state:
...
+
+Sometimes it is necesarry for header to access properties of the screen component such as functions or state.
+
+Let's say we want to create an 'edit contact info' screen with a save button in the header. We want the save button to be replaced by an `ActivityIndicator` while saving.
+
+```js
+class EditInfoScreen extends React.Component {
+  static navigationOptions = ({ navigation }) => {
+    const { params = {} } = navigation.state;
+    let headerRight = (
```
matthamil: Why not use a ternary with `const` instead of reassignment with let?

Jacse: I think using ternary in the return object gets confusing, but we could do:

```js
static navigationOptions = ({ navigation }) => {
  const { params = {} } = navigation.state;
  const headerRight = params.isSaving ? <ActivityIndicator /> : (
    <Button
      title="Save"
      onPress={params.handleSave ? params.handleSave : () => null}
    />
  );
  return { headerRight };
};
```

Would you prefer this?

kelset: I agree with @Jacse in this scenario: since it's an example, it's easier for a reader to follow what's happening if there is a big "if statement" clearly telling you what's going to be further explained - in the phrase after the code block.

docs/guides/Guide-Headers.md
```diff
@@ -79,4 +79,60 @@ Now, the header can interact with the screen route/state:
 header-interaction
 ```
 
+### Header interaction with screen component
+
+Sometimes it is necesarry for header to access properties of the screen component such as functions or state.
```
kelset: typo: necesarry -> necessary

Jacse: fixed

hnordt:
@Jacse could you explain why the button would "flick" without this line:

```jsx
onPress={params.handleSave ? params.handleSave : () => null}
```

?

Without this line the initial result would be `onPress={undefined}` which would display a warning. The warning is what you mean with the word "flick"?

Jacse:
@hnordt Bad choice of words, the initial text talked about the button not being rendered without a function. The line is still important, though: It's poor practice to supply a component with `undefined` when a function is expected and if a user (theoretically) clicks the button before `handleSave` is set *I believe the app crashes.

hnordt:
@Jacse thank you for the explanation!

PR #2825, opened 2017-10-25
Added iconContainerStyle prop to Drawer

brunocascio:
I am trying to remove the icon opacity when an item is inactive. Currently, it is not possible so far.
This PR adds the ability to set custom styles inside of the icon container.


spencercarli:
This looks good to me! @kelset or @matthamil anything I'm missing?

PR #2827, opened 2017-10-25
Fix handle Android back button example in docs

dharik:
I was following the documentation to make the back button work on Android when react-navigation is integrated with redux.  The example code triggers an error!

`dispatch` isn't available in onBackPress, it comes from props!

**Test plan (required)**
I tested on Android. Before the fix, it showed an error. After, no more error!

### DIFF
--- a/docs/guides/Redux-Integration.md
+++ b/docs/guides/Redux-Integration.md
@@ -115,7 +115,7 @@ class ReduxNavigation extends React.Component {
     if (nav.index === 0) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


spencercarli:
`dispatch` is made available on line 14 so there's no need to access it on `this.props`. I'm going to close this since it isn't a necessary update but please don't let that discourage you from making future contributions!

dharik:
Haha oops, that's embarassing!

PR #2833, opened 2017-10-27
Add iPhone X SafeAreaView

davepack:
This adds support for iPhone X to react-navigation via SafeAreaView. 
- Uses a JS approach that can be used in RN v0.49 and lower. 
- Includes `forceInset` prop, which is not available on the native version that will be released in RN v0.50. 
- Adds SafeAreaView to Header, Drawer, and Tab views. 
- Updates NavigationPlayground to provide testing and examples of SafeAreaView use.

This essentially mimics the native version by adding padding on the side of the view that touches a screen edge. Ideally the API introduced here will eventually be migrated into the native version of SafeAreaView where performance will be much better.

### Native SafeAreaView shortcomings:
* Only available in RN v0.50, which will be released maybe a day or two before iPhone X is released.
* Inset is added as padding and cannot be customized as margin (e.g. padding on sides, margin on top).
* Inset cannot be overridden to disable inset on any side.
* Fails with react-navigation tab views (tab screens are laid end-to-end off the screen and SafeAreaView did not add inset properly).

### JS SafeAreaView benefits:
* Works on React Native v0.49 and earlier, devs can start adding iPhone X support now.
* Adds `forceInset` prop to customize the behavior of the inset.
    * Customize the size of the spacing to make it smaller or larger.
    * Disable inset on any side so you can nest SafeAreaViews for a padding + margin effect.
    * Force inset to always show on any side in case there is an unexpected behavior (e.g. animated views will weirdly animate the additional padding when a view hits an edge).
    * Adds the specified padding based on the orientation and disregards whether the view is touching a screen edge or not.

#### Why don't we automatically wrap the screens in SafeAreaView to decrease dev workload?
This will behave like adding a margin to the screen. If background color matches, it won't be noticeable. If background color is different or any kind of scrolling view is used, the SafeAreaView padding will appear like a margin and cover the scrolling content.
#### What should I do about ScrollView and list views?
ScrollView now has `contentInsetAdjustmentBehavior`. See NavigationPlayground App.js for an example of this. https://facebook.github.io/react-native/docs/scrollview.html#contentinsetadjustmentbehavior

### DIFF
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -14,7 +14,8 @@ const { isIPhoneX_deprecated } = DeviceInfo;
 const X_WIDTH = 375;
 const X_HEIGHT = 812;
 
-const { minor } = NativeModules.PlatformConstants.reactNativeVersion;
+const { PlatformConstants = {} } = NativeModules;
+const { minor = 0 } = PlatformConstants.reactNativeVersion || {};
 
 const isIPhoneX = (() => {
   if (minor >= 50) {
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -11,8 +11,13 @@ import {
 import withOrientation from './withOrientation';
 
 const { isIPhoneX_deprecated } = DeviceInfo;
+// See https://mydevice.io/devices/ for device dimensions
 const X_WIDTH = 375;
 const X_HEIGHT = 812;
+const PAD_WIDTH = 768;
+const PAD_HEIGHT = 1024;
+
+const { height: D_HEIGHT, width: D_WIDTH } = Dimensions.get('window');
 
 const { PlatformConstants = {} } = NativeModules;
 const { minor = 0 } = PlatformConstants.reactNativeVersion || {};
@@ -22,13 +27,40 @@ const isIPhoneX = (() => {
     return isIPhoneX_deprecated;
   }
 
-  const { height, width } = Dimensions.get('window');
   return (
     Platform.OS === 'ios' &&
-    ((height === X_HEIGHT && width === X_WIDTH) ||
-      (height === X_WIDTH && width === X_HEIGHT))
+    ((D_HEIGHT === X_HEIGHT && D_WIDTH === X_WIDTH) ||
+      (D_HEIGHT === X_WIDTH && D_WIDTH === X_HEIGHT))
   );
 })();
+
+const isIPad = (() => {
+  if (Platform.OS !== 'ios' || isIPhoneX) return false;
+
+  // if portrait and width is smaller than iPad width
+  if (D_HEIGHT > D_WIDTH && D_WIDTH < PAD_WIDTH) {
+    return false;
+  }
+
+  // if landscape and height is smaller that iPad height
+  if (D_WIDTH > D_HEIGHT && D_HEIGHT < PAD_WIDTH) {
+    return false;
+  }
+
+  return true;
+})();
+
+const statusBarHeight = isLandscape => {
+  if (isIPhoneX) {
+    return isLandscape ? 0 : 44;
+  }
+
+  if (isIPad) {
+    return 20;
+  }
+
+  return isLandscape ? 0 : 20;
+};
 
 class SafeView extends Component {
   state = {
@@ -45,24 +77,12 @@ class SafeView extends Component {
     });
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { isLandscape, verbose } = nextProps;
-    if (verbose) console.log(isLandscape);
+  componentWillReceiveProps() {
     this._onLayout();
   }
 
   render() {
-    const {
-      forceInset = false,
-      isLandscape,
-      children,
-      style,
-      verbose,
-    } = this.props;
-
-    // if (verbose) {
-    //   console.log(isLandscape);
-    // }
+    const { forceInset = false, isLandscape, children, style } = this.props;
 
     if (!Platform.OS === 'ios') {
       return <View style={style}>{this.props.children}</View>;
@@ -88,7 +108,7 @@ class SafeView extends Component {
   _onLayout = () => {
     if (!this.view) return;
 
-    const { isLandscape, verbose } = this.props;
+    const { isLandscape } = this.props;
     const { orientation } = this.state;
     const newOrientation = isLandscape ? 'landscape' : 'portrait';
     if (orientation && orientation === newOrientation) {
@@ -116,12 +136,6 @@ class SafeView extends Component {
 
       while (realX < 0) {
         realX += WIDTH;
-      }
-
-      if (verbose) {
-        console.log(realY, winHeight);
-        console.log('');
-        console.log(realX, winWidth);
       }
 
       const touchesTop = realY === 0;
@@ -199,7 +213,7 @@ class SafeView extends Component {
       }
       case 'vertical':
       case 'top': {
-        return isLandscape ? 0 : isIPhoneX ? 44 : 20;
+        return statusBarHeight(isLandscape);
       }
       case 'bottom': {
         return isIPhoneX ? (isLandscape ? 24 : 34) : 0;
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -122,20 +122,16 @@ class SafeView extends Component {
       let realY = winY;
       let realX = winX;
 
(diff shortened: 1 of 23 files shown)


kelset:
Hey @davepack awesome work!
I've glanced at the changes quickly and I have to say that it looks like it's probably going to be quite heavy on the other PRs (as far as minor merge conflicts, I mean), but worth it nonetheless.

I've seen you have modified also some things here and there about the header behaviour & such, and that's why I've tagged @spencercarli ;D

One last thing: this is basically a new thing that devs need to "add" to their code, right? I mean, they'd need to wrap their Views with `SafeAreaView` in some way... right? 🤔 
Maybe I'm wrong about this, but if I'm right I think it should appear in the docs somewhere.

davepack:
Yeah, I think it'll be good to get several people to weigh in on this.

And yes, we'll need to add this to docs. I figured getting the review process started right away would be good. I can start on docs now while we review this.

kelset:
Awesome.

Tbh I think it's best to wait for @Ashoat to submit his PR with the flow fixes + support for 0.49 + support for the new tab-view version, and then update this lib to align with it, and then merge it.
It's probably a bit of overhead but to me his PR has an higher priority since it clean up the codebase. 

Anyway, yeah, many more reviews of this PR should come from awesome ppl... like @vonovak ;D

gaodeng:
iPhone X support is very urgent. awesome 👍👍👍

src/views/SafeAreaView.js
```diff
@@ -0,0 +1,225 @@
...
+
+  // if landscape and height is smaller that iPad height
+  if (D_WIDTH > D_HEIGHT && D_HEIGHT < PAD_WIDTH) {
+    return false;
+  }
+
+  return true;
+})();
+
+const statusBarHeight = isLandscape => {
```
vonovak: is this a reliable indicator of the status bar height, or do you mean to use https://github.com/facebook/react-native/pull/16478 when it's available?

src/views/SafeAreaView.js
```diff
@@ -0,0 +1,225 @@
...
+    const { orientation } = this.state;
+    const newOrientation = isLandscape ? 'landscape' : 'portrait';
+    if (orientation && orientation === newOrientation) {
+      return;
+    }
+
+    const WIDTH = isLandscape ? X_HEIGHT : X_WIDTH;
+    const HEIGHT = isLandscape ? X_WIDTH : X_HEIGHT;
+
+    this.view.measureInWindow((winX, winY, winWidth, winHeight) => {
```
vonovak: is onLayout returning inaccurate {x, y, width, height}, or what is the reason for calling the `measureInWindow`?

src/views/SafeAreaView.js
```diff
@@ -0,0 +1,225 @@
...
+
+      while (realY < 0) {
+        realY += HEIGHT;
+      }
+
+      while (realX < 0) {
+        realX += WIDTH;
+      }
+
+      const touchesTop = realY === 0;
```
vonovak: is strong equality the right approach or should we rather use `const touchesTop = realY < someSmallNumber`?

src/views/SafeAreaView.js
```diff
@@ -0,0 +1,225 @@
...
+
+    this.view.measureInWindow((winX, winY, winWidth, winHeight) => {
+      let realY = winY;
+      let realX = winX;
+
+      while (realY >= HEIGHT) {
+        realY -= HEIGHT;
+      }
+
+      while (realX >= WIDTH) {
```
vonovak: maybe a dumb question; when does it happen that the realX is greater than width? Also, I might be wrong, but couldn't the `while`s be replaced with modulo?

src/views/SafeAreaView.js
davepack: onLayout returns the component's position relative to parent components, so if a parent is offset from the top by 40 pixels, but the child fills the whole parent, the onLayout in the child will return a y coordinate of 0, even thought it's not at the top of the window.

src/views/SafeAreaView.js
davepack: Not a dumb question, and this one threw me for a loop. It's because in a TabNavigator the views get placed side by side and then an x-offset is changed to move the correct screen into the view port. TabView does it this way so it can easily enable swiping.

davepack: Also, you're right, I'll update it to use modulo. :eyeroll:

src/views/SafeAreaView.js
davepack: This is worth exploring. In my testing this worked as expected, but devs might need a threshold of some sort. I think it's fine as is for now.

src/views/SafeAreaView.js
davepack: Funny, one of the comments on that PR says to rely on SafeAreaView ;-)

The idea right now is that SafeAreaView will eventually handle this in native-land and devs who use SafeAreaView will never have to worry about the status bar again. This is for sure a naive solution, but it is actually an improvement on what React Nav does right now. It does need to get better.

spencercarli:
On an iPhone 7 with iOS 10.3 the icons extend beyond the tab bar. This can be seen in the `NavigationPlayground` example "Stacks in Tabs" (or any demo with tabs)
iphone 7 ios 10 3

spencercarli:
Tabbar bottom's icons on iOS 10.3 (confirmed on iphone se and 7) extend beyond the tabbar

iphone se ios 10 3
iphone 7 ios 10 3

src/views/TabView/TabBarBottom.js
```diff
@@ -179,64 +169,84 @@ class TabBarBottom extends PureComponent<DefaultProps, Props, void> {
...
     borderTopWidth: StyleSheet.hairlineWidth,
     borderTopColor: 'rgba(0, 0, 0, .3)',
-    backgroundColor: '#F7F7F7', // Default background color in iOS 10+
+  },
+  tabBar: {
+    // height: 49, // Default tab bar height in iOS 10+
+    flexDirection: 'row',
+  },
+  tabBarLandscape: {
+    height: 29,
```
spencercarli: On versions of iOS under 11 this should be 49 because the tab items will be stacked.

davepack: Hmm, k, I'll make that change. On a related note, tab icons should also be slightly smaller on iOS 11 in landscape. Doesn't seem easy to implement that as is. @vonovak had a solution for withStatusBar that had a render prop that passed status bar height. I think a future improvement for SafeAreaView would be a render prop that passed the screen size class or something, with which you could update the tab icon size. I'll open an issue at some point.

danieljvdm:
@davepack thanks for this awesome PR! One question, are you experiencing any "jump" when a screen first loads and causes the tab bar height to be too tall and then after a second goes to normal height?
Attached photos.
simulator screen shot - iphone x - 2017-11-02 at 14 50 52
simulator screen shot - iphone x - 2017-11-02 at 14 50 53

kelset:
@danieljvdm `beta.16` has some known issues, can you try pointing at master to see if it's already been fixed by one of the commits we did after the release? Or open a new issue (following the issue template ofc)?

danieljvdm:
@kelset no dice, I'll open an issue

dantman:
The way this was added to `DrawerNavigatorItems` causes breakage when a custom container is used to provide a navigation header or separate drawer items into sections.

screen shot 2017-11-10 at 8 17 41 pm

aphillipo:
Hello @davepack - I have an issue with changing the TabBar background colour and the iPhone X home indicator.  Basically the gap on the SafeAreaView at the header is fine and clear how to style, on the TabBar it's incredibly unclear how to change the background colour from the iOS default colour (I've literally tried everything I can think of).  Thanks for your work on this issue and hopefully there is a simple answer to this!  I can do a pull request to the documentation once I know what to do or if this is a new issue! Thanks!

kelset:
Hey @aphillipo I think it would be better if you open a new issue (following the template) so it's easier for anyone to keep track of this "customization" you are describing ;)

andreyslyusar:
@aphillipo to update TabBar bottom SafeAreaView colour you need to update tabBarOptions for TabBar:
```
export const navigatorConfig = {
	tabBarOptions: {
		style: {
			backgroundColor: 'your colour',
		},
	},
};
```

aphillipo:
@andreyslyusar Legend, works perfectly!

PR #2834, opened 2017-10-27
Update goBack docs

tienf4m:
I think many users don't really understand what `go back anywhere` means. From my understanding it actually means go back from the currently active route. What do you guys think?


kelset:
Thanks, I'll properly go through it tomorrow.

docs/guides/Screen-Navigation-Prop.md
```diff
@@ -90,7 +90,29 @@ class ProfileScreen extends React.Component {
...
+```... 
+navigation.navigate(SCREEN_KEY_A);
+...
+navigation.navigate(SCREEN_KEY_B);
+...
+navigation.navigate(SCREEN_KEY_C);
+...
+navigation.navigate(SCREEN_KEY_D);
+```
+
```
fabriziomoscon: It would be useful to see what is the navigation stack history after each `goBack()` calls.

tienf4m: Also I just realized that this part is a bit misleading, because what you pass into navigate is actually just the route name, but what you want to pass into goBack is actually a key generated by the routers, and you can access it with`navigation.state.key`.
So I suggest we simplify the navigation stack as:
ScreenA -> ScreenB -> ScreenC -> ScreenD
And also explain that the key that we pass into goBack is from navigation.state.key.

spencercarli: That's a good point @tienf4m. Want to update this PR to reflect that?

PR #2843, opened 2017-10-29
Set initialLayout in defaultProps on Android

Minishlink:
This is a follow-up of the last two PRs (#2733 and #2385) that adresses [this problem](https://github.com/react-native-community/react-native-tab-view/issues/312).

This:
* fixes the `initialLayout` prop conflicts
* sets the default `initialLayout` in `defaultProps` rather than on every render (and doesn't mutate props)
* uses the window's width and height to avoid the one frame delay

### DIFF
--- a/src/views/TabView/TabView.js
+++ b/src/views/TabView/TabView.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import React, { PureComponent } from 'react';
-import { View, StyleSheet, Platform } from 'react-native';
+import { View, StyleSheet } from 'react-native';
 import { TabViewAnimated, TabViewPagerPan } from 'react-native-tab-view';
 import type { Layout } from 'react-native-tab-view/src/TabViewTypeDefinitions';
 import SceneView from '../SceneView';
@@ -20,7 +20,6 @@ export type TabViewConfig = {
   tabBarComponent?: ReactClass<*>,
   tabBarPosition?: 'top' | 'bottom',
   tabBarOptions?: {},
-  initialLayout?: { width: number, height: number },
   swipeEnabled?: boolean,
   animationEnabled?: boolean,
   lazy?: boolean,
@@ -38,7 +37,6 @@ type Props = {
   tabBarComponent?: ReactClass<*>,
   tabBarPosition?: 'top' | 'bottom',
   tabBarOptions?: {},
-  initialLayout?: { width: number, height: number },
   swipeEnabled?: boolean,
   animationEnabled?: boolean,
   lazy?: boolean,
@@ -208,9 +206,6 @@ class TabView extends PureComponent<void, Props, void> {
       screenProps: this.props.screenProps,
       style: styles.container,
     };
-    if (Platform.OS === 'android') {
-      props.initialLayout = { width: 1, height: 0 };
-    }
 
     return <TabViewAnimated {...props} />;
   }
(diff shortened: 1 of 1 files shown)


src/views/TabView/TabView.js
```diff
@@ -56,7 +54,18 @@ type Props = {
...
 };
 
-class TabView extends PureComponent<void, Props, void> {
+let defaultInitialLayout = undefined;
+if (Platform.OS === 'android') {
+  // fix for https://github.com/react-native-community/react-native-tab-view/issues/312
+  const { width, height } = Dimensions.get('window');
+  defaultInitialLayout = { width, height };
+}
+
```
vonovak: could you please change this to something like

```
const windowDimensions = Dimensions.get('window')
const defaultInitialLayout = Platform.select({
  ios: undefined,
  android: { width: windowDimensions.width, height: windowDimensions.height },
  android: { width: 1, height: 0 } // or rather this because that's what they say here https://github.com/react-native-community/react-native-tab-view/issues/297
})
```

src/views/TabView/TabView.js
```diff
@@ -20,7 +20,6 @@ export type TabViewConfig = {
   tabBarComponent?: ReactClass<*>,
   tabBarPosition?: 'top' | 'bottom',
   tabBarOptions?: {},
-  initialLayout?: { width: number, height: number },
```
vonovak: keep this here, but change the type to `Layout` which can be imported from `import type { Layout } from 'react-native-tab-view/src/TabViewTypeDefinitions';`

Ken-Lim:
I have been tracking this problem, this fix does introduced an artifact on the initial rendering -  the first tab was push to the left then back to normal position.
Please see the below GIF for the problem
tab_problem

vonovak:
@Ken-Lim thanks for commenting. Does using `{ width: 1, height: 0 }` make it better?

Minishlink:
@vonovak It adresses the same problem indeed. But without the screen dimensions, the tabs' content is hidden during the initial frame. Since it's a different problem, I'll put it back to { width: 1, height: 0 } :)

@Ken-Lim, are you talking about this PR or the current state of master?

src/views/TabView/TabView.js
Minishlink: It's already in the props type 3 lines below ;)

src/views/TabView/TabView.js
```diff
@@ -56,7 +54,14 @@ type Props = {
   },
 };
 
-class TabView extends PureComponent<void, Props, void> {
+class TabView extends PureComponent<$Shape<Props>, Props, void> {
+  static defaultProps = {
+    // fix for https://github.com/react-native-community/react-native-tab-view/issues/312
+    initialLayout: Platform.select({
+      android: { width: 1, height: 0 },
```
vonovak: Please be explicit and include `ios: undefined`. Thanks

src/views/TabView/TabView.js
vonovak: huh, good catch :)

src/views/TabView/TabView.js
Minishlink: I should add `web: undefined` also then?

vonovak: I see your point. Let's keep it.

vonovak:
thanks @Minishlink !

Ken-Lim:
@vonovak, using `{ width: 1, height: 0 }` doesn't helps in my case.

@Minishlink, the behaviour is observed with this PR and current master.

vonovak:
@Ken-Lim you'll probably have to pass your own custom initialLayout

PR #2870, opened 2017-11-01
Allow overrides of route parameters on BACK navigation action

moriyoshi:
It would be very handy, while it might disagree with the navigation concept, if route parameters could be overridden on the dispatch of BACK navigation action where the parameters are used to statefully render the header contents.

This patch adds `params` parameter to `NavigationSetParamsAction` and allows one to do so for StackRouter.

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -22,9 +22,12 @@ const RESET = 'Navigation/RESET';
 const SET_PARAMS = 'Navigation/SET_PARAMS';
 const URI = 'Navigation/URI';
 
-const back = (payload: { key?: ?string } = {}): NavigationBackAction => ({
+const back = (
+  payload: { key?: ?string, params?: ?NavigationParams } = {}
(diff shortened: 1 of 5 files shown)


brentvatne:
@ericvicenti had some thoughts about this, which I will attempt to summarize:

- when idempotent navigation is properly implemented, you should be able to just navigate back with the params, making this PR unnecessary
- if you pass params into goBack, you aren't sure exactly where those params are going (what is the previous screen? you may not know). this isn't necessarily bad because it allows some route to be generic, but it may lead to some confusing behaviors, eg: screen A navigates to B and A uses some param x to calculate its title. screen B at some point calls goBack and passes param x, which clobbers the x param on A. so, screen A needs to have knowledge of what might be passed back to it from B. again, this might not be bad, it's just something we should acknowledge.

ericvicenti:
Ok, closing this out in favor of idempotent navigate action

PR #2880, opened 2017-11-02
fix for breakage in 185289ad6658e07783c807ee7ac29b9e1244fa21.

pvinis:
that commit broke the possibility to have any kind of different separator on the bottom of the header.
this commit restores that.

like before, you can have in `headerStyle` something like `borderBottomWidth: 3, borderBottomColor: 'red'` for a custom separator, or `borderBottomWidth: 0` for no separator.

fixes a regression introduced in 185289ad6658e07783c807ee7ac29b9e1244fa21.

**Test plan (required)**

add what i have above for testing.

**Code formatting**

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -308,6 +308,7 @@ class Header extends React.PureComponent<Props, State> {
     } = options;
     const appBarHeight = Platform.OS === 'ios' ? (isLandscape ? 32 : 44) : 56;
     const containerStyles = [
+      styles.container,
       {
         height: appBarHeight,
       },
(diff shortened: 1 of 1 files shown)


pvinis:
@davepack

davepack:
I have also created a PR: https://github.com/react-community/react-navigation/pull/2886

It seems like unexpected things could potentially happen if you apply the backgroundColor in the container but put styles like border, shadow, and elevation in a child view.

pvinis:
is yours working with changing `borderBottomWidth` and `borderBottomColor`?
does it need these to be outside `headerStyle`?

aforty:
Tested as well, looks good here. `SafeAreaView` really doesn't need a set style.

kelset:
Thanks for this PR, but I think we'll go for #2886 for fixing the header styling behaviour.

kelset:
@pvinis @aforty we worked a new PR which comprehend this change too, this one #2889; I think this can be closed in favor of that one.

brentvatne:
resolved in 1.0.0-beta.17 after merging https://github.com/react-community/react-navigation/pull/2889

PR #2886, opened 2017-11-02
Update header and tab bar with container style prop

davepack:
Background color was not enough customization on the default header and tab bars. This replaces those props with headerContainerStyle and containerStyle on header and tabbar respectively.

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -135,9 +135,9 @@ Style object for the header
 
 Style object for the title component
 
-### `headerBackgroundColor`
+### `headeContainerStyle`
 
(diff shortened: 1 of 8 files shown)


docs/api/navigators/StackNavigator.md
```diff
@@ -135,9 +135,9 @@ Style object for the header
 
 Style object for the title component
 
-### `headerBackgroundColor`
+### `headeContainerStyle`
```
kelset: Two things:

davepack:
Going a different direction now to fix, so I'll close this.

davepack:
new PR is https://github.com/react-community/react-navigation/pull/2889

PR #2889, opened 2017-11-02
SafeAreaView padding from style prop

### LINKED ISSUE #2875: Header elevation styling to remove shadow does not work in beta16
### Current Behavior
Issue first raised in:
https://github.com/react-community/react-navigation/issues/2858

Subsequently, along with other raised issues, all got logged under:
https://github.com/react-community/react-navigation/issues/2854

Which led to the following PR:
https://github.com/react-community/react-navigation/pull/2871

The PR addressed the background color issues, but not the elevation styling on the header.

### Expected Behavior

Should be able to set elevation in:
- the headerStyle of nevigationOptions
- tabBarOptions.style (for TabNavigators)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0.0-beta.16
| react-native     | 0.49.3
| node             | 8.4.0
| npm or yarn      | 0.27.5

### LINKED ISSUE #2888: TabBarBottom changing height during renders on screen mount
### Current Behavior
TabBarBottom initially has a larger height than expected, as shown by the first screenshot. After a second it corrects to the second screenshot. Happens on iPhone X as well as other iPhone devices.

simulator screen shot - iphone x - 2017-11-02 at 14 50 52
simulator screen shot - iphone x - 2017-11-02 at 14 50 53

### Expected Behavior
Expect no change of height during render

### Your Environment

Tested on both beta 16 and latest master (f899b2eb13a48e7f84250f2e6f59ef2e5a5b9085).
Tested on iPhone X simulator and iPhone 8 simulator
React Native 0.49.3

davepack:
SafeAreaView now adds padding specified in the style prop. This eliminates the need to have extra nested Views to add padding.

backgroundColor no longer needs to be specified in a separate prop on stack navigator and tab bar.

should fix #2888 (testing needed)
should also fix #2875


src/views/SafeAreaView.js
```diff
@@ -88,17 +89,13 @@ class SafeView extends Component {
       return <View style={style}>{this.props.children}</View>;
     }
 
-    if (!forceInset && minor >= 50) {
```
brentvatne: so we no longer use react-native's built-in SafeAreaView?

src/views/SafeAreaView.js
```diff
@@ -153,7 +150,19 @@ class SafeView extends Component {
     const { touchesTop, touchesBottom, touchesLeft, touchesRight } = this.state;
     const { forceInset, isLandscape } = this.props;
 
+    const {
```
brentvatne: would be good to have a comment about this destructuring block to clarify why we're setting the default values like this

src/views/SafeAreaView.js
```diff
@@ -195,6 +204,21 @@ class SafeView extends Component {
       });
     }
 
+    // new height/width should only include padding from insets
+    // height/width should not be affected by padding from style obj
+    if (style.height && typeof style.height === 'number') {
```
brentvatne: confused about what is going on here and below

spencercarli:
screenshot_1509660040

Navbar is behind statusbar in the NavigationPlayground example on android

src/views/SafeAreaView.js
davepack: when height is specified on the SafeAreaView, height is supposed to include padding, so the inset padding makes the view much smaller than it's supposed to be. This is the problem I was having with the header. Making the height include the inset padding fixes this, but also doesn't break the expectation that the specified padding from style does not make the component larger than that height.

davepack: also, height can be a string for percent. This doesn't handle that case.

davepack:
@spencercarli that is a known issue https://github.com/react-community/react-navigation/issues/12, also apparently pretty old. SafeAreaView should definitely handle this, but it's not a regression. If there's a quick way to add that in I'll do it, but I haven't looked at that issue yet and I don't have a solution.

src/views/SafeAreaView.js
davepack: Not for this version. It doesn't support padding.

spencercarli:
Ah I see 😄 . I don't know what the fix is either and since it's not a regression I don't think we should block this pr because of it

src/views/SafeAreaView.js
davepack: added

davepack:
I just remembered padding can be a string with percent, I need to figure out how to handle that.

brentvatne:
A couple of subtle, undesirable differences in before/after of this PR:

## Navigation bar with margin (before)

screen shot 2017-11-02 at 6 13 45 pm

## After

screen shot 2017-11-02 at 6 14 37 pm

## Tab bar with margin (before)

screen shot 2017-11-02 at 6 14 09 pm

## After
screen shot 2017-11-02 at 6 14 30 pm

Realistically, a margin on navigation bar and tab bar doesn't make too much sense -- don't know what kind of thing people would expect to be solving by doing this. So it isn't a blocker. However we should continue to flesh out the SafeAreaExample app to add more test cases and see where behavior isn't what we expect, then resolve those.

Further, if you test the app on Android you will notice that behavior is in some cases inconsistent between iOS and Android, we should resolve these wherever it makes sense to do so.

B3rry:
screen shot 2018-03-01 at 1 38 36 pm

@brentvatne: We're currently using a margin on a navigation bar for a modal. Applying a margin to Tab Bar is for sure a use case. Header still seems to not be inheriting `margin-top` as of `react-navigation@1.2.1`. Noticed this inconsistency with regards to sizing. [#3184](https://github.com/react-navigation/react-navigation/issues/3184) has context.

PR #2890, opened 2017-11-03
Safe area example

brentvatne:
Adds an example project for us to test out safe area view in various contexts


kelset:
Awesome! Thanks @brentvatne !

PR #2906, opened 2017-11-05
Disable / enable swipe depending on screen.

cmpayc:
Fix react-community#1760

Problem: if you use the StackNavigator in the TabNavigator and you are at that time on the StackNavigator screen, the swipe will switch the screens to the TabNavigator.

Solution:
Adds a new option in the ```Screen Navigation Options``` in ```TabNavigator```:

**```swipeEnabled ```**
True or false to enable or disable swiping between tabs, if not set then defaults to TabNavigatorConfig option swipeEnabled .


src/views/TabView/TabView.js
```diff
@@ -185,6 +184,9 @@ class TabView extends React.PureComponent<Props> {
 
     const tabBarVisible =
       options.tabBarVisible == null ? true : options.tabBarVisible;
+      
+    const swipeEnabled =
+      options.swipeEnabled == null ? this.props.swipeEnabled : options.swipeEnabled;
```
spencercarli: `swipeEnabled` should have a default prop set to `true` so that this wouldn't be a breaking change.

cmpayc: The value of ```props.swipeEnabled``` is taken from the TabNavigator.js, where it is already set to Presets value (for Android is ```true```, but for iOS is ```false```)

spencercarli: Ah okay great! I had confused myself with the `tabBarVisible` above. Once the docs are added I will merge. Thank you!

cmpayc: Adds a new option in the ```Screen Navigation Options``` in ```TabNavigator```:

**```swipeEnabled ```**
True or false to enable or disable swiping between tabs, if not set then defaults to TabNavigatorConfig option swipeEnabled .

nicholasc:
Is this on hold because of documentation? If so, I am willing to do that.

cmpayc:
Documentation updated. CircleCI errors fixed.

spencercarli:
Thanks!

nicholasc:
Is this going to be part of this week's release or next one?

spencercarli:
It will be part of the next release, which should happen sometime next week.

StepanNaryshkov:
I can not find any examples in the documentation. Can anyone show me how to use this feature or link to the documentation where is this feature?

StepanNaryshkov:
@cmpayc Does this feature disable all my StackNavigator ? For example, I want to use swipe only tabs and if I click on a screen on a tab I want to disable it, it is possible?
Tab(swiped)>mainScreen(swiped)>child of mainScreen(disabled)

PR #2942, opened 2017-11-09
Reduce unnecessary re-rendering by injecting a smart shouldComponentUpdate.

exKAZUu:
Fixes partially #608 (i.e. only for _(point 1)_ described in https://github.com/react-community/react-navigation/issues/608#issuecomment-328635042)

## Motivation

This PR reduces unnecessary re-rendering with Stack Navigator by injecting a smart `shouldComponentUpdate` which judges whether the given scene is active or not.

Please note that I tried to use `StaticContainer` like https://github.com/react-community/react-navigation/issues/608#issuecomment-328689388, but it cannot prevent from re-rendering child components which refers to state managed by Redux.

## Test plan

~I tested this PR in my modified `ReduxExample` project which has a `TextInput` component for updating `state` managed by Redux.~
~I confirmed that only one component was re-rendered when updating state which multiple components refers to by entering texts in the `TextInput`.~
~If needed, I can share my modified `ReduxExample` for testing purpose (but its code is a bit dirty).~

Please see https://github.com/react-community/react-navigation/pull/2942#issuecomment-343527899.

### DIFF
--- a/src/views/CardStack/CardStack.js
+++ b/src/views/CardStack/CardStack.js
@@ -400,6 +400,7 @@ class CardStack extends React.Component<Props> {
               screenProps={screenProps}
               navigation={navigation}
               component={SceneComponent}
+              scene={scene}
             />
           </View>
           {this._renderHeader(scene, headerMode)}
@@ -411,6 +412,7 @@ class CardStack extends React.Component<Props> {
         screenProps={this.props.screenProps}
         navigation={navigation}
         component={SceneComponent}
+        scene={scene}
       />
     );
   }
--- a/src/views/SceneView.js
+++ b/src/views/SceneView.js
@@ -7,12 +7,14 @@ import type {
   NavigationScreenProp,
   NavigationComponent,
   NavigationRoute,
+  NavigationScene,
 } from '../TypeDefinition';
 
 type Props = {
   screenProps?: {},
   navigation: NavigationScreenProp<*>,
   component: NavigationComponent,
+  scene?: NavigationScene,
 };
 
 export default class SceneView extends React.PureComponent<Props> {
@@ -27,8 +29,17 @@ export default class SceneView extends React.PureComponent<Props> {
   }
 
   render() {
-    const { screenProps, navigation, component: Component } = this.props;
+    const { screenProps, navigation, component: Component, scene } = this.props;
 
-    return <Component screenProps={screenProps} navigation={navigation} />;
+    const comp: any = (
+      <Component screenProps={screenProps} navigation={navigation} />
+    );
+    const originalMethod = comp.type.prototype.shouldComponentUpdate;
+    if (originalMethod && scene) {
+      comp.type.prototype.shouldComponentUpdate = function(...args: [*]) {
+        return scene.isActive && originalMethod.apply(this, args);
+      };
+    }
+    return comp;
   }
 }
--- a/src/views/SceneView.js
+++ b/src/views/SceneView.js
@@ -34,10 +34,13 @@ export default class SceneView extends React.PureComponent<Props> {
     const comp: any = (
       <Component screenProps={screenProps} navigation={navigation} />
     );
-    const originalMethod = comp.type.prototype.shouldComponentUpdate;
-    if (originalMethod && scene) {
+    if (scene) {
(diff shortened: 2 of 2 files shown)


exKAZUu:
@kelset I’d like you to review this PR because it is inspired by your suggestion.

exKAZUu:
We can probably test this PR using @RoryKelly’s example (https://github.com/react-community/react-navigation/issues/608#issuecomment-343208937).

exKAZUu:
I've created a test repo: https://github.com/exKAZUu/react-navigation-issues-608 thanks to RoryKelly’s example (https://github.com/react-community/react-navigation/issues/608#issuecomment-343208937).

The test app with this PR shows the following result when pushing the `Dispatch` button on the screen D:
```
got action INCREMENT
D Rendered
```
Also, it seems to keep the current animation behavior.

In contrast, the test app with the current react-navigation shows the following result:
```
got action INCREMENT
A Rendered
B Rendered
C Rendered
D Rendered
```

kelset:
Hey thanks for submitting, I'm not sure this weekend I'll have enough time to review it properly, but I'll take a look asap

exKAZUu:
Thanks! I intend not to rush you and I just want to contribute.

kelset:
Btw I'd like to ping @dantman since he proposed a solution for performances [here](https://github.com/react-community/react-navigation/issues/608#issuecomment-328689388) so he may have an opinion on your approach.

Still not sure I've time to properly review today :(

dantman:
I like dealing with the performance issue by blocking render. However there are some tweaks needed and I'm not quite sure about this exact way of doing it.

First, we don't actually want a sole `isActive` check. We also may want to render "if prev.isActive != next.isActive" and "if prev.params !== next.params". Allowing a single render when `isActive` changes gives the component a single render where it knows it has become inactive, this way if the component is really complex it can stop timers or cleanup things temporarily while it is inactive. And if a setParams dispatch is explicitly made (changing `params`) it's probably a good idea to let that screen render with those new params even if it's not currently visible.

Second, I'm not sure about suddenly exposing `screen` to components.

Finally, I really don't think this way of hacking `shouldComponentUpdate` instead of putting it inside of some form of static container is a good idea.

* Having react-navigation mutate a component like this feels like a really horrible hack.
* This code may not actually work correctly in some situations. Remember that besides React.createClass and ES6 classes, React also supports "Stateless Functional Components" ie: a function that just returns render results instead of being an instanced class. These do not have a Component prototype for you to modify `shouldComponentUpdate` in a way React will actually use it.
* From what I'm reading I believe this code in its current implementation will actually leak memory, slowly degrade performance a bit, and (unless the js engine applies a tail call optimization to the `shouldComponentUpdate`) eventually crash the app with a stack overflow. Every time render() is called this code saves the current `shouldComponentUpdate`, creates a new function that wraps and calls it, and replaces the `shouldComponentUpdate` with it. The second time the SceneView gets a render call (which will be for any navigation state change, not just ones related to this scene), it will do this again, taking the `shouldComponentUpdate` wrapper it created last render and wrapping it again. And it will keep doing this every single time the navigation state changes making a deeper and deeper stack of wrappers as you navigate the app until either the device's memory or js environment's stack limit is exhausted.

I understand this was done to workaround redux tunnelling through the optimization. Though I'm wary of using this hack even for that. I'll need to know more about how redux is causing this problem (how `@connect` or `props.navigation` would do this I'm not entirely certain at the moment). But I'm wary about it because if redux used on the scene view (which is patched with this hack) causes problems that bypass the `shouldComponentUpdate` optimization, then what about children of that screen route that also might use redux and are not patched.

I think there's probably another solution to the redux issue we haven't thought of yet that will pop up when we investigate what the cause of the redux issue is.

brentvatne:
I'm going to close this pull request because @dantman raises many valid concerns. We'll continue to discuss this in https://github.com/react-navigation/react-navigation/issues/608 and hopefully soon that will carry over into a RFC.

exKAZUu:
@dantman Thank you for giving your valuable review comments!

@brentvatne Sorry for leaving WIP code.
I think my reproduction (but bit old) is still useful for you.
https://github.com/exKAZUu/react-navigation-issues-608
If there is something I can help you, please let me know.

Issue #2944, opened 2017-11-09
TabBarBottom no longer animatable


mujavidb:
### Current Behavior
When an animated style is applied to the component, the component throws the same error as when styles are applied to a normal non-animated `View`.

### Expected Behavior
Styles passed to the component should be passed down to the `Animated.View`, allowing the TabBarBottom to be animated.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0.0-beta.19
| react-native     | 0.48.0
| node             | 7.4
| npm or yarn      | 4.6.1

mujavidb:
#2945

mujavidb:
I've added a PR for reference on how I fixed this in my local copy.

PR #2945, opened 2017-11-09
Update TabBarBottom to accept styling for animations

mujavidb:
### Current
When an animated style is applied to the component, the component throws the same error as when styles are applied to a normal non-animated `View`.

### Fix
Styles passed to the component are passed down to the `Animated.View` via the prop `animateStyles`, allowing `TabBarBottom` to be animated.

**Test plan (required)**

Unit test on props passing through to children.


mujavidb:
#2944

kelset:
Thanks for the PR - I think @davepack or @spencercarli can review it.

In the meantime, it looks like this breaks a couple jest tests:

```
FAIL  src/navigators/__tests__/TabNavigator-test.js (9.124s)
  ● Console

    console.warn node_modules/react-native/jest/setup.js:99
      Calling .measureInWindow() in the test renderer environment is not supported. Instead, mock out your components that use findNodeHandle with replacements that don't rely on the native environment.

  ● TabNavigator › renders successfully

    expect(value).toMatchSnapshot()
    
    Received value does not match stored snapshot 1.
    
    - Snapshot
    + Received
    
    @@ -61,10 +61,11 @@
           />
         </View>
       </View>
       <View
         collapsable={undefined}
    +    style={undefined}
       >
         <View
           onLayout={[Function]}
           style={
             Object {
      
      at Object.<anonymous> (src/navigators/__tests__/TabNavigator-test.js:31:18)
      at tryCallTwo (node_modules/promise/lib/core.js:45:5)
      at doResolve (node_modules/promise/lib/core.js:200:13)
      at new Promise (node_modules/promise/lib/core.js:66:3)
      at Promise.resolve.then.el (node_modules/p-map/index.js:46:16)
      at tryCallOne (node_modules/promise/lib/core.js:37:12)
      at node_modules/promise/lib/core.js:123:15

 FAIL  src/views/__tests__/TabView-test.js (9.622s)
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
      
      Check the render method of `ScrollView`.
          in Unknown (created by ScrollView)
          in ScrollView (created by TabViewPagerScroll)
          in TabViewPagerScroll (created by TabViewAnimated)
          in View (created by View)
          in View (created by TabViewAnimated)
          in TabViewAnimated (created by TabView)
          in TabView (created by withCachedChildNavigation(TabView))
          in withCachedChildNavigation(TabView)
    console.warn node_modules/react-native/jest/setup.js:99
      Calling .measureInWindow() in the test renderer environment is not supported. Instead, mock out your components that use findNodeHandle with replacements that don't rely on the native environment.

```

Can you modify your code to pass those tests?

mujavidb:
Done

davepack:
Although this fixes something that SafeAreaView has broken, it’s also a breaking change to the api. I think I’d rather find a way to make either Animated.SafeAreaView, or just roll the support for safe area directly into React Native’s View component. I’m leaning towards the latter as we surface all these issues with using SafeAreaView.

That said, this is a good work around that I’ll approve, but I think we need to make sure we don’t need it anymore before final RC of react navigation.

kelset:
Ok I'll leave a comment for myself about this being a breaking change (ping to @spencercarli too)

PR #2946, opened 2017-11-09
Await promise from Transitioner onTransitionStart

jamesreggio:
**Motivation**

This small PR enables a custom transitioner (using the `Transitioner` component) to delay the transition animation by returning a promise from `onTransitionStart`. The transition animation occurs after the promise resolves (or immediately, if not promise is returned).

This small feature enables the custom transitioner to perform some asynchronous work with the new scene before the animation begins (e.g., wait for an `onLayout` callback). It also enables the transitioner to delay the transition until certain data or images load.

In our app, we built a transitioner for variable-height 'sheets' that pop up from the bottom of the screen. In order to animate the presentation of the sheet correctly and consistently, we need to first measure the height of the sheet. With this patch, our transitioner can return a promise from `onTransitionStart` and then resolve the promise when the new sheet's height has been obtained, thus triggering the animation.

Not that it helps to elucidate much, but here's a GIF demonstrating this transitioner in action:

2017-11-09-banter-bottom-sheet

**Test plan (required)**

We've been using this patch for the past 4 months off a private branch to power the above custom transitioner. Given that there are not many tests for the `Views` directory, I hope this is sufficient. I'm happy to receive guidance on how to write a test for this.


kelset:
This looks awesome, thanks for submitting!

I think maybe @vonovak can provide a more robust feedback than mine ^^

jamesreggio:
@vonovak — any time to take a look?

vonovak:
Hmm, I see this can be useful to delay the `Animated.parallel(animations).start(this._onTransitionEnd);` part, but am unsure about the API change - if we go ahead with this, `onTransitionEnd` should also have this possibility. What does @spencercarli think about this?

jamesreggio:
@vonovak, thanks for thinking about it.

Regarding `onTransitionEnd`, what behavior happens after the `onTransitionEnd` callback? From inspecting the `_onTransitionEnd` implementation, it appears that the only thing that after the callback is the invocation of any transitions that were issued and enqueued while the current one executes.

I'm struggling to think of a reason for delaying an unrelated transition, but I'm happy to implement the same promise-awaiting logic for `onTransitionEnd` if you'd prefer the APIs to be symmetrical in their behavior.

vonovak:
@jamesreggio yeah, good point. Let's wait for Spencer's thoughts.

spencercarli:
Cool! A few thoughts

1. Even though adding it to `onTransitionEnd` may seem pointless I think, for the sake of consistency, it would be nice to add.
2. Would this have any breaking changes? Does a non-promise function automatically resolve with `await`? I can't remember 😄

vonovak:
@spencercarli a for 2, good question. I'm fairly certain that if you do not return anything, it will behave the as if you returned a promise, eg. the return value (undefined) will be wrapped in `Promise.resolve()`.

> The transition animation occurs after the promise resolves (or immediately, if not promise is returned).

I'm not sure it will be **immediate**, it would be nice to do some manual testing to make sure everything looks the same if a non-promise is returned.

jamesreggio:
@vonovak is correct — if you `await` a non-promise, it evaluates to an immediately resolved promise.

However, that does means that execution is deferred until the next tick (`setImmediate` on RN) per the promise spec. As I've learned with my RN app, waiting for `setImmediate` can sometimes take a while, so I'll revise the code to be truly immediate if no promise is returned.

jamesreggio:
@vonovak, @spencercarli — I think everything has been addressed.

src/views/Transitioner.js
```diff
@@ -172,12 +172,17 @@ class Transitioner extends React.Component<Props, State> {
...
-        nextProps.onTransitionStart(
+    this.setState(nextState, async () => {
+      if (nextProps.onTransitionStart) {
+        const result = nextProps.onTransitionStart(
           this._transitionProps,
           this._prevTransitionProps
         );
+
+        if (result instanceof Promise) {
+          await result;
```
spencercarli: Wrap in a try/catch blog? We should handle a rejected promise (currently we get a warning)

simulator screen shot - iphone 8 - 2017-11-28 at 11 37 57

src/views/Transitioner.js
```diff
@@ -231,9 +236,18 @@ class Transitioner extends React.Component<Props, State> {
...
-        this.props.onTransitionEnd(this._transitionProps, prevTransitionProps);
+    this.setState(nextState, async () => {
+      if (this.props.onTransitionEnd) {
+        const result = this.props.onTransitionEnd(
+          this._transitionProps,
+          prevTransitionProps
+        );
+
+        if (result instanceof Promise) {
+          await result;
```
spencercarli: Same comment in regards to catching a rejected promise

jamesreggio:
@spencercarli — I'm not sure it's a good idea to just swallow rejected promises. That should only happen if there's a bug in the user's code, and the unhandled rejection handler is the perfect way for them to see that.

spencercarli:
You're completely right - going to merge this as soon as CI is done. Thanks for adding this ability!

spencercarli:
I'm not sure why codecov is saying there is a decreased code coverage as there are no tests for the Transitioner as it stands 🤔

PR #2950, opened 2017-11-10
SafeAreaView - Fix for Web usage

rmevans9:
With the recent change to add SafeAreaView there was a change that broke projects using `react-navigation` inside `react-native-web`. The use of `DeviceInfo` and more specifically the property `isIPhoneX_deprecated` caused exceptions anytime `SafeAreaView` is used with `react-native-web`. This meant that `Header` couldn't be used without `SafeAreaView` being fixed.

**Test plan (required)**
Test that the functionality still works correctly in native by opening a `StackNavigator` with a header on both native (iOS or Android) and Web (via `react-native-web`)

### DIFF
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -11,7 +11,6 @@ import {
 } from 'react-native';
 import withOrientation from './withOrientation';
 
-const { isIPhoneX_deprecated } = DeviceInfo;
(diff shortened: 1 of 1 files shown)


kelset:
Thanks @rmevans9 for the PR and @vonovak for reviewing 🙏

PR #2958, opened 2017-11-10
Added a guide under intermediate concepts.

shubhnik:
A guide for integrating redux and react-navigation with authentication flow. 

Motivation: The official redux integration documentation is very very brief. There are other tutorials which I personally felt carry unnecessary complexity. Then I couldn't found any tutorial for authentication flow with redux integrated. I have tried to write this article from the ground up.


kelset:
Awesome, thanks for submitting!

PR #2967, opened 2017-11-13
Move SafeAreaView out of DrawerNavigatorItems to DrawerSidebar.

### LINKED ISSUE #2963: SafeAreaView in DrawerNavigatorItems causes extra spacing between things when DrawerNavigatorItems is not the ONLY thing in the drawer
### Current Behavior
#2833 added a `SafeAreaView` to every `DrawerNavigatorItems` instance. Now when you customize the drawer to either add a header or footer to it or to split up the items into multiple sections you get extra padding in between your sets of drawer items and other sets / headers or footers.

screen shot 2017-11-10 at 8 17 41 pm

### Expected Behavior
`SafeAreaView` should only be used once in the drawer content container, not in the `DrawerNavigatorItems` component. And custom implementations should be able to customize it because sometimes the contents aren't supposed to be inset (eg: the status bar should overlap the drawer header).

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0.0-beta.19
| react-native     | 0.50.3
| node             | v8.9
| npm or yarn      | 1.3.2

davepack:
The SafeAreaView was breaking custom contentContainers in drawers. This fixes #2963.

### DIFF
--- a/src/navigators/DrawerNavigator.js
+++ b/src/navigators/DrawerNavigator.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import React from 'react';
-import { Dimensions, Platform } from 'react-native';
+import { Dimensions, Platform, ScrollView } from 'react-native';
 
 import createNavigator from './createNavigator';
 import createNavigationContainer from '../createNavigationContainer';
@@ -9,6 +9,7 @@ import TabRouter from '../routers/TabRouter';
 import DrawerScreen from '../views/Drawer/DrawerScreen';
 import DrawerView from '../views/Drawer/DrawerView';
 import DrawerItems from '../views/Drawer/DrawerNavigatorItems';
+import SafeAreaView from '../views/SafeAreaView';
 
 import NavigatorTypes from './NavigatorTypes';
 
@@ -25,13 +26,21 @@ export type DrawerNavigatorConfig = {
 
 const { height, width } = Dimensions.get('window');
 
+const defaultContentComponent = (props: *) => (
+  <ScrollView alwaysBounceVertical={false}>
+    <SafeAreaView forceInset={{ top: 'always', horizontal: 'never' }}>
+      <DrawerItems {...props} />
+    </SafeAreaView>
+  </ScrollView>
+);
+
 const DefaultDrawerConfig = {
   /*
    * Default drawer width is screen width - header width
    * https://material.io/guidelines/patterns/navigation-drawer.html
    */
   drawerWidth: Math.min(height, width) - (Platform.OS === 'android' ? 56 : 64),
-  contentComponent: DrawerItems,
+  contentComponent: defaultContentComponent,
   drawerPosition: 'left',
   drawerBackgroundColor: 'white',
   useNativeAnimations: true,
--- a/src/views/Drawer/DrawerSidebar.js
+++ b/src/views/Drawer/DrawerSidebar.js
@@ -36,6 +36,7 @@ type Props = {
   contentOptions?: {},
   screenProps?: {},
   style?: ViewStyleProp,
+  drawerPosition?: 'left' | 'right',
 };
 
 /**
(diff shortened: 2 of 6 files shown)


dantman:
Thanks for the attempt but there are a few problems.

Firstly, one of my drawer items ends up with extra spacing. This is added by a SafeView inside the individual drawer items. I have a feeling that the `touchesBottom` calculation is not reliable enough. The automatic vertical padding should be unnecessary for the drawer items since the container already handles the vertical padding, so you can probably just update this SafeView with `vertical: never`.
drawer-pr-mid

A separate issue which will be a bug even when you're not using a custom drawer. As you can see by using `horizontal: always` instead of `[openingSideOfTheDrawer]: always, [opposingSide]: never` there is padding on the right side of the drawer item that doesn't belong. This bug will be silent and hard to know is there, but will show up confusingly for people when there is a drawer item with a long enough title, on an iPhone X, in landscape mode.
drawer-pr-landscape

Here's the other problem. 
drawer-pr-top
drawer-pr-bottom

As you can see, by putting the SafeView outside of the ScrollView/custom drawer content the entire content and scroll view is inset. The standard behaviour for a navigation header is for the header's background to extend into the status bar, the header component itself is responsible for adding the safe area padding, not the drawer. Additionally the bottom of the scroll view is inset so that drawer items get cut off suddenly at the end of the safe area. Apple's guidelines seem to imply that drawer items, tables, and other scroll view content should be visible under the indicator; you add padding to the inside of the scroll view so they aren't overlapped when you scroll all the way up, not inset the scroll view itself.
[designing for iphone x](https://developer.apple.com/videos/play/fall2017/801/).

Instead of being part of the DrawerSidebar the default `contentComponent` should be `(props) => <ScrollView><SafeArea><DrawerItems {...props} /></SafeArea></ScrollView>` (with the same extra props you're using) and the `contentComponent` documentation example should be updated. This way the safe area is inside the scroll view where it belongs, and anyone replacing it with a custom one can control what the safe area applies to.

davepack:
Thanks for the feedback, latest commits should take care of those issues.

I didn't do a great job handling all the possible view customization in rnav my first time through adding SafeAreaView.

davepack:
Also, I changed the default contentComponent to be a ScrollView with `alwaysBounceVertical={false}`. Does that seem better than requiring a custom component just to have scrolling?

dantman:
Works great.

Yes, the default drawer should have a scroll view. ;) I'd consider that a different react-navigation bug we fixed as a side effect.

Now we just need to update the documentation at https://reactnavigation.org/docs/navigators/drawer#Example%3A

src/navigators/DrawerNavigator.js
```diff
@@ -25,13 +26,21 @@ export type DrawerNavigatorConfig = {
 
 const { height, width } = Dimensions.get('window');
 
+const defaultContentComponent = (props: *) => (
+  <ScrollView alwaysBounceVertical={false}>
```
brentvatne: should we set `flex: 1` on the scrollview? I believe we want it to take up the full drawer space available to it even if only one or two items

src/views/Drawer/DrawerSidebar.js
```diff
@@ -110,21 +113,20 @@ class DrawerSidebar extends React.PureComponent<Props> {
     const { state } = this.props.navigation;
     invariant(typeof state.index === 'number', 'should be set');
     return (
-      <View style={[styles.container, this.props.style]}>
```
brentvatne: looks like we're not using styles.container or `this.props.style` anymore, intentional?

brentvatne:
a couple of comments but otherwise looks good, nice one

src/navigators/DrawerNavigator.js
davepack: Seems to set `flex: 1` by default, at least on iOS. In testing it looked fine.

src/views/Drawer/DrawerSidebar.js
davepack: Ah, that was a mistake. Originally replaced with SafeAreaView, then removed SafeAreaView. It should be put back in.

PR #3030, opened 2017-11-22
Add NavigationContainer and NavigationContainerProps types

### LINKED ISSUE #2933: Update to react-native@0.50 and flow-bin@0.56
I know a lot of effort went into this, but I still have a lot of flow errors on react-navigation and i have no idea what i should do anymore.

I am using flow version 0.56 with react-native 0.50.x and react-navigation@1.0.0-beta.19 and atom ide-flowtype (latest version)

bildschirmfoto 2017-11-08 um 15 05 55

I also get a lot of errors in test-files where "expect" or "it" is not defined. What's wrong with my configuration?

Ashoat:
Adding types for these fixes the errors we're seeing when using `TabNavigator` and `StackNavigator`. We should finally be able to close #2933 now.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -228,9 +228,9 @@ export type NavigationScreenDetails<T> = {
   navigation: NavigationScreenProp<NavigationRoute>,
 };
 
-export type NavigationScreenOptions = {|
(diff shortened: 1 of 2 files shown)


kelset:
Awesome work @Ashoat 🙏

PR #3037, opened 2017-11-23
Set sdkVersion to 23.0.0

tab:
This small PR fixes "React Native version mismatch" error in ReduxExample.

rn-redux-example-error


kelset:
@spencercarli or @davepack is the right way to update the expo app? Can't recall

spencercarli:
Thanks!

PR #3038, opened 2017-11-23
Pass props down into SafeAreaView's nested View

neilkimmett:
- this was causing a problem where we were setting `pointerEvents` on a `SafeAreaView` and it wasn't being respected


neilkimmett:
lol nope this totally breaks `SafeAreaView`, gonna do a new PR that only passes along `pointerEvents`.

PR #3041, opened 2017-11-24
Issue #2794 -  Fixing iPad iOS 11 Tab Bar Bottom Behavior

SteffeyDev:
## Problem Being Solved

This addresses Issue #2794, which pointed out that when the iOS 11 tab bar changes were made in this library, they were not tested on iPad, especially considering multitasking in iPad, and thus broke the UI.

When the iPad was in landscape, the tab bar was shorter (29pt), when that behavior should only happen on an iPhone in landscape, and when the iPad app was in split-screen with another app, horizontal tabs were used, even though they didn't actually fit and normal tabs should have been used.  Apple-made apps were used for comparison of how it should act.

## Solution & Explanation
This PR changes the way horizontal tabs are handled in TabBarBottom to match the behavior Apple uses on iOS 11 as closely as possible.

Here is a breakdown of the use cases:
- iPhone in Portrait: Use normal tabs (labels beneath icon)
- iPhone in Landscape: Use horizontal tabs (labels beside icon) with shortened tab bar 
- iPad in any orientation: Use horizontal tabs with normal height tab bar initially, but switch to normal tabs after render if the horizontal tabs don't fit.

The iPad mechanic works as such: the tabs are initially rendered as horizontal, and the width of each tab bar item is measured.  If the measurements indicate that the items are overlapping, then switch to the normal tabs.  Usually, the horizontal tabs will overlap when using the app in split screen mode or app-over-app mode, where the width is that of an iPhone but the device is technically still in landscape.

## Notes

The changes was tested on Android as well to ensure that the behavior there was not changed.

The only thing about this I may want to change is the way it is handled when an iPad app is launched with a small width.  With the current way I have implemented it, the tabs are rendered horizontally, and then changed shortly after to vertical.  Should I hide the tabs (opacity: 0) on iPad until after I determine what type of tabs should be used?

Hopefully the rest of the changes make sense, let me know of any issues.


SteffeyDev:
I don't know how to use codecov, is that something I need to learn and work on to pass these tests or is that something you take care of?

src/views/TabView/TabBarBottom.js
```diff
@@ -99,19 +112,18 @@ class TabBarBottom extends React.PureComponent<Props> {
...
-    }
 
     if (typeof label === 'string') {
       return (
         <Animated.Text
-          style={[styles.label, { color, marginLeft, marginTop }, labelStyle]}
+          style={[
+            styles.label,
+            { color },
+            labelStyle,
```
spencercarli: User-specified styles should always be the last argument so that they can override anything.

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +169,12 @@ class TabBarBottom extends React.PureComponent<Props> {
       this.props.getTestIDProps && this.props.getTestIDProps(scene);
     return testIDProps;
   };
+  shouldUseHorizontalTabs() {
+    const { useHorizontalTabs } = this.state;
+    const { isLandscape } = this.props;
+    if (isPad) return canUseHorizontalTabs && useHorizontalTabs;
+    else return canUseHorizontalTabs && isLandscape;
```
spencercarli: This type if syntax isn't used elsewhere in the project. Please use 

```javascript
if (isPad) {
  ...
} else {
  ...
}

File: src/views/TabView/TabBarBottom.js [hunk 3]
```diff
@@ -220,17 +248,29 @@ class TabBarBottom extends React.PureComponent<Props> {
...
-                  {this._renderLabel(scene)}
+                <Animated.View style={[styles.tab, { backgroundColor }]}>
+                  <View
+                    onLayout={evt => {
+                      const { useHorizontalTabs, viewWidth } = this.state;
+                      if (
+                        viewWidth > 0 &&
+                        useHorizontalTabs &&
+                        viewWidth - 20 <
+                          evt.nativeEvent.layout.width * routes.length
```
spencercarli: Can you add some comments as to what you're doing here and why? For example, where is that 20 coming from?

spencercarli:
Codecov is there to check if the coverage of tests has changed at all. If you run `yarn run test` on your machine you should have a `coverage` directory which will have an index.html in it. If you look at that and click the files you're modifying you can see spots where cases where the tests don't hit.

I typically don't like to merge PRs that reduce overall test coverage.

SteffeyDev:
@spencercarli I think I addressed your requested changes, let me know.  I am worried about the performance of onLayout as well, but don't know of a better way to get the tab bar width and make sure that we don't have overlapping tab bar items.  Do you have a suggestion of a better way to approach this.

I did do some more testing today and changed the process I use to determine if we should be using horizontal tabs to be more consistent (there was a race condition before).

As for test coverage, do I need to learn your testing framework and figure out how to test to make sure that the horizontal tabs are only used in the correct situations?

SteffeyDev:
@spencercarli Still waiting for your feedback when you get the chance, about how to proceed

SteffeyDev:
@spencercarli Still waiting on instructions on how to move forward, and @davepack still waiting on review.  I know everyone is busy, but what I am proposing is not even a feature, but a bug fix that impacts all iPads running iOS 11 (which is most of them by this point).  Let me know what I need to do to keep this moving forwards.

kelset:
Hey @SteffeyDev, sorry but winter is the time of the year when everyone is was less available to work on open source - moreover Dave stepped out of the project. I'll try to ping them privately but can't assure anything.

src/views/TabView/TabBarBottom.js
```diff
@@ -179,14 +206,19 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+        : styles.tabBarRegular,
       style,
     ];
 
     return (
-      <Animated.View style={animateStyle}>
+      <Animated.View
+        style={animateStyle}
+        onLayout={evt => {
+          this.setState({ viewWidth: evt.nativeEvent.layout.width });
```
satya164: `layout` is already passed as a prop to `TabBarBottom`

src/views/TabView/TabBarBottom.js
```diff
@@ -53,11 +54,23 @@ type Props = {
...
+  viewWidth: number,
+  maxTabItemWidth: number,
+};
+
 const majorVersion = parseInt(Platform.Version, 10);
 const isIos = Platform.OS === 'ios';
-const useHorizontalTabs = majorVersion >= 11 && isIos;
+const canUseHorizontalTabs = majorVersion >= 11 && isIos;
+const isPad =
+  Dimensions.get('window').height / Dimensions.get('window').width < 1.6;
```
satya164: We avoid using `Dimensions` API in the library because `Dimensions` represents the layout of the whole app rather than the container. The navigator could be contained in a smaller container view and using the `Dimensions` API breaks that use case.

SteffeyDev:
@satya164 I like your use of flexWrap to change from bottom to side faster, I will look into using that.  The only issue I see is that when the text is next to the icon, it needs to have a larger font size by about 4 points.  Is there a way to do that automatically as well?

The main use case is going into split-screen mode on iPad running iOS 11, where the tabs need to go from being horizontal to vertical once the screen width gets too small.

src/views/TabView/TabBarBottom.js
SteffeyDev: It is? I don't see it in the Prop types.  Is it called something else?

SteffeyDev:
@satya164 Another holdup with your approach that I just found is that the tab bar icon is positioned absolutely, and both the inactive and active are on top of each other, so you can't actually use a simple flexWrap to layout one relative and one absolute item.  How do you suggest I move forward with this in mind?

SteffeyDev:
@satya164 Another Issue I found with wrapping it is that certain tabs with shorter labels could be laid out horizontally while ones with longer labels would be laid out vertically in the same tab bar.  This is obviously broken.

In defense of my approach, I always start iPhone as vertical layout (most apps are opened in portrait to start) and the iPad as horizontal (most apps open full screen to start), so this heuristic approach minimizes post-mount changes seen by users.  That being said, if there was any other way to get this behavior without waiting on an asynchronous callback, I'm all ears.

I also want to point out that this is still a bug fix; currently, if you move an iPad into split screen mode while in the horizontal orientation, all of the tab bar labels overlap and are unreadable.  Though this does change things post-layout, at least it would move the labels underneath the icons so you could still read them.  If this were a feature I am adding, I would be more hesitant to use the callback, but I'm replacing a broken feature with one that is functional, just not ideal.

@satya164 Let me know what you think

src/views/TabView/TabBarBottom.js
SteffeyDev: I am using the Dimensions not as layout constraints, but as a device detection mechanism.  Knowing the device powers my heuristic-based layout approach (best-guess), and once the callback comes in I re-layout if necessary.  See my latest comment in the main thread for details.

satya164:
>  the text is next to the icon, it needs to have a larger font size by about 4 points. Is there a way to do that automatically as well

I don't think flexbox can be used to it

> certain tabs with shorter labels could be laid out horizontally while ones with longer labels would be laid out vertically in the same tab bar. This is obviously broken.

Fixable by specifying a width for the lab labels. This could be hardcoded to some value based on the tab bar's width or ask the user to specify it.

> I am using the Dimensions not as layout constraints, but as a device detection mechanism. Knowing the device powers my heuristic-based layout approach

Device dimesnions don't matter. Only the container's dimesnions do. So any heuristic which tries to guess things can produce wrong results and a janky experience.

I don't think trying to guess what's the layout here is the best approach. It's impossible to know the layout synchronously and it's impossible to adjust the layout without shifting the content for certain cases, which takes away from the native experience. It's most likely that the user knows the dimesnions of the `TabNavigator`. `TabNavigator` already allows the user to pass a `initialLayout` which tells you the dimensions of the tabview. It's possible to reuse this info to determine what to do.

I think the best way to move forward is to add an option (`tabBarAdaptive: true`) which allows the developer to explicitely enable the behaviour to adapt the tab items to the container layout. The `layout` prop passed to the `TabBar` should be used to do all the calculations and the developer should specify the `initialLayout` prop to avoid shifiting of content when adaptive mode is enabled. It's not the best API, but at least it provides the best experience for the user. And for most people `initialLayout` is just `{ width: Dimensions.get('window').width, height: whatever }`.

SteffeyDev:
@satya164 Thanks for getting back to me.  Per your note on the use of Dimension, I would like to point out that it is used elsewhere in the library (in DrawerNavigator.js) for similar purposes, but I agree it is not ideal.

I am still confused how to get the `layout` prop, though I do see the `initialLayout` prop available, so I can use that to set it up initially.  I find that my approach does work fine when changing widths, it is only janky on initial render, so if I can use the `initialLayout` to set the initial render correctly, we should be fine from there.

I will go ahead and add a `tabBarAdaptive` option that defaults to true, so that it can be easily disabled if desired.  Because this is an iOS 11 only feature natively, should I restrict this to iOS 11 devices or allow it on Android as well?

@satya164 Let me know about that `layout` prop, and I'll get the new version going.

SteffeyDev:
Ok, I pushed the latest version that forgoes the use of `onLayout` in favor of props, and allows the feature to be disabled if desired.

Still waiting to hear about:

- [ ] Use of 125pt threshold
- [ ] Use of Dimensions as device detection mechanism
- [ ] Naming of new option
- [ ] Availability of horizontal tabs
- [ ] Final code review

See last comment for details on these items.

Once these are resolved, we should be good to merge.

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,16 @@ type Props = {
   tabStyle?: ViewStyleProp,
   showIcon?: boolean,
   isLandscape: boolean,
+  initialLayout?: Layout,
```
satya164: There is no need for an `initialLayout` prop. If an `initialLayout` prop is provided to `TabNavigator`, the `layout` prop will contain that data.

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,16 @@ type Props = {
...
+  layout: Layout,
+  enableHorizontalTabs: boolean,
```
satya164: Props are usually named as `scrollEnabled` etc. Also, I'm not sure `horizontalTabs` is the correct name since it's the labels inside the tabs which are horizontal, and to me horizontal or vertical tabs feel like the tabs themselves are arranged in that direction. I'd just name this prop something like `adapative` or `responsive` unless you have a better name suggestion.

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,16 @@ type Props = {
...
 };
 
 const majorVersion = parseInt(Platform.Version, 10);
 const isIos = Platform.OS === 'ios';
-const useHorizontalTabs = majorVersion >= 11 && isIos;
+const canUseHorizontalTabs = majorVersion >= 11 && isIos;
+const isPad =
```
satya164: Remove usage of `Dimensions` entirely

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +164,30 @@ class TabBarBottom extends React.PureComponent<Props> {
...
   };
+  shouldUseHorizontalTabs() {
+    const { routes } = this.props.navigation.state;
+    const {
+      isLandscape,
+      initialLayout,
+      layout,
+      enableHorizontalTabs,
+    } = this.props;
+    const maxTabBarItemWidth = 125;
```
satya164: Probably also read the `tabStyle` prop and use the `width` or `maxWidth` property if provided

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +164,30 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+    let tabBarWidth;
+    if (layout && layout.width) tabBarWidth = layout.width;
+    else if (initialLayout && initialLayout.width)
```
satya164: Please use the same code style as rest of the lib

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +164,30 @@ class TabBarBottom extends React.PureComponent<Props> {
       this.props.getTestIDProps && this.props.getTestIDProps(scene);
     return testIDProps;
   };
+  shouldUseHorizontalTabs() {
```
satya164: Add new line before this and prefix the method with an underscore

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +164,30 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+    const { routes } = this.props.navigation.state;
+    const {
+      isLandscape,
+      initialLayout,
+      layout,
+      enableHorizontalTabs,
+    } = this.props;
+    const maxTabBarItemWidth = 125;
+    let tabBarWidth;
+    if (layout && layout.width) tabBarWidth = layout.width;
```
satya164: `layout.width` always exists. The check is not necessary

src/views/TabView/TabBarBottom.js
```diff
@@ -157,6 +164,30 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+    else if (initialLayout && initialLayout.width)
+      tabBarWidth = initialLayout.width;
+    else return isPad;
+    if (!enableHorizontalTabs) return false;
+    if (isPad) {
+      return (
+        canUseHorizontalTabs &&
```
satya164: Instead of this variable `canUseHorizontalTabs`, use it as the default value of `enableHorizontalTabs` prop, so the user can override it

src/views/TabView/TabBarBottom.js
SteffeyDev: Ah, good to know

src/views/TabView/TabBarBottom.js
SteffeyDev: I see why horizontalTabs could be confusing, I also feel that adaptive/responsive is too generic and you may want those options to do something else (non iOS 11 specific) in the future.  That being said, I can't think of a better name.

satya164:
> it is only janky on initial render

Yeah, that's what I'm worried about.

> should I restrict this to iOS 11 devices or allow it on Android as well

Let's set it to true by default on iOS > 11 and allow user to override the behaviour

> Or should I make that a configurable option

If we read the `tabStyle` to if a width is specified there, it should be fine

> Continue to use Dimension as a way to detect if we are on an iPad or iPhone

Using `Dimensions` to set the default value of the prop sounds fine

src/views/TabView/TabBarBottom.js
SteffeyDev: How can I detect device (whether or not is tablet) without use of Dimension?

src/views/TabView/TabBarBottom.js
SteffeyDev: check is not for existence, but for non-zero.  If layout.width is 0, this will return false and I can short-circuit.

satya164: `layout && layout.width` checks for both falsy `layout` and falsy `width`. `layout` will always be an object and the `width` property will always exit

src/views/TabView/TabBarBottom.js
SteffeyDev: Checking for width works, but flow check can't handle `maxWidth` as a property of `tabStyle`. I get errors like `property 'maxWidth'. Property not found in Array|Boolean
Number.`

satya164: You need to flatten the style property with `StyleSheet.flatten` before reading properties on it. Also, the properties can be percentage strings, in which case, you need to calculate the numeric value.

SteffeyDev:
@satya164 Ok, renamed option to `adaptiveTabs`. Is that what you were looking for?

Also pushed next changes for review

SteffeyDev:
To address the usage of Dimension, I now use `layout.height` to check for `isHeightConstrained`, knowing that most phones in landscape have a height < 500 and most tablets have a height > 500 in either orientation.  This means that `isPad` the constant derived from Dimension, is only used for the first frame if initialLayout is not provided.

SteffeyDev:
Note: the build is failing due to some import issue, as far as I'm concerned the issue is not related to my changes.

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,15 @@ type Props = {
...
   isLandscape: boolean,
+  layout: Layout,
+  adaptiveTabs: boolean,
 };
 
 const majorVersion = parseInt(Platform.Version, 10);
 const isIos = Platform.OS === 'ios';
-const useHorizontalTabs = majorVersion >= 11 && isIos;
+const canUseHorizontalTabs = majorVersion >= 11 && isIos;
+const isPad =
```
satya164: Can you change this variable to `isTablet` or `isIPad`?

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,15 @@ type Props = {
   tabStyle?: ViewStyleProp,
   showIcon?: boolean,
   isLandscape: boolean,
+  layout: Layout,
+  adaptiveTabs: boolean,
```
satya164: I think just `adaptive` is fine. `Tabs` is redundant? Or change to `adaptiveLabels`

src/views/TabView/TabBarBottom.js
```diff
@@ -51,11 +53,15 @@ type Props = {
...
 };
 
 const majorVersion = parseInt(Platform.Version, 10);
 const isIos = Platform.OS === 'ios';
-const useHorizontalTabs = majorVersion >= 11 && isIos;
+const canUseHorizontalTabs = majorVersion >= 11 && isIos;
```
satya164: Change this variable name to `isIOS11`

src/views/TabView/TabBarBottom.js
```diff
@@ -67,6 +73,7 @@ class TabBarBottom extends React.PureComponent<Props> {
     showLabel: true,
     showIcon: true,
     allowFontScaling: true,
+    adaptiveTabs: canUseHorizontalTabs,
```
satya164: Change the default value to `isIOS11 && isIPad`

src/views/TabView/TabBarBottom.js
```diff
@@ -158,6 +164,36 @@ class TabBarBottom extends React.PureComponent<Props> {
...
 
+  _shouldUseHorizontalTabs() {
+    const { routes } = this.props.navigation.state;
+    const { isLandscape, layout, adaptiveTabs, tabStyle } = this.props;
+
+    if (!adaptiveTabs) {
+      return false;
+    }
+
+    let maxTabBarItemWidth = 125;
```
satya164: Extract this to the top scope as a constant

src/views/TabView/TabBarBottom.js
```diff
@@ -158,6 +164,36 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+    if (tabStyle) {
```
satya164: This doesn't handle styles created with `StyleSheet.create` or percentage styles. Change this to:

```js
let maxTabBarItemWidth;

const flattenedTabStyle = StyleSheet.flatten(tabStyle);

if (flattenedTabStyle) {
  if (typeof flattenedTabStyle.width === 'number') {
    maxTabBarItemWidth = flattenedTabStyle.width;
  } else if (typeof flattenedTabStyle.width === 'string' && flattenedTabStyle.endsWith('%')) {
    const width = parseFloat(flattenedTabStyle.width);
    if (Number.isFinite(width)) {
      maxTabBarItemWidth = layout.width * (width / 100);
    }
  } else if (typeof flattenedTabStyle.maxWidth === 'number') {
    maxTabBarItemWidth = flattenedTabStyle.maxWidth;
  } else if (typeof flattenedTabStyle.maxWidth === 'string' && flattenedTabStyle.endsWith('%')) {
    const width = parseFloat(flattenedTabStyle.maxWidth);
    if (Number.isFinite(width)) {
      maxTabBarItemWidth = layout.width * (width / 100);
    }
  }
}

if (!maxTabBarItemWidth) {
  maxTabBarItemWidth = 125;
}
```
 Extract it to a function to simplify.

src/views/TabView/TabBarBottom.js
SteffeyDev: Apple uses horizontal tabs on iPhones as well by default, so I'm just going to put the default as isIOS11

SteffeyDev:
@satya164 Ok, all changes made, except for making default `isIOS11 && isTablet`, instead default is just `isIOS11` because iPhones running iOS 11 also use horizontal tabs by default when in landscape.

SteffeyDev:
@spencercarli Could you look back over your requested changes when you get the chance?  I think I addressed all of them.

spencercarli:
Yargh, sorry for dropping the ball on reviewing this. Thanks for picking it up, @satya164! And thank you very much @SteffeyDev for your work (and patience 😄 )

satya164:
@SteffeyDev this was reverted because it broke layout on iPhone.

SteffeyDev:
@satya164 can you please be more specific what you mean by "broke"?  That doesn't give me much to go off of. It works fine in my app (which is live on the app store)

satya164:
I don't know. @brentvatne tested it

spencercarli:
simulator screen shot - iphone x - 2018-01-12 at 09 41 09

I _think_ I see where the issue was occurring. If you've got two tabs it will always use horizontal tabs, even when it should use vertical. The screenshot is "Stacks over Tabs" example in demo app.

Also, I'm not sure if this is true as a result of this PR, the vertical padding when landscape seems off.

SteffeyDev:
Technically that shouldn't happen, but because you guys didn't want me to use Dimensions for device detection, I had to treat tablets and phone the same. Due to that, if there is space to lay out horizontally, then it will, regardless of width.

The only concession I can think of making is adding an isWidthConstrained check, to disable horizontal tabs in small width panels (probably < 500 again). 

How would you feel about this change?

PR #3046, opened 2017-11-24
Pass pointerEvents through in SafeAreaView

neilkimmett:
- this was causing us a problem where where `pointerEvents: 'box-none'` wasn't working on a `SafeAreaView` so the view was capturing touches it shouldn't

### DIFF
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -111,6 +111,7 @@ class SafeView extends Component {
         ref={c => (this.view = c)}
         onLayout={this._onLayout}
         style={safeAreaStyle}
+        pointerEvents={this.props.pointerEvents}
       >
         {this.props.children}
       </View>


neilkimmett:
This breaks some snapshot tests, cause it adds `pointerEvents: undefined` to a bunch of views. Whats the best way around this:
a) re-record the snapshots
b) only set the `pointerEvents` prop if its not undefined
?

leethree:
I think `SafeAreaView` should pass `rest` parameter to the inner `View`, not just `pointerEvents`.

neilkimmett:
@leethree yeah I tried doing that (in https://github.com/react-community/react-navigation/pull/3038) but it totally broke everything 😕

ericvicenti:
This looks good, but #3038 would probably also work, if the `{...this.props}` was placed before the other props.

For snapshot tests, you can run `yarn test-update-snapshot` and it will re-record the new values into the codebase. `pointerEvents: undefined` everywhere is fine I think.

brentvatne:
closing because no response. if someone wants to take this over again it would be appreciated

neilkimmett:
Ah I'm sorry, I stopped working on the project that used this so I totally slept on it. Will try to find to take another look.

pjivers-bom:
Any update on this? It's causing me issues too.

PR #3048, opened 2017-11-25
Fix flowtype error. We are allowed to have just a single child component node here

mikelambert:


Ashoat:
This is already fixed on master: https://github.com/react-community/react-navigation/blob/3484c8ad1367bcc6498a47c5aac8319f6c075508/src/views/TouchableItem.js

PR #3064, opened 2017-12-01
Upgrade react-native-tab-view

satya164:
With this update, the `lazy` option has been removed, so it's a breaking change.

- It didn't work great with more than 2 tabs
- It's not very performant to listen to all events during the gestures since we're trying to move away towards native gestures.
- It basically triggered a re-render of the whole tab view mid-gesture, which could cause jank

I understand that it's still desirable to lazily load tabs, so I think it's time we implement [`focus` events](https://github.com/react-community/react-navigation/issues/51) so that this can be implemented in userland. The approach I'd go with is this:

- Implement a placeholder component for each tab screen based on a `loaded` state
- When a tab screen is focused, toggle the `loaded` state and replace placeholder component with actual content

### DIFF
--- a/docs/api/navigators/TabNavigator.md
+++ b/docs/api/navigators/TabNavigator.md
@@ -88,7 +88,6 @@ The route configs object is a mapping from route name to a route config, which t
 - `swipeEnabled` - Whether to allow swiping between tabs.
 - `animationEnabled` - Whether to animate when changing tabs.
 - `configureTransition` - a function that, given `currentTransitionProps` and `nextTransitionProps`, returns a configuration object that describes the animation between tabs.
-- `lazy` - Whether to lazily render tabs as needed as opposed to rendering them upfront.
 - `initialLayout` - Optional object containing the initial `height` and `width`, can be passed to prevent the one frame delay in [react-native-tab-view](https://github.com/react-native-community/react-native-tab-view#avoid-one-frame-delay) rendering.
 - `tabBarOptions` - Configure the tab bar, see below.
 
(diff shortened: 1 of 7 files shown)


douglasjunior:
This is a serious problem, a feature should not be removed before the new one is implemented.

Please go back to the version where lazy is stable.

satya164:
> Please go back to the version where lazy is stable.

You are free to stay on older version. I need the library to move forward and improve performance on lower end devices.

`lazy` was never stable. There are several issues with it including bad performance. Besides, people use it only because React Navigation doesn't have hooks for events such as `focus`.

Either I can work on trying to get the feature work properly or work on improving more important parts of the library. I choose to do the later. react-native-tab-view is an open source library on which I work in my free time and don't get paid to work on, and I need to optimize how I spend my time on it for a better user experience overal.

kelset:
> Please go back to the version where lazy is stable.

@douglasjunior this PR has not been merged yet, so no reason to jump in to just be negative about a PR moving the lib forward.

Moreover I agree with @satya164, it would be better to finally implement the focus feature to replace the lazy; this way it should also improve performances for the Stack Nav.

satya164:
> but tbh we should be able to merge also the isFocused implementation before doing a new release

yeah, won't be too much work to implement

shilch:
Please adjust the version according to https://semver.org/ next time.
NPM put the line `"react-navigation": "^1.0.0-beta.22"` in the package.json. This version did support the `lazy` option. However, after reinstalling the node_modules it did not work any longer and I had to look for the error and found this PR here.

PR #3094, opened 2017-12-07
Don't use params from unrelated action when initializing nav state

Ashoat:
`TabRouter.getStateForAction` has some code at the start to initialize an empty nav state before handling any action. This initialization is achieved by passing `INIT` actions to all child routes. However, the code strangely also passes these child routes the params that were given to the action. The initialization should be separate from the processing of the action.

### DIFF
--- a/src/routers/__tests__/TabRouter-test.js
+++ b/src/routers/__tests__/TabRouter-test.js
@@ -215,7 +215,6 @@ describe('TabRouter', () => {
     const navAction = {
       type: NavigationActions.NAVIGATE,
       routeName: 'Baz',
-      params: { foo: '42', bar: '43' },
     };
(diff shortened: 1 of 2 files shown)


src/routers/__tests__/TabRouter-test.js
```diff
@@ -215,7 +215,6 @@ describe('TabRouter', () => {
     const navAction = {
       type: NavigationActions.NAVIGATE,
       routeName: 'Baz',
-      params: { foo: '42', bar: '43' },
```
Ashoat: Specifying params to a `NAVIGATE` call that hits a router is non-sensical, as a router cannot take params. Indeed, the only reason this had any effect in the first place was because of the weird way the removed code behaved.

PR #3124, opened 2017-12-14
Safe area view: support animated style values

steobrien:
### Motivation
https://github.com/react-community/react-navigation/pull/2833/files#diff-7d1ace708bab0c8203594be64846ebecL318 appears to have broken a previously supported use case: passing animated values (e.g. an interpolated value) with `headerStyle`.

### Changes
- Make `SafeAreaView` use an `Animated.View` in place of a regular `View`.
- Update some (unrelated) test snapshots, which appear to have diverged

I believe this fixes https://github.com/react-community/react-navigation/issues/3108


kelset:
cc @davepack or @spencercarli (?)

spencercarli:
Looks good!

PR #3137, opened 2017-12-16
Fix issue with SET_PARAMS causing navigation

ghuh:
This is a fix for https://github.com/react-community/react-navigation/issues/1711 where setting params on a tap is causing the tab view to go to the that tab even if it was not the current active tab.

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -190,6 +190,12 @@ export default (
             ...state,
             routes,
           };
+        } else {
+          // Don't navigate on SET_PARAMS
+          // Fix for https://github.com/react-community/react-navigation/issues/1711
+          return {
+            ...state,
+          };
         }
       }
       if (activeTabIndex !== state.index) {


kelset:
Seems to be working fine, tested in my prod app.

@satya164 I think this can be merged, do you mind double checking?

ericvicenti:
I think this change will break things. By returning early from this spot in the function, you will not reach the code which allows child routers to handle the action (https://github.com/react-navigation/react-navigation/pull/3137/files#diff-eabbcca3a86b04810423baa86b608f46L209)

If you have a StackNavigator inside a TabNavigator, I think this will prevent SET_PARAMS from working on any of the screens in the stack.

The example code to reproduce #1711 is pretty confusing to me (https://snack.expo.io/HJmosSpcW). I'm not sure why the params are being set after mounting. Can you post a simple example of your bug, and explain what you're trying to do, so I can help find an alternative solution?

normdoow:
I am having a similar issue to this. I'm not using the Tab Navigator however. Instead I have a StackNavigator nested inside a DrawerNavigator. Stacks set up like this...
```const innerDrawerStack = StackNavigator({
    Home: { screen: Menu },
    CardsView: { screen: CardsView },
    Feedback: { screen: FeedbackView }
}, {
    navigationOptions: ({navigation}) => ({
        headerLeft: <HamburgerButton navigation={ navigation } />,
        headerStyle: {
            backgroundColor: colors.primaryColor,
        },
        headerTitleStyle: {
            fontWeight: 'bold',
            fontSize: 16,
            color: 'white'
        },
        gesturesEnabled: false
    }),
    initialRouteName: 'Home'
})

const DrawerStack = DrawerNavigator({
    InnerStack: { screen: innerDrawerStack } //inner drawer stack adds the header since drawers don't have a header
},{ contentComponent: CustomDrawerView })```

my component `Home` calls `this.props.navigation.setParams({credit: resp.val.toFixed(2)});` to set the navigation options dynamically for that component. This works fine when navigating to the component normally. But when calling reset to clear the stack. like this...
```resetNavToHome(props) {     //pass in the props of the component to use for navigation
        return props
               .navigation
               .dispatch(NavigationActions.reset({
                    index: 0,
                    actions: [
                        NavigationActions.navigate({ routeName: 'DrawerStack'})
                    ]
                }));
    }```
This will take me to the drawerstack which then goes to the nested initial route `Home` but when i call `setParams` this time it will restart the d`Home` component in an infinite loop and will freeze.

I believe part of the issue with this is the nested stacks. almost like the setParams is dealing with the wrong stack since it just got reset. I think this is the same issue as the TabNavigator issues. If you have any more questions I would be glad to try and give more info. I think this is still an issue. I would greatly appreciate a fix or work around as I need this dynamic capability for my header.

PR #3151, opened 2017-12-19
Update DrawerNavigator.js

HunDunDM:
Set the default value of the DrawerNavigatorConfig correctly.


HunDunDM:
sad, unfamiliar with `flow`

PR #3152, opened 2017-12-19
Set the default value of the DrawerNavigatorConfig correctly.

HunDunDM:
If `DrawerNavigatorConfig` has been used before, the default `drawerOpenRoute` `drawerCloseRoute` `drawerToggleRoute` will not be correct.


HunDunDM:
I'm not sure if the original code is inherently problematic. I do not know how to solve it. But I think this change is no problem.........

wachunei:
How can we help getting this through? 
As fas as I know, example project was fixed by adding these configs instead of fixing this library, which I believe should be done instead and prioritize.

HunDunDM:
I do not understand why the examples made such changes. Is this expected?

oroundo-tech:
This is great solution. I have came to this problem, and found issue my self, and thought: why haven't they put those things in DefaultDrawerConfig, this makes no sense. I wasted too much time finding were the problem is. Now I see this PR and now I'm asking myself: why this PR has not been merged already. I can help if needed, just to get this trough. I've already fixed my code, but for the sake of other users and project itself.

satya164:
@kelset this PR looks good

kelset:
Awesome, I've run the update branch, if circleCi passes feel free to merge :)

HunDunDM:
I refer to `src/views/TabView/TabView.js` to change the type of `DrawerViewProps` in `src/views/Drawer/DrawerView.js`. Now circleCi has passed.

PR #3191, opened 2017-12-27
Fix drawer toggle

rgovind92:
Drawer does not toggle as per [this](https://github.com/react-community/react-navigation/issues/2760) issue. `this._drawer.state.drawerShown` seems to be available only on iOS?

**Test plan (required)**

Works on both iOS and Android (Issue didn't exist in iOS in the first place)


GantMan:
I have the code updated --- so just the double equals to a triple (our JavaScript curse)  please.

rgovind92:
Changed double equals to a triple in fix_drawer_toggle branch.

Issue #3206, opened 2018-01-02
Custom navigation prop and redux integration


rodan-lewarx:
### Current Behavior
- When integrating react-navigation with redux (as described in documentation) I found out an issue. I tried to pass custom navigation prop = I didnt want to use addNavigationHelpers and I wanted to construct navigate, setParams and goBack actions on my own. I expected that such modified navigation prop object when passed to top level router would be available in all the screens. This didnt happen. I have found out that even custom props get overwritten in `react-navigation/src/views/CardStack/CardStack.js` 

on line 405 `const { navigation } = this._getScreenDetails(scene);` as `this._getScreenDetails` calls `addNavigationHelpers` when screendetails are not saved and this navigation gets passed into SceneView

### Expected Behavior
- I was expecting that this (line 405 of Cardstack.js) would behave like 

`const navigation = this.props.navigation || this._getScreenDetails(scene).navigation;`

 instead. So that if there is a custom prop passed it will always get distributed further down to SceneView and its context.
- Am I missing something. Or is it okay to make PR like this?

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation  | 1.0.0.-beta22
| react-native  |  0.50.4
| node         |  v8.9.1 
| yarn     | 1.3.2

kelset:
👋 there.

Yeah sure open a PR, so it can be tested and further discussed :)

rodan-lewarx:
@kelset I have created the PR for testing here #3208 

So you can check it out.

I dont think this is a feature request but a bug, because [Redux integration docs](https://reactnavigation.org/docs/guides/redux) say that:

> Once you pass your own navigation prop to the navigator, the default navigation prop gets destroyed. You will most probably pass the navigation prop's properties that you want to access. Normally state and dispatch properties are passed to the navigator. You will learn how to pass those properties further in this guide. Since you have destroyed the default props, if you try to invoke something you have not explicitly passed down, it won't work. So, if you didn't pass dispatch to the navigator and only passes state than you can't access dispatch further in your Components.

Which is not happening in this case and in the end the navigation prop gets always overwritten in the screen.

kelset:
Awesome, thanks.

(and yeah I actually tagged it as a bug earlier, but then I was a bit concerned that would be a bit "too harsh", dunnow)

brentvatne:
@Ashoat - not really sure what to think about this one

Ashoat:
[...]

PR #3208, opened 2018-01-02
Overwrite navigation props only if they are not received as a prop

rodan-lewarx:
**Issue**
This refers to: #3206 

[React-navigation documentation on Redux integrations](https://reactnavigation.org/docs/guides/redux) states in point 2 that: 

> Once you pass your own navigation prop to the navigator, the default navigation prop gets destroyed.

I dont think that is what is happening. I tried the redux integration like in the example and when I pass my own navigation prop to top level connected navigator (stacknavigator in my case) without using `addNavigationHelpers` function (I wanted to use custom navigate, goBack functions) these got always overwritten along the way. I was able to log the exact spot and it happened here in `CardStack.js` because `this._getScreenDetails` calls `addNavigationHelpers` internally and creates new navigation prop that gets passed around.

**How this can be replicated?**
Use stacknavigator as in connect to redux example. And insert custom navigation prop (for example with custom navigation, goBack, etc function or for test purposes with entirely new key and value, for example `'test': 'test'` ). Then try to console log received props on the screens and you will find that these custom values are not there.

**How this can be tested**
After this one change the values are correctly propagated in screens

### DIFF
--- a/src/views/CardStack/CardStack.js
+++ b/src/views/CardStack/CardStack.js
@@ -64,7 +64,7 @@ type Props = {
 
 /**
  * The max duration of the card animation in milliseconds after released gesture.
- * The actual duration should be always less then that because the rest distance 
+ * The actual duration should be always less then that because the rest distance
  * is always less then the full distance of the layout.
  */
 const ANIMATION_DURATION = 500;
@@ -402,7 +402,8 @@ class CardStack extends React.Component<Props> {
     SceneComponent: NavigationComponent,
     scene: NavigationScene
   ): React.Node {
-    const { navigation } = this._getScreenDetails(scene);
+    const navigation =
+      this.props.navigation || this._getScreenDetails(scene).navigation;
     const { screenProps } = this.props;
     const headerMode = this._getHeaderMode();
     if (headerMode === 'screen') {


rodan-lewarx:
Let me know if more informations/examples are needed. Maybe there is something I dont see and this has other consequences. Maybe this is expected behavior and not a bug. But then I think  the documentation in redux integration needs to be changed.

rodan-lewarx:
@satya164 Thank you for comments.

So the screen navigation prop passed to sceneview and screens is different by design from the navigation prop passed to navigators when connecting them to redux (as in the example in documentation)? The difference is due to the state, which should be different?

What do you think is the best way to make available custom functions on navigation prop in the screens? So that if I define some other helpers or overwrite some functions in navigation passed to navigator when integrating with redux that this navigation will be available on screens?

Is it something you suggested in second paragraph: `addNavigationHelpers({ ...this.props.navigation, state: scene.route })` if it will be updated for all types of navigators? Would it be ok to do it without the `addNavigationHelper` part (if I would want to pass overwritten navigate, goBack actions then use of addNavigationHelpers would overwrite those) so just `{ ...this.props.navigation, state: scene.route }` ?

satya164:
> The difference is due to the state, which should be different

Yes, the navigator's `navigation.state` includes the state of the navigator, which is the list of the children routes, index etc. while screen's `navigation.state` includes the state of the screen, which is the particular route and its children.

> What do you think is the best way to make available custom functions on navigation prop in the screens?

I think overwriting behavior of the default methods shouldn't be allowed, and you should avoid it for to avoid confusion.

To provide additional methods on the `navigation` prop, you can wrap the navigator/screens in an HOC to do that, like I did in https://github.com/satya164/react-navigation-addons. Don't think there's a better way right now.

> Is it something you suggested in second paragraph

That'd be the way to do it with the current PR, but I'm not sure it's a good idea.

I think better to go with the HOC approach for now.

rodan-lewarx:
> I think overwriting behavior of the default methods shouldn't be allowed, and you should avoid it for to avoid confusion.

@satya164 I was thinking this was possible based on the docs for redux-integration that say:

> Once you pass your own navigation prop to the navigator, the default navigation prop gets destroyed. You will most probably pass the navigation prop's properties that you want to access. Normally state and dispatch properties are passed to the navigator. You will learn how to pass those properties further in this guide. Since you have destroyed the default props, if you try to invoke something you have not explicitly passed down, it won't work. So, if you didn't pass dispatch to the navigator and only passes state than you can't access dispatch further in your Components.

I wanted to overwrite these because I didnt want someone to call the default actions instead of the custom ones I am using. But it seems that this is not possible.

brentvatne:
cc @ericvicenti

ericvicenti:
I think this change would cause some problems because, as Satya pointed out, there is a big difference between the parent navigation prop that goes into CardStack, and the child navigation prop that is passed to the child

It’s not super clear what the original issue is. Can you maybe post some code that demonstrates the issue, then we can help you find a solution?

rodan-lewarx:
Sorry for delay in answering.

The original issue was that when connecting to redux I didnt want to use addNavigationHelpers when creating the navigation prop, so I could inject my own functions for navigate, reset, etc. But I couldnt use that in the screen because the navigation prop got overwritten along the way with addNavigationHelpers used.

In the docs there was mention that I was responsible for making the navigation prop so that it would be available on screens, but it gets overwritten along the way.

PR #3267, opened 2018-01-12
Issue #2794 - Tab bar fix - Continued from #3041

SteffeyDev:
This is a continuation of #3041, please look there for reasoning and description.

This further modifies the conditional checks on horizontal tabs, using this algorithm:

```js
const isHeightConstrained = layout.height < 450;
const isWidthConstrained = layout.width < 450;
if (isWidthConstrained) {
  return false;
} else if (isHeightConstrained) {
  return true;
} else {
  return canFit;
}
```
where `canFit` is true if we can fit horizontal tabs safely, false if not.

@spencercarli @satya164 Let me know if this makes sense as an approach, or what other ideas you have.


SteffeyDev:
@satya164 I know everyone is busy, but I opened this MR the same day that you reverted my original one, and I would appreciate some sort of status update so I can gauge when this will go in.  I would love to use the latest version of your library in my app but am stuck until these fixes go through.

brentvatne:
@SteffeyDev - sorry we are in the middle of doing various other high priority tasks for the future of react-navigation, we'll get to this. until then you can use a fork with this code :)

src/views/TabView/TabBarBottom.js
```diff
@@ -158,6 +165,68 @@ class TabBarBottom extends React.PureComponent<Props> {
...
+    if (!adaptive) {
+      return false;
+    }
+
+    let tabBarWidth = layout.width;
+    if (tabBarWidth === 0) {
+      return isTablet;
+    }
+
+    const isWidthConstrained = layout.width < 450;
```
brentvatne: can we just use horizontal tabs 1) on ipad always 2) on iphone when in landscape?

no need for magic numbers like 450 here

SteffeyDev: No, because on tablets in split screen mode, the app could be as wide as it would be on a phone, and there is no space for horizontal tabs (this was the original issue #2794).  We need to be able to handle all sorts of split-screen configurations, Platform.isPad is useless for this because it can't detect split-screen.  I believe split-screen on Android is becoming more popular as well.

SteffeyDev: Please check original conversion in #3041 for more details on motivation for change and methodology chosen.

SteffeyDev:
@brentvatne Ignore previous comments, in a similar fashion to #3251 I have moved to `Platform.isPad` as the determinator of whether we should layout on a has-space basis or should use device orientation.  Let me know what you think.

PR #3280, opened 2018-01-16
added hiddenTabs property for TabBarBottom component

Arekkusuva:
Now you can pass `hiddenTabs` property (array with tabs indexes) to `TabBarBottom` component. That allows you to hide certain tabs dynamically.

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -195,8 +195,9 @@ class TabBarBottom extends React.PureComponent<Props> {
           forceInset={{ bottom: 'always', top: 'never' }}
         >
           {routes.map((route: NavigationRoute, index: number) => {
-            if (hiddenTabs.indexOf(index) > - 1) return null;
-
+            if (hiddenTabs.indexOf(index) > -1) {
+              return null;
+            }
             const focused = index === navigation.state.index;
             const scene = { route, index, focused };
             const onPress = getOnPress(previousScene, scene);
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -51,6 +51,7 @@ type Props = {
   tabStyle?: ViewStyleProp,
   showIcon?: boolean,
   isLandscape: boolean,
+  hiddenTabs: Array<number>,
 };
 
 const majorVersion = parseInt(Platform.Version, 10);
@@ -67,6 +68,7 @@ class TabBarBottom extends React.PureComponent<Props> {
     showLabel: true,
     showIcon: true,
     allowFontScaling: true,
+    hiddenTabs: [],
   };
 
   _renderLabel = (scene: TabScene) => {
@@ -171,6 +173,7 @@ class TabBarBottom extends React.PureComponent<Props> {
       animateStyle,
       tabStyle,
       isLandscape,
+      hiddenTabs,
     } = this.props;
     const { routes } = navigation.state;
     const previousScene = routes[navigation.state.index];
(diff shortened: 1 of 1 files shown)


satya164:
Why do you need to render a tab if there's no way to access it?

Arekkusuva:
@satya164 What do you mean?

satya164:
Please explain the use case behind the PR. This PR hides the entry in the tab bar while the page is still rendered. Why do you need this feature?

Arekkusuva:
@satya164 I needed ability to hide certain tabs from TabBar dynamically. And i believe that's the fastest way to do this. Also you can set `lazy` property to true if you don't want to render hidden tabs.

Arekkusuva:
@satya164 And I believe we can control the rendering of certain pages with standard methods of this library and React. `lazy` property only one of several methods to do this.

satya164:
> I needed ability to hide certain tabs from TabBar dynamically

What's the use case that requires hiding tab items from tab bar, but not from the rendered pages?

> Also you can set lazy property to true if you don't want to render hidden tabs.

`lazy` option was removed. Also, that's not how it used to work. It was used to load tabs laziliy until you access them for the first time, and they still stay rendered after you navigate away.

Anyway, can you describe the use case? I know you want to hide an item from tab bar, but "why"? It's not possible to decide if this is a reasonable change if you don't explain the use case.

nikasvan:
@satya164 Use case for this would be:

Using StackNavigator, TabNavigator and DrawerNavigator in the app and trying to show TabBar inside DrawerNavigator screens. Here is my [Stackoverflow question](https://stackoverflow.com/questions/47889379/making-tabnavigators-tabbar-always-visible-on-every-route-of-the-app) which I have asked 4-5 times before this and got no answer from anyone. 

@Arekkusuva Thanks for this, I did something similar to this and it should be merged. Just one thing though did you make it to work for android(in TabBarTop.js, I think)?

Arekkusuva:
@satya164 I use this in react native project to allow users customize UI.

Arekkusuva:
@nikasvan If you know how to implement this better, then do it please. Since I'm not that good at this library.

nikasvan:
@Arekkusuva I was asking this because I have not been able to do it for Android yet:D. 

If you would like your PR to be merged most important rule is to make sure that it works both for iOs and Android as well!

Arekkusuva:
@nikasvan Well, i’ll check it tomorrow. Also i will try to remove hidden tabs from rendered pages.

satya164:
> Using StackNavigator, TabNavigator and DrawerNavigator in the app and trying to show TabBar inside DrawerNavigator screens.

You need to have a custom TabBar if you want to collapse multiple items to one item. Just hiding items won't solve your use case. You still need a custom tab bar to be able to show a "more" button. And if you are making your custom tab bar, it's straightforward to hide some items.

> I use this in react native project to allow users customize UI.

That still doesn't explain the use case. Why do you want to hide tab items when the pages are still rendered?

Arekkusuva:
@satya164 I’ll try to do this later. I mean pages.

nikasvan:
So Found the solution for Android. The thing is that for Android react-navigation is using [react-native-tab-view](https://github.com/react-native-community/react-native-tab-view). So to make it work what needs to change is in _react-native-tab-view/src/TabBar.js_ on line 449 just after components `return` statement add `this.props.hideIndex.indexOf(i) > -1 ? null : <TouchableItem.......` http://prntscr.com/i45qr8

brentvatne:
closing this, please make the required change in react-native-tab-view if they will accept it :) then open a pr here and follow the pull request template

```

**Test plan (required)**

The code must pass tests.

**Code formatting**

```

brentvatne:
oh and of course thank you so much for taking the time to open this in the first place :) sorry we didn't come to a solution that we can land here, looking forward to the follow up if you're still motivated

PR #3282, opened 2018-01-16
Add example with navigation reset message

pietro909:
This PR adds another example in the NavigationPlayground which shows **how to reset a nested navigator** (Stack into Tab). This is among the most asked questions, and I had to find out on my own.

See the screencast and notice how:
1. start from _Home_ tab
2. navigate three times to the _Profile Screen_ 
3. enter _Settings_ tab
4. go back to _Home_ and see that the navigation stack was reset to the first screen

reset-message-example

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -23,6 +23,7 @@ import Drawer from './Drawer';
 import MultipleDrawer from './MultipleDrawer';
 import TabsInDrawer from './TabsInDrawer';
 import ModalStack from './ModalStack';
+import ResetNestedNavigators from './ResetNestedNavigators';
 import StacksInTabs from './StacksInTabs';
 import StacksOverTabs from './StacksOverTabs';
 import SimpleStack from './SimpleStack';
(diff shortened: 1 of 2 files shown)


kelset:
LGTM

brentvatne:
@ericvicenti - do you think this is useful to add to playground?

ericvicenti:
I think the example is a bit hard to understand. Its very hard to tell what is happening, or when you would want this in an app.

This seems to demonstrate functionality of the tab bar, more than anything else. I didn't know this API even existed: 

```
tabBarOnPress: ({ previousScene, scene, jumpToIndex }) => {
```

Maybe this could be framed as an advanced tab bar behavior example, rather than having a whole example for this one specific type of navigation

Could we spice things up with a real use case, or have a puppies tab and kittens tab or something?

pietro909:
The whole point of this PR was to provide a working example of a navigation reset message, which is missing. There's plenty of related questions (even here in the issues) and I couldn't find _one_ working example. Even the docs are unclear on this point.

> when you would want this in an app

Reset one of the tabs while navigating to another one, so that when you come back it will show the `initialRoute` rather than the last page you visited.

> an advanced tab bar behavior example

I don't think that this is an _advanced_ behavior: it is a straightforward feature of a navigation framework.

brentvatne:
hey @pietro909 - this seems like something we might want to add to `TabNavigator` if it is common enough. probably the way I'd phrase it is to unmount the tab entirely if you switch away from it.

as for providing more clear information about reset, I agree this would be great. we have a lot of work to do on the ergonomics around reset. I created an issue on our docs page about this: https://github.com/react-navigation/react-navigation.github.io/issues/37

PR #3289, opened 2018-01-17
Export index of scene for navigation state props.js

kishanmundha:
Exporting index of current scene in navigation state props

Sometime we need current route index and based on index we take some decision. In my case, I want to show my custom left component if there is no back route otherwise back button should show. There is no way to find route index inside screen component header navigation props.

This solve our issue by checking current index.

### DIFF
--- a/src/views/CardStack/CardStack.js
+++ b/src/views/CardStack/CardStack.js
@@ -145,13 +145,13 @@ class CardStack extends React.Component<Props> {
         dispatch: navigation.dispatch,
         state: {
           ...scene.route,
-          index: scene.index
+          index: scene.index,
         },
       });
       screenDetails = {
         state: {
           ...scene.route,
-          index: scene.index
+          index: scene.index,
         },
         navigation: screenNavigation,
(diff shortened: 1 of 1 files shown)


joshjens:
yes please! i need this 👍

ericvicenti:
Did you test this on the navigation playground app? It breaks all of the tabs examples, because they are rendered under the CardStack. The problem is that the index of the parent card stack is overriding the index of the child navigation state.

@joshjens, why do you need this? There is probably another way to fix your problem

kishanmundha:
Hi @ericvicenti,

I didn't test on navigation playground app and I'm not using navigation tab so I'm not aware about issue with tab, but we need solution for the same.

Can you provide a example to get current route index in navigation props. I explain my issue in pull request. I'm rendering left back button based on current route index.

jemmyphan:
I need this feature as well and since we only need it on headerLeft, so i try it with another approach. https://github.com/jemmyphan/react-navigation/commit/1220681630cf6a561405bf3ff437ed027d9d1612
how do you think about that?

PR #3304, opened 2018-01-18
Pass prevTransitionsProps to transitionConfig

MatthieuLemoine:
[StackNavigator documentation](https://reactnavigation.org/docs/navigators/stack#StackNavigatorConfig) says that the provided `transitionConfig` function (allowing to override default transitions) will be passed `transitionProps` and `prevTransitionProps` as arguments. But this isn't always the case : 

`transitionProps` and `prevTransitionProps` are only passed to compute `transitionSpec`.
https://github.com/react-navigation/react-navigation/blob/b221afc80b223d88ae6adffcebbaec67b46f6215/src/views/CardStack/CardStackTransitioner.js#L68-L74

For all the other properties, two empty objects are passed instead : 
https://github.com/react-navigation/react-navigation/blob/41b587ca657d486a545e796e58a22e4b836d3311/src/views/CardStack/CardStack.js#L431-L442

This is a bummer when you need to have access to the previous transition in `screenInterpolator` for example.

This PR makes the `transitionConfig` API consistent by always passing `transitionProps` and `prevTransitionsProps`.

**Test plan (required)**

Test : ✅ 
Flow : ✅ 

**Code formatting**

ESLint : ✅


brentvatne:
@MatthieuLemoine - can you rebase this? happy to review next week

MatthieuLemoine:
@brentvatne rebased

src/views/CardStack/CardStack.js
```diff
@@ -85,7 +85,7 @@ class CardStack extends React.Component {
     if (props.screenProps !== this.props.screenProps) {
       this._screenDetails = {};
     }
-    props.scenes.forEach(newScene => {
```
brentvatne: shouldn't have a type annotation here. we removed flow from codebase

brentvatne:
@ericvicenti - can you review when you get a chance? I'm not too familiar with this part of the code base yet

PR #3305, opened 2018-01-18
HeaderBackButton animation performance

michaelduminy:
**Motivation**
On Android the ripple animation on the header's default back button is not as smooth as it could be. This is obvious in Dev mode where the animation proceeds very slowly and only upon completion does the action occur. Wrapping the `onPress` function in `InteractionManager` allows the animation to be as performant as possible.

**Test plan**
It's a simple enough PR. Not sure much testing is required here. I tested this with a complex app on a physical Android device and on iOS via Expo Snack. No problems.

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -11,6 +11,7 @@ import {
   StyleSheet,
   View,
   ViewPropTypes,
+  InteractionManager,
 } from 'react-native';
 
 import HeaderTitle from './HeaderTitle';
@@ -100,7 +101,7 @@ class Header extends React.PureComponent<Props, State> {
   }
 
   _navigateBack = () => {
-    this.props.navigation.goBack(null);
+    InteractionManager.runAfterInteractions(this.props.navigation.goBack);
   };
 
   _renderTitleComponent = (props: SceneProps): ?React.Node => {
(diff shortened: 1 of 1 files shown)


src/views/Header/Header.js
```diff
@@ -100,7 +101,7 @@ class Header extends React.PureComponent<Props, State> {
   }
 
   _navigateBack = () => {
-    this.props.navigation.goBack(null);
+    InteractionManager.runAfterInteractions(this.props.navigation.goBack);
```
satya164: Replace this with `requestAnimationFrame` instead

michaelduminy: Done 👍  Though I am curious why.

ericvicenti:
Does `requestAnimationFrame` fix this issue as well as the interaction manager?

Does this result in any noticeable delay? If not then I think this change is good

michaelduminy:
@ericvicenti in my testing `requestAnimationFrame` and `InteractionManager` perform equally well. At first I was hesitant because it seems more idiomatic to use `InteractionManager` but will defer to the collaborators on this.

satya164:
Using `InteractionManager` will be a bad experience because it'll wait till all interactions are finished before calling the method, e.g. if an animation is running when you press the back button, it will wait till that animation is finished, which could be several seconds depending on the animation.

`requestAnimationFrame` works because it makes calling the function async, which means the button animation won't wait till the function returns.

michaelduminy:
@satya164 Thanks for explaining, great point!

PR #3316, opened 2018-01-20
Fix the section link for Reset action

pranavcode:
**Motivation**
Sectional link to Reset section fails due to case difference.

**Test plan**
As this is a view change on the doc, test plan does not apply.


stereodenis:
@pranavcode I made this change, cause in github it should be in lowercase, but on a website it should be with first uppercase letter

PR #3323, opened 2018-01-22
Export UnwrappedHeader so Header can be customized easier

CrazyPython:

If an end-user wants to customize the HeaderTitle, this PR exports a Header class that has not had `withOrientation` applied to it.

Example: 
```js
class EditableHeaderTitle extends React.Component {
    constructor(props) {
        super(props)
        this.state = { text: props.initialText }
    }

    render() {
        return (
            <View pointerEvents={'auto'} style={{ height: 30, width: 500 }}>
                <TextInput
                    style={{
                        flex: 1,
                        fontSize: Platform.OS === 'ios' ? 17 : 20,
                        fontWeight: Platform.OS === 'ios' ? '600' : '500',
                        color: '#000',
                        textAlign: Platform.OS === 'ios' ? 'center' : 'left',
                        marginHorizontal: 16,
                        autoCapitalize: 'words',
                    }}
                    onChangeText={(text) => {
                        this.setState({ text })
                        this.props.onChangeText()
                    }}
                    value={this.state.text}
                />
            </View>
        )
    }
}

class EditableHeader extends UnwrappedHeader {
    _renderTitleComponent = (props: SceneProps): ?React.Node => {
        // $FlowFixMe
        const details = this.props.getScreenDetails(props.scene)
        const headerTitle = details.options.headerTitle
        if (React.isValidElement(headerTitle)) {
            return headerTitle
        }
        const titleString = this._getHeaderTitleString(props.scene)

        const titleStyle = details.options.headerTitleStyle
        const color = details.options.headerTintColor
        const allowFontScaling = details.options.headerTitleAllowFontScaling

        // On iOS, width of left/right components depends on the calculated
        // size of the title.
        const onLayoutIOS =
            Platform.OS === 'ios'
                ? (e: LayoutEvent) => {
                    this.setState({
                        widths: {
                            ...this.state.widths,
                            [props.scene.key]: e.nativeEvent.layout.width,
                        },
                    })
                }
                : undefined

        const RenderedHeaderTitle = EditableHeaderTitle
        return (
            <RenderedHeaderTitle
                onLayout={onLayoutIOS}
                allowFontScaling={allowFontScaling == null ? true : allowFontScaling}
                style={[color ? { color } : null, titleStyle]}
            >
            </RenderedHeaderTitle>
        )
    }
}

export default withOrientation(EditableHeader)

// ...

static navigationOptions = {
        header: ((props: *) => <EditableHeader onChangeText={console.warn} {...props} />),
}
```

If we inherit directly off of Header as wrapped by withOrientation, the overridden methods won't be used. 

- [x] Prefer **small pull requests**. 

**Test plan (required)**
The only changes made are:
```diff
-class Header extends React.PureComponent<Props, State> {
+export class UnwrappedHeader extends React.PureComponent<Props, State> {
    static defaultProps = {
      leftInterpolator: HeaderStyleInterpolator.forLeft,
      titleInterpolator: HeaderStyleInterpolator.forCenter,
 @@ -405,4 +405,4 @@ const styles = StyleSheet.create({
    },
  });
  
-export default withOrientation(Header);
+export default withOrientation(UnwrappedHeader);
```

That being said, the example renders:

I am using the same library without a modified header in the rest of my app, and it doesn't crash elsewhere.

I ran `yarn test`. All tests passed.

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -48,7 +48,7 @@ const STATUSBAR_HEIGHT = Platform.OS === 'ios' ? 20 : 0;
 const TITLE_OFFSET = Platform.OS === 'ios' ? 70 : 56;
 
 type Props = HeaderProps & { isLandscape: boolean };
-class Header extends React.PureComponent<Props, State> {
+export class UnwrappedHeader extends React.PureComponent<Props, State> {
   static defaultProps = {
     leftInterpolator: HeaderStyleInterpolator.forLeft,
     titleInterpolator: HeaderStyleInterpolator.forCenter,
@@ -405,4 +405,4 @@ const styles = StyleSheet.create({
   },
 });
 
-export default withOrientation(Header);
+export default withOrientation(UnwrappedHeader);


brentvatne:
hi! I love the idea to make the header more extensible, but I am a bit concerned about an approach that depends on subclassing the header -- might as well just fork it in this case. i'm not sure what the best approach is here but we should discuss it more in a rfc. if you have an idea, please submit to https://github.com/react-navigation/rfcs or open an issue there to discuss it

CrazyPython:
@brentvatne hi, can you clarify your concerns on subclassing the header? Different ideologies have opinions on whether Conventional OO is good or not, but this is the simplest solution. The prototype chain was created to allow inheritance. 

**Forking is not a viable option.** Initially my solution was to copy all of the code from Header.js. When I tried removing unnecessary methods and inheriting, but `withOrientation` prevented that. I didn't want to modify the default header class. 

Open for extension, closed for modification.
---
I don't see why an RFC is nessecary. I changed two lines of code in a way that does not affect the rest of the codebase whatsoever. Can we discuss it here instead?

satya164:
> Conventional OO is good or not, but this is the simplest solution. The prototype chain was created to allow inheritance.

You shouldn't use inheritance to compose React components, the way to compose them is to render them and to pass props to customize them.

Regardless, our components are not written with OOP in mind. The methods and properties you're overriding/accessing are internal and might change in any point, we could change it to a function component tomorrow, rewrite the implementation or wrap it in an HOC. Exposing the internals makes it very difficult for us to make any kind of change to the component without breaking people's code. The current contract is simple, you have React component and you pass props to customize that. It's a pretty good contract which gives us flexibility over code while keeping breaking changes to minimum.

> I don't see why an RFC is nessecary.

Because that's how we do things, we make RFCs before changes so that we can arrive on a good implementation.

> I changed two lines of code in a way that does not affect the rest of the codebase whatsoever.

The two lines of changes will massively affect the maintenance cost of the Header component.

CrazyPython:
> It's a pretty good contract which gives us flexibility over code while keeping breaking changes to minimum.
Maybe, but it is bad for extensibility. Again, if I want to modify Header at all, I have to copy all of the code from the Header file.

> while keeping breaking changes to minimum.
People are already forking the repository and adding their own breaking changes.

I propose this structure:
- create an `AbstractHeader` class that has the simplest possible contract. 
- create a `LeftRightHeader` class that is the current Header and inherits off of `AbstractHeader`
- when you want to modify `LeftRightHeader` in a way that breaks functionality, simply make a new `AbstractHeader` and make that one the default. You can mark `LeftRightHeader` as deprecated.

satya164:
Please create an RFC!

> constructor with arguments and a render()

That's what a React component is for, why is a constructor better than `<Header {...someProps} />`? If you need to customize something and it's not possible, suggest a prop.

> People are already forking the repository and adding their own breaking changes.

When people fork the repo and make changes, it doesn't affect our maintenance cost.

Anyway, please open an RFC.

CrazyPython:
An RFC can't be changed. I want to have a discussion before having one. A constructor is better because it can be adapted. Your users will do things that you don't have the time to add props for or could even imagine. `AbstractHeader`'s methods should all raise an error with the message "method is not implemented in abstract class". AbstractHeader is a contract: it contains only arguments, Flow annotations, and an error raised in case anyone tries to use AbstractHeader directly.

> If you need to customize something and it's not possible, suggest a prop.
Yes, but most of us don't have time to wait for a prop to be implemented.
> When people fork the repo and make changes, it doesn't affect our maintenance cost.
Yes. They don't get updates. Most people using this repo fork it. They don't do it direct, so backward compatibility is being broken either way.

satya164:
> An RFC can't be changed

RFC means "Request for comments", it can be changed. That's the sole purpose of it, ask for feedback and update the proposal to address them.

> Your users will do things that you don't have the time to add props for or could even imagine.

Sure, and that'll not be officially supported. If you don't care about breaking your code on updating the library, you can do whatever you want.

> AbstractHeader is a contract: it contains only arguments

How is "arguments" any different from props?

Anyway, please make an RFC if you want to discuss. I'm unsubscribing from this thread.

Issue #1363, opened 2017-05-03
[RFC] API for event emitter


satya164:
### Motivation

- There are a lot of things that can happen in a navigator which the user might want to get notified of. Right now there is no standard way of doing that.
- Many times, the user needs to call an instance method in the screen component when something happens in the tabbar/header, currently the only way is to do `setParams`, which is messy.

### Use cases
- Scroll to top on tab press (#314)
- Scroll to top on header title press
- Refresh data in screen on header button press
- Save button (#145)
- Navigation events (#51)
etc.

### Inspiration

ExNavigation has a built in event emitter https://github.com/expo/ex-navigation/blob/master/example/components/EventEmitterExample.js#L95

Disregarding the use case in the given example, which is better done with `params`, it's nice how easy it is to interact with the current screen from the tab bar/header/drawer to do one off tasks like refreshing, scrolling, saving etc. which don't actually need to change the state.

### API

The API is simple, to attach a listener:

```js
this.props.navigation.addListener('tabPress', callback)
```

To remove a listener:

```js
this.props.navigation.removeListener('tabPress', callback)
```

To emit an event:

```js
this.props.navigation.emit('tabPress', data)
```

### Technical details

The events are always scoped to the screen where they are defined. Dispatching an event in one screen will not cause another screen to receive the events.

The listeners are stored in the navigator and hence don't need to be reattached when the `navigation` prop changes, which matches how people will expect it to work.

cc @ericvicenti @skevy @grabbou @DanielMSchmidt

DanielMSchmidt:
First of all I like this approach of RFCs and also the proposed change. It is IMHO not overengineered and provides value for a lot of use cases. I have still some maybe a bit nooby questions:

- will we ignore the return value of the callback or should there be an.option to cancel these events or something similar? 
- will we dispatch redux actions for these events as well?
- what information will we pass to the callback?
- what additional cost of maintaining this do you expect?
- where will we need to touch the docs?

Hope this feedback helps more than it is idiotic 😇

satya164:
> will we ignore the return value of the callback or should there be an.option to cancel these events or something similar?

It won't be possible to cancel all events, so I think we should ignore the return value. However it;s still possible to pass a function in the data to cancel the event if necessary (like `e.preventDefault`) etc.

> will we dispatch redux actions for these events as well?

I'd say no. Since these are aimed at cross-communication in a single screen. People can dispatch redux actions in response to an event if the want.

> what information will we pass to the callback?

[...]

PR #3345, opened 2018-01-23
Navigation Event Subscriptions

ericvicenti:
Implements the following event subscription system, as proposed in https://github.com/react-navigation/rfcs/pull/1

# Summary

Add an event emitter to the navigation prop that emits focus/blur events for the route.

This proposal has been pioneered by @satya164: https://github.com/react-navigation/react-navigation/issues/1363

This will resolve longstanding requests for such functionality, as seen here: https://github.com/react-navigation/react-navigation/issues/51

# Basic example

In this example, we focus the first text input when the screen has completed transitioning:

```
componentDidMount() {
  this._sub = this.props.navigation.addListener(
    'didFocus',
    this._focusFirstTextInput
  );
}
componentWillUnmount() {
  this._sub.remove();
}
```

# Motivation

Screens need to know about navigation changes while they are mounted. Consider the following use cases:

### Data and Media subscription

Screens must know when to request live data or when to enable/disable resource-intensive APIs. Media should be able to auto-pause at correct times.

### Unmounting

Tabs and cards may need to unmount and re-mount contents when they are focused.

May be useful in conjunction with removeClippedSubviews and utilities like https://github.com/SoftwareMansion/react-native-resource-saving-container

### Keyboard dismissal

KB must dismiss when scene _starts_ to transition back (mayBlur)

### StatusBar Re-configuration

Screens may want to re-configure the status bar when the become active or blurred

### Input autofocus

Inputs want to open the keyboard after the screen finishes transition

### Related Concepts for Future work and exploration:

_Back button handling_ - Screens may want to prevent back actions based on internal state

_Refocus_ - On extra tab press, navigate to top of stack, then scroll to top on next press

# Detailed design

## Event System

An event emitter will be provided as part of each screen navigation prop. The events will be emitted by the navigator as changes happen to the navigation state.

`this.props.navigation.addListener` will be a `NavigationEventSubscriber`:

```
export type NavigationEventSubscriber = (
  eventName: string,
  eventHandler: (payload: NavigationEventPayload) => void
) => {
  +remove: () => void,
};
```

## Event Payload

Each event will emit a payload of the same shape, with the following type:

```
type NavigationEventPayload = {|
  +state: NavigationState,
  +lastState: NavigationState,
  +action: NavigationAction,
|};
```

This shows the previous and current state of the navigation route, and the action that was fired.

## Top-Level "action" Event

The "action" event will be emitted for _every_ navigation prop, when an action attempts to change the navigation state. The event only fires for the navigation state of the particular screen/navigator.

In some cases, the navigation state has not changed and `props.navigation.state === newState`. One example of this is the 'maySwipeBack' action fired by the stack navigator, which will be implemented in a following proposal.

## Navigator Events

The following events will available to all screens inside of navigators. Navigators will implement these additional actions by subscribing to its own `onAction` event from `props.navigation.addListener`.

### willFocus

Called when a screen will become focused, after it is mounted. Will be re-called when returning to the screen. Useful for starting to subscribe to new data.

onWillFocus may be called extra times, which is considered 'refocusing', which is most commonly tapping on a tab that is already selected. On iOS this extra focus may result in scroll-to-top behavior.

### didFocus

Called after the transition of focusing the screen. May be useful for input autofocus that waits for the transition to complete before focusing, or any other animation that should happen when the screen finishes becoming active.

### willBlur

The screen will be popping or another screen may become active. At this point screens may unsubscribe from heavy data subscriptions or APIs.

### didBlur

The screen is fully invisible now. This action will NOT fire when a screen is unmounted, for example at the end of a pop transition.

## "Is Transitioning" State

The nav state will be augmented with the `isTransitioning` flag to notify subscribers when a navigation animation is happening. The new structure of a navigation state (/route) is:

```
type NavigationState = {
  index: number,
  routes: Array<NavigationRoute>,
  isTransitioning: boolean,
};
```

When an animated transition begins, the index/routes change, and isTransitioning is toggled to true. If it does not switch to true, the navigation happens immediately without animation.

There will be a new navigation action for transition completion: 'CompleteTransitionAction', which toggles the `isTransitioning` boolean to false. This is done by StackNavigator, or any navigator once the transition completes.

This augmented state and new completion action is useful for navigators to provide the correct focus/blur events to children.

## Immediate Navigation Actions

The "Navigate" and "Back" actions will automatically set `isTransitioning: true`. To override this behavior and request an immediate transition without animation, you can provide a new `immediate: true` option to the navigate and back actions, and the action will leave `isTransitioning` as false.


dantman:
Can we include the eventName as `type` in the `NavigationEventPayload`. It would be polite to enable the standard pattern of subscribing a single handler to multiple events and conditioning on `event.type` in the case where the code in question has stuff that might be repeated if done in separate handlers.

For example:
```js
_handleFocusChanged(event) {
  this.setState({isFocused: event.type !== 'willBlur'});
}

componentWillMount() {
  this._sub1 = this.props.navigation.addListener('willBlur', this._handleFocusChanged.bind(this));
  this._sub2 = this.props.navigation.addListener('willFocus', this._handleFocusChanged.bind(this));
}
```

satya164:
Yeah, was thinking this too but couldn't come up with a use case. Thanks @dantman

Jonovono:
This looks great! Have you thought about how you would hook it up into redux? Instead of having to do this listening on every screen component, it would be cool to have it as a prop in redux and be like `this.props.isFocused` of some sort.

ericvicenti:
@dantman @satya164 , I agree that will come in handy! I'll add `type` to NavigationEvent 😄 

@Jonovono, this may slightly complicate the setup for redux apps. Apps with redux must manually fire the top level 'action' event (which causes all of the downstream focus/blur events). Otherwise these events will never fire. That said, redux users can already use a wide variety of techniques to handle state changes!

Regardless of Redux, I think `props.isFocused` will come in handy for lots of people, so we should offer a HOC to consume these events and provide that new prop.

src/TypeDefinition.js
```diff
@@ -140,16 +147,17 @@ export type PossiblyDeprecatedNavigationAction =
  * The state for the root navigator of our app represents the whole navigation
  * state for the whole app.
  */
+
+// TODO: rename NavigationState to NavigationParentState
```
brentvatne: do we want to address renaming this before landing this PR or is something else blocking it?

src/TypeDefinition.js
```diff
@@ -172,8 +180,16 @@ export type NavigationLeafRoute = {
   params?: NavigationParams,
 };
 
+// A route which describes a navigator state ("the route for a navigator")
 export type NavigationStateRoute = NavigationLeafRoute & NavigationState;
 
+// NavigationRoute is either a screen in a navigator (leaf) | a navigator (state)
+export type NavigationRoute = NavigationLeafRoute | NavigationStateRoute;
+
+// TODO: enable this after renaming NavigationParentState
```
brentvatne: same comment as above re: renaming NavigationParentState

src/TypeDefinition.js
```diff
@@ -404,11 +436,13 @@ export type NavigationDispatch = (
 
 export type NavigationProp<S> = {
   +state: S,
+  addListener: NavigationEventSubscriber,
```
brentvatne: let's maybe call this `NavigationNavigatorProp` (and as a follow-up, consider removing the prefix `Navigation` throughout our types)

src/getChildEventSubscriber.js
```diff
@@ -0,0 +1,137 @@
...
+): NavigationEventSubscriber {
+  const actionSubscribers = new Set();
+  const willFocusSubscribers = new Set();
+  const didFocusSubscribers = new Set();
+  const willBlurSubscribers = new Set();
+  const didBlurSubscribers = new Set();
+
+  const getChildSubscribers = evtName => {
+    switch (evtName) {
+      case 'action':
```
brentvatne: note: in follow-up pr, we should make sure that an event fires for mayBlur here (or whatever we want to call it) so we can hide keyboard etc when we start dragging

src/getChildEventSubscriber.js
```diff
@@ -0,0 +1,137 @@
...
+        return didBlurSubscribers;
+      default:
+        return null;
+    }
+  };
+
+  const emit = (eventName, payload) => {
+    const subscribers = getChildSubscribers(eventName);
+    subscribers &&
+      subscribers.forEach(subs => {
```
brentvatne: nit: `subs` sounds plural and it tripped me up a bit, might be better to make it `sub` or `subscriber`

src/getChildEventSubscriber.js
```diff
@@ -0,0 +1,137 @@
...
+        // $FlowFixMe - Payload should probably understand generic state type
+        subs(payload);
+      });
+  };
+
+  let isSelfFocused = false;
+
+  const cleanup = () => {
+    upstreamSubscribers.forEach(subs => subs && subs.remove());
```
brentvatne: same nit as above re subs

src/getChildEventSubscriber.js
```diff
@@ -0,0 +1,137 @@
...
+
+  const upstreamEvents = [
+    'willFocus',
+    'didFocus',
+    'willBlur',
+    'didBlur',
+    'action',
+  ];
+
+  const upstreamSubscribers = upstreamEvents.map(eventName => {
```
brentvatne: this maps to nulls and should return result of `addListener`

src/getChildEventSubscriber.js
brentvatne: also as discussed, this is never called. additionally, we should change upstreamSubscribers to lazy initialize when we add a listener and clean up when all listeners are removed, and this should be repeatable

src/getChildEventSubscriber.js
```diff
@@ -0,0 +1,137 @@
...
+        emit('willBlur', childPayload);
+      }
+      if (!isFocused && didFinishTransitioning && isSelfFocused) {
+        emit('didBlur', childPayload);
+        isSelfFocused = false;
+      }
+    });
+  });
+
+  return (eventName, eventHandler) => {
```
brentvatne: this is where we would handle lazily adding upstream subscribers, and we'd augment the `remove` function to cleanup

src/navigators/StackNavigator.js
```diff
@@ -64,7 +65,11 @@ export default (
       cardStyle={cardStyle}
       transitionConfig={transitionConfig}
       onTransitionStart={onTransitionStart}
-      onTransitionEnd={onTransitionEnd}
+      onTransitionEnd={(lastTransition, transition) => {
+        const { state, dispatch } = props.navigation;
```
brentvatne: state is unused

src/routers/StackRouter.js
```diff
@@ -285,22 +300,20 @@ export default (
 
       if (action.type === NavigationActions.BACK) {
         const key = action.key;
-        let backRouteIndex = null;
+        let backRouteIndex = state.index;
```
brentvatne: what is the purpose of this change?

src/routers/TabRouter.js
```diff
@@ -76,6 +76,7 @@ export default (
         state = {
           routes,
           index: initialRouteIndex,
+          isTransitioning: false,
```
brentvatne: probably later we should integrate this properly in tab-view to handle animating tabs

src/views/CardStack/CardStack.js
```diff
@@ -144,6 +145,10 @@ class CardStack extends React.Component<Props> {
       > = addNavigationHelpers({
         dispatch: navigation.dispatch,
         state: scene.route,
+        addListener: getChildEventSubscriber(
```
brentvatne: looking forward to moving this all out of view code in near future

src/routers/TabRouter.js
ericvicenti: yes, absolutely!

src/routers/StackRouter.js
ericvicenti: explained in person

src/navigators/StackNavigator.js
ericvicenti: willrm

src/getChildEventSubscriber.js
ericvicenti: yeah, exactly.

Fixed in https://github.com/react-navigation/react-navigation/issues/3354

src/getChildEventSubscriber.js
ericvicenti: yeah, will fix. even though cleanup is not used lol

src/TypeDefinition.js
ericvicenti: no thanks!

benevbright:
Pretty excited to see how it goes!

yuyao110120:
About willBlur event, did you consider the return of the gesture?  I need dismiss keyboard, if it showed.

ericvicenti:
Hi @yuyao110120, I'm not sure exactly about that use case. Maybe you can play with the version in master, and file an issue if something doesn't work? Please include an example of the code!

timurridjanovic:
This is currently in 1.0.0-beta.27, but when I install that version with yarn I still don't have addListener on the navigation prop.

ericvicenti:
This is not in 1.0.0-beta.27 because it just landed 2 days ago. Can you try installing from master instead?

timurridjanovic:
Yeah I can do that temporarily. Is there somewhere I can check to know when it will be released to update the version once that happens?

ericvicenti:
We will definitely publish a new version within the next week or two. You can keep an eye on our commit log for the new version, or periodically check the npm page to see the bump

noxee:
Is this supposed to work with all Navigator types? I've installed from master and I've tried to subscribe to the `didFocus`  event in a `TabNavigator` but it doesn't look like it's being triggered.

This is the code I'm using:
```
import React from 'react';
import { TabNavigator } from 'react-navigation';
import { StyleSheet, Text, View } from 'react-native';

class Home extends React.Component {
  state = {
    focused: false
  };

  componentDidMount() {
    this._sub = this.props.navigation.addListener('didFocus', this.setFocus);
  }

  componentWillUnmount() {
    this._sub.remove();
  }

  setFocus = () => {
    console.log('home focused');
    this.setState({ isFocused: true });
  };

  render() {
    return (
        <Text style={{ color: this.state.isFocused ? 'red' : 'blue' }}>
          Home
        </Text>
      </View>
    );
  }
}

class Feed extends React.Component {
  state = {
    focused: false
  };

  componentDidMount() {
    this._sub = this.props.navigation.addListener('didFocus', this.setFocus);
  }

  componentWillUnmount() {
    this._sub.remove();
  }

  setFocus = () => {
    console.log('feed focused');
    this.setState({ isFocused: true });
  };

  render() {
    return (
        <Text style={{ color: this.state.isFocused ? 'red' : 'blue' }}>
          Feed
        </Text>
      </View>
    );
  }
}

export const Tabs = TabNavigator({
  Home: {
    screen: Home
  },
  Feed: {
    screen: Feed
  }
});

export default class App extends React.Component {
  render() {
    return <Tabs />;
  }
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center'
  }
});
```

timurridjanovic:
to install from master all I need to do is yarn add https://github.com/react-navigation/react-navigation.git ? 

I'm getting an error saying addListener is not a function. Is there a build step or something I'm missing?

noxee:
@timurridjanovic you may need to remove your `node_modules` and then run `yarn install`

timurridjanovic:
I tried it, still getting the same error.
It comes from the src/views/CardStack/CardStack.js file.
navigation.addListener is undefined:
`  _getScreenDetails = scene => {
    const { screenProps, navigation, router } = this.props;
    let screenDetails = this._screenDetails[scene.key];
    if (!screenDetails || screenDetails.state !== scene.route) {
      //navigation.addListener is undefined
      const screenNavigation = addNavigationHelpers({
        dispatch: navigation.dispatch,
        state: scene.route,
        addListener: getChildEventSubscriber(
          navigation.addListener,
          scene.route.key
        ),
      });
      screenDetails = {
        state: scene.route,
        navigation: screenNavigation,
        options: router.getScreenOptions(screenNavigation, screenProps),
      };
      this._screenDetails[scene.key] = screenDetails;
    }
    return screenDetails;
  };
`

timurridjanovic:
From what I can tell, addListener is supposed to be set in src/createNavigationContainer, but this._isStateful() returns false in my case and it never gets set.

timurridjanovic:
I think it's because I'm using react navigation with redux and setting my own navigation, which is why this._isStateful() returns false:       
```
<AppNavigator
          navigation={addNavigationHelpers({ dispatch, state: nav })} />
```
Is there a way I could set addListener as well?

ericvicenti:
@noxee, I'm fixing that issue now 😄 

@timurridjanovic, I'm guessing you have a redux integration. This means that you'll need to provide an event subscriber yourself as a part of the top-level navigation prop. You can see how we set that up here: https://github.com/react-navigation/react-navigation/blob/master/src/createNavigationContainer.js#L189

We will actually be switching to "fbemitter" here soon, which will reduce the amount of boilerplate. (cc @brentvatne) Of course I am also planning to fix the documentation for everybody else using ReactNav with Redux.

sandropoluan:
Great!

But one thing,
The 'didBlur' event not fire when first time I click another screen. It will fire on second attempt.

towfiqi:
Just tested. Not working for me, My understanding is this code will only run when the specific tab is focused/tapped.  But its running on app start, without tapping that specific tab:

```
class Map extends Component {
  componentDidMount(){
        this._sub = this.props.navigation.addListener( 'didFocus',
            console.log('This Should Only Show up when Map tab is  is focused!')
          );
  }
```

I can see a `addListener` method in the `this.props.navigation`, so I am pretty sure I am using the latest version.

qftgtr:
It seems the new added action `COMPLETE_TRANSITION` is not triggered when using a gesture to go back?

ask007learning:
Events doesn't get fired. I am using redux in my app. Is there a working example with redux ? I could't find it in docs.

ericvicenti:
@towfiqi, can you provide more information on how you reached that `TypeError`? Does it happen in the navigation playground? Or maybe you could provide sample code?

@qftgtr You're right, there is a bug when swiping back! I have a fix for that coming in #3420

@ask007learning, have you seen the new docs on redux integration? There is a new middleware available, which will allow events to work in redux apps

ask007learning:
It is working now, my redux configurations were off. Followed the new redux integration instructions at https://reactnavigation.org/docs/redux-integration.html

qftgtr:
@ericvicenti I see you guys have released a new update for the fix. Thanks a lot!

qftgtr:
@ericvicenti I tried the update (beta.29) but it seems the problem is still there. I am using StackNavigator in my app and what I tested is push two screens in and swipe one back. The `COMPLETE_TRANSITION` action is still not triggered, and the back button on the header doesn't respond.

patlux:
> Just tested. Not working for me, My understanding is this code will only run when the specific tab is focused/tapped. But its running on app start, without tapping that specific tab:

I have the same behavior. Is that normal or a bug?

brentvatne:
@patlux - you need to follow the updated https://reactnavigation.org/docs/redux-integration.html

ericvicenti:
@qftgtr. The complete transition action never happens when you swipe back, because technically the transition happens immediately when the gesture finishes. I don't know why the back button would become unresponsive. Can you post a new issue for that, and include sample code, or explain how it can be replicated in the navigation playground example?

DeanmvSG:
Is there a demo of this working within a mixed Stack/Tab Navigator setup?  My setup is as follows:

StackNavigator (Login or Logged In) -> Tab Navigator (DashboardStack, SettingsStack etc) -> Each tab then has its own Stack Navigator (Dashboard Page).

I've implemented the above setup for `willFocus` and `willBlur` onto the Dashboard page and it works for when moving through the DashboardStack, but not when moving from one Tab to another via the Tab Navigator.

a-sane:
@DeanmvSG  I had the same problem. In my case, this has happened because I'm using redux-immutable. So if you also using redux-immutable you should use `state.get('nav')` instead of `state.nav` as it written in https://reactnavigation.org/docs/redux-integration.html manual

DeanmvSG:
@a-sane I've not got Redux in this project (haven't wrapped my head around it yet! 😁), I imagine it is a setup thing rather than the library itself...

timurridjanovic:
Works like a charm! Happy days! Only issue I'm having is didFocus doesn't get triggered on initial page load, is that normal?

brentvatne:
@timurridjanovic - that should be fixed in the latest version

slorber:
I created an HOC to provide an "isFocused" prop to screens, using the new event system and it seems to work fine so I made a PR: https://github.com/react-navigation/react-navigation/pull/3512

Until merged for those needing this feature here is a simple version that could work for you:

```javascript
const provideIsFocused = Comp => {
  class IsFocusedProvider extends React.Component {
    state = {
      isFocused: false,
    };

    componentDidMount() {
      this.subs = [
        this.props.navigation.addListener("didFocus", () => this.setState({isFocused: true})),
        this.props.navigation.addListener("willBlur", () => this.setState({isFocused: false})),
      ];
    }

    componentWillUnmount() {
      this.subs.forEach(sub => sub.remove());
    }

    render() {
      return (
        <Comp isFocused={this.state.isFocused} {...this.props}/>
      )
    }
  }

  return IsFocusedProvider;
};
```

bradywatkinson:
I believe I am having a problem related to this change:

I am using redux and have followed the new redux-integration guide.

When I navigate to a screen, the complete_transition is firing (and updating the state) correctly, however, it is somehow causing the complete_transition action to fire again (and thus creating an infinite loop)

ericvicenti:
Hi Brady! Sorry to hear about that. Can you file an issue with detailed information about your navigation setup, and if possible, some example code that we can use to reproduce the issue? I’m particularly interested in the log of events that were dispatched, and the before/after navigation state. Thanks!

bradywatkinson:
Just reading back through this thread:

@towfiqi: you have to use a callback.
```
class Map extends Component {
  componentDidMount(){
        this._sub = this.props.navigation.addListener( 'didFocus', () => (
            console.log('This Should Only Show up when Map tab is  is focused!')
          ));
  }
```

wachunei:
I was having the same issue and my mistake was that I was passing the middleware creator a wrong piece of the state, my navigation reducer is at `state.navigation` and I was passing `state.nav` as it was shown on the guide.

```js 
const middleware = createReactNavigationReduxMiddleware('root', state => state.navigation);
```

swingywc:
Is there any way to addListener to Tab Navigator without using redux? Btw the addListener works like a charm in Stack Navigator 👍  Thank you so much!

gameunite:
Is the fix for the didFocus not being called at start already included here?

pensierinmusica:
@ericvicenti @brentvatne @skevy is this in the docs yet? Couldn't find it, maybe I'm missing something.. Thx!

endoooo:
@ericvicenti this feature is so cool (and much needed)! 😄 

I'd like to use it to trigger some custom entry and exit animations that will be controled by the screen component, instead of configuring custom navigators with Transitioners. but in order to replace the Transitioner completely, I need a way to interpolate some styles with the position/progress available in the transitionProps and prevTransitionProps.

the event handler payload could pass a reference to those animated values, couldn't? what do you think?

pensierinmusica:
Hi! Does anybody know how to fix this? https://github.com/react-navigation/react-navigation/issues/3893

Events don't seem to work when using `getStateForAction()` in connection with Redux.

Thx!

@ericvicenti

Palisand:
@ericvicenti Why wasn't the navigation state augmented with `isTransitioning`?

Palisand:
Also, here's @slorber's HOC with flow types:
```javascript
// @flow

import React, {Component} from 'react';
import type {ComponentType} from 'react';
import type {NavigationScreenProp, NavigationEventSubscription} from 'react-navigation';

export default function withFocusState<Props: {navigation: NavigationScreenProp<*>}>(
  Screen: ComponentType<Props>,
): ComponentType<Props> {
  return class extends Component<Props, {isFocused: boolean}> {

    state = {
      isFocused: false
    };

    subs: Array<NavigationEventSubscription>;

    componentDidMount() {
      const {navigation} = this.props;
      this.subs = [
        navigation.addListener('didFocus', () => {
          this.setState({isFocused: true});
        }),
        navigation.addListener('willBlur', () => {
          this.setState({isFocused: false});
        }),
      ];
    }

    componentWillUnmount() {
      this.subs.forEach(sub => sub.remove());
    }

    render() {
      return <Screen isFocused={this.state.isFocused} {...this.props} />
    }
  }
}
```

slorber:
> @ericvicenti Why wasn't the navigation state augmented with isTransitioning?

I think it's because we don't want by default to trigger unnecessary re-renderings to screens that don't necessarily need to listen to that prop. Most users won't use this, and it could have a perf impact for all users.

Palisand:
Good point.

brentvatne:
@pensierinmusica - if that's still a problem please create an issue on https://github.com/react-navigation/react-navigation-redux-helpers

dantman:
@jsina I'm not sure what your issue is, but if it's something that you could use `componentWillReceiveProps` to implement then it sounds like something you could just move to `componentDidUpdate`.

benevbright:
'willFocus' with Tab has a bug. 
If I force to change current tab by `this.props.navigation.navigate('tab1');` and go back to original tab, 'willFocus' of original screen is not fired.
If I change tab by touching tab buttons, it works well.

brentvatne:
@benevbright - tracking in https://github.com/react-navigation/react-navigation/issues/3511

PR #3350, opened 2018-01-25
Remove flow

brentvatne:
**Test plan (required)**

Run the tests and the playground


brettdh:
@ericvicenti Is there some discussion somewhere that explains how/why this decision was made?

I don't have an opinion about it for this project, but I was planning to use flow for a different project, so I'm wondering what's prompting its removal in react-navigation, and what you'd do differently if you chose to adopt it again in the future.

kelset:
Short version is that keeping up with Flow slowing us down, and flow-types are a better approach.

brettdh:
@kelset thanks for the info - I don't think I understand the distinction you're making, though. I assume there's a separate thing from Flow called flow-types that I'm just unfamiliar with (and couldn't track down with a quick search of either google or this repository).

kelset:
I was referring to [flow-typed](https://github.com/flowtype/flow-typed)

satya164:
@brettdh hey, longer version, we got rid of flow since it required extra work for every React Native release. We are planning to have to have type definitions for the public API which should be sufficient for consumers.

brettdh:
@kelset thanks for clarifying. I'm not sure I see flow-typed as comparable to Flow - rather, it seems complementary. That is, if you're using Flow and you have libraries that aren't, flow-typed is a great way to get the benefits of Flow for your use of that library interface. (I can definitely see the benefit of having a flow-typed definition for react-navigation's API, though.)

@satya164 thanks. Your longer version is still pretty short :) If anyone has the time, I'd love to read an extended blog post about:
- Why you initially decided to adopt Flow
- What benefits it provided
  - Bonus if there were specific bugs discovered/avoided
- What extra work you had to do because of Flow
  - Bonus for specific examples
- How you weighed the tradeoffs and came to the decision to drop Flow
- What future improvements to the Flow/react-native ecosystem could alleviate this
  - For instance, what if react-native itself had a flow-typed interface definition? Then (I think) you'd just have to bump the interface definition version.
- How you'd evaluate such a tool today, knowing what you know now

brentvatne:
not speaking for everybody involved in the decision but i think my time is better spent on improving react-navigation than writing a post like that (which could take days to do it well). additionally, i don't think our experience is much different from other people who have used flow and stopped, so you can probably find other information out there.

i'm a big fan of static types and in particular reasonml but a big factor here is that most people using react native do not use flow, or they use typescript, and this raised the barrier to entry to contributing and understanding the codebase. an key design decision of react navigation is that it should be written in js as much as possible so that most react native developers can understand it, this is an important distinction over using native navigation libraries where you need to understand objc/java in order to debug. so by using flow, often with some pretty advanced features to be able to create the api that we wanted, we made it more difficult for non-flow users to understand. the tradeoff being that by removing flow, it's more difficult for flow users to understand. we hope to mitigate this by providing a type interface file. additionally, we're going to lean more heavily on tests.

here's a slack discussion around the decision, which misses out on some of the irl discussions but still gives a decent overview.

screen shot 2018-01-26 at 10 34 42 am
screen shot 2018-01-26 at 10 34 58 am
screen shot 2018-01-26 at 10 35 11 am
screen shot 2018-01-26 at 10 35 22 am

brettdh:
@brentvatne that's a great amount of context - thanks!

AlicanC:
Why couldn't you just use `any` and `$FlowFixMe` whenever you couldn't or didn't want to type something?

Don't you have to maintain a separate type definition now?

AlicanC:
Also, the "people don't use Flow", "people can't use Flow" arguments could be reasonable if you needed to learn Flow to consume this library, but you don't.

You took an action that discourages skilled developers and encourages less skilled developers. Isn't it obvious that this will result in more code, but less code quality? Do you really want this?

We lowered the skill requirement in our latest interviews and some developers we hired did more bad than good because they were taking too much of our skilled developers' time in reviews.

Developers who can't figure out how to type or fix lint issues are really no good. This doesn't mean they are bad developers, but type systems and linters have great benefit on code quality and developers should spend some time learning them, so should your contributors.

satya164:
@AlicanC libraries with flow aren't the same as apps. `react-navigation` needs to support multiple versions of `react-native` and as a result, multiple versions of flow. flow has breaking changes every release, which make it impossible to support multiple versions unless your codebase is really small.

> Why couldn't you just use any and $FlowFixMe whenever you couldn't or didn't want to type something?

Because when your flow version doesn't match with what `react-navigation` uses, Flow will throw type errors for the library code, and you can't use `any` and `$FlowFixMe` for library code.

> Don't you have to maintain a separate type definition now?

A separate type definition is much easier to maintain since we need to have definitions for only the public API, which is quite small. Also with a separate definition, we can have multiple definitions for multiple flow versions.

I disagree with this sentiment that developers who can't fix type errors aren't good. Having used Flow extensively, the errors messages can be extremely confusing and impossible to know how to fix them.

Anyway, we tried to solve a problem we had, considered the alternatives and arrived at the current solution. If you think you can provide a better solution to the problem, please open an RFC in https://github.com/react-navigation/rfcs

brentvatne:
> Developers who can't figure out how to type or fix lint issues are really no good. This doesn't mean they are bad developers, but type systems and linters have great benefit on code quality and developers should spend some time learning them, so should your contributors.

- @Ashoat is very familiar with Flow and it took him a week to get react-navigation compatible with the latest update for the last react-native release.
- @ericvicenti worked at Facebook for years, where they use Flow all over the place. While pairing with him on some improvements to this library we were able to make the changes we wanted very quickly, and we knew they were correct, but figuring out the types due to several layers of generics was tricky and, as far as we could tell, only slowing us down and making the code more difficult for people to understand.

> Why couldn't you just use any and $FlowFixMe whenever you couldn't or didn't want to type something?

What's the point of having Flow if you use `any` whenever you run into a complex type? @ericvicenti describes `any` as "flow poison" and I tend to agree. By doing this we get most of the disadvantages of using Flow and less of the advantages.

> You took an action that discourages skilled developers and encourages less skilled developers. Isn't it obvious that this will result in more code, but less code quality? Do you really want this?

Flow is different enough from TypeScript that developers with a TypeScript background wanting to contribute would run into obstacles that may prevent contribution.

Everyone knows how to write tests though. And so the decision to remove Flow is accompanied by a requirement that new pull requests include tests (unless it makes no sense, which is sometimes the case) and don't regress on code coverage.

Like @satya164 said, we are open to a RFC with alternative solutions to the problems we mention in this thread. And as I said earlier, we don't know yet if this will end up being the right solution, "we'll continue to evaluate, and if we find that we were getting more value from flow than we realized then adding back flow or typescript to this codebase is still on the table."

Issue #1790, opened 2017-06-07
TabRouter has no initialRouteParams option.


simonemarinelli:
It would be nice to have the initialRouteParams like in StackRouter.

kelset:
Closing since the `initialRouteName` param exists now.

alexisbronchart:
@kelset `initialRouteName` exists but not `initialRouteParams`. Could you reopen?

kelset:
Ah gotcha. Sure.

MrHubble:
@kelset I believe this can be closed again please as it was added in [1.0.0-beta.28](https://github.com/react-navigation/react-navigation/releases/tag/v1.0.0-beta.28). Thanks.

PR #3352, opened 2018-01-25
allow initialRouteParams for TabRouters

superburrito:
Allow the passing in of `initialRouteParams` into `TabRouter`'s config since `StackRouter` already has that feature. The request was filed [here](https://github.com/react-navigation/react-navigation/issues/1790).

First PR, please let me know if I've missed something!

### DIFF
--- a/src/routers/TabRouter.js
+++ b/src/routers/TabRouter.js
@@ -12,6 +12,7 @@ export default (routeConfigs, config = {}) => {
 
   const order = config.order || Object.keys(routeConfigs);
   const paths = config.paths || {};
+  const initialRouteParams = config.initialRouteParams;
   const initialRouteName = config.initialRouteName || order[0];
   const initialRouteIndex = order.indexOf(initialRouteName);
   const backBehavior = config.backBehavior || 'initialRoute';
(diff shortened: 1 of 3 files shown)


brentvatne:
looks good to me, can you rebase?

superburrito:
@brentvatne Sure 🙂

src/routers/TabRouter.js
```diff
@@ -38,18 +39,21 @@ export default (routeConfigs, config = {}) => {
       let state = inputState;
       if (!state) {
         const routes = order.map(routeName => {
+          const includeInitialParams = routeName === initialRouteName;
```
satya164: Will be cleaner to do `const params = routeName === initialRouteName ? initialRouteParams : undefined;` here and then just use `params` below.

src/routers/TabRouter.js
```diff
@@ -69,6 +73,9 @@ export default (routeConfigs, config = {}) => {
             params: {
               ...route.params,
               ...params,
+              ...(route.routeName === initialRouteName
+                ? initialRouteParams
+                : {}),
```
satya164: Can use `null` here instead of empty object

superburrito: Didn't realise this, thanks!

src/routers/TabRouter.js
superburrito: Was unsure about letting the `params` key exist as `params: undefined`... will make the changes.

PR #3363, opened 2018-01-25
Bring back TypeDefinition.js and re-add Flow to NavigationPlayground

Ashoat:
We should keep TypeDefinition around so Flow users can import the types, and we should keep Flow in NavigationPlayground so we can flag when PRs break types.


examples/NavigationPlayground/package.json
```diff
@@ -12,13 +12,13 @@
   },
   "dependencies": {
     "expo": "^24.0.2",
-    "react": "16.0.0",
-    "react-native": "^0.51.0",
+    "react": "16.2.0",
+    "react-native": "^0.52.0",
```
ericvicenti: can you either undo this change, or make the same version change for the redux example and the main library `package.json`? I think we should keep all the versions in sync

PR #3378, opened 2018-01-29
Allow swipeEnabled be a function

maxusr:
Allow swipeEnabled to be a function with state param.


ericvicenti:
This looks good. Can you use prettier to format the code, then we can merge?

PR #3381, opened 2018-01-29
Fix incorrect easing functions for CardStack

Andreyco:
**Motivation**
Fix unintentional error

**Solution**
1. `Animated.timing` function accepts `TimingAnimationConfig` which expect `easing` (or dont expect at all) function with following signature: `(value: number) => number`
[source code](https://github.com/facebook/react-native/blob/8547b7e11163d545b7b99d4bdd063ef71129d62c/Libraries/Animated/src/Easing.js#L82)
2. under the hood, `TimingAnimation` falls back to `easeInOut` function, if `easing` is not specified
`this._easing = config.easing !== undefined ? config.easing : easeInOut();`
[source code](https://github.com/facebook/react-native/blob/8547b7e11163d545b7b99d4bdd063ef71129d62c/Libraries/Animated/src/animations/TimingAnimation.js#L61)
3. by mistake passing `Easing.linear()` results in following
`typeof Easing.linear(); // undefined` which makes statement in step 2 fall back to `easeInOut`

This commit makes passing `easeInOut` function explicit and fixes existing issue and several Flow warnings

**Test plan (required)**
run `flow` command to check for potential warnings


src/views/CardStack/CardStack.js
```diff
@@ -38,6 +38,14 @@ import TransitionConfigs from './TransitionConfigs';
 
 const emptyFunction = () => {};
```
ericvicenti: why not simply:

```js
const EaseInOut = Easing.inOut(Easing.ease);
...
easing: EaseInOut,
```

There is no need to over-optimize this

Andreyco: Sure, just did what `Animated` does under the hood (see source in the links above).
I am happy to change that, however.

Also, I am thinking about another change to types, will post the idea when I get to desktop but it probably deserves another PR

brentvatne:
looks good to me if we integrate the change suggested by @ericvicenti above!

Andreyco:
Guys, done... I see there are conflicts with recent removal of Flow types...

PR #3384, opened 2018-01-30
Fix NavigationEvent types

Ashoat:
dca37627a243b191fda5856aa6ce9eb1ee4957cf broke the types, but it wasn't noticed because Flow wasn't in the CI. This fixes the types; separate PR coming to add Flow to the CI.

### DIFF
--- a/src/TypeDefinition.js
+++ b/src/TypeDefinition.js
@@ -430,7 +430,7 @@ export type NavigationEventPayload = {
 
 export type NavigationEventCallback = (payload: NavigationEventPayload) => void;
 
-export type NavigationEventListener = {
+export type NavigationEventSubscription = {
   remove: () => void,
 };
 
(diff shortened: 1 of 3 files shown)


Ashoat:
Just pushed another commit that refines the callback type to match what @ericvicenti gave me the other day

src/TypeDefinition.js
```diff
@@ -414,7 +414,23 @@ export type NavigationProp<S> = {
...
+export type NavigationEventPayload = {
+  type: EventType,
+  action: NavigationAction,
+  state: NavigationState,
+  lastState: NavigationState,
+};
+
+export type NavigationEventCallback = (payload: NavigationEventPayload) => void;
+
+export type NavigationEventListener = {
```
ericvicenti: Rather than a listener, I think this is usually referred to as the 'subscription'

Other than that, I think this is ready to merge!

PR #3385, opened 2018-01-30
Add NavigationPlayground Flow test to CI

Ashoat:
In order to make sure that contributors don't making breaking changes to our Flow types without corresponding updates to the libdefs, we should make sure these breaking changes get caught by CI. Right now our libdefs are any-typed, but I'll put up a PR later that will refine the libdef (located at `flow-typed/npm/react-navigation_vx.x.x.js`).

### DIFF
--- a/examples/NavigationPlayground/package.json
+++ b/examples/NavigationPlayground/package.json
@@ -8,7 +8,7 @@
     "eject": "react-native-scripts eject",
     "android": "react-native-scripts android",
     "ios": "react-native-scripts ios",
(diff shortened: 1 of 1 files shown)


Ashoat:
(This is failing CI because `examples/NavigationPlayground/js/SimpleTabs.js` is mistyped, which is fixed by #3384)

ericvicenti:
Woohoo! I'm excited because our new approach focuses on flow-as-a-feature, providing direct value for users rather than the implmentors!

PR #3388, opened 2018-01-30
fix statusHeight on non-iphoneX devices in landScape

uragecz:
Problem in landScape on non-iPhoneX devices. There was no statusBar height. In landScape mode it returned 0.

screen shot 2018-01-30 at 11 42 43

### DIFF
--- a/src/views/SafeAreaView.js
+++ b/src/views/SafeAreaView.js
@@ -76,11 +76,7 @@ const statusBarHeight = isLandscape => {
     return isLandscape ? 0 : 44;
   }
 
-  if (isIPad) {
-    return 20;
-  }
-
-  return isLandscape ? 0 : 20;
+  return 20;
 };
 
 const doubleFromPercentString = percent => {


src/views/SafeAreaView.js
```diff
@@ -76,11 +76,7 @@ const statusBarHeight = isLandscape => {
     return isLandscape ? 0 : 44;
   }
 
-  if (isIPad) {
```
ericvicenti: Why did you delete this? I think iPad landscape mode still has 20px status bar

uragecz: Because its unnecessary part of code... now we always return 20px..

uragecz: And the test is unnecessary too... all non-iPhone X devices have status bar, only iPhone X in landscape mode doesn't... so it works

ericvicenti: Oops, I misread that. So, its always 20px? What about smaller phone sizes or older versions of iOS? If you think this is safe then lets merge!

uragecz: Ehm, i just found it was some bug on my simulator... because the status bar should hide on landscape mode.. so your code is correct, i apologize

PR #3390, opened 2018-01-30
Encapsulate DrawerSidebar in the SafeAreaView.

jainkuniya:
So that proper padding styles can be applied to the drawer content.

**motivation** 
Without using `SafeAreaView` component in `DrawerView`, proper padding styles are not applied. Thus drawer content was overlapping status bar on iPhone X.

From [this commit](https://github.com/jainkuniya/react-navigation/commit/f2f4e5ce44c76b8273ad6082e8e5866fc8ebe7dd#diff-1ef5b8f9d10c4acbaf92697f3abd0041) I was expecting `SafeAreaView` component in the `DrawerSidebar`, but can't find it. So submitting this PR. 

**Test plan (required)**
Test on iPhone X

### DIFF
--- a/src/navigators/__tests__/__snapshots__/DrawerNavigator-test.js.snap
+++ b/src/navigators/__tests__/__snapshots__/DrawerNavigator-test.js.snap
@@ -82,158 +82,172 @@ exports[`DrawerNavigator renders successfully 1`] = `
     }
   >
     <View
+      collapsable={undefined}
+      onLayout={[Function]}
       style={
-        Array [
-          Object {
-            "flex": 1,
-          },
-          undefined,
-        ]
+        Object {
+          "flex": 1,
+          "paddingBottom": 0,
+          "paddingLeft": 0,
+          "paddingRight": 0,
+          "paddingTop": 20,
+        }
       }
     >
-      <RCTScrollView
-        DEPRECATED_sendUpdatedChildFrames={false}
-        alwaysBounceHorizontal={undefined}
-        alwaysBounceVertical={false}
-        onContentSizeChange={null}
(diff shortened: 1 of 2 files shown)


jainkuniya:
Hi @brentvatne, @ericvicenti
Thoughts on this :)

brentvatne:
@jainkuniya - as you pointed out in that other commit, it's already used: https://github.com/jainkuniya/react-navigation/commit/f2f4e5ce44c76b8273ad6082e8e5866fc8ebe7dd#diff-f7142895fdff1c837104ff540353add0R31

I don't want to add it outside of here because then when you render a custom content view, you don't have full control over it -- the `SafeAreaView` is forced on you. Are you using a custom content view? If so, just add a `SafeAreaView` around it in your code

jainkuniya:
Yes, we are using custom content view. 
Actually we have tabs in the drawer, which touches bottom but don't touch top. So bottom safe area are automatically handled via `SafeAreaView` wrapper in `TabView`. I was like to handle top safeArea as a wrapper in the Drawer component in `react-navigation` itself.
Else we need to handle only top safeArea around our Drawer component. Which I thinks will not be more generic. :)
can we give an optional prop in the Drawer component, if user wants to handle it manually? Which can be further extended to `TabView` too, if users wants to handle `SafeArea` manually in the tabView too. :)

brentvatne:
@jainkuniya - I'd prefer to leave it up to the user to choose to use `SafeAreaView` or not in that component itself, but we're open to a RFC for it! https://github.com/react-navigation/rfcs

until then, I'll close this issue, but I do really appreciate you taking the time to open it and try to improve the library for other users! thanks!

PR #3391, opened 2018-01-30
Add Flow libdef

Ashoat:
This PR introduces a Flow library definition for this project.

Currently it's in `examples/NavigationPlayground/flow-typed/react-navigation.js`, but I wonder if that location is too easy to miss for contributors, who will have to keep the libdef updated if they make any changes to public interfaces. Anybody have a better idea of where to place it?


examples/NavigationPlayground/flow-typed/react-navigation.js
```diff
@@ -0,0 +1,1007 @@
+// @flow
+
+import * as React from 'react';
+import { Animated, Image } from 'react-native';
+import type {
```
Ashoat: @calebmer, you mentioned on Slack that it shouldn't be possible to import types from package files, and should only be possible to import from other libdefs. However, both this import and the `Layout` import below seem to work. Is this something that will break once I try it on `flow-typed`?

If so, what is the best practice here? The only thing I can think of is to copy-paste the types in, but that seems icky, especially for `StyleObj` which seems to change fairly often. Related: https://github.com/flowtype/flow-typed/issues/631

ericvicenti:
This looks great but I definitely agree it shouldn't be in examples, that would be confusing. Feel free to create some new top level thing to put typey things like this

Ashoat:
I moved the libdef to a new `flow` folder at the root.

Please hold off on merging - I want to make sure that the type imports I'm using work in the `flow-typed` environment first.

Ashoat:
Here is the corresponding `flow-typed` PR: https://github.com/flowtype/flow-typed/pull/1794

The import seem to work there as well, but let's wait and see what the maintainers say before merging this.

Ashoat:
Okay, I've removed all import statements. Let's wait on `flow-typed` PR to land first, but this should be ready to go.

Ashoat:
`flow-typed` PR has landed! Ready for merge

Andreyco:
Guys, libdef is not pulled by `flow-typed`. Looks like versioning does not work with `-beta.x` releases.

Ashoat:
@Andreyco I think that's because the current latest release still has Flow types in the package. Try install master and then running `flow-typed`. Alternately, you can directly install the libdef with `flow-typed install react-navigation@1.0.0`.

Andreyco:
@Ashoat wow, haven't knew of this feature! `flow-typed install react-navigation@1.0.0` works good.

arnoldbird:
This install ran OK but did not resolve my problems:

`Alternately, you can directly install the libdef with **flow-typed install react-navigation@1.0.0**.`

I came to this issue from https://github.com/flowtype/flow-typed/issues/631
But I also have [this other issue](https://github.com/react-navigation/react-navigation/issues/3640), which may be related.

PR #3392, opened 2018-01-30
[BREAKING] New createNavigator API

ericvicenti:
Implementing the new navigator API, as proposed here: https://github.com/react-navigation/rfcs/pull/7/files

Overall this is a huge improvement to the structure of the codebase and the re-usability of the navigation views.

This has exposed a race condition after navigation on stack, where events are currently broken:
- Navigate twice in a stack from A->B->C:
    - The willFocus for C is skipped
    - The didBlur of A happens after the C transition

Also, this breaks the flow types, which need an overhaul. It should be easier to fix once we land the actions cleanup: https://github.com/react-navigation/rfcs/pull/18


examples/NavigationPlayground/js/CustomTabs.js
```diff
@@ -66,19 +65,14 @@ const CustomTabBar = ({ navigation }) => {
   );
 };
 
-const CustomTabView = ({ router, navigation }) => {
+const CustomTabView = ({ sceneDescriptors, navigation }) => {
   const { routes, index } = navigation.state;
-  const ActiveScreen = router.getComponentForRouteName(routes[index].routeName);
+  const descriptor = sceneDescriptors.find(d => d.key === routes[index].key);
```
brentvatne: I guess we need to update this to an object based on rfc

src/react-navigation.web.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
 
   // Navigators
   get createNavigator() {
-    return require('./navigators/createNavigator').default;
+    return require('./navigators/createNavigatorDEPRECATED').default;
```
brentvatne: is this intentional?

src/navigators/createNavigator.js
```diff
@@ -1,19 +1,49 @@
...
+          state: route,
+          navigation: childNavigation,
+        };
+      });
+
+      return (
+        <NavigatorView
+          screenProps={screenProps}
+          navigation={navigation}
+          navigationConfig={navigationConfig}
```
satya164: It'll be nice to accept the `navigationConfig` as props to the component too. Which overrides the config defined statically.

src/react-navigation.web.js
brentvatne: (this file doesn't seem to exist)

brentvatne:
@ericvicenti - I think a small rebase is required (re: StackRouter-test) and some snapshot re-recording

src/navigators/createNavigator.js
brentvatne: this seems reasonable to me, but I'm curious what @ericvicenti thinks.

one thing that strikes me as a bit weird about passing props in to navigators is that there is not really a good way to do this for anything but the root navigator. the only thing I can think of is:

```js
@connect(getSomeReduxState)
class MyNavigatorWithProps extends React.Component {
  static router = OriginalNavigator.router;

  render() {
    return (
      <OriginalNavigator
          myProp={this.props.someThingFromReduxState}
       />
    );
  }
}
```

is this what you had in mind @satya164? or do you think this is mainly just useful on the root navigator?

brentvatne:
> Since this is going to be a breaking change, maybe worth rethinking the router to initialize its state when rendered instead of when the navigator is created so that the options for the router can be specified as props rather than needing to specify them statically?

commented above about this already, basically this sounds interesting and at least in the case of setting a different initial route name on the root navigator is useful. I wonder what the tradeoffs are.

I think we can address this in a follow-up PR -- this one has been in the works for a while and ideally we should land it and kick the tires asap. agree that batching them into the same major release may make sense here.

src/react-navigation.web.js
ericvicenti: whoops, will fix. At some point I had just renamed it, but we do want to change the external API here.

src/navigators/createNavigator.js
ericvicenti: I can't think of immediate use cases for it, but I agree. It makes a lot of sense to let props override the navigationConfig here.

ericvicenti:
> Since this is going to be a breaking change, maybe worth rethinking the router to initialize its state when rendered instead of when the navigator is created so that the options for the router can be specified as props rather than needing to specify them statically?
>
> It would make it possible to specify initial route and params when rendering the component instead of statically.

I'd like to hear more about the motivation behind this. Can you write up an RFC, @satya164? I agree it would make sense to bundle in the same major version

src/navigators/createNavigator.js
brentvatne: one concern I have is that we need to be sure that the props actually *can* be changed dynamically. things like `initialRouteName` aren't a big deal because you never expect them to change after render... but it's not clear that it is sane for us to support changing most navigator config props dynamically.

src/react-navigation.web.js
brentvatne: ping @ericvicenti  ;)

src/views/TabView/TabView.js
```diff
@@ -22,12 +21,9 @@ class TabView extends React.PureComponent {
...
   _renderScene = ({ route }) => {
-    const { screenProps } = this.props;
-    const childNavigation = this.props.childNavigationProps[route.key];
-    const TabComponent = this.props.router.getComponentForRouteName(
-      route.routeName
-    );
-
+    const { screenProps, descriptors } = this.props;
+    const descriptor = descriptors[route.key];
+    const TabComponent = descriptor.getComponent();
```
brentvatne: the props passed in to `ResourceSavingSceneView` need to come from the navigationConfig config &mdash; `lazy`, `removeClippedSubviews`, `animationEnabled`, `swipeEnabled` etc aren't passed in directly as props to `TabView` anymore, they are provided via `props. navigationConfig`.

you can verify that this isn't working as expected by going to the stacks in tabs example on iOS and noticing that you can swipe back and forth between the tabs

src/views/Drawer/DrawerView.js
```diff
@@ -49,111 +47,127 @@ export default class DrawerView extends React.PureComponent {
...
+      );
+      tabDescriptors[route.key] = {
+        key: route.key,
+        getComponent,
+        options,
+        state: route,
+        navigation: childNavigation,
+      };
+    });
+
```
brentvatne: somewhere here the `navigationConfig` for drawer icons isn't being threaded through correctly, I believe. you can repro by going to the drawer example in playground and seeing that there are no icons, but in the one published from master there are

ericvicenti:
Thanks for finding those issues! Wiilfix soon

aksonov:
Guys, I'm very excited to see this PR was implemented and merged. Now I'm trying to figure out how to change RNRF to work with it :)

wluxion:
> This has exposed a race condition after navigation on stack, where events are currently broken:

Not seeing a separate issue for this -- is it being tracked/worked on? Still broken in rc2.

brentvatne:
cc @ericvicenti ^

PR #3393, opened 2018-01-31
Idempotent navigate with key for StackRouter

ericvicenti:
The navigation action will now always behave idempotently when called with a key. When pushing, it will push a route with that key. If the route already exists, it will navigate to that route and set the params.

This is a commonly requested feature, and will address the needs of #2334 #135 #1313, #2400, #2578

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -37,6 +37,9 @@ const navigate = createAction(NAVIGATE, payload => {
   if (payload.action) {
     action.action = payload.action;
   }
+  if (payload.key) {
+    action.key = payload.key;
+  }
   return action;
 });
 
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -19,10 +19,26 @@ export default function(navigation) {
         NavigationActions.back({ key: actualizedKey })
       );
     },
-    navigate: (routeName, params, action) =>
-      navigation.dispatch(
-        NavigationActions.navigate({ routeName, params, action })
-      ),
+    navigate: (navigateTo, params, action) => {
+      if (typeof navigateTo === 'string') {
+        return navigation.dispatch(
+          NavigationActions.navigate({ routeName: navigateTo, params, action })
+        );
+      }
+      invariant(
+        typeof navigateTo === 'object',
+        'Must navigateTo an object or a string'
+      );
+      invariant(
+        params == null,
+        'Params must not be provided to .navigate() when specifying an object'
+      );
+      invariant(
+        action == null,
+        'Child action must not be provided to .navigate() when specifying an object'
+      );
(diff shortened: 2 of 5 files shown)


dann1609:
@ericvicenti please include same solution to  NavigationActions. Other way redux user can get your solution.

brentvatne:
look at the diff @dann1609, you can use this with redux, just include the key in the action...

dann1609:
I did and didn't work.

ericvicenti:
@dann1609, can you please file a detailed issue with the exact example code that demonstrates the problem? Otherwise we can’t help you.

dann1609:
I do my navigation using the next code:

`dispatch(NavigationActions.navigate(
                {
                    routeName: route,
                    params: params,
key:key
                }
            ))`

App navigate without problem, but next page (in stack navigator) set a random key, not the key provided. So, the next time I call the same dispatch. Navigation will create the same page since key is not the same.

lucasleelz:
Using key may lead to the following situation:
In this environment the sources for assign MUST be an object. This error is a performance optimization and not spec compliant.

marsonmao:
I tried supplying the key of the current screen to navigate, it works and prevents multiple stacking with multiple touching, but it only works once. If I want to navigate one more level deeper then it won't do it. 

I think the reason is that once I supply the current key, the router propagates/memorizes it so that I cant navigate twice since the key remains the same at the second navigate.

To be more specific the navigation structure looks like this:
```
[screen1]: { Screen1 }
[screen2]: { Screen2 }
[screen3]: { Screen3 }

```
in Screen1:
`dispatch(NavigationActions.navigate({ routeName: 'screen2', key: this.props.navigation.state.key}));` is doing fine. But if I do it again in Screen2:
`dispatch(NavigationActions.navigate({ routeName: 'screen3', key: this.props.navigation.state.key}));` has no effect, since the 2 keys are the same. 

The `this.props.navigation.state.key` are from different components (Screen1 and Screen2) so they should be different, which is the case If I do not supply the key, then each route key ends with the counter++ so that each key is unique.

So the question is, how do I correctly use this new feature? Thanks.

marsonmao:
@ericvicenti Hope you got some time to take a look 😄

ericvicenti:
I can only help if you post a MVCE of your problem: https://stackoverflow.com/help/mcve 

Otherwise it’s not clear what is going on

sonaye:
I've been looking for something like this for ages.

@marsonmao If you are using a StackNavigator, the first route by default will have a random key unless you specify it in the StackNavigator configs `initialRouteKey: 'Home'` (https://github.com/react-navigation/react-navigation/pull/3540).

jkdf2:
Does the documentation include any examples of what makes for a good key?

PR #3397, opened 2018-02-01
Handle the navigate action as first action with nested StackRouter

matthargett:
Fixes issue discovered when implementing deep linking in an application. Handles the navigate action with params and nested StackRouter as the first action. Thanks to my colleague @muramasa for tracking down the root cause, making a fix, and writing a unit test that reproduces this issue.

**Test plan (required)**
Verified internal application bug was fixed with manual test. Added unit test that failed `jest test` without the fix, and verified new test (and all other tests) pass with the fix in place:
```
Test Suites: 16 passed, 16 total
Tests:       124 passed, 124 total
Snapshots:   6 passed, 6 total
Time:        2.882s
Ran all test suites.
``` 

**Code formatting**

```yarn run test``` and pre-commit hooks passed.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -24,6 +24,13 @@ function behavesLikePushAction(action) {
   );
 }
 
+// Gives us a single place to generate a unique
+// key if they didn't provide one
+function addKeyToRoute(key, route) {
+  route.key = key || `Init-${generateKey()}`;
+  return route;
+}
+
 export default (routeConfigs, stackConfig = {}) => {
   // Fail fast on invalid route definitions
   validateRouteConfigMap(routeConfigs);
@@ -108,11 +115,11 @@ export default (routeConfigs, stackConfig = {}) => {
               isTransitioning: false,
               index: 0,
               routes: [
-                {
-                  ...action,
+                addKeyToRoute(action.key, {
+                  routeName: action.routeName,
+                  params: action.params,
                   type: undefined,
-                  key: `Init-${_getUuid()}`,
-                },
+                }),
               ],
             };
           }
@@ -124,12 +131,11 @@ export default (routeConfigs, stackConfig = {}) => {
             isTransitioning: false,
             index: 0,
             routes: [
-              {
+              addKeyToRoute(action.key, {
                 params: action.params,
                 ...childRouter.getStateForAction(childAction),
                 routeName: action.routeName,
-                key: `Init-${_getUuid()}`,
-              },
+              }),
             ],
           };
         }
@@ -149,12 +155,11 @@ export default (routeConfigs, stackConfig = {}) => {
           ...(action.params || {}),
           ...(initialRouteParams || {}),
         };
-        route = {
+        route = addKeyToRoute(action.key, {
           ...route,
           routeName: initialRouteName,
-          key: `Init-${generateKey()}`,
           ...(params ? { params } : {}),
-        };
+        });
         // eslint-disable-next-line no-param-reassign
         state = {
           key: 'StackRouterRoot',
@@ -271,23 +276,21 @@ export default (routeConfigs, stackConfig = {}) => {
             };
           }
         }
-        const key = action.key || generateKey();
+
         if (childRouter) {
           const childAction =
             action.action || NavigationActions.init({ params: action.params });
-          route = {
+          route = addKeyToRoute(action.key, {
             params: action.params,
             // merge the child state in this order to allow params override
             ...childRouter.getStateForAction(childAction),
-            key,
             routeName: action.routeName,
-          };
+          });
         } else {
-          route = {
+          route = addKeyToRoute(action.key, {
             params: action.params,
-            key,
             routeName: action.routeName,
-          };
+          });
         }
         return {
           ...StateUtils.push(state, route),
@@ -331,11 +334,11 @@ export default (routeConfigs, stackConfig = {}) => {
               routeToPush = navigatedChildRoute;
             }
             if (routeToPush) {
-              return StateUtils.push(state, {
+              const route = addKeyToRoute(action.key, {
                 ...routeToPush,
-                key: generateKey(),
                 routeName: childRouterName,
               });
+              return StateUtils.push(state, route);
             }
           }
         }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/routers/StackRouter.js [hunk 1]
```diff
@@ -100,22 +100,41 @@ export default (routeConfigs, stackConfig = {}) => {
...
-          childRouters[action.routeName] !== undefined
+          childRouter !== undefined
         ) {
+          if (childRouter === null) {
+            return {
+              isTransitioning: false,
+              index: 0,
+              routes: [
+                {
+                  ...action,
```
ericvicenti: rather than putting in the whole action and overriding type, can we instead just explicitly use the routeName and params from the action?

Also we should use the key from the action, if it is provided

src/routers/__tests__/StackRouter-test.js
```diff
@@ -856,6 +856,139 @@ describe('StackRouter', () => {
...
+    //       routeName: 'main',
+    //       routes: [
+    //         {
+    //           index: 0,
+    //           key: 'Init-id-0-40',
+    //           params: { code: 'test', foo: 'bar', id: '4' },
+    //           routeName: 'profile',
+    //           routes: [
+    //             {
+    //               key: 'Init-id-0-39',
```
ericvicenti: I fixed this test so that keys are actually predictable, so you can actually use this commented out stuff instead

src/routers/__tests__/StackRouter-test.js
```diff
@@ -856,6 +856,139 @@ describe('StackRouter', () => {
...
+            code: 'test',
+            foo: 'bar',
+          },
+        },
+      },
+    });
+
+    expect(state && state.index).toEqual(0);
+    expect(state && state.isTransitioning).toEqual(false);
+    /* $FlowFixMe */
```
ericvicenti: flow is gone, please rm!

src/routers/StackRouter.js
```diff
@@ -100,22 +100,41 @@ export default (routeConfigs, stackConfig = {}) => {
...
+            action.action || NavigationActions.init({ params: action.params });
           return {
             isTransitioning: false,
             index: 0,
             routes: [
               {
-                ...action,
-                type: undefined,
+                params: action.params,
+                ...childRouter.getStateForAction(childAction),
```
ericvicenti: yup this makes sense

src/routers/StackRouter.js
jayphelps: 👍

src/routers/__tests__/Routers-test.js
```diff
@@ -174,8 +179,8 @@ test('Supports lazily-evaluated getScreen', () => {
     immediate: true,
     routeName: 'Qux',
   });
-  expect(state1.routes[0].key).toEqual('Init-id-0-4');
-  expect(state2.routes[0].key).toEqual('Init-id-0-5');
+  expect(state1.routes[0].key).toEqual('Init-id-0');
+  expect(state2.routes[0].key).toEqual('Init-id-1');
```
jayphelps: I noticed these tests were relying on the global key counter instead of resetting beforeEach with _TESTING_ONLY_normalize_keys

src/routers/__tests__/StackRouter-test.js
jayphelps: Done. Also added #3439 to remove the reset of these in the codebase.

src/routers/StackRouter.js
jayphelps: Based on a conversation @ericvicenti and I had on Slack, it should always be using `action.key` if available, not just in this specific place. PR updated to reflect that.

src/routers/__tests__/StackRouter-test.js
```diff
@@ -1007,6 +1007,126 @@ describe('StackRouter', () => {
...
+    });
+
+    expect(state).toEqual({
+      index: 0,
+      isTransitioning: false,
+      key: 'StackRouterRoot',
+      routes: [
+        {
+          index: 0,
+          isTransitioning: false,
```
jayphelps: The fact that these route objects has `isTransitioning` and `index` seemed weird to me. Seems like it happens because some code paths treat spread the state over the route, like https://github.com/react-navigation/react-navigation/blob/fae1f71fb3f3fcbf7ad8ffdb9e0f5f1a840bf660/src/routers/StackRouter.js#L123

I'm new to the codebase, so still groking things. Lmk if my suspicions are applicable.

src/routers/StackRouter.js
```diff
@@ -24,6 +24,13 @@ function behavesLikePushAction(action) {
   );
 }
 
+// Gives us a single place to generate a unique
+// key if they didn't provide one
+function addKeyToRoute(key, route) {
+  route.key = key || `Init-${generateKey()}`;
+  return route;
+}
+
```
jayphelps: At first I did this in a single place, at the start of `getStateForAction`

```js
  const key = action.key || `Init-${generateKey()}`;`
  // etc..
```

So that I could things like this:

```js
return {
  routeName: action.routeName,
  params: action.params,
  type: undefined,
  key
},
```

The trouble with that approach was that my testing showed that not all codepaths actually need any key at all, and but in those cases the global `generateKey` counter would still be incremented. Perhaps that's OK, but it made a few tests have unexpected key values (e.g. test writer assumes we're at 0, but we're at 1).

`getStateForAction` is rather complex, so I wanted to touch as little as possible to avoid introducing bugs in codepaths not currently tested.

src/routers/StackRouter.js
```diff
@@ -102,23 +109,38 @@ export default (routeConfigs, stackConfig = {}) => {
       // Set up the initial state if needed
       if (!state) {
         let route = {};
-        if (
-          behavesLikePushAction(action) &&
-          childRouters[action.routeName] !== undefined
-        ) {
+        const childRouter = childRouters[action.routeName];
+        if (behavesLikePushAction(action) && childRouter !== undefined) {
```
jayphelps: I had to rebase some changes that caused a conflict here, double check this is still the right behavior plz.

src/routers/__tests__/StackRouter-test.js
ericvicenti: isTransitioning is a feature of the stack router. Each route in the stack may be a navigation state of another router, so we need to treat them as such, when the routeName matches

ericvicenti:
I went and cleaned up this work in https://github.com/react-navigation/react-navigation/pull/3474 . Thanks for identifying and fixing this! Apologies for screwing up attribution

PR #3399, opened 2018-02-01
Redux middleware for new event system

Ashoat:
The benefits of this approach (in contrast with current implementation in `ReduxExample`) are:

1. Less boilerplate for Redux users.
2. It guarantees that every action will trigger the listeners.


src/react-navigation.js
```diff
@@ -88,4 +88,13 @@ module.exports = {
   get withNavigation() {
     return require('./views/withNavigation').default;
   },
+
+  // Redux integration
+  get constructReactNavigationReduxMiddleware() {
```
ericvicenti: naming of "construct" sorta conflicts with the "create" naming used elsewhere in the library

examples/ReduxExample/src/reducers/index.js
```diff
@@ -54,7 +50,6 @@ function auth(state = initialAuthState, action) {
 }
 
 const AppReducer = combineReducers({
-  lastAction,
```
ericvicenti: hah, thanks! I knew this was a bad idea

brentvatne:
@Ashoat - could we move the redux specific functionality into another library on this org? `react-navigation-redux-helpers` or something, which exports `createReactNavigationReduxMiddleware` and `createReduxBoundAddListener`. this ensures that anything that we can do with redux here we could potentially do with another state management library and generally that the interface we're hooking into is extensible

Ashoat:
Okay, I've published `react-navigation-redux-helpers` to npm. This PR now just updates the docs and `ReduxExample`.

PR #3400, opened 2018-02-01
StackRouter Reset Action should match state key

ericvicenti:
There are a ton of issues relating to reset, because there is no way to match the key of the state that you are setting. This change makes that possible.

We should also probably add a way to do this from the navigation helpers so that you can do `props.navigation.reset({key, ...`

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "url": "git@github.com:react-navigation/react-navigation.git",
     "type": "git"
   },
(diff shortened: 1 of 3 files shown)


brentvatne:
as we have discussed before we need to re-think reset, but this is a good move in the right direction for the current implementation / design!

ericvicenti:
Yeah, I think this will help a lot of use cases, without breaking anybody :-D

PR #3401, opened 2018-02-01
New StackRouter Push Action

ericvicenti:
New action behaves like "navigate" on a stack, but does not accept a key because it always pushes


ericvicenti:
Tests arefailing cause i forgot to re-record snapshots. Still ready to be reviewed

PR #3403, opened 2018-02-01
StackRouter PopToTop Action

ericvicenti:
A commonly requested feature for the StackRouter, which often forces people to use the reset action.

This supports an "immediate" param, which may not work until the transitioner handles immediate transitions.

This adds `popToTop` to the navigation helpers, which are technically available even when there is no stack router in use. Eventually I think we should move away from "addNavigationHelpers" to address this problem


brentvatne:
> Eventually I think we should move away from "addNavigationHelpers" to address this problem

👌

PR #3408, opened 2018-02-02
Add left-to-right dismiss gesture to modal

pyankoff:
I needed a way to dismiss StackNavigator modal with a left-to-right gesture. Might not be the best way to implement it, but `gestureDirection: inverted` already creates weird bottom-to-top dismiss gesture on modals, so adding `rotated` option looks net positive.

### DIFF
--- a/docs/api/navigators/StackNavigator.md
+++ b/docs/api/navigators/StackNavigator.md
@@ -160,7 +160,7 @@ Object to override the distance of touch start from the edge of the screen to re
 
 #### `gestureDirection`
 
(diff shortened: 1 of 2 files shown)


ericvicenti:
We are already thinking about ways to make this more customizable, and "direction" isn't really sufficient. So, rather than fixing this in a one-off way, I think we should draft up a RFC for a more holistic approach

brentvatne:
agree with @ericvicenti! fwiw you don't have to use modal stack type for this, just set the headerMode to screen (if you want headers, otherwise none) and it will transition from right to left. modal just changes the transition type.

PR #3411, opened 2018-02-02
Fix failure link and makes well expressed of ChatScreen

SaveYourTime:
1. Updates the link at line 44 correctly.
2. Add quick look for ChatScreen Component, in case of not well expressed.


brentvatne:
hi! sorry for this awkward state the website is in right now -- we are in the middle of transitioning to a new website and this particular doc no longer exists there! https://github.com/react-navigation/website

brentvatne:
also -- thank you so much for submitting this change, i look forward to more in the future! if we hadn't just re-written this section i would have absolutely accepted it

PR #3414, opened 2018-02-02
[BREAKING] Default to bottom tabs

brentvatne:
It's confusing to have such a significant difference between layouts on iOS and Android when using TabNavigator. Further, the top tabs pattern seems to be becoming increasingly less common (anecdotally, from my experience) and more apps are just using bottom tabs. This PR makes the default behavior the same on iOS and Android.

**Test plan (required)**

Run the playground app.

@satya164 pointed out that we should make the default bottom tab bar look more like the material design guidelines on Android, I agree with this and will add follow-up commits for it.

### DIFF
--- a/src/navigators/TabNavigator.js
+++ b/src/navigators/TabNavigator.js
@@ -45,14 +45,14 @@ const TabNavigator = (routeConfigs, config = {}) => {
 };
 
 const Presets = {
-  iOSBottomTabs: {
+  BottomTabs: {
     tabBarComponent: TabBarBottom,
     tabBarPosition: 'bottom',
     swipeEnabled: false,
     animationEnabled: false,
     initialLayout: undefined,
   },
-  AndroidTopTabs: {
+  TopTabs: {
     tabBarComponent: TabBarTop,
     tabBarPosition: 'top',
     swipeEnabled: true,
@@ -80,10 +80,9 @@ const Presets = {
  *```
  */
 TabNavigator.Presets = {
-  iOSBottomTabs: Presets.iOSBottomTabs,
-  AndroidTopTabs: Presets.AndroidTopTabs,
-  Default:
-    Platform.OS === 'ios' ? Presets.iOSBottomTabs : Presets.AndroidTopTabs,
+  BottomTabs: Presets.BottomTabs,
+  TopTabs: Presets.TopTabs,
+  Default: Presets.BottomTabs,
 };
 
 export default TabNavigator;


brentvatne:
TODO:

components-bottomnavigation-spec-fixed1

https://material.io/guidelines/components/bottom-navigation.html#bottom-navigation-specs

ferrannp:
^ we wanted to add such component in Paper https://github.com/callstack/react-native-paper/issues/18 but not there yet. I'd be great if we get it here 😄

Minishlink:
If you make an implementation of the Android material bottom navigation, it would be great that React Navigation renders this BottomNavigationBar component if tabBarPosition is undefined or bottom, and renders the previous implementation, TabBarTop, when tabBarPosition is top :)

brentvatne:
> If you make an implementation of the Android material bottom navigation, it would be great that React Navigation renders this BottomNavigationBar component if tabBarPosition is undefined or bottom, and renders the previous implementation, TabBarTop, when tabBarPosition is top :)

seems reasonable

timomeh:
@brentvatne If https://github.com/timomeh/react-native-material-bottom-navigation could be useful for you, I'd be happy to help.

brentvatne:
that looks fantastic @timomeh! can you open a PR that switches to this by default on android, and also makes it possible to opt-in to it on ios? one tricky thing is to try to share many of the config options between the two

timomeh:
Yeah, sharing the configs will be a bit tricky. I initially developed the BottomNavigation as a standalone component, and later added a Wrapper-Component for react-navigation, which I admittedly just hacked in there.

I'm planning an update of the BottomNavigation in the next few days (also, it's about time that I add some tests). After that, I can take a look at the implementation of the current TabBar and open a WIP PR.

brentvatne:
sounds like a good plan, i'm excited for it!

timomeh:
A short status update: I am currently improving the Bottom Navigation to be more stable, more performant and easier to customize, which in turn will make the react-navigation implementation easier and better to maintain.

Progress can be tracked here: https://github.com/timomeh/react-native-material-bottom-navigation/pull/76

After that I will follow up with a WIP PR.

satya164:
Closing this since we've decided to split tab navigator into different navigators for prevent confusion. Current progress: https://github.com/react-navigation/react-navigation-tabs

timomeh:
@brentvatne @satya164 I'm a bit confused what your plans are for this feature. Should I follow up with a PR in react-navigation-tabs, or is there no interest anymore?

satya164:
@timomeh we include [bottom navigation component from react-native-paper](https://callstack.github.io/react-native-paper/bottom-navigation.html) in [react-navigation-tabs](https://github.com/react-navigation/react-navigation-tabs). you can provide react navigation integration in your repo if you want!

PR #3440, opened 2018-02-06
StackNavigator Replace Action

ericvicenti:
The long awaited action to replace a route in StackNavigator!

Has a nearly identical API to navigate, except the key is mandatory. Currently the behavior will keep the same key intact, and simply allow changing of the routeName, params, and child router action.

Alternatively, we could decide that replacement means the key changes. In that case, the action would need to be able to specify the new key, in addition to the key of the route being replaced. Would this be named "replaceKey"?

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -1,7 +1,6 @@
 // @flow
 
 declare module 'react-navigation' {
-
   /**
    * First, a bunch of things we would love to import but instead must
    * reconstruct (mostly copy-pasted).
@@ -68,6 +67,8 @@ declare module 'react-navigation' {
 
     // The action to run inside the sub-router
     action?: NavigationNavigateAction,
+
+    key?: string,
   |};
 
   declare type DeprecatedNavigationNavigateAction = {|
@@ -130,8 +131,9 @@ declare module 'react-navigation' {
     type: 'Reset',
     index: number,
     key?: ?string,
-    actions:
-      Array<NavigationNavigateAction | DeprecatedNavigationNavigateAction>,
+    actions: Array<
+      NavigationNavigateAction | DeprecatedNavigationNavigateAction
+    >,
   |};
 
   declare export type NavigationUriAction = {|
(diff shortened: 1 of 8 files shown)


ericvicenti:
cc @Ashoat, It seems that the test is failing because of type definition in flow-typed not lining up with this new API. Whats the best way to handle new features or breaking changes like this?

Ashoat:
1. The file ‘flow/react-navigation.js` (the “libdef”) must be updated such that running `flow` in `examples/NavigationPlayground` produces no errors.
2. Once you have the final updated version of the libdef, you should create a PR on `flow-typed` with the new version of the libdef.

flow/react-navigation.js
```diff
@@ -209,9 +239,8 @@ declare module 'react-navigation' {
     params?: NavigationParams,
   };
 
-  declare export type NavigationStateRoute =
-    & NavigationLeafRoute
-    & NavigationState;
+  declare export type NavigationStateRoute = NavigationLeafRoute &
```
Ashoat: I find this reads less clearly than the original format, and it doesn't do as good of a job at preserving `git blame` attribution.

flow/react-navigation.js
```diff
@@ -144,9 +146,37 @@ declare module 'react-navigation' {
     uri: string,
   |};
 
+  declare export type NavigationReplaceAction = {|
+    +type: 'Navigation/REPLACE',
```
Ashoat: As far as I understand, there are no reasons to make any of these properties covariant. Can you remove the `+` signs please?

Ashoat: Here's a good primer on property variance: https://flow.org/blog/2016/10/04/Property-Variance/

ericvicenti: Hmm.. why not? We used this all over the place at FB to make properties read-only

It would be pretty crazy if users were allowed to do `action.type = 'Overridden'`

flow/react-navigation.js
ericvicenti: I don't really control what prettier does here. Maybe we are using different versions or configurations of prettier?

ericvicenti:
Yeah, we run prettier on everything, right?

brentvatne:
@Ashoat - can you write up some steps in contributing.md about how to update flow interface? we should be clear about whether we use prettier on that or not also

Ashoat:
> Hmm.. why not? We used this all over the place at FB to make properties read-only
>
> It would be pretty crazy if users were allowed to do action.type = 'Overridden'

Oh, interesting... I guess it does have that effect. I wonder if there is a difference between `$ReadOnly<>` and marking properties covariant.

Should we mark all of our actions' properties as covariant, instead of just the new ones you added here?

> Yeah, we run prettier on everything, right?

Generally support that idea. Little sad that the Flow syntax isn't better.

I had to add the `flow` directory in `.eslintignore` on the libdef file, so Prettier isn't running on precommit. It might be possible to get eslint working with the libdef syntax... first step would be to bring back the Flow stuff in `.eslintrc`. Not familiar if there's a way to get Prettier to run on precommit without eslint, but @ericvicenti seems to have gotten it to run.

> @Ashoat - can you write up some steps in contributing.md about how to update flow interface? we should be clear about whether we use prettier on that or not also

In terms of the how, [my text above](https://github.com/react-navigation/react-navigation/pull/3440#issuecomment-363413699) covers it. If you need more I might be able to help over the weekend.

lfkwtz:
@ericvicenti Is there a reason that key is required on this action (since it is optional when replace is used on the helper)?

I set it to just pull the last index if key isn't found on a local branch, which works the same way as the helper.

I'll open a PR if you're alright with an optional key on this.

ericvicenti:
The key is required because we wont know which route should be replaced. On the helper its fine because we have access to the route and can pass the key in directly.

lfkwtz:
But this actions just loops through the routes to find the index of the route matching the provided key.

```const routeIndex = state.routes.findIndex(r => r.key === action.key);```

So what'd be the harm in just pulling off the last route in state.routes if the key isn't found?

lfkwtz:
I can write some tests but essentially I was thinking something as simple as this --
https://github.com/lfkwtz/react-navigation/blob/4d86fa73aa4bf8d0f09116eaa9bfeef4508c180f/src/routers/StackRouter.js#L412

Even if we have nested stack navigators -- if we don't pass in a key param, all we need to do is pull the last index off the current navigator. Otherwise, the method continues as it did originally and recursively looks for the key.

PR #3450, opened 2018-02-06
Fix StackRouter Replace Key Behavior

ericvicenti:
Replace should actually provide new keys on the replaced route, use ‘newKey’ on the action if you want to define the new route key


zaiz91:
@ericvicenti Should the newKey attribute not be added to addNavigationHelpers as well?

ericvicenti:
Sure, I think that makes sense. I'll try to do it with the upcoming actions overhaul for v2

PR #3452, opened 2018-02-07
Sync and switch SafeAreaView with standalone

davepack:
React-native-safe-area-view has existed for a while in tandem with the local version in react-navigation. This makes the standalone package the one source of truth for future changes. Changes have been imported to the standalone and released as v0.6.0.

Tests pass and a visual test was done using NavigationPlayground and SafeAreaExample example apps.


Ashoat:
Wondering, would it be worth putting up a PR for the improved `SafeAreaView` on the mainline React Native repo? Probably doesn't make sense if you're still iterating, but figured it would be nice to get it out to everyone using React Native.

davepack:
@Ashoat Totally, that was the original goal. I think the optimal solution for SafeAreaView needs to be native with the `forceInset` option. Unfortunately I'm not in a position to do it anytime soon (and I'd have to learn some objective c first). Fortunately `forceInset` helps handle the jank of a non-native implementation.

PR #3455, opened 2018-02-07
[BREAKING] Decode URI encoded string of deep link

uiureo:
# problem
example:

```js
StackNavigator({
  search: {
    screen: SearchScreen,
    path: 'search/:keyword'
  }
})
```

When you pass encoded URI like `app://search/2018%2F02%2F07` for deep link, it passes raw value `params: { keyword: '2018%2F02%2F07' }`  to `SearchScreen`. 

It's not consistent with the behavior when calling `.navigate('search', { keyword: '2018/02/07' })` from JavaScript. 

Also, it has trouble when you use multi-byte language characters such as Japanese.
(ex:  https://ja.wikipedia.org/wiki/%E3%82%A6%E3%82%A3%E3%82%AD%E3%83%9A%E3%83%87%E3%82%A3%E3%82%A2 )

I think it would be better to pass decoded params to screen.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -521,7 +521,15 @@ export default (routeConfigs, stackConfig = {}) => {
         }
         const nextResult = result || {};
         const paramName = key.name;
-        nextResult[paramName] = matchResult;
+
+        let decodedMatchResult;
+        try {
+          decodedMatchResult = decodeURIComponent(matchResult);
+        } catch (e) {
+          // ignore `URIError: malformed URI`
+        }
+
+        nextResult[paramName] = decodedMatchResult || matchResult;
         return nextResult;
       }, queryParams);
 
--- a/src/routers/__tests__/StackRouter-test.js
+++ b/src/routers/__tests__/StackRouter-test.js
@@ -1355,6 +1355,28 @@ describe('StackRouter', () => {
     );
   });
 
+  test('URI encoded string get passed to deep link', () => {
+    const uri = 'people/2018%2F02%2F07';
+    const action = TestStackRouter.getActionForPathAndParams(uri);
+    expect(action).toEqual({
+      routeName: 'person',
+      params: {
+        id: '2018/02/07',
+      },
+      type: NavigationActions.NAVIGATE,
+    });
+
+    const malformedUri = 'people/%E0%A4%A';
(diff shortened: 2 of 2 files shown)


brentvatne:
@ericvicenti - seems fine to me

src/routers/StackRouter.js
```diff
@@ -521,7 +521,15 @@ export default (routeConfigs, stackConfig = {}) => {
         }
         const nextResult = result || {};
         const paramName = key.name;
-        nextResult[paramName] = matchResult;
+
+        let decodedMatchResult;
+        try {
+          decodedMatchResult = decodeURIComponent(matchResult);
```
brentvatne: is there any way this could be decoded to empty string (other than empty string), which would then cause `decodedMatchResult || matchResult` to return `matchResult`?

uiureo: No, I think.
When it has a null character, `decodeURIComponent('%00')` returns `'\u0000'`.

src/routers/__tests__/StackRouter-test.js
```diff
@@ -1355,6 +1355,28 @@ describe('StackRouter', () => {
     );
   });
 
+  test('URI encoded string get passed to deep link', () => {
+    const uri = 'people/2018%2F02%2F07';
+    const action = TestStackRouter.getActionForPathAndParams(uri);
+    expect(action).toEqual({
+      routeName: 'person',
```
satya164: uri has `people`, but `routeName` is `person`, is it correct?

ericvicenti:
Thanks for fixing this, seems like an important improvement!

src/routers/__tests__/StackRouter-test.js
uiureo: yes https://github.com/react-navigation/react-navigation/blob/2187d8fe660a35a579e971cf237634a3bfb60afb/src/routers/__tests__/StackRouter-test.js#L69-L72

uiureo:
Thanks!

esbenvb:
Apologies for waking up this topic again.

I just experienced that the URL query params that are converted into a the `params` that my screen receives, are not being URL decoded by default. I guess that should happen though, that's why I started searching in the issues of this repo and came across this PR. 

Is it on purpose that this does only decode named path-elements and not the query params, and what are the reasons behind this decision?

PR #3462, opened 2018-02-07
Add iconStyle property to tabBarOptions for iOS

SoundBlaster:
This property is Android-only now, but is very useful for styling icons.

I need a same look of TabBar on both platforms, but without this changes I can not do it. Also this changes provide a more consistent and solid API for customising TabBar.

### DIFF
--- a/src/views/TabView/TabBarBottom.js
+++ b/src/views/TabView/TabBarBottom.js
@@ -93,6 +93,7 @@ class TabBarBottom extends React.PureComponent {
       renderIcon,
       showIcon,
       showLabel,
+      iconStyle,
     } = this.props;
     if (showIcon === false) {
       return null;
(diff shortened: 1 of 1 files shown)


src/views/TabView/TabBarBottom.js
```diff
@@ -105,7 +106,11 @@ class TabBarBottom extends React.PureComponent {
         inactiveTintColor={inactiveTintColor}
         renderIcon={renderIcon}
         scene={scene}
-        style={showLabel && this._shouldUseHorizontalTabs() ? {} : styles.icon}
+        style={
+          showLabel && this._shouldUseHorizontalTabs()
```
satya164: I'd change this to 

```js
[showLabel && this._shouldUseHorizontalTabs() ? styles.icon : null, iconStyle]
```

brentvatne:
looks like you need to update jest snapshots

src/views/TabView/TabBarBottom.js
SoundBlaster: Prettier is very capricious 😅

ericvicenti:
You can run `yarn jest -u` to record new snapshots and hopefully the tests will pass

brentvatne:
this belongs on https://github.com/react-navigation/react-navigation-tabs now :O

Issue #3430, opened 2018-02-05
black screen is cased by push and pop quickly


musicode:
## The issue tracker is reserved for bug reports only.

1. have you tried updating to the latest release?

   yes

2. does it happen only on Android/iOS?

  iOS, I dont know how about android

3. does it happen on both Debug and Release modes?

   both

4. does it happen with Debugger On or Off? Both?

   off

5. Can you provide a GIF or a repo or a snack to reproduce this issue?

I found that when i push and pop **quickly**, it will happen at the last pop action.

![image](https://user-images.githubusercontent.com/2732303/35776760-6adce642-09dd-11e8-8685-8f852f346483.png)

![image](https://user-images.githubusercontent.com/2732303/35776770-8dc71c72-09dd-11e8-836b-6f834c4db7e3.png)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0.0-beta.29
| react-native     |  0.52.0
| node             |   v8.9.0
| npm or yarn      |  5.5.1

kelset:
Without code it's hard to investigate.

brentvatne:
indeed we need some way to reproduce this as I wasn't able to on my own, can you post some code and steps to follow please @musicode? thanks for the report!

musicode:
[this](https://user-images.githubusercontent.com/2732303/35776770-8dc71c72-09dd-11e8-836b-6f834c4db7e3.png) is a screenshot gif, it's so frequently, i need help

kelset:
Again, @musicode , it's **code** that is needed because it's really hard to investigate without.

You can use [Snack Expo](snack.expo.io) to provide us with a simple repro.

brentvatne:
@musicode - as @kelset suggested, please create a snack where we can reproduce this. it's possible if you're using redux that something is wrong with the redux config. also be sure to update to the latest version. i can't reproduce this using the same steps you did in your gif on the playground app: https://expo.io/@react-navigation/NavigationPlayground

musicode:
@brentvatne i don't use redux.

I set `transitionConfig` like below, it' ok. I think the default transitionConfig has some bugs.

```js
StackNavigator(
    routeConfig,
    {
      headerMode: 'screen',
      backTitle: null,
      transitionConfig: () => ({
        transitionSpec: {
          duration: 300,
          easing: Easing.out(Easing.poly(4)),
          timing: Animated.timing,
        },
        screenInterpolator: sceneProps => {
[...]

PR #3469, opened 2018-02-08
StackRouter block actions while transitioning

ericvicenti:
The most straightforward fix for two issues is to block all navigation actions while mid-transition of a stack navigator. This will fix:

The double-navigate on double tap issue, because the first navigation will start the transition and the second action will be ignored.

Will fix the buggy header experience that you can see when going back and forward to a different route quickly. This happens because the next navigate action happens before the completion action. After the fix, the navigate action will be ignored, the user will tap again, and will see a good transition

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -45,6 +45,9 @@ const navigate = createAction(NAVIGATE, payload => {
   if (payload.key) {
     action.key = payload.key;
   }
+  if (payload.immediate) {
+    action.immediate = payload.immediate;
+  }
   return action;
 });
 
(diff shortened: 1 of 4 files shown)


musicode:
[#3430](https://github.com/react-navigation/react-navigation/issues/3430) is still not ok

brentvatne:
@musicode - we need to be able to reproduce it. I have been unable to in all of my attempts. please explain how we can do it with the NavigationPlayground app or provide a failing test for us

Amurmurmur:
I can still performs a double click and open 2 screens one after the other in a StackNavigator

brentvatne:
we reverted this. see https://github.com/react-navigation/rfcs/issues/16 for discussion

Issue #3471, opened 2018-02-08
Can not switch tab until navigate back animation ends


helele90:
### Current Behavior

It seems like a bug. [this is video](https://www.youtube.com/watch?v=LHCu-Cm8tS4&feature=youtu.be)

If I switch back to tab2 immediately after returning, It will automatically switch to tab1.
If I wait a minute switch to tab2, there is no problem.
All interfaces may be triggered.

[Snack Expo](https://snack.expo.io/r1ye4TK8M)

### Expected Behavior

 The last time in the video is Expected Behavior.

### How to reproduce

It seems that the navigation animation does not end before switching tabs, it will fire.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0
| react-native     | 50.1
| node             | v8
| npm or yarn      | npm 6.4

kelset:
Can you please provide a repro with [Snack Expo](https://snack.expo.io/)?

helele90:
@kelset  [this is Snack Expo](https://snack.expo.io/ryilXTKIf)

brentvatne:
fixed and released 1.0.1, thanks for the report with the video and the snack! ❤️

kgdev:
I believe the same fix exists for the StackRouter. Should we apply the same fix to StackRouter?

dazweeja:
Agree with @kgdev. The same bug still exists in StackRouter, but in that case it's a simple one-line fix - change line 545 in master from:

`action.type === NavigationActions.SET_PARAMS`

to:

`action.type === NavigationActions.SET_PARAMS || action.type === StackActions.COMPLETE_TRANSITION`

PR #3473, opened 2018-02-08
Make TabRouter handle COMPLETE_TRANSITION in a child router without switching active index

brentvatne:
# Motivation

Resolves https://github.com/react-navigation/react-navigation/issues/3471

# Test plan

Run tests, try to reproduce the issue in the tabs in stacks example in NavigationPlayground.

# Follow-up

https://github.com/react-navigation/rfcs/issues/12


ericvicenti:
Thank you!!

PR #3474, opened 2018-02-08
Fix StackRouter initial navigate child router behavior and key consistency

ericvicenti:
This problem was found and fixed by @matthargett and @jayphelps in #3397. I’m just rebasing and cleaning a few things up

Also shoutout to @muramasa who wrote the test to diagnose this

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -101,23 +101,32 @@ export default (routeConfigs, stackConfig = {}) => {
       // Set up the initial state if needed
       if (!state) {
         let route = {};
-        if (
-          behavesLikePushAction(action) &&
-          childRouters[action.routeName] !== undefined
-        ) {
+        const childRouter = childRouters[action.routeName];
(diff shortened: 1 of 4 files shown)


src/routers/StackRouter.js
```diff
@@ -101,23 +101,30 @@ export default (routeConfigs, stackConfig = {}) => {
...
       if (!state) {
         let route = {};
-        if (
-          behavesLikePushAction(action) &&
-          childRouters[action.routeName] !== undefined
-        ) {
+        const childRouter = childRouters[action.routeName];
+        if (behavesLikePushAction(action) && childRouter !== undefined) {
+          let childState = {};
+          if (childRouter !== null) {
```
brentvatne: we only execute this code if `childRouter !== undefined` (the above conditional), is it necessary to check that it's not null here? if so we should add a comment

ericvicenti: Sure I'll add a comment. This is the main fix from #3397. A child router is null if there is a route there but no router. Previously we had not been allowing the child router to handle the initial action

Issue #3477, opened 2018-02-09
screen freezes after calling navigate function with 'key'


ryanix:
### Current Behavior

- What code are you running and what is happening?
  Inside StackNavigator I had three screens registered like this
```javascript
const Stack = StackNavigator({
  Home: {
    screen: HomeScreen
  },
  Home1: {
    screen: HomeScreen1
  },
  Home2: {
    screen: HomeScreen2
  }
},{
  headerMode: 'none',
});
```
on each screen I had a button which will navigate to next screen when pressed, on the last screen I had a button navigate to the first screen
When I navigate back to 'Home' screen on 'Home2' screen with 
```javascript
navigation.navigate({routeName, key}) 
```
I used a global variable routeNameList to record the route key of each stack,

routeNameList looks like this { routeName: key }

it successfully navigate back, but the screen freeze,  button no longer responds to click

- Include a screenshot if it makes sense.

### Expected Behavior

- What do you expect should be happening?
Screen should not freeze
- Include a screenshot if it makes sense.

### How to reproduce

full code is attached:
```javascript
import React, { Component } from 'react';
import {  AppRegistry,View,Button} from 'react-native';
import {StackNavigator} from 'react-navigation';


var routeNameList = {}
function navigateTo(nav,route) {
  let key = routeNameList[route]
  if (key) {
    nav.navigate({routeName: route, key:key})
  } else {
    nav.navigate(route)
  }
}

function register(props) {
  let route = props.navigation.state;
  let {routeName, key} = route;
  routeNameList[routeName] = key
}


class HomeScreen extends React.Component {
  render() {
    register(this.props)
    return (
      <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
        <Button title='Jump To Home1'  onPress={()=> navigateTo(this.props.navigation,'Home1')}/>
      </View>
    );
  }
}

class HomeScreen1 extends React.Component {
  render() {
    register(this.props)
    return (
      <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
        <Button title='Jump To Home2'  onPress={()=> navigateTo(this.props.navigation,'Home2')}/>
      </View>
    );
  }
}

class HomeScreen2 extends React.Component {
  render() {
    register(this.props)
    return (
      <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
        <Button title='Jump back To Home'  onPress={()=> navigateTo(this.props.navigation,'Home')}/>
[...]

PR #3478, opened 2018-02-09
Get rid of unused routes when navigating back with key

brentvatne:
# Motivation

Fixes https://github.com/react-navigation/react-navigation/issues/3477

# Test plan

- Run tests (added one)
- Try the example in NavigationPlayground


MrHubble:
@brentvatne Thank you for fixing this so quickly! It's very much appreciated.

PR #3486, opened 2018-02-09
Make Router({ RouteName: Component }) a valid way to configure a router

brentvatne:
# Motivation

When getting started, and often until deep linking is required, people use the following style of route config:

```js
const MyStack = StackNavigator({
  ScreenA: { screen: MyScreen },
  ScreenB: { screen: MyOtherScreen },
  ScreenC: { screen: MyNestedNavigator },
});
```

This PR doesn't change that (it's still valid), but it also makes the following valid:

```js
const MyStack = StackNavigator({
  ScreenA: MyScreen,
  ScreenB: MyOtherScreen,
  ScreenC: MyNestedNavigator,
});
```

# Test plan

Run tests, try playground.

I don't know how to update the type interface for this &mdash; I get a bunch of errors with the following:

```
  declare export type NavigationScreenRouteConfig =
    | NavigationComponent
    | {
        screen: NavigationComponent,
      }
    | {
        getScreen: () => NavigationComponent,
      };
```

cc @Ashoat @ericvicenti

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -96,39 +96,19 @@ const ExampleInfo = {
   },
 };
 const ExampleRoutes = {
-  SimpleStack: {
-    screen: SimpleStack,
-  },
-  SimpleTabs: {
-    screen: SimpleTabs,
-  },
-  Drawer: {
-    screen: Drawer,
-  },
+  SimpleStack: SimpleStack,
+  SimpleTabs: SimpleTabs,
+  Drawer: Drawer,
   // MultipleDrawer: {
   //   screen: MultipleDrawer,
   // },
-  TabsInDrawer: {
-    screen: TabsInDrawer,
-  },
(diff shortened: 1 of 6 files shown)


Ashoat:
Just checked the CI result and that's pretty much exactly what the Flow errors say:

```js
Error: js/App.js:246
                          v--------------
246: const AppNavigator = StackNavigator(
247:   {
248:     ...ExampleRoutes,
...:
263: );
     ^ call of `StackNavigator`
 99:   SimpleStack: SimpleStack,
                    ^^^^^^^^^^^ intersection. This type is incompatible with
                                                   v
335:   declare export type NavigationRouteConfig = {
336:     navigationOptions?: NavigationScreenConfig<*>,
337:     path?: string,
338:   } & NavigationScreenRouteConfig;
```

brentvatne:
@Ashoat - thanks for the tip! that fixed it

PR #3487, opened 2018-02-10
Cleanup around StackRouter

brentvatne:
- Make StackRouter a bit easier to read, update and add some comments
- Add an explicit test around stack and navigate

# Motivation

I was experimenting with the code and improving my understanding of it. Made a few changes here in the process so hopefully it's a bit easier for the next person to understand.

# Test plan

Run the tests. This should not have any impact on functionality.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -51,6 +51,61 @@ export default (routeConfigs, stackConfig = {}) => {
   const pathsByRouteNames = { ...stackConfig.paths } || {};
   let paths = [];
 
+  function getInitialState(action) {
+    let route = {};
+    const childRouter = childRouters[action.routeName];
+
+    // This is a push-like action, and childRouter will be a router or null if we are responsible for this routeName
+    if (behavesLikePushAction(action) && childRouter !== undefined) {
+      let childState = {};
+      // The router is null for normal leaf routes
+      if (childRouter !== null) {
+        const childAction =
+          action.action || NavigationActions.init({ params: action.params });
+        childState = childRouter.getStateForAction(childAction);
+      }
+      return {
+        key: 'StackRouterRoot',
+        isTransitioning: false,
+        index: 0,
+        routes: [
+          {
+            params: action.params,
+            ...childState,
+            key: action.key || generateKey(),
+            routeName: action.routeName,
+          },
+        ],
+      };
+    }
+
+    if (initialChildRouter) {
+      route = initialChildRouter.getStateForAction(
+        NavigationActions.navigate({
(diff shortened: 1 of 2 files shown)


src/routers/StackRouter.js
```diff
@@ -348,6 +303,60 @@ export default (routeConfigs, stackConfig = {}) => {
         }
       }
 
+      // Handle pop-to-top behavior. Make sure this happens after children have had a chance to handle the action, so that the inner stack pops to top first.
+      if (action.type === NavigationActions.POP_TO_TOP) {
+        if (state.index === 0) {
+          return {
+            ...state,
```
ericvicenti: this looks crazy, without an explanation of why its happening

src/routers/StackRouter.js
```diff
@@ -100,63 +155,11 @@ export default (routeConfigs, stackConfig = {}) => {
...
-          routeName: initialRouteName,
-          key: action.key || generateKey(),
-        };
-        state = {
-          key: 'StackRouterRoot',
-          isTransitioning: false,
-          index: 0,
-          routes: [route],
-        };
+        return getInitialState(action);
```
ericvicenti: I like this. I wonder if we should make this file way more functiony, in the name of comprehension. Create functions for each action and attempt to get `getStateForAction` down to ~100 lines or something

src/routers/StackRouter.js
```diff
@@ -348,6 +303,60 @@ export default (routeConfigs, stackConfig = {}) => {
         }
       }
 
+      // Handle pop-to-top behavior. Make sure this happens after children have had a chance to handle the action, so that the inner stack pops to top first.
+      if (action.type === NavigationActions.POP_TO_TOP) {
```
ericvicenti: we should probably fix this to accept a key at some point soon, same with the others

src/routers/__tests__/StackRouter-test.js
```diff
@@ -513,6 +513,39 @@ describe('StackRouter', () => {
...
+      {
+        type: NavigationActions.NAVIGATE,
+        routeName: 'Corge',
+      },
+      state
+    );
+
+    expect(state2.index).toEqual(1);
+    expect(state2.routes[1].index).toEqual(1);
+    expect(state2.routes[1].routes[1].index).toEqual(1);
```
ericvicenti: might want to verify "Corge" route name for clarity here

src/routers/StackRouter.js
brentvatne: this actually didn't change but I can add a comment here

PR #3494, opened 2018-02-10
Remove Linking/BackHandler from PlatformHelpers.

raarts:
The Linking component is now supported by react-native-web. It also provider BackHandler as a dummy components, so there's no need to supply dummy replacements ourselves.

So this fixes the warning, and React Navigation on react-native-web should now be able to pick up the initialUrl too.

### DIFF
--- a/src/PlatformHelpers.web.js
+++ b/src/PlatformHelpers.web.js
@@ -1,9 +1,3 @@
-export const Linking = {
-  addEventListener: () => {},
-  removeEventListener: () => {},
-  getInitialURL: () => Promise.reject('Unsupported platform'),
-};
+import { BackHandler } from 'react-native';
 
-export const BackHandler = {
-  addEventListener: () => {},
-};
+export { BackHandler };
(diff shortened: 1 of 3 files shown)


src/PlatformHelpers.web.js
```diff
@@ -1,9 +1,3 @@
-export const Linking = {
-  addEventListener: () => {},
-  removeEventListener: () => {},
-  getInitialURL: () => Promise.reject('Unsupported platform'),
-};
+import { BackHandler } from 'react-native';
```
ericvicenti: I know that everybody rewrites `react-native` into `react-native-web` in the webpack config, but it just feels so hacky to be apparently importing mobile code from a web module </rant>

satya164: Why keep this file? It does the same thing as the other one. Isn't renaming `PlatformHelpers.native.js` to `PlatformHelpers.js` should address the issue?

PR #3495, opened 2018-02-10
Run Prettier on Flow libdef on precommit

Ashoat:
My first thought was to do it through `eslint`, which already runs on `precommit`. However, `eslint` has a custom parser, and the only way to get it to process flow-typed files is with `babel-eslint`, but `babel-plugin-transform-flow-strip-types` can't handle the libdef-specific syntax `declare export type`.

Since we can't get `eslint` to run Prettier for us, I just set up `lint-staged` to call it on the libdef file separately. I'm don't feel too strongly about running Prettier on the libdef, so if this approach is too messy/brittle let me know and we can scuttle it.

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -332,12 +332,12 @@ declare module 'react-navigation' {
     navigationOptions?: ?NavigationScreenConfig<Options>,
   };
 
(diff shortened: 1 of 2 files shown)


brentvatne:
lgtm, thanks!

PR #3496, opened 2018-02-11
Allow modification of SafeAreaView props

ArsenyYankovsky:
**Motivation**:

So my use case is the following. I have this offline thing on top of my screens:

rsz_screenshot_1518292144

Which I achieve by doing following in my App.js:

```
    <SafeAreaView/>
    <NetworkStatus isOffline={this.state.isOffline}/>
    <RootNavigation />
```

This works fine on Android, but on iOS I get some additional padding on the header which looks like this:

rsz_2s6a5205_k

By introducing new `safeAreaViewProps` option we allow users to override forceInset property which solves my use case.

**Test plan**

I used following code to fix my issue described above:

```
    Story: {
      screen: StoryScreen,
      navigationOptions: ({navigation}) => ({
        title: "Story",
        safeAreaViewProps: { forceInset: {}}
      }),
    },
```

For the screens with unmodified options default behavior is applied. I also tested it on Android which led to the same results.

All tests pass and eslint doesn't report any errors.


brentvatne:
can you rebase this? looks like there is a conflict.

also, this needs to be added to the flow/react-native.js file and docs.

as an aside, and I think you pointed this out in a previous issue -- I'm not sure that this is a good solution in the long-term but if it solves problems for people for now that's great and we can revisit it when @ericvicenti and I revisit the problem of safe areas in the coming months.

ArsenyYankovsky:
This is more flexible as the PR I submitted before, it allows users to access SafeAreaView properties. Where should I add the docs? Couldn't find the place in the repo.

brentvatne:
thanks @ArsenyYankovsky! can you possibly rebase again? I made some changes to header recently that resulted in conflicts.

as for docs -- https://github.com/react-navigation/react-navigation.github.io

ArsenyYankovsky:
I've rebased repo and created a pull request for the documentation repo https://github.com/react-navigation/react-navigation.github.io/pull/70.

src/views/Header/Header.js
```diff
@@ -479,10 +479,13 @@ class Header extends React.PureComponent {
       safeHeaderStyle,
     ];
 
+    const { safeAreaViewProps } = options;
+
     return (
       <SafeAreaView
         forceInset={{ top: 'always', bottom: 'never' }}
         style={containerStyles}
+        {...safeAreaViewProps}
```
brentvatne: I thought about this some more and I think we probably don't want people to inject any random prop in here, they could accidentally break the layout in their app. I think instead we should only pay attention to the `forceInset` prop. so we would change it to something like the following:

```
const { safeAreaViewProps = {} } = options;
const forceInset = safeAreaViewProps.forceInset || { top: 'always', bottom: 'never' }; 

return (
  <SafeAreaView
    forceInset={forceInset}
    style={containerStyles}
  >
```

dhirendrarathod2000:
I am also facing the same problem.
The code that is suggested under code review works for me.

Any ETA when this will get released?

Thanks
D

ArsenyYankovsky:
@dhirendrarathod2000 I will do the changes @brentvatne suggested tonight.

src/views/Header/Header.js
satya164: Maybe it makes more sense to have the prop named as `headerForceInset` instead of `safeAreaViewProps` since we're not passing all props here and `SafeAreaView` is an implementation detail.

dhirendrarathod2000: or we can put safe header style inside that too.
Does that sound right?
So that all the new props we add in the future we can add under that object.

brentvatne: @satya164 - makes sense. let's do what @satya164 suggested

ArsenyYankovsky:
I've updated both this and documentation pull request (https://github.com/react-navigation/react-navigation.github.io/pull/70)

dhirendrarathod2000:
@brentvatne Can you please approve/merge this?

Issue #2046, opened 2017-07-02
Improvement: navigation.state.params should be empty object instead of undefined


tyleralves:
It is inconvenient that the `navigation.state.params` property is undefined when no route params are passed in. In components that rely on a property of params, one must add an additional check that params exists, before checking that the desired property exists:

`const photoType = (navigation.state.params && navigation.state.params.photoType) || 'self';`

If the `params` property was an empty object rather than undefined, it would be much nicer:

`const photoType = navigation.state.params.photoType || 'self';`

jurajkrivda:
This is my approach to this problem


```
import { path } from 'ramda';
import { withProps } from 'recompose';

withProps(props => ({
  photoType: path([ 'navigation', 'state', 'params', 'photoType' ], props),
}))
```

nonameolsson:
@jurajkrivda Nice solution! But as @tyleralves says, I also think that it should be an empty object. Then it isn't necessary with a third-party library to perform such check.

brentvatne:
yeah this seems pretty reasonable

peterpme:
@tyleralves , @nonameolsson, @jurajkrivda #3510 recently merged in offers:

```
this.props.navigation.getParam('name', 'default value')
```

https://reactnavigation.org/docs/navigation-prop.html#getparam-get-a-specific-param-value-with-a-fallback

Thanks!

brentvatne:
i'd suggest using `getParam` - there's a lot of discussion in this rfc about it: https://github.com/react-navigation/rfcs/pull/21

PR #3510, opened 2018-02-13
Add getParam navigation helper

peterpme:
Handling `navigation.state.params.name` when `params` doesn't exist results in an error. 

Initially, I was going to create an empty object.
After reading #2046, I went with the suggestion of creating a `getParam` helper instead.

**Test plan (required)**
- Set a parameter
- Get the parameter

- Get an undefined parameter

### DIFF
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -62,11 +62,10 @@ export default function(navigation) {
     },
 
     getParam: (paramName, defaultValue) => {
-      const param =
-        navigation.state.params && navigation.state.params[paramName];
+      const params = navigation.state.params;
 
-      if (param != null) {
-        return param;
+      if (params && paramName in params) {
+        return params[paramName];
       }
 
       return defaultValue;
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -61,6 +61,12 @@ export default function(navigation) {
       return navigation.dispatch(NavigationActions.setParams({ params, key }));
     },
 
+    getParam: (paramName, defaultValue) => {
+      const param =
+        navigation.state.params && navigation.state.params[paramName];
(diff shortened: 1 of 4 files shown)


src/addNavigationHelpers.js
```diff
@@ -61,6 +61,12 @@ export default function(navigation) {
       return navigation.dispatch(NavigationActions.setParams({ params, key }));
     },
 
+    getParam: (paramName, defaultValue) => {
+      const param =
+        navigation.state.params && navigation.state.params[paramName];
+      return param || defaultValue;
```
dantman: You should have an `if` condition testing for the presence of the `paramName` in `params` instead of doing an or. In the current implementation if params is `params: {foo: 0}` then `getParam('foo')` will return `undefined` instead of `0`.

peterpme: :+1: thanks!

dantman:
`in` would probably be better than testing for `undefined`/`null`. If someone explicitly sets `{foo: null}` in their params, then they probably don't want `getParam('foo', 0)` returning `0` or `getParam('foo')` changing that `null` to `undefined`.

peterpme:
What if `params` is undefined?

`_.get` returns `undefined` if no argument is set
`_.get({}, 'b', null)` will return `null`

`getParam('foo', 0)` should return 0
`getParam('foo')` // undefined
`getParam('foo', null)` // null

I'm all for giving `in` a shot, but I think then I think it'd be nice if `params` was an empty object by default

Let me know what you think!

dantman:
`params` being undefined is the same as a specific param not being a property of the object.

So something like:
```js
const params = navigation.state.params;

if ( params && paramName in params ) {
  return params[paramName];
}

return defaultValue;
```

peterpme:
Great suggestion, thanks!

PR #3512, opened 2018-02-13
add withNavigationFocus HOC

slorber:
As discussed in other issues, we should be able to know easily which screen has the focus.

This HOC idea has been mentioned here by @satya164 and @ericvicenti 
- https://github.com/react-navigation/react-navigation/issues/51#issuecomment-276003658
- https://github.com/react-navigation/react-navigation/issues/51#issuecomment-278761705
- https://github.com/react-navigation/react-navigation/pull/3345#issuecomment-360260067

Here is a basic implementation using the new lifecycle hooks to provide this feature

### DIFF
--- a/src/views/withNavigationFocus.js
+++ b/src/views/withNavigationFocus.js
@@ -23,7 +23,7 @@ export default function withNavigationFocus(Component) {
         ];
       } else {
         console.warn(
-          'withIsFocused wrapped component did not receive navigation from props, so isFocused prop swill remain false'
+          'withNavigationFocus wrapped component did not receive navigation from props, so isFocused prop will remain false'
         );
       }
     }
--- a/src/views/withNavigationFocus.js
+++ b/src/views/withNavigationFocus.js
@@ -1,35 +1,45 @@
 import React from 'react';
+import propTypes from 'prop-types';
 import hoistStatics from 'hoist-non-react-statics';
+import invariant from '../utils/invariant';
 
 export default function withNavigationFocus(Component) {
   class ComponentWithNavigationFocus extends React.Component {
     static displayName = `withNavigationFocus(${Component.displayName ||
       Component.name})`;
 
+    static contextTypes = {
+      navigation: propTypes.object.isRequired,
+    };
+
     state = {
       isFocused: false,
     };
 
     componentDidMount() {
-      if (this.props.navigation) {
-        this.subs = [
-          this.props.navigation.addListener('didFocus', () =>
-            this.setState({ isFocused: true })
-          ),
-          this.props.navigation.addListener('willBlur', () =>
-            this.setState({ isFocused: false })
-          ),
-        ];
-      } else {
-        console.warn(
-          'withNavigationFocus wrapped component did not receive navigation from props, so isFocused prop will remain false'
-        );
-      }
+      const navigation = this.getNavigation();
+      this.subs = [
+        navigation.addListener('didFocus', () =>
+          this.setState({ isFocused: true })
+        ),
+        navigation.addListener('willBlur', () =>
+          this.setState({ isFocused: false })
+        ),
+      ];
     }
 
     componentWillUnmount() {
-      this.subs && this.subs.forEach(sub => sub.remove());
+      this.subs.forEach(sub => sub.remove());
     }
+
+    getNavigation = () => {
+      const navigation = this.props.navigation || this.context.navigation;
+      invariant(
+        !!navigation,
+        'withNavigationFocus can only be used on a view hierarchy of a navigator. The wrapped component is unable to get access to navigation from props or context.'
+      );
+      return navigation;
+    };
 
     render() {
       return (
--- a/src/views/withNavigationFocus.js
+++ b/src/views/withNavigationFocus.js
@@ -18,7 +18,7 @@ export default function withNavigationFocus(Component) {
 
     componentDidMount() {
       const navigation = this.getNavigation();
-      this.subs = [
+      this.subscriptions = [
         navigation.addListener('didFocus', () =>
           this.setState({ isFocused: true })
         ),
@@ -29,7 +29,7 @@ export default function withNavigationFocus(Component) {
     }
 
     componentWillUnmount() {
-      this.subs.forEach(sub => sub.remove());
+      this.subscriptions.forEach(sub => sub.remove());
     }
 
     getNavigation = () => {
--- a/src/views/withNavigationFocus.js
+++ b/src/views/withNavigationFocus.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import propTypes from 'prop-types';
 import hoistStatics from 'hoist-non-react-statics';
 
 export default function withNavigationFocus(Component) {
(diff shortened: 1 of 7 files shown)


slorber:
just added an example to playground app: TabsWithNavigationFocus

src/views/withNavigationFocus.js
```diff
@@ -0,0 +1,46 @@
...
+  class ComponentWithNavigationFocus extends React.Component {
+    static displayName = `withNavigationFocus(${Component.displayName ||
+      Component.name})`;
+
+    state = {
+      isFocused: false,
+    };
+
+    componentDidMount() {
+      if (this.props.navigation) {
```
ericvicenti: I think it would be great if we could use the navigtion from context if the user forgets to pass it in as a prop. Take a look at the other HOC, `withNavigation`.

src/views/withNavigationFocus.js
```diff
@@ -0,0 +1,46 @@
...
+        this.subs = [
+          this.props.navigation.addListener('didFocus', () =>
+            this.setState({ isFocused: true })
+          ),
+          this.props.navigation.addListener('willBlur', () =>
+            this.setState({ isFocused: false })
+          ),
+        ];
+      } else {
+        console.warn(
```
ericvicenti: Once we start accepting context from this HOC, I think it would best to have an invariant here instead of warning- things would need to be very broken.

src/views/withNavigationFocus.js
```diff
@@ -0,0 +1,46 @@
...
+    componentWillUnmount() {
+      this.subs && this.subs.forEach(sub => sub.remove());
+    }
+
+    render() {
+      return (
+        <Component
+          {...this.props}
+          isFocused={this.state.isFocused}
+          ref={this.props.onRef}
```
satya164: This will not work for function components. We need to check if the component is a function component or not before attaching the ref, and only attach the ref to class components.

On another note, I have seen a `getWrappedInstance` method to be a more common way to expose the wrapped component's ref, so maybe better to use it instead of `this.props.onRef` for better ecosystem compatibility.

satya164: Here is an example HOC where I do it:
https://github.com/callstack/react-native-paper/blob/master/src/core/withTheme.js#L84
https://github.com/callstack/react-native-paper/blob/master/src/core/withTheme.js#L114

ericvicenti: We already used onRef somewhere else cc @brentvatne (although I don't feel strongly either way)

satya164: cc @brentvatne

slorber: I prefer `getWrappedInstance` compared to `this.props.onRef` as it is used in react-intl and react-redux (but it requires a `{withRef: true}` config on the HOC. I used onRef because it's already what is available on `withNavigation` so didn't want to use something different and keep the API consistent.


Also worth noting that I don't like much the api of react-intl as it does not play well with `compose` because the config is 2nd arg. I prefer allowing `withNavigationFocus(config)(Comp)` and withNavigationFocus(Comp)` instead of withNavigationFocus(Comp,config)`. What do you think about it?

satya164: > it requires a {withRef: true} config on the HOC

Yeah, feels kinda unnecessary, but no idea why it's that way.

> I prefer allowing withNavigationFocus(config)(Comp)

+1, but what config?

slorber:
I updated the PR to get navigation from both props and context

src/views/withNavigationFocus.js
satya164: Do we need really need a config? Isn't it enough to detect if a component is function component or class? Or there are other issues with attaching ref which I'm missing?

satya164: `onRef` is used here https://github.com/react-navigation/react-navigation/blob/67f98b69eb141d9adbc579fb5256914fd5b66ccc/src/views/withNavigation.js#L20

ericvicenti:
As for the `getWrappedInstance` vs `props.onRef` question, lets move that to a follow-up issue/PR/RFC, because we're already invested in `props.onRef` for now because of `withNavigation`. I don't have a strong preference either way, so I'll leave it up to y'all, but I believe that `onRef` requires a tiny bit less boilerplate for people.

slorber:
fixed the flow errors on playground, tell me if there's anything else to do

brentvatne:
@slorber - can you open a PR on the website repo with an API reference for this? http://github.com/react-navigation/website

brentvatne:
@slorber - it turns out that this isn't a reliable way to find the navigation state. it will only work if it is wrapping a navigation screen. if you add it to some component in the screen, and that component is rendered after the parent screen component is mounted, this will not work until you navigate to and away from the screen where it's rendered. this is because it depends on navigation events, and those are only fired when things change

satya164:
Maybe a `this.props.navigation.isFocused()` method will be helpful to get the current focused state? Users can get the initial focus state by using it in `componentWillMount`

slorber:
I understand that problem, do you have any solution to propose?

What about a new navigator config `withNavigationFocus: true`, and making the HOC private/lib internal. This way all screens will be wrapped by the HOC and we can be sure they won't be mounted after the focus event. This HOC makes mostly sense at the root so navigator config can be an appropriate place.

@satya164 I wouldn't use such `isFocused` function. I've used such child component querying parent from function prop and this get really confusing because if you use this on render (and be certain people will try that) you will be confused because the component won't render on focus change if the component is pure and function is stable. I tricked myself with that in the past.

satya164:
Another option is to always call the listener with the current value as soon as you attach it.

slorber:
I like this idea (it's what History.listen does to provide immediate history location), but shouldn't we create a `onFocusChange` or something instead of calling `didFocus` immediately?

brentvatne:
I don't like calling `didFocus` immediately, this leads to IMO some confusing behaviors

src/views/withNavigationFocus.js
```diff
@@ -0,0 +1,56 @@
...
+  class ComponentWithNavigationFocus extends React.Component {
+    static displayName = `withNavigationFocus(${Component.displayName ||
+      Component.name})`;
+
+    static contextTypes = {
+      navigation: propTypes.object.isRequired,
+    };
+
+    state = {
+      isFocused: false,
```
brentvatne: as per my most recent comment, if we exposed `isFocused()` then this would just be:

```
constructor(props, context) {
  super();
  const navigator = props.navigator || context.navigator;
  this.state = {
    isFocused: navigation.isFocused(),
  };
}
```

slorber: So `isFocused` is a render-unsafe fn that we would use only to init HOC state? Couldn't this lead to users trying to using it directly and being confused when no re-renders happen? Would this function be documented for general usage?

brentvatne: i think if they use it in a function and are confused by no re-renders happening then they have misunderstood how react works, that's not really a problem of the library

slorber: that's true, and it still can be advised to use the HOC anyway.
I'll try to check soon how to provide that `isFocused` func

PR #3526, opened 2018-02-15
Header transition presets with approximate support for UIKit transition style

brentvatne:
# Motivation

Some people want the header to look closer to the iOS defaults, where the back button and back label animate separately, and the back label transitions smoothly into the title. Additionally, the previous back title slides under the back icon and is masked by it. This PR gets us closer to that but it's not perfect.

Most non-Apple apps on my phone don't use this style, though. For example, the Instagram, Twitter, and Facebook apps all cross-fade the header components in-place, with no translation. So I changed the default transition preset to do that. To enable the iOS style header transition preset, you must be using iOS (`MaskedViewIOS` is naturally iOS only, given the name. Having a masked view component on Android would be useful, but it would be super weird to use this transition style on Android), then in your `StackNavigator` config you can add `headerTransitionPreset: 'uikit',`. Open to suggestions for a better name :)

I also fixed the header title weight, which was slightly off for iOS11 styles.

# Test plan

Try out playground on iOS and Android. The Simple Stack example uses the `uikit` style.

You can check out a video of it in action here: http://url.brentvatne.ca/T6jOzE
You can also try out the playground app deployed from this branch here: https://expo.io/@react-navigation/NavigationPlayground?release-channel=header-transition-presets


brentvatne:
@ericvicenti - can you review when you have a sec?

nico1510:
Hey @brentvatne I just wanted to let you know that this part:
```diff
   _renderLeft(props) {
+    const { options } = this.props.getScreenDetails(props.scene);
+
+    if (props.scene.index === 0) {
+      return null;
+    }
```
breaks my custom headerLeft component. I have a navigation structure like this: (abbreviated)
```javascript
const navigationOptions = ({ navigation }: NavigationScreenConfigProps) => ({
    headerLeft: <HeaderBackButton tintColor="white" onPress={() => navigation.goBack(null)} />,
});

const RootNavigator = DrawerNavigator({
    albumsRoute: { screen: StackNavigator({ albums: { screen: AlbumsScreen, navigationOptions } }) },
});
```
(Note: the inner StackNavigator is only needed in order to show a header since DrawerNavigator has no header)

brentvatne:
thank you @nico1510! this is fixed in https://github.com/react-navigation/react-navigation/commit/f0acaddf057020ace17428c570f6c329cf68c38c and included in 1.1.0-rc.3

PR #3528, opened 2018-02-15
Also accept numbers as navigate action key

grsmvg:
This is convenient when using numerical id's as keys. For now I'm using `key: item.id.toString()`.

### DIFF
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -54,7 +54,9 @@ export default function(navigation) {
      */
     setParams: params => {
       invariant(
-        navigation.state.key && typeof navigation.state.key === 'string',
+        navigation.state.key &&
+          (typeof navigation.state.key === 'string' ||
+            typeof navigation.state.key === 'number'),
         'setParams cannot be called by root navigator'
       );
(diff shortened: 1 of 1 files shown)


ericvicenti:
I'm fine with this change conceptually, but it looks incomplete. Can we consider the following?

- Make sure number keys work in the rest of the library
- Test various numerical keys in the navigation playground
- Write/update tests
- Update flow definition file in `flow/`, to include more than `string` as key

brentvatne:
@grsmvg - I'd be interested in seeing this change land, do you have any motivation to push it through or should we take it over?

grsmvg:
@brentvatne I'm spread over a few projects at the moment. It's just a few lines of code here and there, but I'm not familiar with the project structure yet, so it would take me a bit too much time :) Would be nice if someone else could pick this up!

brentvatne:
@grsmvg - sounds good, I'll take it over soon, thanks for letting me know

ericvicenti:
This is pretty low-priority.

If somebody wants do clean this up as discussed above, we'd happily accept a new pull request for it!

Issue #2488, opened 2017-08-29
Why TabNavigator render all sub screens at same time?


guoliang1206:
Four sub screen in the TabNavigator, don't you think it is the extra overhead  about render all of the sub screen at same time?
1. Does there has a way render single sub screen every time.
2.If 1 is not, and each of my sub screens has  a lot of data, how to handle the  hidden danger，e.g. lag, suspend?

![2017-08-29 1 31 22](https://user-images.githubusercontent.com/12846676/29805825-8f942cbc-8cbe-11e7-9ac0-d3080ba52e42.png)

buncismamen:
we got lazy options 
`lazy - whether to lazily render tabs as needed as opposed to rendering them upfront`
https://github.com/react-community/react-navigation/blob/master/docs/api/navigators/TabNavigator.md

guoliang1206:
@buncismamen  how to use Lazy?  what params should inside of lazy? can you provide an example?

buncismamen:
```
const MyApp = TabNavigator({
  Home: {
    screen: MyHomeScreen,
  },
  Notifications: {
    screen: MyNotificationsScreen,
  },
}, {
  lazy: true,
  tabBarPosition: 'top',
  animationEnabled: true,
  tabBarOptions: {
    activeTintColor: '#e91e63',
  },
});
```

guoliang1206:
@buncismamen  thanks , it works , I have wrote it in the wrong place before.

Noitidart:
@buncismamen if they are lazy loaded, do they unload when the tab is changed? It doesnt seem to unload for me.

buncismamen:
@Noitidart I think its not, they are not as smart as flatlist

maybe contributor have better explanation for this one

buncismamen:
@Noitidart idk if this component is as smart as flatlist, but I think its not, maybe contributor have better words and explanation for this, maybe you could create another issue or looking through the code.

Noitidart:
Thanks very much @buncismamen for such a fast reply!! :)

thomaskleiven:
@buncismamen I cannot seem to find the lazy option in the docs. Can I specify this property somewhere else? Thank you.

Noitidart:
@thomaskleiven here is example:

```
const HubNavigator = TabNavigator(
    TABS,
    {
        initialRouteName: INITIAL_ROUTE_NAME,
        backBehavior: 'initialRoute',
        swipeEnabled: false,
        lazy: true,
[...]

PR #3538, opened 2018-02-17
Add back support for lazy tabs and use removeClippedSubviews

brentvatne:
# Motivation

Support for the `lazy` prop was removed from react-native-tab-view. This has been a significant pain point for users (https://github.com/react-navigation/react-navigation/issues/2488). This PR implements it within react-navigation itself. Additionally, it handles applying `removeClippedSubviews` and re-positioning the children of the tab such that they will be clipped when the tab isn't active. This helps with memory usage, and is particularly helpful on Android.

# Test plan

`removeClippedSubviews` is a bit flakey on iOS, but I found that it worked reliably if it was only applied when the screen is not visible. In the event that users run into problems with this, they can toggle it off with `removeClippedSubviews: false` in the tab navigator config. It should work well on Android.


satya164:
Does it work with sliding tabs at all? The main reason it was removed because trying to do this resulted in buggy behavior which I didn't have enough time to research on and fix.

In addition, rendering a new screen mood transition affected the indicator animation and makes it laggy. When switching to a tab with animation will show uninitialized screens in between, which is another undesirable effect.

For the best user experience, users could show a placeholder (like there FB app loading placeholder) and then update the content after switching the tab, which will ensure no frames are dropped and the experience stays smooth.

satya164:
Also not sure how the resource saving container works with animated tabs, but will you be willing to contribute it to react-native-tab-view? I think it's a useful feature to have.

Also there were issues with removeClippedSubviews on iOS with listviews. Are they fixed now? I always had to disable it in my tabs on iOS. Otherwise the content of scroll view in the hidden tab will never render.

TheTekton:
When using `removeClippedSubviews`, the FlatList, SectionList, and VirtualizedList components have a [recordInteraction](https://facebook.github.io/react-native/docs/sectionlist.html#recordinteraction) method that can be used in `componentDidMount` (see [issue 1831](https://github.com/facebook/react-native/issues/1831#issuecomment-331734761)). Works well for the initial render, just not ideal.

What kind of performance trade-offs are there with using something like this versus just setting the render to null for the tab scene?

TheTekton:
Animation between tabs seems to work fine on android if the initial render disables `removeClippedSubviews` and state is reset in componentDidMount to re-enable it for unfocused/inactive screens. Didn't really test on iOS, as animations look odd with TabNavigator anyways. 😆

I tested this within tab scenes, but might look something like this here:

```
constructor(props) {
  // ...
  this.state = {
    // ...
    isInitialRender: true
  };
}

componentDidMount() {
  this.forceRemoveClippedSubviews();
}

forceRemoveClippedSubviews = () => {
  this.setState({ isInitialRender: false });
}

render() {
  // ...
  const { initialRender } = this.state;
  let isRemovingClippedSubviews = false;
  if (!initialRender) {
      isRemovingClippedSubviews = removeClippedSubviews
    } else {
      isRemovingClippedSubviews = !initialRender && !visible && removeClippedSubviews;
    }
  }
  return (
      <View
        style={styles.container}
        collapsable={false}
        removeClippedSubviews={isRemovingClippedSubviews}
      >
        {/* ... */}
      </View>
    );
}
```

PR #3540, opened 2018-02-17
use initialRouteName as key when initializing StackRouter

nico1510:
When initializing the state for a StackRouter the keys for the inital routes are generated instead of using the routeName like it is done in the TabRouter https://github.com/react-navigation/react-navigation/blob/9beb32ecaca56f7a8df3443002b5b73ee6a415a2/src/routers/TabRouter.js#L61-L65. This makes it impossible to navigate to an exisiting (the initial one) screen with the newly added [key](https://github.com/react-navigation/react-navigation/pull/3393) property without pushing a new screen.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -146,7 +146,7 @@ export default (routeConfigs, stackConfig = {}) => {
           ...route,
           ...(params ? { params } : {}),
           routeName: initialRouteName,
-          key: action.key || generateKey(),
+          key: action.key || initialRouteName,
         };
         state = {
           key: 'StackRouterRoot',


nico1510:
if this change makes any sense I will fix the tests. They fail currently because they always expect the generated keys.

brentvatne:
what happens if you use the stacknavigator in two places?

nico1510:
Hm I can't really think of a use case where that would make sense... Could you provide one maybe ?

brentvatne:
we currently do not limit people from using a stacknavigator in multiple places. if we were to land this pr we are essentially imposing a new restriction on people. the fact that tabnavigator apparently doesn't handle this is unfortunate. one example for how this might happen right now: you have a tabnavigator for admins and a tabnavigator for normal users, perhaps they share the same "notifications" tab, which points to a notifications stack.

nico1510:
Not sure if understand that correctly but... 
Then these 2 Stackrouters should be probably mounted under a distinct routeName. Otherwise it would already not be clear on where to navigate. If they are mounted under a distinct routeName then the keys would also be different.

brentvatne:
@nico1510 - can we add `initialRouteKey` config option instead? we may deprecate it in the future but with this at least we have a non-breaking change and still solve the issue at hand

nico1510:
Yes sure I can add that :)

nico1510:
@brentvatne the functionality is now behind a config flag and I added a test case.

flow/react-navigation.js
```diff
@@ -405,6 +405,7 @@ declare module 'react-navigation' {
     initialRouteParams?: NavigationParams,
     paths?: NavigationPathsConfig,
     navigationOptions?: NavigationScreenConfig<*>,
+    initialRouteKey?: 'initialRouteName' | 'generated',
```
brentvatne: this typedef doesn't seem correct. I think it's just `initialRouteKey?: string`

src/routers/StackRouter.js
```diff
@@ -92,11 +92,16 @@ export default (routeConfigs, stackConfig = {}) => {
...
     };
+    const { initialRouteKey } = stackConfig;
     route = {
       ...route,
       ...(params ? { params } : {}),
       routeName: initialRouteName,
-      key: action.key || generateKey(),
+      key:
+        action.key ||
+        (initialRouteKey && initialRouteKey === 'initialRouteName'
```
brentvatne: we shouldn't special case this, if you want it to be the same as the initial route name then just pass in a key that is the same as the initial route name

nico1510:
@brentvatne ah now I see what you meant. I updated the PR

BrandonSmith:
@nico1510 @brentvatne With the addition of `initialRouteKey`, how is it intended to be used? Seems `StackNavigator` fails to pass it down to `StackRouter`.

nico1510:
Oops yes you're right it's not passed. Should be an easy fix though. I can create a PR for this or you can create one yourself if you find the time 😁

BrandonSmith:
No problem. I've got a local StackNagivator with the fix. I'll get a PR with the needed change.

PR #3546, opened 2018-02-19
Fix react-native-web support for #3526

raarts:
https://github.com/react-navigation/react-navigation/pull/3526 introduced use of MaskedViewIOS component. I created a stub in PlatformHelpers.web.js for it, and inverted a Platform test in the code.

**Test plan (required)**

No UI changes, code passes tests.

### DIFF
--- a/src/PlatformHelpers.web.js
+++ b/src/PlatformHelpers.web.js
@@ -1,8 +1,8 @@
 import React from 'react';
-import { BackHandler, Text } from 'react-native';
+import { BackHandler, View } from 'react-native';
 
 const MaskedViewIOS = () => {
-  <Text>MaskedViewIOS not available on Reacht Native Web</Text>;
+  <View>{this.props.children}</View>;
 };
 
 export { BackHandler, MaskedViewIOS };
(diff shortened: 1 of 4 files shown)


src/PlatformHelpers.web.js
```diff
@@ -1,3 +1,8 @@
-import { BackHandler } from 'react-native';
+import React from 'react';
+import { BackHandler, Text } from 'react-native';
 
-export { BackHandler };
+const MaskedViewIOS = () => {
+  <Text>MaskedViewIOS not available on Reacht Native Web</Text>;
```
ericvicenti: Is this the correct fallback behavior? Shouldn’t we just let masking silently fail, so that the stack header doesn’t get messed up with this text?

raarts: Fixed.

src/PlatformHelpers.web.js
```diff
@@ -1,3 +1,8 @@
-import { BackHandler } from 'react-native';
+import React from 'react';
+import { BackHandler, View } from 'react-native';
 
-export { BackHandler };
+const MaskedViewIOS = () => {
+  <View>{this.props.children}</View>;
```
satya164: This should return the element. Change it to:

```js
const MaskedViewIOS = () => <View>{this.props.children}</View>;
```

PR #3556, opened 2018-02-20
Improve `Header` performance a bit

grabbou:
I have been investigating `<Header />` component performance once again today with `render` cycles in particular. I have observed that during `push` phase, a typical Header in NavigationPlayground re-renders 3 times on iOS and 2 times on Android.

The first time is obvious on both platforms since that's when we add a new scene to an array. Second on iOS was also self-explanatory - we measure title width and other layout parameters in order to provide a better animation. 

What got me thinking was the last render cycle that didn't have a clear origin. After digging around I've found it is caused by a `scenes` array changing when transition finishes. What is surprising, it's just the reference to the object that changes, the content (items inside) stay the same.

I have found out that this is caused by `Transitioner` looping through an array of scenes attempting to remove stale scenes. Since scene becomes stale when you `pop` from it, this is obviously a noop when `pushing` a new route. That, obviously, causes an extra render cycle since `filter` produces same, but with a different pointer, object.

### DIFF
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -190,9 +190,7 @@ class Transitioner extends React.Component {
        * pushing a new route). As a result, components that rely on this prop
        * might enter an unnecessary render cycle.
        */
-      scenes: this.state.scenes.length === scenes.length ?
-        this.state.scenes
-        : scenes,
+      scenes: this.state.scenes.length === scenes.length ? this.state.scenes : scenes,
     };
 
     this._transitionProps = buildTransitionProps(this.props, nextState);
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -185,12 +185,13 @@ class Transitioner extends React.Component {
       ...this.state,
       /**
        * Array.prototype.filter creates a new instance of an array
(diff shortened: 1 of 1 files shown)


grabbou:
Another idea for improving Header performance and its general maintenance from a future perspective is to split it into two separate implementations, one for iOS and the other one for Android.

The reason for that architectural move is that these two platforms have significantly different Headers with different behaviours (on iOS, header can be shared across many different screens and have a complex animation between multiple elements) whereas on Android, it's simple `flex` layout in most cases.

Having two separate implementations would not only make it easier to optimise each of the platforms in a more efficient way but also work on iOS animations w/o being worried that Android might get affected by a mistake.

Not sure if that's RFC candidate or I can just keep working in that space, I'll be awaiting feedback. @ericvicenti @brentvatne

brentvatne:
thanks! :D

brentvatne:
> Having two separate implementations would not only make it easier to optimise each of the platforms in a more efficient way but also work on iOS animations w/o being worried that Android might get affected by a mistake.

Agreed, if they are significantly different in terms of implementation we should split them into separate components.

PR #3562, opened 2018-02-21
First pass at default params object

peterpme:
A start to the conversation that took place here
https://github.com/react-navigation/rfcs/pull/21


ericvicenti:
This looks like a fix to the action creator, but the navigation state is defined by StackRouter.getStateForAction. I think we'd probably want to fix it there. In case somebody creates actions by hand, we still want to make sure the params object is there

peterpme:
I knew I was missing something! Thank you for looking this over 😄

PR #3583, opened 2018-02-23
fix typespec of `init` action creator

yordis:
The `init` action creator doesn't not required `payload` parameter.

https://github.com/react-navigation/react-navigation/blob/2bb91a6740e2f1ade012906dc03c972ba6375cc3/src/NavigationActions.js#L24

Notice the `default` value


yordis:
cc: @Ashoat

Ashoat:
Please edit the name and description of this PR so people know what's going on.

brentvatne:
looks fine to me

Issue #3588, opened 2018-02-23
TabNavigator default double press behaviors


brentvatne:
Collapsing the following issues into one:
- [x] scroll to top - https://github.com/react-navigation/react-navigation/issues/2955
- [x] pop to top - https://github.com/react-navigation/react-navigation/issues/3570

brentvatne:
done: https://react-navigation.canny.io/feature-requests/p/scroll-to-top-of-current-tab-when-it-is-pressed-again-in-tabnavigator

FrenchMajesty:
Following up, It would be nice to see something for "Pop to Top" as well.

brentvatne:
@FrenchMajesty - that happens automatically if you have a stack within a tab

octodhruv:
> @FrenchMajesty - that happens automatically if you have a stack within bottom tabs

@brentvatne could you please expand on this? Or perhaps show a working example? That would be super helpful :)

farhadeidi:
I’ve created a gist to solve this problem:

https://gist.github.com/farhadeidi/d9f3af865d4f58d9e5c0c9dc4cc4ee6d

PR #3589, opened 2018-02-24
Pop to top on TabNavigator when tapping already focused tab icon

brentvatne:
# Motivation

The expected behavior for tab bars is that when you have a stack within them and tab the icon for the active tab, then it should pop that stack to the top. This adds that behavior, partially solving https://github.com/react-navigation/react-navigation/issues/3588. The next step is to scroll a scrollview to top if the user taps again after popping to top.

# Test plan

Try out navigation playground Stacks in Tabs.

# Concerns

- ~I don't think this is actually a shippable solution in its current form due to the following edge case: if there are no stacks within the TabNavigator, but the TabNavigator is nested within another StackNavigator, then the pop to top action would be handled by the parent. I think we need to change `popToTop` to optionally accept a key.~ &mdash; I updated popToTop to accept a key and StackRouter to respect it.
- Maybe there's too much router logic in the view?
- Should there be an option to disable this? Or should people just override the onTabPress (or whatever) option and make it call jumpToIndex directly?

### DIFF
--- a/src/NavigationActions.js
+++ b/src/NavigationActions.js
@@ -57,6 +57,7 @@ const pop = createAction(POP, payload => ({
 const popToTop = createAction(POP_TO_TOP, payload => ({
   type: POP_TO_TOP,
   immediate: payload && payload.immediate,
+  key: payload && payload.key,
 }));
 
 const push = createAction(PUSH, payload => {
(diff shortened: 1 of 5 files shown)


JoseVf:
There's a way to disable this behavior or any workaround to avoid this?

brentvatne:
I would recommend against it because that's the expected behavior, but you can do it by providing `tabBarOnPress` to `TabNavigator` `tabBarOnPress={({jumpToIndex, scene}) => jumpToIndex(scene.index)}`

JoseVf:
I don't want it just because I'm making like a checklist app, and if the user taps on the focused tab again by accident will loose all the info that he had... This is a great improvement 😄 👍, but also will be great to have an option to disable it taking in count that specific cases which isn't useful at all 👌

brentvatne:
cool. yeah you can disable it using the technique i mentioned above

PR #3593, opened 2018-02-24
Fix bug #3279 “Action within RESET doesn't trigger”

serhiipalash:
https://github.com/react-navigation/react-navigation/issues/3279

If you pass nested action within reset actions it has no effect.

| software         | version
| ---------------- | -------
| react-navigation | 1.1.2
| react-native     | 0.52

screen shot 2018-02-24 at 12 20 38 pm

Bug examples:
snack: https://snack.expo.io/H11t9nRvz
repo: https://github.com/serhiipalash/react-navigation-bug-3279-example

Bug screenshots:
when you press "Reset with nested action"

screen shot 2018-02-24 at 12 26 01 pm

you see new stack like there was no nested action

screen shot 2018-02-24 at 12 23 44 pm

Expected behaviour:
If I pass nested action within reset actions it should trigger changes in child router.

screen shot 2018-02-24 at 12 34 56 pm

Fix:

screen shot 2018-02-24 at 12 31 22 pm

Go to `fix-3279` branch in bug example repo to test this fix
https://github.com/serhiipalash/react-navigation-bug-3279-example/tree/fix-3279

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -386,21 +386,28 @@ export default (routeConfigs, stackConfig = {}) => {
           // undefined on either the state or the action
           return state;
         }
-        const resetAction = action;
+        const newStackActions = action.actions;
 
         return {
           ...state,
-          routes: resetAction.actions.map(childAction => {
-            const router = childRouters[childAction.routeName];
+          routes: newStackActions.map(newStackAction => {
+            const router = childRouters[newStackAction.routeName];
+
             let childState = {};
+
             if (router) {
+              const childAction =
+                newStackAction.action ||
+                NavigationActions.init({ params: newStackAction.params });
+
               childState = router.getStateForAction(childAction);
             }
+
             return {
-              params: childAction.params,
+              params: newStackAction.params,
               ...childState,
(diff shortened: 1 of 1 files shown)


brentvatne:
- can you add test coverage for this?
- can you add an example to navigation playground that demonstrates this?

thanks!

serhiipalash:
> can you add an example to navigation playground that demonstrates this?

@brentvatne do you have any instructions how to do it?

Isn't snack example not enough? 

> Bug examples:
snack: https://snack.expo.io/H11t9nRvz

ericvicenti:
I do think we should have some tests for this, and also add it to the navigation playground so that we can test this ourselves in the future. But I think this fix is good, so I'm happy to land now and do those as followups.

@serhiipalash, can you send another PR, with a unit test for this reset bug in `StackRouter-test`, and also help add your example code to the navigation playground? (This will encourage other maintainers to test the functionality before making any big changes to the library.)

serhiipalash:
> can you send another PR, with a unit test for this reset bug

@ericvicenti  I'll try to do it on this weekend.

PR #3594, opened 2018-02-24
Fix RFC link

jozan:
As the title says. Thanks.


satya164:
Thank you for the PR!

PR #3603, opened 2018-02-26
Add separate type for addNavigationHelpers options

timwangdev:
The `addNavigationHelpers` option definition doesn't share the same signature as `NavigationProp<S>`. It now updated according to https://reactnavigation.org/docs/redux-integration.html use cases and addNavigationHelpers.js source.

Related PR for DefinitelyTyped: [DefinitelyTyped/DefinitelyTyped#23916](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23916)

**Test plan (required)**

Pass CI.


ericvicenti:
Hi @timwangdev, I'm excited about your contribution, but this isn't a great time to fix this. We're hopefully about to kill `addNavigationHelpers` in the next major release:

https://github.com/react-navigation/rfcs/pull/18

ericvicenti:
We are already planning on deleting addNavigationHelpers in the next major version, so I don't think we will need this.

https://github.com/react-navigation/react-navigation/pull/3619

@timwangdev, I'd love your help on correctly flow-typing the new actions API. Its a bit more complicated than the old addNavigationHelpers, but it should be much more flexible in the future

PR #3618, opened 2018-02-28
Drawer Router

ericvicenti:
Previously the drawer was a confusing configuration of two tab routers. This introduces a new and simple Drawer router


brentvatne:
ahhh this looks so much better!

one that may arise is that, as far as i know, there's no 'built-in' way to deep link to a screen with an open drawer. this feels like an edge case that almost nobody would care about but i've been wrong before

ericvicenti:
Yeah, I can see how that use case might come in handy. We could definitely augment the router's getActionForPathAndParams, so that a certain param (or path) causes the drawer to be open.

browniefed:
Once upon a time this was a thing in netflix.
If you navigated to a non-drawer item from the drawer then came back the drawer would still be open.
Not sure if this is actually possible in react-navigation?

Issue #3184, opened 2017-12-25
StackNavigator Header adds SafeAreaView inset padding even when nested


davidgraves:
Expected behavior:
If StackNavigator is rendering a header at the top of the screen, it should add the appropriate inset for the status bar or iphone X safe area. If it is rendering a header in a view that is already inset from the top of the screen it should NOT add the inset padding.

Actual behavior:
Header.js hard-codes <SafeAreaView forceInset={{top:'always'}} />. This is appropriate only when the StackNavigator is at the top of the screen. In all other cases it is incorrectly adding 20 or 44 pixels (iphone or iphone X) of padding to the header.

Steps to Reproduce:
Embed a StackNavigator in a view that is already protected from the safe area. Try to prevent your navigation bar from being too large by 20 or 44 points of wasted padding.

SafeAreaView already knows if it is at the top of the screen. But Header.js specifies "forceInset: {top: 'always'}".  If SafeAreaView already handles this case, Header.js should not override its behavior. If SafeAreaView does not correctly handle this case, it should be fixed.

kelset:
I think it's duplicate of one of these: https://github.com/react-navigation/react-navigation/issues/3173 or https://github.com/react-navigation/react-navigation/issues/3072

Please check them and if so, close this and comment on the other one.

davidgraves:
This is a duplicate. closing.

attitude:
@davidgraves, this is still valid. It's not a duplicate of any of the mentioned issues and should not be closed. `header.js` introduces top padding even when nested in other View that doesn't touch top screen edge.

bramus:
What @attitude says: this is not a duplicate.

bramus:
Until this gets reopened, because it really is no duplicate, you can bypass it by manually overriding the `containerStyle` in the `transitionConfig` of the **nested** navigator:

```
const routeConfigs = { … };
const NestedNavigator = StackNavigator(routeConfigs, {
	headerMode: 'screen',
	transitionConfig: () => ({
		containerStyle: {
			marginTop: -40,
		}
	}),
});
```

(Of course only set this conditionally: when it's an iPhoneX (see [`react-native-iphone-x-helper`](https://github.com/ptelad/react-native-iphone-x-helper) and when the device is in portrait mode)

B3rry:
@davidgraves, I agree, with @attitude and @bramus, this is not a duplicate.

See this view with a marginTop of '35' for an example:
[...]

PR #3633, opened 2018-03-01
Remove hard-coded forceInset from `SafeAreaView` in `Header`

B3rry:
Since `SafeAreaView` is contex aware with regards to its position, hard coding `forceInset` will cause unremovable padding in some insances.

# Context
[`react-community/react-native-safe-area-view`](https://github.com/react-community/react-native-safe-area-view)'s`SafeAreaView` determines the View's location in relation to the device's viewport in order to determine wether or not to add padding based upon `touchesTop`, … `touchesRight` within [`measureInWindow`](https://github.com/react-community/react-native-safe-area-view/blob/b84d241a8bc215ef956ff3b3a56875e2a04667c5/index.js#L152).

# Motivation
Given this context, `Header.js`'s hard-coding of a `forceInset` `top` value of `always` will cause `SafeAreaView`'s padding to be applied to all instances of `Header`. This leads to additional padding being added to instances of Header, even if they have an offset applied, such as `marginTop`.

While [#3184](https://github.com/react-navigation/react-navigation/issues/3184) was flagged as a duplicate and closed (I believe in error), this PR attempts to fix this rendering problem.

# Test Plan
This bug can be replicated by adding a `marginTop` value of > 0 to `navigationOption`'s `headerStyle`. Expected behavior would be the removal of `SafeAreaView`'s top padding. Encountered behavior is the application of `SafeAreaView`'s top padding, even overriding instances where a different padding is applied. This breaks from the anticipated behavior of inheriting padding when supplied, as mentioned in the comments of PR [#2889](https://github.com/react-navigation/react-navigation/pull/2889). Once applied, `paddingTop` values supplied to `headerStyles` for screens that do not touch the top of the viewport should be properly inherited.

## Bugged View Before Modification: (MarginTop of 36px applied)
screen shot 2018-03-01 at 1 38 36 pm

## Bugged View After Modification: (MarginTop of 36px applied)
screen shot 2018-03-01 at 2 27 43 pm

## Control View Before Modification: (No MarginTop applied)
screen shot 2018-03-01 at 2 33 25 pm

## Control View After Modification: (No MarginTop applied)
screen shot 2018-03-01 at 2 31 42 pm

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -474,9 +474,7 @@ class Header extends React.PureComponent {
     ];
 
     return (
-      <SafeAreaView
-        style={containerStyles}
-      >
+      <SafeAreaView style={containerStyles}>
         <View style={StyleSheet.absoluteFill}>{options.headerBackground}</View>
         <View style={{ flex: 1 }}>{appBar}</View>
       </SafeAreaView>
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -475,7 +475,6 @@ class Header extends React.PureComponent {
 
     return (
       <SafeAreaView
-        forceInset={{ top: 'always', bottom: 'never' }}
         style={containerStyles}
       >
         <View style={StyleSheet.absoluteFill}>{options.headerBackground}</View>


brentvatne:
hi there! thanks for the PR! this is handled in https://github.com/react-navigation/react-navigation/pull/3496 in a way that lets you instead customize the insets, which i prefer over removing the forced insets entirely due to some limitations with the existing safeareaview implementation. in particular, i believe that the change you propose here will cause undesirable behavior when you transition a view from bottom to top (eg: showing a modal with that type of transition).

B3rry:
@brentvatne thanks for the response. I’m still believe the default should not be hard coded. If the intent of SafeAreaView is to be context aware, why remove that pattern of behavior in the default implementation?

brentvatne:
@B3rry - I think we could do something clever like collapse the margin into the SafeAreaView inset [similar to what we do with padding](https://github.com/react-navigation/react-navigation/commit/a62ad18b3192be4d386a25fcd9dcabf3fbcf318d). perhaps the inset padding could be `paddingTop = Math.max(currentPaddingTopValue - marginTopOrVertical, 0)` 
- `marginTopOrVertical` because we should support `marginVertical` which has less precedence than `marginTop`
- we'd need to work out how this behaves if the user provides both a custom `paddingTop` and `marginTop`

PR #3634, opened 2018-03-01
Add SwitchNavigator

brentvatne:
# Motivation

Authentication flows are kind of difficult to implement currently, and given that 1) they're something that just about every app has and 2) they're usually built early on in the process, we need to make auth flows super easy to implement. This PR introduces SwitchNavigator, which allows you to easily jump from one screen to another, and it unmounts the previous screen and resets its route, and has no "back" behavior by default. This is all done just by calling `navigate`. I've added an example for how to do this to the NavigationPlayground.

This PR is intended to solve the problem immediately for 1.x users. I will open another PR with this rebased against master (now home to changes that won't land until 2.0) and with further more radical changes including making TabRouter and DrawerRouter composed of a SwitchRouter (as switch is the more generic concept).

# Test plan

- Try NavigationPlayground
- [x] Add test coverage for `resetOnBlur` behavior of `SwitchRouter`, which is the only novel addition on top of the `TabRouter` base that I copy and pasted (as mentioned above, this copy and paste job will be cleaned up for 2.x)

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -25,6 +25,7 @@ import MultipleDrawer from './MultipleDrawer';
 import TabsInDrawer from './TabsInDrawer';
 import ModalStack from './ModalStack';
 import StacksInTabs from './StacksInTabs';
+import SwitchWithStacks from './SwitchWithStacks';
 import StacksOverTabs from './StacksOverTabs';
 import StacksWithKeys from './StacksWithKeys';
 import SimpleStack from './SimpleStack';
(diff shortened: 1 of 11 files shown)


ScreamZ:
Best PR of the project :D

Thanks

alvelig:
> making TabRouter and DrawerRouter composed of a SwitchRouter (as switch is the more generic concept).
👍

brentvatne:
@alvelig - done!

https://github.com/react-navigation/react-navigation/blob/bdda6fa5be2cbdef2a1f4f5b3a65fa4518003c27/src/routers/TabRouter.js
https://github.com/react-navigation/react-navigation/blob/bdda6fa5be2cbdef2a1f4f5b3a65fa4518003c27/src/routers/DrawerRouter.js

alvelig:
That was very-very quick LOL

Issue #2082, opened 2017-07-06
How to change active tab style in TabNavigator? 


bhargavjoshi:
Right now the library provides way to provide labelStyle
 
```
labelStyle: {
    fontWeight: "bold"
}
```

is there any way it supports activeTab styling?

ixje:
The TabNavigatorConfig supports [tabBarOptions](https://reactnavigation.org/docs/navigators/tab#tabBarOptions-for-TabBarBottom-default-tab-bar-on-iOS) like `activeTintColor` and `activeBackgroundColor`

Another way is via e.g. `tabBarIcon` of `navigationOptions` like so: 
```
		navigationOptions: {
			tabBarLabel: 'Give Feedback',
			tabBarIcon: ({ tintColor, focused }) => (
				<FAIcons
					name={focused ? 'comment' : 'comment-o'}
					size={24}
					style={{ color: tintColor }}
				/>
			),
		},
```
you can use the `focused` for conditional styles as well

spencercarli:
Hi! In an effort to get the hundreds of issues on this repo under control I'm closing issues tagged as questions. Please don't take this personally - it's simply something we need to do to get the issues to a manageable point. If you still have a question I encourage you to re-read the docs, ask on StackOverflow, or ask on the #react-navigation Reactiflux channel. Thank you!

hellenzhou:
_getTopBar = () => {
        let tops = {};
        this.topParams.map((item, index) => {
            tops[`tab${index}`] = {
                screen: props => <TaskContent labelItem={item} {...props} />,
                navigationOptions: {

                    tabBarLabel: ({ tintColor, focused }) =>
                        focused ? (<Text style={{ fontWeight: 'bold' }} >{item}</Text>) :
                            (<Text style={{ fontWeight: 'normal' }} >{item}</Text>),
                }
            }
        })
        return tops;
    }

    render() {
        const TabNavigator = createAppContainer(createMaterialTopTabNavigator(this._getTopBar(),
            {
                'lazy': false,

                initialLayout: {
                    height: 0,
                    width: Dimensions.get('window').width,
                },

                tabBarOptions: {
                    pressOpacity: 0,
                    tabStyle: { ...styles.tabStyle },
                    activeTintColor: "red",

                    labelStyle: { ...styles.labelStyle },
                    // showIcon: true,  
                    // activeTabStyle: {
                    //     fontWeight: 'bold',
                    //     backgroundColor: 'red',
                    // },

[...]

PR #3635, opened 2018-03-01
Adding activeLabelStyle prop

julesmoretti:
This is to create the ability to override default appearance like regular font vs bold when using the TabBarTop component. 

Issue #2082

thank you.

I was trying to make the active label bold compared to others while using the TabBarTop, I could have created some custom tabBarIcon but that would have been the wrong place for it.

**Test plan (required)**

Could not find method to do this. edited it on my end and it works.

The code must pass tests.

**Code formatting**

### DIFF
--- a/src/views/TabView/TabBarTop.js
+++ b/src/views/TabView/TabBarTop.js
@@ -22,6 +22,7 @@ export default class TabBarTop extends React.PureComponent {
       inactiveTintColor,
       showLabel,
       upperCaseLabel,
+      activeLabelStyle,
       labelStyle,
       allowFontScaling,
     } = this.props;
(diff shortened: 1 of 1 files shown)


julesmoretti:
@satya164 - Thank you for your response. To me, the confusion is in the component section naming. 

This is the way I visually break down the TabNavigator. So can you help shine some light on what 1 to 5 are called in the styling?

screen shot 2018-03-02 at 10 53 51 am

satya164:
> So can you help shine some light on what 1 to 5 are called in the styling

The problem is not naming. The problem is that there are multiple states between the active and inactive state while it's animating. Having styles for active and inactive state leaves out all the intermediate states.

julesmoretti:
Ok so what is your recommendation for having the font become bold when tab is active?

And what are the various states?

julesmoretti:
@satya164 ?

ericvicenti:
This project is moving over to https://github.com/react-navigation/react-navigation-tabs

Feel free to experiment with this feature in that repo.. I think it should be possible to render both labels and fade the opacity between them as the route becomes focused.

PR #3645, opened 2018-03-02
More specific injected Navigation props in Flow definition

EdwardDrapkin:
Fixed the navigation props type to be a tad more specific since it leaving it as just NavigationState created a seemingly-impossible-to-resolve situation with type parameters.


Ashoat:
Can you explain a bit more about the conflicting types issue? It seems like if you use `withNavigation` with a non-route ‘NavigationState` (ie. route container state) set as the context variable `navigation` then this refined type would be incorrect. Is this just not something that can happen?

Ashoat:
Actually, this makes sense. `withNavigation` will always pull `SceneView`'s `navigation` context prop, which will always correspond to a specific route, and thus be a `NavigationStateRoute`.

Issue #3636, opened 2018-03-02
Frequently call navigation.setParams, memory is getting higher and higher.


helele90:
```
import React, { Component } from 'react';
import { View, StyleSheet, AppRegistry, ScrollView } from 'react-native';
import { StackNavigator } from "react-navigation"; // 1.0.0

class Page extends Component {

  static navigationOptions = ({navigation}) => ({
    title: `Page1`
  });

  render() {
    return (
      <View style={styles.container}>
        <ScrollView style={{flex: 1}} scrollEventThrottle={16} onScroll={this.onScroll} >
          <View style={{backgroundColor: 'red', height: 2000}} />
        </ScrollView>
      </View>
    );
  }

  onScroll = () => {
    console.debug('onScroll');
    this.props.navigation.setParams({});
  }

}

const App = StackNavigator({
    Page: {
    screen: Page,
  },
});

const styles = StyleSheet.create({
  container: {
    flex: 1
  },
});

export default App;
AppRegistry.registerComponent("hello", () => App);
```

### Current Behavior

- What code are you running and what is happening?
Every time I call navigation.setParams, the memory gets higher.
Memory grows faster in my project.
- Include a screenshot if it makes sense.
Application Start
<img width="622" alt="2018-03-02 09 49 12" src="https://user-images.githubusercontent.com/4125469/36879733-28476e74-1e00-11e8-8d57-ff2d2bcb5783.png">
Call thousands navigation.setParams({})
<img width="488" alt="2018-03-02 09 54 02" src="https://user-images.githubusercontent.com/4125469/36879764-42c6e6bc-1e00-11e8-868c-c532fea04675.png">
A few minutes later, not call navigation.setParams({})
<img width="490" alt="2018-03-02 10 00 54" src="https://user-images.githubusercontent.com/4125469/36879846-a2819714-1e00-11e8-905e-c693efe58da1.png">
### Expected Behavior

- What do you expect should be happening?
- Include a screenshot if it makes sense.

### How to reproduce

use my example code

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.2.1
| react-native     | 50.1
| node             | 8
| npm or yarn      | 5.4

JulianKingman:
I'm having a similar problem, maybe some sort of memory leak? My memory usage went as high as 15gb 😮

brentvatne:
this was resolved in latest version. see referenced prs such as https://github.com/react-navigation/react-navigation/pull/3649

[...]

PR #3649, opened 2018-03-03
Cache child event subscriptions (1.x branch)

brentvatne:
# Motivation

I tracked down the memory leak reported in https://github.com/react-navigation/react-navigation/issues/3636 using Chrome memory tab and found that it was a result of creating new subscribers anytime a `StackNavigator` state changes. ~This PR only fixes it in `StackNavigator`, it is likely also a problem with other navigators but I don't want to invest too much time into those because I've already fixed it for all of them on master (for the upcoming 2.0 release, see: https://github.com/react-navigation/react-navigation/pull/3648). It would be nice for someone to follow up a PR to apply these same changes to other navigators.~ I believe this also applies to other navigators and made the appropriate changes here. This PR does not clean up upstream subscribers when a navigator unmounts, however, but that is an existing issue with minimal impact and which is fixed in https://github.com/react-navigation/react-navigation/pull/3648.

# Test plan

Try running the example code from https://github.com/react-navigation/react-navigation/issues/3636 with this branch and you will see that memory usage does not simply climb forever as you scroll up and down.

### DIFF
--- a/src/getChildEventSubscriber.js
+++ b/src/getChildEventSubscriber.js
@@ -4,6 +4,7 @@
  * Based on the 'action' events that get fired for this navigation state, this utility will fire
  * focus and blur events for this child
  */
+
 export default function getChildEventSubscriber(addListener, key) {
   const actionSubscribers = new Set();
   const willFocusSubscribers = new Set();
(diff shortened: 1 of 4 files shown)


brentvatne:
With updating params every 1ms, before:

screen shot 2018-03-02 at 6 08 53 pm

After:

screen shot 2018-03-02 at 6 09 38 pm

PR #3650, opened 2018-03-03
Fix: tabBar icons are not visible.

jainkuniya:
**Motivation** for making this change. 

Icons in tabBar are not visible. So I took this as challenge and find this quick fix. :)
In our project we have updated to `"react-navigation": "1.3.2"` and facing this issue, so I thought of fixing it in the library itself. 
Let me know changes :)

**Test plan (required)**

Clone this https://github.com/jainkuniya/sample-tab-navigation.git and run `yarn install` or `npm install`, and run app.

With 
```
    "react": "16.2.0",
    "react-native": "^0.54.0",
    "react-native-vector-icons": "^4.5.0",
    "react-navigation": "1.3.2"
```

app will look like this
screen shot 2018-03-03 at 5 41 25 pm

icons are not visible

**On this PR**
screen shot 2018-03-03 at 5 42 36 pm

### DIFF
--- a/src/views/TabView/TabBarIcon.js
+++ b/src/views/TabView/TabBarIcon.js
@@ -53,12 +53,11 @@ const styles = StyleSheet.create({
     // We render the icon twice at the same position on top of each other:
     // active and inactive one, so we can fade between them:
     // Cover the whole iconContainer:
-    position: 'absolute',
-    top: 0,
-    left: 0,
-    right: 0,
-    bottom: 0,
     alignItems: 'center',
+    alignSelf: 'center',
+    height: '100%',
     justifyContent: 'center',
+    position: 'absolute',
+    width: '100%',
   },
 });


borisyankov:
I can confirm this issue with React Native 0.54
Good job!

brentvatne:
I think this is caused by a regression with react-native layout in 0.54: https://snack.expo.io/BymTJkcOM -- if you copy this code and run it locally on react-native 0.54 you see different results.

it seems like children of absolutely positioned views need explicit width and height set. this is inconsistent with web -- i tried this example with react-native-web and it worked

screen shot 2018-03-04 at 1 14 31 pm

we'll need to create an upstream issue for this, but in the meantime this PR does solve the problem.

mrharel:
when will this be released?

brentvatne:
this evening

franckmaurin:
Is this fix applied to the v2 too ? Icons are not visible.

brentvatne:
@franckmaurin - please create an issue following template if you are observing that behavior

PR #3651, opened 2018-03-03
Add isFocused helper to navigation and fix withNavigationFocus (1.x branch)

brentvatne:
# Motivation

`withNavigationFocus` didn't initialize with the correct state, it always assumed not focused. This PR adds a helper to get the focused state and uses that to properly initialize the state in `withNavigationFocus`. More discussion in https://github.com/react-navigation/react-navigation/pull/3512

In a follow-up PR I'll port this to master (the 2.0-track branch), which should be a good amount less code thanks to only adding navigation helpers in one place.

# Test plan

Run playground example. Tests pass.

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -114,10 +114,10 @@ const ExampleInfo = {
     name: 'Animated Tabs Example',
     description: 'Tab transitions have custom animations',
   },
-  // TabsWithNavigationFocus: {
-  //   name: 'withNavigationFocus',
-  //   description: 'Receive the focus prop to know when a screen is focused',
-  // },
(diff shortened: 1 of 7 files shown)


brentvatne:
cc @slorber

AndrewIngram:
@brentvatne 

This causes something to break that works in 1.3.2. I'm using `withNavigationFocus` around a screen component and get this error in the constructor, the error comes from the `isFocused()` call, but I haven't been able to dig further than that (the stack trace isn't useful, unfortunately):

simulator screen shot - iphone x - 2018-03-06 at 13 30 00

brentvatne:
@AndrewIngram - are you using redux? please create a new issue with minimal repro

PR #3659, opened 2018-03-05
Fix typespec of back action creator

yordis:
https://github.com/react-navigation/react-navigation/blob/c6301abaed51ef7ac2aa370ed18d66e39ca9e15d/src/NavigationActions.js#L21 doesn't required `params`


yordis:
@Ashoat I need this for a PR on the redux helper repo

cc: @ericvicenti

PR #3668, opened 2018-03-05
feat(context): refactor passing navigation context

### LINKED ISSUE #3524: Header buttons in root navigation stack don't receive navigation context
### Current Behavior
Currently if you are building a custom element to control navigation from a header button, IF you attempt to use the withNavigation HOC on that custom element, the navigation property is undefined.

Here is a brief example :

```
// navigation.js
StackNavigator({
	'Profile': {
		key: 'profile',
		navigationOptions: {
			headerRight: ( <CustomElement /> )
		}
});

// CustomElement.js
....
export default withNavigation(CustomElement);
```

### Expected Behavior

- Would expect the context in the withNavigation HOC pass in the navigator prop to my custom element.  

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.0.3
| react-native     | 0.53
| node             | 8.6.0
| npm or yarn      | 1.3.2

thymikee:
**Summary**

This is a refactor proposal for passing `navigation` through context, using the new API through `create-react-context`. 
The change is motivated by Header being rendered outside SceneView (which was a context provider for `navigation`), not getting the `navigation` prop when it was rendered as a root view, even though wrapped with `withNavigation` HOC.

Fixes #3524

**Test plan**

I'll yet need to come up with proper testing for this.

cc @satya164 as it was his idea for the context overhaul and I'm sure he has more background on the topic.


brentvatne:
hi @thymikee! can you rebase against master? I pushed a change to add `isFocused` to master (had it on 1.x branch already)

brentvatne:
waiting for review from @satya164

satya164:
I'll review today

thymikee:
@brentvatne what do you think, should we go in direction proposed by @satya164?

brentvatne:
yeah that sounds good to me

brentvatne:
@thymikee - ping :) would love to get this in for 2.0

thymikee:
I've rebased and added a test. I think we can implement what @satya164 said as a followup, as I'm not 100% sure how to approach this and if I actually understand the problem correctly.

src/views/withNavigationFocus.js
```diff
@@ -36,25 +39,17 @@ export default function withNavigationFocus(Component) {
...
-        'withNavigationFocus can only be used on a view hierarchy of a navigator. The wrapped component is unable to get access to navigation from props or context.'
-      );
-      return navigation;
-    };
-
     render() {
       return (
         <Component
           {...this.props}
+          navigation={this.props.navigation}
```
satya164: This should not be needed

src/views/Drawer/DrawerScreen.js
```diff
@@ -12,11 +13,9 @@ class DrawerScreen extends React.PureComponent {
     const descriptor = descriptors[routes[index].key];
     const Content = descriptor.getComponent();
     return (
-      <SceneView
-        screenProps={screenProps}
-        component={Content}
-        navigation={descriptor.navigation}
-      />
+      <NavigationProvider value={descriptor.navigation}>
```
satya164: Will be nice to move this wrapper to `SceneView` to avoid repetition

src/views/withNavigationFocus.js
```diff
@@ -2,26 +2,29 @@ import React from 'react';
...
 export default function withNavigationFocus(Component) {
   class ComponentWithNavigationFocus extends React.Component {
     static displayName = `withNavigationFocus(${Component.displayName ||
       Component.name})`;
 
-    static contextTypes = {
-      navigation: propTypes.object.isRequired,
-    };
-
     constructor(props, context) {
```
satya164: There shouldn't be a `context` argument anymore.

satya164:
Also need to address:

> For header and tab, we probably want to wrap individual items such as headerLeft, headerRight, tabBarIcon etc. with the HOC to provide the same navigation prop as the screen.

thymikee:
>For header and tab, we probably want to wrap individual items such as headerLeft, headerRight, tabBarIcon etc. with the HOC to provide the same navigation prop as the screen.

Do you say they should be wrapped with `withNavigation` by default, so one don't have to do it anymore?

satya164:
> Do you say they should be wrapped with withNavigation by default, so one don't have to do it anymore?

With NavigationProvider so correct navigation is available

satya164:
@brentvatne let's merge it now since it's an improvement over previous code anyway (support for React 16.3.1 due to new context API). we can address other stuff later.

Issue #686, opened 2017-03-14
How to dismiss modal containing a StackNavigator


davideweaver:
I have a few **StackNavigators** with the root having `mode: "modal"`...

```
const AddStack = StackNavigator({
  Add1: { screen: Add1Screen },
  Add2: { screen: Add2Screen },
  Add3: { screen: Add3Screen }
}, {
  mode: "screen"
});

const Root StackNavigator({
  MainStack: { screen: MainStack },
  AddStack: { screen: AddStack }
}, {
  mode: "modal"
})
```

The `AddStack` modal is a 3-step series of screens. When the user is on the third screen (`AddScreen3`) I would like them to click Save and have the modal go away (and be back on `MainStack`).

It looks like the right way to do it would be using **NavigationActions.reset**, but I can't figure it out.

Can anyone shed some light on the right way to do it?

davideweaver:
After looking into this more, I can achieve the results I want by swiping down on `Add3Screen`. However I need to be able to do it with a button click. Swiping down dispatches the following...

```
action Object {type: "Navigation/BACK", key: "id-1489519010411-0"}
```

It looks like that key is being generated automatically. Is there a way to assign or lookup the key so I can use it programmatically?

davideweaver:
I've figured out a solution until I find a better one. The solution is based on the previous comment about dispatching a "Navigation/BACK" action with the key of the modal StackNavigator screen. I use the routeName to find the key, then dispatch the action. In my case the routeName is `AddStack`.

**navActions.js** (partial)
```
export function dismissModal(modalRouteName) {
  return async (dispatch, getState) => {
    var modalKey = findRouteKey(getState(), modalRouteName);
    if (modalKey) {
      dispatch(NavigationActions.back({key: modalKey}));
    }
  }
}
```

This goes in my screen component...
```
dispatch(navActions.dismissModal("AddStack")
```
see the full **navActions.js** at [https://gist.github.com/davideweaver/34d701c402c199676916d471916f9dca](https://gist.github.com/davideweaver/34d701c402c199676916d471916f9dca)

Maybe we can add a helper like this to **NavigationActions** in the future.

blaues0cke:
Thank you, @davideweaver. Your code helped me a lot to understand my very similiar problem. Actually my problem was that the modal was closed whenever I hit the "back" button inside the child `StackNavigator`.

[...]

PR #3669, opened 2018-03-06
Add dismiss helper, made possible by also adding carefullyGetParent

brentvatne:
# Motivation

Sometimes you want to be able to dismiss a nested stack, for example if you have a root stack that goes from your main app to a modal stack, you might want to go back to the main app when you're n screens deep in the modal stack. People came up with a pretty good solution for this in a previous discussion on the issues here: https://github.com/react-navigation/react-navigation/issues/686#issuecomment-292225313 - I just adapted this to `addNavigationHelpers`.

As for `navigation.carefullyGetParent`, we dicussed it on https://github.com/react-navigation/rfcs/issues/27.

# Test plan

Run tests, try it in playground.

### DIFF
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -49,7 +49,8 @@ class MyNavScreen extends React.Component<MyNavScreenProps> {
         />
         <Button onPress={() => navigation.popToTop()} title="Pop to top" />
         <Button onPress={() => navigation.pop()} title="Pop" />
-        <Button onPress={() => navigation.goBack(null)} title="Go back" />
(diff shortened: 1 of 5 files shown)


brentvatne:
@ericvicenti - agreed, will update this PR for something like `dangerouslyGetParent`. `dismiss` only makes sense in the contact of a stack I believe, but it's a bit trickier than that -- it only makes sense when the grandparent navigator is a stack: stack -> stack/tab/whatever -> screen

ScreamZ:
What do you think of https://github.com/react-navigation/react-navigation/issues/686#issuecomment-371075602 ?

brentvatne:
2.0 only yes. your solution there looks like a good one for now! :)

PR #3670, opened 2018-03-06
Flow type: SafeAreaView doesn't require children

leethree:
`children` prop should be optional.

See https://github.com/flowtype/flow-typed/commit/4ec24a5b93eb832b5879b239900174029d92d13b#commitcomment-27932697


davepack:
Actually yeah, I've used that for when I want just a simple spacer on the top or the bottom of the screen.

Issue #3629, opened 2018-03-01
Navigation options, unable to provide styling for custom header back image


spaceye:
Hi!

There's no way to provide style for custom Header Back image, `navigationOptions.headerBackImage` — think of width, height, margins, etc. The whole idea of custom image is pretty much useless without it.
We could add `navigationOptions.headerBackImageStyle` prop and pass it down to corresponding Header back button components as, say `iconStyle` or `imageStyle`. Just like we do with `navigationOptions.headerBackTitle` and supporting `navigationOptions.headerBackTitleStyle`.

* I'm filing this issue rather than making a feature request since I believe it's a bug in this context.

Thanks!

brentvatne:
`headerBackImageProps` is probably better in case you need to add other things.. but at this point, why even expose this when we you could just use `headerLeft` instead?

shyaniv7:
I used this login and it works great for me

```
navigationOptions: ( { navigation } ) => ( {
    headerStyle: {
      backgroundColor: colors.white
    },
    headerLeft: <BackButton navigation={navigation} />,
  } )
```

BackButton.js wrapping our custom Icon component
```
const BackButton = ( { navigation } ) => {
  const onPress = () => navigation.goBack();

  return (
    <Icon
      color={colors.black}
      containerStyle={styles.backButton}
      name="chevron-left"
      onPress={onPress}
      size={20}
    />
  );
};
```

spaceye:
@brentvatne, @shyaniv7 Thanks for your replies guys! However it would mean that I'll have to reimplement `HeaderBackButton` cause it's more than just an image there, correct? So in this case it's even more strange that we allow for custom image and at the same time force people to certain geometry in the icon's styling which beats the purpose, does it not? No offense just trying to better communicate my problem with it.

brentvatne:
@SpaceK33z - yeah there's not too much to it though https://github.com/react-navigation/react-navigation/blob/master/src/views/Header/HeaderBackButton.js

ikhilko:
@brentvatne - thank you for your response! But it really looks strange when you allow to pass custom icon/image for header back button, but it's has hardcoded style (geometry) - so `headerBackImage` prop becomes useless in this case. 
Maybe you can revisit the code of the `HeaderBackButton` component and allow for this new `imageStyle` property or custom Image component (with styles) as a property just like `headerLeft`.

brentvatne:
[...]

PR #3680, opened 2018-03-07
Reworked headerBackImage navigation prop and rendering

spaceye:
**Motivation**

Current headerBackImage navigation property allows for image customization but confines any such custom image to certain hardcoded geometry within corresponding components (i.e. `HeaderBackImage`).

I've prepared with PR according with the discussion that we've had in #3629.

**Test plan**

I've added a new demo suite for this prop, ran it in Expo for both platforms, changed documentation on this prop as while we leave it unchanged we do change its meaning (resource -> element / component).

**iOS**
screen shot 2018-03-07 at 5 01 43 pm
screen shot 2018-03-07 at 5 01 47 pm
screen shot 2018-03-07 at 5 01 49 pm

**Android**
screen shot 2018-03-07 at 5 05 31 pm
screen shot 2018-03-07 at 5 05 41 pm
screen shot 2018-03-07 at 5 05 44 pm


brentvatne:
thanks! this seems much better. could you possibly clean up the playground example to remove all of the listener code and just focus on the back button aspect of it? prefer not to add any unnecessary code to the example

spaceye:
@brentvatne, much obliged and yep, done so!

PR #3683, opened 2018-03-07
Allow passing null to onNavigationStateChange prop

Leeds-eBooks:
As seen here: react-navigation/react-navigation#360, we need to be able to pass null to turn off logging.

This PR mirrors https://github.com/flowtype/flow-typed/pull/1912, which will be overridden when this repo is merged into flow-typed.


Ashoat:
Thanks @Leeds-eBooks!

brentvatne:
> As seen here: react-navigation/react-navigation#360, we need to be able to pass null to turn off logging.

that's not true. it's disabled by default

PR #3690, opened 2018-03-08
fix: Dimensions.addEventListener & removeEventListener in old version…

Creemli:
… of react-native

Explain the motivation for making this change. What existing problem does the pull request solve?

Fix the error about Dismensions.addEventListener when use this package with react-native@0.42 and before. Since the peerDependencies defined in package.json are:

 "peerDependencies": {
    "react": "*",
    "react-native": "*"
  }
Test plan (required)

The code must pass tests.

Code formatting

### DIFF
--- a/src/views/withOrientation.js
+++ b/src/views/withOrientation.js
@@ -14,11 +14,13 @@ export default function(WrappedComponent) {
     }
 
     componentDidMount() {
-      Dimensions.addEventListener('change', this.handleOrientationChange);
+      Dimensions.addEventListener &&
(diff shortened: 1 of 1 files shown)


brentvatne:
why not update to a more recent version of react native? this 0.42 is nearly a year old at this point

brentvatne:
I think it's not worth supporting very old versions of react-native, perhaps we need to be explicit about the version that we support. Sorry I am going to close this PR though it is certainly appreciated and we will discuss further on this issue: https://github.com/react-navigation/rfcs/issues/30. cc @ericvicenti

PR #3695, opened 2018-03-08
Add Back Image Style props

luco:
Adding a style prop for Back Image. I wasn't able to style my custom `backImage`, so I modified to accept it. You can change by using this:

```
...navigationOptions = {
  headerBackImageStyle: {
    borderWidth: 1,
    width: 21
  }
}
```


brentvatne:
hey luco! I think I prefer this PR: https://github.com/react-navigation/react-navigation/pull/3680

it makes `headerBackImage` a react element, which allows you to then provide a style.

thank you for the PR though!

PR #3701, opened 2018-03-09
Pass initialRouteKey into StackRouter (#3540)

BrandonSmith:
The `initialRouteKey` stack config option was introduced in 1.5.0 in the `StackRouter`. However, it is not being passed into `StackRouter` from `StackNavigator`.


brentvatne:
can you rebase master? thanks!

BrandonSmith:
@brentvatne I presume you want the fix in `master` as well as `1.x`? As a separate PR?

PR #3710, opened 2018-03-10
Fix isFocused from withNavigationFocus is a function

### LINKED ISSUE #3709: isFocused from withNavigationFocus is a function on first render
IsFocused from withNavigationFocus is a function on first render, before becoming a boolean.

Here is a log of the typeof the prop inside the render function of a component

```js
console.log('isFocused',
  typeof this.props.isFocused,
  typeof this.props.isFocused === 'function' ?
    this.props.isFocused() :
    this.props.isFocused,
);
```

image

This seems to be the culprit:

image

## How to reproduce
See the snack

https://snack.expo.io/SkKrJpltf

## Versions:

react-navigation 1.5.0

dozoisch:
- when in a tab container, isFocused is a function on first render an boolean afterwards
- closes #3709

---

I think this is the culprit. I made a pr to make it simpler to fix. If you think the issue should be fix in another way, just close this! 3709 is still valid :)

### DIFF
--- a/src/withCachedChildNavigation.js
+++ b/src/withCachedChildNavigation.js
@@ -31,7 +31,7 @@ export default function withCachedChildNavigation(Comp) {
       });
     }
 
-    _isRouteFocused = route => () => {
+    _isRouteFocused = route => {
       const { state } = this.props.navigation;
       const focusedRoute = state.routes[state.index];
       return route === focusedRoute;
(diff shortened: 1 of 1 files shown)


ericvicenti:
This looks like a good fix. cc @satya164

Maybe we could avoid confusion like this by avoiding the "bind" keyword in the implementation, and use fat arrows instead.

```js
        isFocused: this._isRouteFocused.bind(this, scene.route),
```

The above feels way more confusing than this:

```js
        isFocused: () => this._isRouteFocused(scene.route),
```

satya164:
Yeah, let's do what @ericvicenti suggested!

dozoisch:
@satya164 @ericvicenti done!

PR #3716, opened 2018-03-11
State Persistence

ericvicenti:
Knocking down a big item on the v2 roadmap!!

Adds a new state persistence mechanisms to all of the navigators via createNavigationContainer

There are two new props that you can provide to a navigation container: `storageKey` and `renderLoading`.

`storageKey` is the string key to be used with AsyncStorage while saving and persisting navigation state. To disable persistence, set to null.

`renderLoading` allows you to render something while the navigator re-hydrates its state and loads its initial URL. By default this returns null, but Expo users will want to render a <AppLoading /> component for smooth app launches

There is also functionality in this PR to observe errors that come from re-hydrating state, and gracefully recover by dispatching an init action.

Also this revises the init action to *reset* the navigation state, rather than preserve the previous state.


src/createNavigationContainer.js
```diff
@@ -41,15 +48,33 @@ export default function createNavigationContainer(Component) {
...
       };
     }
 
+    _renderLoading() {
+      // don't use a default prop for renderLoading, because it fires an invariant for controlled navigators
+      return this.props.renderLoading ? this.props.renderLoading() : null;
+    }
+    _getStorageKey() {
+      // a null storageKey prop tells the container to NOT persist, so we explicitly check for undefined
+      if (this.props.storageKey === undefined) {
```
brentvatne: I think state persistence should be opt-in rather than opt-out -- more apps will *not* use it than those that will.

src/createNavigationContainer.js
```diff
@@ -166,9 +234,9 @@ export default function createNavigationContainer(Component) {
 
     // Per-tick temporary storage for state.nav
 
-    dispatch = action => {
-      if (!this._isStateful()) {
-        return false;
+    dispatch = async action => {
```
brentvatne: I don't think we should change this to be async, we depend on it returning `true` or `false` in some situations and users might as well.

ericvicenti: good point, will fix

src/createNavigationContainer.js
ericvicenti: sure. I'll rename the prop to `persistenceStorageKey` and make it default to null

brentvatne:
we might want to flag this as experimental initially because I imagine that we will iterate on the `renderLoading` api. for example, if you already have some loading screen rendered for another step in app loading process, you will want to keep it rendered rather than render the navigator with the `renderLoading` api. so we may want to have some way to imperatively load data so we have the data available when we mount navigator and can then do it synchronously like usual

ericvicenti:
Yeah, how would you like to mark this as experimental?

brentvatne:
maybe something like this?

```
persistenceExperimental={{
  storageKey: '..',
  renderLoading: () => <View />
}}
```

brentvatne:
or

maybe something like this?

```
```
  persistenceKeyExperimental: '..',
  renderLoadingExperimental: () => <View />
```

**Comment by brentvatne** [comment]
or

```
  persistenceKey='..'
  renderLoadingExperimental={() => <View />}
```

jamsch:
Just a note: React Native has always had a bug with `AsyncStorage.get()` when remote debugging is enabled. It only works the first time the app is launched on remote debugging and subsequent refreshes won't work.

PR #3718, opened 2018-03-12
Swap addListener out for isFocused prop on ResourceSavingSceneView (1.x)

motiz88:
This is a backport of 81a86fa091b34d3b7f349596eb00cee924d1d31f (#3700) from `master`.

I don't have an isolated test case to share, unfortunately, nor a full analysis of why this fix is correct, but I am upgrading an existing Redux + `react-navigation` based app from `1.0.0-beta.21` to `1.5.1` (yeah, I know) and this fixes a problem I ran into.

Basically, I have a `TabNavigator` at index `0` of a customised `StackNavigator` (something like the "[less pushy `navigate`](https://github.com/react-navigation/rfcs/blob/master/text/0004-less-pushy-navigate.md)" proposal btw). If I stack something at index `1`, then `navigate` from that to a _different_ tab within the lower `TabNavigator`, it appears that the `ResourceSavingSceneView` hosting that tab doesn't get the `action` event it's listening for, never sets its state to `visible: true`, and hence stays blank. Normal navigation between tabs, with the tab bar, works as expected.

Meanwhile, the code which is currently on `master` applies almost as-is and sidesteps the dependency on event propagation behaviour - which itself may or may not be correct, I don't really know; [this line](https://github.com/react-navigation/react-navigation/blob/1d573bc246c4182c4aa1603857487e1a81340655/src/getChildEventSubscriber.js#L119) seems to be skipped under some conditions which correlate with `action` events not hitting the tab.


package.json
```diff
@@ -7,8 +7,7 @@
     "url": "git@github.com:react-navigation/react-navigation.git",
```
satya164: Can you revert the changes here?

motiz88: @satya164 - done. Note that these extra changes were from `master` though.

PR #3742, opened 2018-03-13
Fix memory leak in TabView-test.js

thymikee:
**Summary**

I've noticed there's a leak in the test suite (process hanging) caused by not mocked timer in TabViewPagerScroll used by default on iOS. 
FYI, the leak was detected by running `jest --detectLeaks` flag.

**Test plan (required)**

Tests are green.


brentvatne:
thanks!

Issue #3746, opened 2018-03-14
isFocused gives error


potasium:
- `<TouchableOpacity style={ styles.button } onPress={() => console.log(this.props.navigation.isFocused()) }>`
- `<TouchableOpacity style={ styles.button } onPress={() => console.log(this.props.navigation.isFocused) }>`

### Current Behavior
![ekran resmi 2018-03-14 11 56 47](https://user-images.githubusercontent.com/36793879/37392316-d5e93c42-277e-11e8-95eb-80d47292870f.png)

`ƒ (route) {
        var state = _this.props.navigation.state;
        var focusedRoute = state.routes[state.index];
        return route === focusedRoute;
      }`

### Expected Behavior

- true / false

| software         | version
| ---------------- | -------
| react-navigation | 1.5.2
| react-native        | 0.53.3
| node                    | 8.9.4
| npm                     | 4.6.1

brentvatne:
v1.5.3

EyMaddis:
@brentvatne this is not on the releases page https://github.com/react-navigation/react-navigation/releases

RossKinsella:
@brentvatne Lets get 1.5.3 as the latest release on the releases page. Its currently set to 1.5.2, which has this issue in it. Thanks for you work on this great library

Issue #3729, opened 2018-03-13
withNavigationFocus is buggy


slicejunk:
in views/CardStack/CardStack.js navigation is taken from props
```
  _isRouteFocused = route => {
    const { state } = this.props.navigation;
    const focusedRoute = state.routes[state.index];
    return route === focusedRoute;
  };
```
but navigation is not passed to CardStack component in  views/CardStack/CardStackTransitionner.js:
```
      <CardStack
        screenProps={screenProps}
        headerMode={headerMode}
        headerTransitionPreset={headerTransitionPreset}
        mode={mode}
        router={router}
        cardStyle={cardStyle}
        transitionConfig={transitionConfig}
        transitionProps={props}
        prevTransitionProps={prevProps}
      />
```

maybe the right code should be this one?
```
  _isRouteFocused = route => {
    const { transitionProps: { navigation: { state } } } = this.props;
    const focusedRoute = state.routes[state.index];
    return route === focusedRoute;
  };
```

### Current Behavior
![simulator screen shot - iphone se - 2018-03-13 at 08 37 14](https://user-images.githubusercontent.com/5887100/37328251-c69a3ba0-2699-11e8-965f-b621b2cf3555.png)

### Expected Behavior

- no bug

### How to reproduce

just use @withNavigationFocus anywhere in a StackNavigator

### Your Environment
- react navigation : v1.4 -> v1.5.2
- react-native : *
- node: *

lassecapel:
We experience the same issue. With nested StackNavigators in TabNavigator.

lucasleelz:
same question.

brentvatne:
hi there! sorry you're having an issue. as the issue template explains, we require that you provide a runnable example that reproduces the bug on https://snack.expo.io or in a github repository. please try to minimize the superfluous code and focus only on reproducing the bug.

you might wonder, why are we strict about this? the reason is that there are only a small number of maintainers on this project and there are hundreds of people who report issues. if it takes us 20 minutes on average to reproduce each issue and we go through 10 issues each day, that's about half of our day just trying to reproduce bugs, without even beginning to investigate solutions. compare that to 20 minutes for each person who reports an issue.

please create a new issue with this and i'd be happy to review it!

slicejunk:
@brentvatne This bug is not a functional issue but a mistake in one line of code, you would spend more time playing with an example than reading my comment.
I let someone else (@lassecapel , @lucasleelz ?) to create a new issue as required

[...]

PR #3749, opened 2018-03-14
fix for withNavigationFocus

slicejunk:
fix for https://github.com/react-navigation/react-navigation/issues/3729 , https://github.com/react-navigation/react-navigation/issues/3746


brentvatne:
thanks! released in v1.5.3

PR #3750, opened 2018-03-14
Fix issue with withNavigationFocus HOC.

ScreamZ:
Hello,

trying to use `withNavigationFocus` HOC, but seems not working because state was not reachable.

Based on the rest of the call, I figured out that this code is working as expected.

Am I right ? @brentvatne, i Let you squash merge this if you are okay.

Regards


brentvatne:
thanks! this was resolved in https://github.com/react-navigation/react-navigation/pull/3749

Issue #3686, opened 2018-03-07
React Navigation 3.0 wishlist


brentvatne:
This isn't a roadmap, and it's too far out to be concrete (we haven't released 2.0 yet), but I thought I'd just share what I have in mind. This is all subject to change and we'll need to put out RFCs for them, but they roughly outline what problems I'd like to solve in the months after 2.0.

- [ ] react-native-gesture-handler on cardstack and drawer
- [ ] better story around modals (full screen and partial overlays)
- [ ] better story around layout (status bar translucent/opaque/height, safe areas, translucent header/tabbar)
- [ ] export a scrollview component that scrolls to top automatically when tapping tab icon and already at first screen in stack for that tab. it should also handle insets with translucent headers and tabbars. 
- [ ] audit performance: be able to explain why we do each render on every re-render of a screen or header 
- [ ] status bar config in navigationOptions
- [ ] screen orientation in navigationOptions

melihberberolu:
When tapping tab icon and already at not a first screen goBack one screen in tab's stack.

benevbright:
You know what? In my eyes, these thoughts should be a roadmap of 2.0. They are all key feature and things we all hope. Good job. It's really good to know.

But I understand as the roadmap of 2.0 is to make the basement much stable and solid and then these things go onto there.

brentvatne:
> When tapping tab icon and already at not a first screen goBack one screen in tab's stack.

this already happens

Issue #3685, opened 2018-03-07
React Navigation 2.0 roadmap


brentvatne:
### Planned features

- [x] New createNavigator API (https://github.com/react-navigation/react-navigation/pull/3392)
- [x] Real DrawerRouter (https://github.com/react-navigation/react-navigation/pull/3618)
- [x] 'dismiss' helper that calls goBack with the parent key (https://github.com/react-navigation/react-navigation/pull/3669)
- [x] `dangerouslyGetParent` helper (https://github.com/react-navigation/rfcs/issues/27)
- [x] Make stack actions bubble like navigate (https://github.com/react-navigation/react-navigation/pull/3617)
- [x] Less pushy Navigate behavior (https://github.com/react-navigation/rfcs/pull/25)
- [x] Minimize React 16.3 warnings by moving away from lifecycle methods where possible (unable to do this for Transitioner.js, needs more refactoring for this to work)
- [x] Extract tab navigator to separate library that exports `BottomTabNavigation`, `MaterialBottomTabNavigation` and `MaterialTopTabNavigation`. Add deprecated lib for the old behavior. Add examples to NavigationPlayground.
- [x] Shallow navigation options (rfc merged) (https://github.com/react-navigation/rfcs/pull/26)
- [x] New Drawer router and actions
- [x] State persistence (https://github.com/react-navigation/react-navigation/pull/3716)
- [x] New Actions API (https://github.com/react-navigation/react-navigation/pull/3619 and https://github.com/react-navigation/rfcs/pull/18)

### Documentation updates

- [x] Shorthand w/o `screen`
- [x] Limitations section of docs
- [x] Persistence
- [x] Update Fundamentals section
- [x] Update any Snacks
- [x] Explain `key` (what it means, why)
- [x] Navigate vs push
- [x] Describe how `navigationOptions` are resolved and how to set them directly on navigators
- [ ] Update Tab navigation docs
- [ ] Modular action creators
- [ ] Update navigation prop docs

### Unresolved questions

- [x] Should `pop` and `popToTop` actually bubble up?

### Bugs to fix

- [x] Verify that we avoid race conditions around Transitioner navigation completion
- [ ] Update Flow definition

### Bonus

- [x] Make navigation playground look much nicer on android
- [x] Docs for navigator ref at top level (https://github.com/react-navigation/react-navigation.github.io/pull/85)

### Post-2.0 followup

- [ ] Add link to CS50 lecture
- [ ] Add link to React Amsterdam talk
[...]

PR #3751, opened 2018-03-14
Experiment: use react-native-gesture-handler/DrawerLayout

brettdh:
**Motivation**

I found myself wanting to create a `DrawerNavigator` with some custom styling:
- Make it appear that the drawer is underneath the view, rather than on top of it
- Style the overlay that hides the non-drawer content
  - Specifically, set it to a color with alpha zero so as to remove it entirely

I found that `react-native-gesture-handler`'s `DrawerLayout` component supports exactly these features, as props. I was about to attempt to create my own custom drawer navigator with this `DrawerLayout`, when I noticed a few things:
- The 3.0 Roadmap (and a handful of other issues) mention a plan to move to using `react-native-gesture-handler/DrawerLayout` in place of the current polyfill. 
- After looking at the source for the existing `DrawerNavigator`, I surmised that it would be very simple to swap out `DrawerLayout`, since the API only contains additions.

However, I didn't see a clear place to discuss my intention here, since:
- The roadmap issue is limited to collaborators
- I couldn't find any other issues PRs discussing this plan
- I didn't see any existing branches going down this road
- It could be an RFC, but the code changes are so small that it seems easier to discuss as a PR

Feel free to close this PR and redirect me to a more appropriate avenue, or reject in favor of an implementation that's already in the works, or something else. I just wanted to start the conversation, and it turned out to be nearly as simple a change as I expected. (I was also hoping that this might help this functionality to be released sooner, since I would really like to use it 🙂)

**Implementation Notes**

Besides swapping out the `DrawerLayout` component, I made these notable changes:
- Don't call `_drawer.{open,close}Drawer()` when the drawer is already {open, closed}
  - If the drawer opens or closes because of non-programmatic user action (e.g. swipe from left, or tap outside drawer), the `onDrawer{Open,Close}` event would cause the drawer to animate a second time with no user action.
  - The `suppressAnimation` approach here feels a bit hacky; I welcome suggestions for another way to make an action only update the nav state without subsequently calling `openDrawer()` or `closeDrawer()`; i.e. distinguishing between.
- Use `onDrawerStateChanged` instead of `onDrawer{Open,Close}`
  - Previously, the navigation state was not being updated until the drawer animation finished
  - This prevented e.g. reopening the drawer while the close animation was still happening
  - Changed to update nav state on the `Settling` state

**Test plan (required)**

- Tests pass
  - I haven't yet updated the `DrawerNavigator` test, because I wasn't sure how to let Jest use a native module - or mock the right subsets of it to make the test still meaningful.
- NavigationPlayground looks good on iOS/Android
  - Behaves same as before
- New props are passed down correctly and have the desired effect
  - Not tested exhaustively; tested with `drawerType`, `hideStatusBar`, and `overlayColor`

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
   ],
   "peerDependencies": {
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "react-native-gesture-handler": "^1.0.0"
   },
   "dependencies": {
     "clamp": "^1.0.1",
@@ -36,7 +37,6 @@
     "prop-types": "^15.5.10",
     "react-lifecycles-compat": "^3",
     "react-native-drawer-layout-polyfill": "^1.3.2",
-    "react-native-gesture-handler": "^1.0.0-alpha.41",
     "react-native-safe-area-view": "^0.7.0",
     "react-navigation-deprecated-tab-navigator": "1.2.0",
     "react-navigation-tabs": "0.2.0"
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "prop-types": "^15.5.10",
     "react-lifecycles-compat": "^3",
     "react-native-drawer-layout-polyfill": "^1.3.2",
+    "react-native-gesture-handler": "^1.0.0-alpha.41",
     "react-native-safe-area-view": "^0.7.0",
     "react-navigation-deprecated-tab-navigator": "1.2.0",
     "react-navigation-tabs": "0.2.0"
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "prettier-eslint": "^8.8.1",
     "react": "16.2.0",
     "react-native": "^0.52.0",
+    "react-native-gesture-handler": "^1.0.0",
     "react-native-vector-icons": "^4.2.0",
     "react-test-renderer": "^16.0.0"
   },
--- a/src/navigators/createDrawerNavigator.js
+++ b/src/navigators/createDrawerNavigator.js
@@ -41,6 +41,10 @@ const DefaultDrawerConfig = {
   drawerPosition: 'left',
   drawerBackgroundColor: 'white',
   useNativeAnimations: true,
+  drawerType: 'front',
+  hideStatusBar: false,
+  statusBarAnimation: 'slide',
+  overlayColor: 'black',
 };
 
 const DrawerNavigator = (routeConfigs, config = {}) => {
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -21,6 +21,7 @@ import { SafeAreaView, createStackNavigator } from 'react-navigation';
 import CustomTabs from './CustomTabs';
 import CustomTransitioner from './CustomTransitioner';
 import Drawer from './Drawer';
+import StyledDrawer from './StyledDrawer';
 import MultipleDrawer from './MultipleDrawer';
 import TabsInDrawer from './TabsInDrawer';
 import ModalStack from './ModalStack';
@@ -62,6 +63,10 @@ const ExampleInfo = {
   Drawer: {
     name: 'Drawer Example',
     description: 'Android-style drawer navigation',
+  },
+  StyledDrawer: {
+    name: 'Styled Drawer Example',
+    description: 'Drawer navigation using extra styling',
   },
   StackWithHeaderPreset: {
     name: 'UIKit-style Header Transitions',
@@ -133,6 +138,7 @@ const ExampleRoutes = {
   SwitchWithStacks,
   SimpleTabs: SimpleTabs,
   Drawer: Drawer,
+  StyledDrawer: StyledDrawer,
   // MultipleDrawer: {
   //   screen: MultipleDrawer,
   // },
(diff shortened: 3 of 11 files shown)


brettdh:
...hrm, the test failures are flow-related - I forgot to add types for the new `config` items. Odd that it didn't fail locally when I ran `yarn test`.

brettdh:
Also, apparently the above QR code points to the right commit-based release channel, but since the flow step in CircleCI failed, it's never been published to - but loads anyway. See https://github.com/expo/expo/issues/1039

brentvatne:
Awesome @brettdh! This is on the 3.0 wishlist (https://github.com/react-navigation/react-navigation/issues/3686) and I really appreciate you getting this going! I don't have time right now to review this properly because I want to prioritize wrapping up the 2.0 roadmap (https://github.com/react-navigation/react-navigation/issues/3685) first, but I will come back to this when that's all ready.

satya164:
Maybe it'll be nice to move drawer to a separate repo along with this change.

src/NavigationActions.js
```diff
@@ -112,12 +112,15 @@ const completeTransition = createAction(COMPLETE_TRANSITION, payload => ({
 
 const openDrawer = createAction(OPEN_DRAWER, payload => ({
   type: OPEN_DRAWER,
+  suppressAnimation: false,
```
satya164: I think better to have the same behavior in both test and real code.

src/navigators/createDrawerNavigator.js
```diff
@@ -41,6 +41,10 @@ const DefaultDrawerConfig = {
   drawerPosition: 'left',
   drawerBackgroundColor: 'white',
   useNativeAnimations: true,
+  drawerType: 'front',
```
satya164: I know it's from the gesture handler repo, but it's unclear what `drawerType` means

src/NavigationActions.js
brettdh: I'm not sure what behavior you're referring to here.

src/navigators/createDrawerNavigator.js
brettdh: `drawerType` is one of many existing passthrough props, all of which have the same names as the props they get passed through to on `DrawerLayout`. I'm inclined to maintain that, so that there's no ambiguity about what happens with those props.

package.json
```diff
@@ -36,6 +36,7 @@
     "prop-types": "^15.5.10",
     "react-lifecycles-compat": "^1.0.2",
     "react-native-drawer-layout-polyfill": "^1.3.2",
+    "react-native-gesture-handler": "^1.0.0-alpha.41",
```
brentvatne: we should make this a peer dependency

brentvatne:
one main concern I have with the current behavior of the drawer layout is that we need to re-render so much of the app in order to transition from closed -> open and vice versa. for example, looking at the profiling timeline of this (in dev mode, in chrome) reveals that it takes 65ms in a pretty basic case because it re-renders both the drawer and the scene being rendered. this isn't an issue specific to this pr, however

package.json
brettdh: Is there a good way to handle the [different version requirements](https://github.com/kmagiera/react-native-gesture-handler#supported-configurations) for different versions of react-native, or would we just set the loosest spec here (`^1.0.0-alpha.29`) and document the required versions? Should we also make the react-native requirement stricter (`>=0.47`)? Or is there a way I'm unaware of to tie two peerDep versions together?

Tangentially related: we may also want to consider adding support for using the `GestureHandler` included in Expo.

brentvatne: @brettdh - sorry I haven't had time to look back at this until now. I think we should only release this when rngh has left alpha, at which point we can assume that it will follow semver and we can support major version 1

brettdh: @brentvatne Makes sense to me, though of course I'll be sad for the delay. What about the Expo considerations? Or would it not matter, since [the `expo` package depends on `react-native-gesture-handler`](https://github.com/expo/expo-sdk/blob/master/package.json#L59) anyway, thereby satisfying the peer dependency?

brentvatne: > Or would it not matter, since the expo package depends on react-native-gesture-handler anyway, thereby satisfying the peer dependency?

precisely.

> Makes sense to me, though of course I'll be sad for the delay.

I suppose what we could do here is specify a peerDependency that is satisfied by both pre-release and uh, release versions (what's the proper term for this anyhow? :P) I don't anticipate that the API for drawer will change significantly, so it should be safe.

brettdh: Updated, rebased, pushed.

brentvatne:
@brettdh - I added you as a collaborator on https://github.com/react-navigation/react-navigation-drawer-navigator - let's pull this project out of react-navigation core and work on it there. This will allow us to make a bunch of changes to it without breaking 2.x functionality, and still use it alongside react-navigation@2.0.0. It's also the direction I want to move navigators in general (eg: react-navigation-tabs)

brettdh:
@brentvatne works for me :) I'll leave it to you to do the initial extraction, though I'm happy to figure out the right `filter-branch` for my changes once that's done.

brentvatne:
let's move this over to https://github.com/react-navigation/react-navigation-drawer :)

PR #3752, opened 2018-03-14
Kanalasumant/upgrade to react16.3

kanalasumant:
Pushing for support React v16.3 and React Native v0.54 and up by replacing componentWillMount with UNSAFE_componentWillMount and componentWillReceiveProps with UNSAFE_componentWillReceiveProps

### DIFF
--- a/examples/NavigationPlayground/js/App.js
+++ b/examples/NavigationPlayground/js/App.js
@@ -153,7 +153,7 @@ class MainScreen extends React.Component<any, State> {
     scrollY: new Animated.Value(0),
   };
 
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
     Asset.fromModule(
       require('react-navigation/src/views/assets/back-icon-mask.png')
     ).downloadAsync();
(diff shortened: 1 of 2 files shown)


kanalasumant:
I seriously wish I had the time to test the repo using static getDerivedStateFromProps everywhere instead of componentWillReceiveProps. But for now UNSAFE methods might work.

ericvicenti:
Nope, this will break for older versions of RN, which do not implement these methods

kanalasumant:
@ericvicenti but the package.json has peerDependencies of * of react-native. That means from now on only the latest version i.e 0.54 will be installed right. and that supports these methods. I'm trying to use the Context API in my project and the only way to use it is to upgrade my react and react-native, but for doing that react-navigation and other libraries need to support these unsafe methods too, hence the motivation for a change. Please let me know what can I do to fix this.

arnsa:
@kanalasumant you could probably create two files, one with, for example, getDerivedStateFromProps method and the other one with componentWillReceiveProps. Then check React.version value and include the according file. 

But anyways, I am pretty sure there must be a better solution, because this looks kind of dirty to me.

brentvatne:
mostly fixed this already in other commits on master. only remaining thing to fix is `Transitioner.js`

PR #3763, opened 2018-03-15
Prevent duplicate drawer events (OPEN_DRAWER/CLOSE_DRAWER)

sraka1:
When calling `openDrawer`/`closeDrawer` on the new Drawer router, this also triggers `onDrawerOpen`/`onDrawerClose` on the `DrawerLayout` which in turn calls `openDrawer`/`closeDrawer` again (without need). This is especially annoying when managing the state with redux.

This change makes sure that you are opening a closed drawer and vice-versa. I couldn't find a good reason for opening an already opened drawer, so I think this change breaks nothing. 

Similar logic was already present before the Drawer refactor.

Drawer behaviour was tested and it seems it works fine (still works perfectly fine for gestures and all other non-user code triggered drawer toggling scenarios).

Slightly unrelated - a change in the docs for the Drawer component is quite warranted IMHO. I will try to create another PR there if time permits.


sraka1:
The duplicate events don't actually _do anything_, since in DrawerRouter this check is [also performed](https://github.com/react-navigation/react-navigation/blob/bdda6fa5be2cbdef2a1f4f5b3a65fa4518003c27/src/routers/DrawerRouter.js#L25). 

It's basically just when an explicit action is triggered (i.e. via a button), the event callback (onDrawerOpen/onDrawerClose) is also triggered, dispatching the additional OPEN_DRAWER/CLOSE_DRAWER action...

I can make an example, but hopefully the above clears it up 🙃

ericvicenti:
Tests are failing on master and I'm pretty sure this doesn't break anything, so I'm landing now even though the tests are failing

Issue #3674, opened 2018-03-06
Provide an option to use default transition when overloading transition configuration 


jdgreenberger:
This issue is a feature suggestion. The [feature guidelines](https://reactnavigation.org/docs/contributing.html#suggesting-a-feature) suggested I create a feature request in the issues. Please let me know if there is a more preferable location. 

Right now, if you want to specify a custom transition, you must overload the `transitionConfig` param. There are situations however, where one may want to conditionally display a custom transition, and use the default in other situations. Example:
```javascript
screenInterpolator: (props) => {
  if (props.scene.route.routeName === 'someView') {
    return customTransition(props);
  } else {
    return defaultScreenTransition(props);
  }
}
```
 One alternative, based on this [issue](https://github.com/react-navigation/react-navigation/issues/2493), is to reach into the SRC and pull out the default transition you need. This solution is not ideal because:
1) the `forHoritzontal` transition does not reflect the actual default transitions which use a different transition for iOS/android 
2) reaching into the src is bug prone if one ever intends to upgrade

I'm proposing `transitionConfig` should take an option to return to the default transition, or expose the default transition function as a top level library export.

jdgreenberger:
Any update on this? I can try to submit a PR if the owners agree with the suggestion.

brentvatne:
@ericvicenti - any thoughts here?

ericvicenti:
I think it makes sense for us to export the defaults from the library so that people can override it conditionally like this.

ericvicenti:
A PR would be appreciated!

ayushmad:
I am not sure it should be part of the same PR if not I can create a new one. It would also be nice to have _StackViewStyleInterpolar_ exposed. 
My thinking was if Interpolator were exposed then consumer could pick more the interpolator needed for there screenTransition. 

**Use Case:**
Defaults really work really well for the app(its an IOS only app) I am working on except in one screen I wanted to add the `forFadeFromBottomAndroid` in IOS.
I was planning on implementing my own and saw that this already exists and wanted to use it directly.
Currently I am going down and pulling the interpolator from the library.


* I wanted to check if this makes sense or if you guys want consumers to use build there own interpolator.
- I can create a pull request if needed.



_Side Note:_
For the Example above where you only want to add transitions for specific screen. I get around it by not providing a config at all. Based on the code it always falls back to the default config.
```
transitionConfig: () => {
   if (props.scene.route.routeName === 'someView') {
        return customTransitionConfig;
[...]

PR #3766, opened 2018-03-16
Export default transition config in stack navigator

jdgreenberger:
Addresses https://github.com/react-navigation/react-navigation/issues/3674 by exporting the default transition config for stack navigation. I'm not sure where the best place to export is — I tried in the navigation, but that seemed lower level than a screen transition. Also looking for suggestion as to the best place to test.

### DIFF
--- a/src/react-navigation.js
+++ b/src/react-navigation.js
@@ -145,4 +145,10 @@ module.exports = {
   get withOrientation() {
     return require('./views/withOrientation').default;
   },
+
+  // Default Settings
+  get defaultStackViewTransitionConfig() {
+    return require('./views/StackView/StackViewTransitionConfigs')
+      .defaultTransitionConfig;
+  },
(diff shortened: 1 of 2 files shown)


ericvicenti:
Can you write some documentation that demonstrates how this is used? You can just comment it for now, and we will move it to the website when this lands.

rreorr:
any progress on this? I would love to make use of it as I am replacing some transitions but do not wish to overwrite all of them. The ability to import the default config would make this very easy to do.

jdgreenberger:
It appears in the latest version of the library that a custom `transitionConfig` function has access to props, so there is a simple solution.
```
transitionConfig: (props) => {
   if (props.scene.route.routeName === 'someView') {
        return customTransitionConfig;
   }
   return {}; // will automatically return to the default
}
```

PR #3776, opened 2018-03-17
Lazy loading placeholder element for TabNavigator

kevinsperrine:

It adds a `lazyPlaceholder` prop that can take a react element to render in place of null to give the user a better experience on tab change.

**Test plan (required)**
As there aren't any tests currently detailing the lazy loading feature, I wasn't sure how to thoroughly test this, but am willing to make this more robust with a little direction.

I re-used my reproduction repo from an open issue, and created a lazyPlaceholder branch to show this in action. https://github.com/kevinsperrine/react-navigation-dynamic-redux-tab-test/tree/lazyPlaceholder I've ran this in expo on both platforms. 
lazyplaceholder

### DIFF
--- a/src/navigators/TabNavigator.js
+++ b/src/navigators/TabNavigator.js
@@ -20,6 +20,7 @@ const TabNavigator = (routeConfigs, config = {}) => {
     tabBarPosition,
     tabBarOptions,
     lazy,
+    lazyPlaceholder,
     removeClippedSubviews,
     swipeEnabled,
     animationEnabled,
@@ -34,6 +35,7 @@ const TabNavigator = (routeConfigs, config = {}) => {
     <TabView
       {...props}
       lazy={lazy}
+      lazyPlaceholder={lazyPlaceholder}
       removeClippedSubviews={removeClippedSubviews}
       tabBarComponent={tabBarComponent}
       tabBarPosition={tabBarPosition}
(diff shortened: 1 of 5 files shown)


ericvicenti:
Hi there! We have moved development on the tab view over to this other repo:

https://github.com/react-navigation/react-navigation-tabs

Plus, I think in the new version, the swipable tabs are never lazy, so this change probably won't be needed.

cc @satya164

satya164:
Yeah. `TabNavigator` is split into 2 navigators in the new version, and one more is added.

`BottomTabNavigator`, `MaterialBottomTabNavigator`: These don't have swipe gestures and always lazy. The advantage is faster initial render.

`MaterialTopTabNavigator`: This one has a swipe gesture, so it's never `lazy` as it provides poor UX. However, the way you implemented a placeholder is exactly what I wanted to users to do with swipeable tabs. I think this can live in userland. Probably a HOC:

```js
withLazyTab(ScreenComponent, {
  placeholder: PlaceholderComponent,
})
```

The new focus events make it easy to implement this.

kevinsperrine:
@satya164 I presume you're talking about the focus events exampled here: https://github.com/react-navigation/react-navigation/blob/master/examples/NavigationPlayground/js/TabsWithNavigationFocus.js Thanks!

PR #3795, opened 2018-03-19
Allow setting explicit gestureDirection for more control

n1ru4l:

**Motivation**

I need more control over how the gestures work. Right now when overwriting the `transitionConfig.screenInterpolator`, the gestures mismatch the animation. By changing the `navigationOptions.gestureDirection` (in a backwards compatible way) to support the values `up`, `down`, `left` and `right` this allows to configure the direction, where `default` and `inverted` would (in combination with `mode`) fail to _fix_ the user experience.

**Test plan (required)**

I will try to setup an expo snack

**Code formatting**

Will do when finished

### DIFF
--- a/src/views/StackView/StackViewLayout.js
+++ b/src/views/StackView/StackViewLayout.js
@@ -207,10 +207,14 @@ class StackViewLayout extends React.Component {
       mode,
     } = this.props;
     const { index } = navigation.state;
-    const isVertical = mode === 'modal';
     const { options } = scene.descriptor;
+    const isVertical = mode === 'modal'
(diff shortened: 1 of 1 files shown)


rjacobskind:
@n1ru4l Thanks for submitting this PR! I actually built on it a bit and submitted a new PR so that users can swipe up/down as swipe gestures (as you implemented), but also to have the screen transition in the direction of the swipe gesture. Take a look below

https://github.com/react-navigation/react-navigation/pull/3868

brentvatne:
closing in favor of #3868 which still needs us to review it

PR #3806, opened 2018-03-20
Implement paths on SwitchRouter

ericvicenti:
This is a redo of #1688, now moved to SwitchRouter, and I added a test

Thanks go to @faceyspacey for identifying and fixing this!


src/routers/StackRouter.js
```diff
@@ -209,7 +209,7 @@ export default (routeConfigs, stackConfig = {}) => {
         if (action.type !== NavigationActions.PUSH && lastRouteIndex !== -1) {
           // If index is unchanged and params are not being set, leave state identity intact
           if (state.index === lastRouteIndex && !action.params) {
-            return state;
```
brentvatne: was this intentionally included int his pr?

Issue #2732, opened 2017-10-11
Transition from headerless screen to screen with header jerks content on first screen down


mschipperheyn:
### Current Behavior
I have two screens, one is without a header and one is with a header. The content on screen one is centered vertically. 

When I navigate from screen 1 to screen 2, during the transition, the content from screen 1 jumps down to the position it would have had if the screen did have a header.

### Expected Behavior
The content on screen 1 should stay in place.

### Your Environment
| react-navigation | 1.0.0-beta.13
| react-native     |  0.49.3
| node             | 8.5.0
| npm or yarn      | yarn

## code
```
const ScreenStack = StackNavigator(
    { ...routez.drawerRoutez, ...routez.loginRoutez },
    {
        initialRouteName: 'Loading',
        headerMode: 'float',
    },
);

const PrimaryNav = DrawerNavigator(
    {
        DrawerStack: {
            screen: ScreenStack,
        },
    },
    {
        headerMode: Platform.OS === 'ios' ? 'float' : 'screen',
        navigationOptions: ({ navigation }) => ({
            headerStyle: { backgroundColor: 'green' },
            gesturesEnabled: false,
            headerLeft: (
                <IconButton
                    icon="menu-1"
                    onPress={() => {
                        if (navigation.state.index === 0) {
                            navigation.navigate('DrawerOpen');
                        } else {
                            navigation.navigate('DrawerClose');
                        }
                    }}
                />
            ),
        }),
    },
);

export default PrimaryNav;
```

```
class LoginScreen extends React.Component {
}

LoginScreen.navigationOptions = {
    header: null,
    drawerLockMode: 'locked-closed',
};

class RegisterScreen extends React.Component {
}

const Title = () => <FormattedMessage id="ui.registration" />;

RegisterScreen.navigationOptions = ({ navigation: { state: { params = {} } } }) => ({
    drawerLockMode: 'locked-closed',
    title: <Title />,
    headerRight: (
        <HeaderSaveButton
            formName={FORM}
            dispatch={params.dispatch}
            submitting={params.submitting}
            title={params.title}
        />
    ),
});

```

helios1138:
you can give header a height and set the same value as negative margin on root view

mschipperheyn:
Yeah, that would be a workaround. But it does seem to me like a bug.

kelset:
Thanks for reporting. Is it related in any way to https://github.com/react-community/react-navigation/pull/2520?

heyjared:
[...]

PR #3821, opened 2018-03-23
Smoothly transition header visibility in Stack

skevy:
### Summary

This addresses #2732 (which also happens to be the top issue on canny.io).

Transitioning between screens with and without headers when `headerMode` is set to "screen" works just fine, but the current experience when `headerMode` is set to "float" is very broken. The goal of this PR is to replicate the behavior of UIKit on iOS: when a screen without a header is pushed, the header on the previous screen is pushed off the screen to the left, and the individual components of the header do not animate. When a screen with a header is pushed _after_ a screen without one, we slide the new header on the screen from right to left, in sync with the push animation, again while avoiding animating the individual header components.

I also updated the UIKit transition example in Navigation Playground to demonstrate this.

Here's a gif of the animation from a native iOS app:

gif of native ios transition

### Test Plan:

Try the UIKit transition example in Navigation Playground.

### DIFF
--- a/src/views/Header/HeaderStyleInterpolator.js
+++ b/src/views/Header/HeaderStyleInterpolator.js
@@ -1,13 +1,37 @@
 import { Dimensions, I18nManager } from 'react-native';
 import getSceneIndicesForInterpolationInputRange from '../../utils/getSceneIndicesForInterpolationInputRange';
 
-const crossFadeInterpolation = (first, index, last) => ({
-  inputRange: [first, index - 0.9, index - 0.2, index, last],
-  outputRange: [0, 0, 0.3, 1, 0],
+function hasHeader(scene) {
+  if (!scene) {
+    return true;
+  }
+  const { descriptor } = scene;
+  return descriptor.options.header !== null;
+}
+
+const crossFadeInterpolation = (scenes, first, index, last) => ({
+  inputRange: [
+    first,
+    first + 0.001,
+    index - 0.9,
+    index - 0.2,
+    index,
+    last - 0.001,
+    last,
+  ],
+  outputRange: [
+    0,
+    hasHeader(scenes[first]) ? 0 : 1,
+    hasHeader(scenes[first]) ? 0 : 1,
+    hasHeader(scenes[first]) ? 0.3 : 1,
+    hasHeader(scenes[index]) ? 1 : 0,
+    hasHeader(scenes[last]) ? 0 : 1,
+    0,
+  ],
 });
 
 /**
- * Utility that builds the style for the navigation header.
+ * Utilities that build the style for the navigation header.
  *
  * +-------------+-------------+-------------+
  * |             |             |             |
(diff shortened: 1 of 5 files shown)


src/views/Header/HeaderStyleInterpolator.js
```diff
@@ -1,13 +1,37 @@
...
+  const { descriptor } = scene;
+  return descriptor.options.header !== null;
+}
+
+const crossFadeInterpolation = (scenes, first, index, last) => ({
+  inputRange: [
+    first,
+    first + 0.001,
+    index - 0.9,
+    index - 0.2,
```
ericvicenti: These numbers scare me. Not sure what the alternative is, but it would be nice if the cross fading felt less hardcoded.

Of course these numbers were already here, so this change seems reasonable

src/views/Header/HeaderStyleInterpolator.js
```diff
@@ -17,6 +41,37 @@ const crossFadeInterpolation = (first, index, last) => ({
...
+  const index = scene.index;
+
+  const width = layout.initWidth;
+  const rtlMult = I18nManager.isRTL ? -1 : 1;
+  const translateX = position.interpolate({
+    inputRange: [first, index, last],
+    outputRange: [
+      rtlMult * (hasHeader(scenes[first]) ? 0 : width),
+      rtlMult * (hasHeader(scenes[index]) ? 0 : isBack ? width : -width),
+      rtlMult * (hasHeader(scenes[last]) ? 0 : -width),
```
ericvicenti: Was this tested with RTL? This constraint is can be easy to forget about.. maybe we should add some info to the contributors guide to help people test edge cases.

src/views/StackView/StackViewLayout.js
```diff
@@ -433,11 +445,22 @@ class StackViewLayout extends React.Component {
       screenInterpolator &&
       screenInterpolator({ ...this.props.transitionProps, scene });
 
+    // If this screen has "header" set to `null` in it's navigation options, but
+    // it exists in a stack with headerMode float, add a negative margin to
+    // compensate for the hidden header
```
ericvicenti: hmm, so the whole card is moved up, interesting. This is probably the most reasonable thing for now, but I wonder if we should instead make the header hight padding some explicit thing in the future, and in this case it would just be 0.

src/views/Header/HeaderStyleInterpolator.js
skevy: yah -- i didn't mess with that too hard. The thing that I added were the extra control points of `first + .001` and `last - .001` to the interpolation

src/views/StackView/StackViewLayout.js
skevy: yah, that's probably fair. Maybe this can be addressed in future changes to Stack.

src/views/Header/HeaderStyleInterpolator.js
skevy: I did test with RTL, though unfortunately I think the swipe to go back gesture is broken in RTL mode.

cbjs:
can this update solve my problem. 
https://github.com/react-navigation/react-navigation/issues/3889

I tries newest 1.5.9, the problem still exists.

brentvatne:
yes but this only exists in the 2.0 branch

wvteijlingen:
Thanks, this is a great improvement! Quick question though: Is it possible to have these smooth transitions when using a custom header component? I'm currently using 1.x with a custom header, but it seems that this you don't get this smooth transition when using a custom header in 2.x.

ericvicenti:
Nope this won't work for custom headers, but you can try copying the
changed code into your header. We will try to make animated header
customization easier in the future.

janicduplessis:
@skevy @ericvicenti Awesome work on this! However there seems to be an issue when navigating between 2 screens with `header: null`. Let me know if you need a repro case.

animated gif-source

src/views/StackView/StackViewLayout.js
janicduplessis: I'm seeing some flickers sometimes on app load where a screen without header is offset by the header height, not 100% sure but it might be caused by this because we measure the header view, I think we'd have to use the same value that's used to add the header height padding like @ericvicenti mentionned.

janicduplessis: I can confirm this is the issue, using state does help a bit but does not solve the issue. The only solution is to have the header height synchronously. It is exposed as a constant on the Header component but this will break custom header components if we use that.

wvteijlingen:
@ericvicenti Too bad it doesn't work with custom headers, but I understand given the current header layout.

Would it be possible to separate the rendering of the header into two components? An outer container view that manages the visibility transitions, and the real header that is rendered into that container. That makes is easy to render a custom header component, and still have smooth visibility transitions.

ragunaru:
I'm on 2.2.0 and the header transition bug is still not fixed?

brentvatne:
@ragunaru - it sounds like you may be seeing a different issue because this is indeed working as expected. please create a new issue and follow the template and i'll be happy to look into it!

Issue #3598, opened 2018-02-25
Dynamic nested TabNavigator: cannot read property 'foreach' of undefined


raarts:
### Current Behavior

This probably is a feature request.

Swapping out a nested tabnavigator for another one on the fly does not work, and results in:

(And I'm still not well-versed enough in React Navigation to create a PR for this..)

```
react-dom.development.js?cada:8305 The above error occurred in the <withCachedChildNavigation(TabView)> component:
    in withCachedChildNavigation(TabView) (at TabNavigator.js:34)
    in Unknown (at createNavigator.js:13)
    in Navigator (at createNavigationContainer.js:227)
    in NavigationContainer (at WebMainTabNavigator.js:33)
    in AppRootTabNavigator (at SceneView.js:17)
    in SceneView (at ResourceSavingSceneView.js:62)
    in div (created by View)
    in View (at ResourceSavingSceneView.js:55)
    in div (created by View)
    in View (at ResourceSavingSceneView.js:46)
    in ResourceSavingSceneView (at TabView.js:32)
    in div (created by View)
    in View (at TabViewPagerPan.js:251)
    in div (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent (at TabViewPagerPan.js:238)
```

### Expected Behavior

I need it to work.

### How to reproduce

https://snack.expo.io/@raarts/cannot-read-property-'foreach'-of-undefined

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.1.2
| react-native     | 0.52.0
| node             | 8.9.4
| npm or yarn      | 1.3.2

Tommos0:
I have the same issue:
https://snack.expo.io/S1XkMPTOf

It happens when a Navigator gets passed the `navigator` prop from a parent element that got this passed because it is a screen of a parent Navigator.
e.g.
```
screen: SubNav /> // OK
screen: (props) => <SubNav navigation={props.navigation} /> // Crash
```

Tommos0:
I do needs this kind of behaviour. The situation is that my SubNav is not a direct child of MainNav, eg.
```
MainNav
--|View
--|--|Header
--|--|SubNav
```
I'm gonna solve the problem by having seperate Navigators, each with its own state. It will however complicate navigation from/to different components in my app (as the right `navigation` object will have to be used).

Tommos0:
Mentions of same:
https://github.com/react-navigation/react-navigation/issues/335#issuecomment-290226709
[...]

PR #3823, opened 2018-03-23
throw readable error when wrong navigation prop is passed

Tommos0:
See https://github.com/react-navigation/react-navigation/issues/3598

### DIFF
--- a/src/navigators/createNavigator.js
+++ b/src/navigators/createNavigator.js
@@ -42,6 +42,11 @@ function createNavigator(NavigatorView, router, navigationConfig) {
       const { routes } = state;
 
       const descriptors = {};
+      if (typeof routes === 'undefined') {
+        throw TypeError(
+          'No "routes" found in navigation state. Did you try to pass the navigation prop of a React component to a Navigator child? See https://github.com/react-navigation/react-navigation/issues/3598#issuecomment-375622188'
+        );
+      }
       routes.forEach(route => {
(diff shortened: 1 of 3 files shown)


src/navigators/createNavigator.js
```diff
@@ -42,6 +42,11 @@ function createNavigator(NavigatorView, router, navigationConfig) {
       const { routes } = state;
 
       const descriptors = {};
+      if (typeof routes === 'undefined') {
+        throw TypeError(
+          'No "routes" found in navigation state. Did you try to pass the navigation prop of a React component to a Navigator child? See https://github.com/react-navigation/react-navigation/issues/3598#issuecomment-375622188'
```
ericvicenti: rather than referring to a comment in an issue, can we link to the docs instead?

This might be an appropriate location: https://next.reactnavigation.org/docs/custom-navigators.html#navigator-navigation-prop

Or you could improve a different doc, and link to that

ericvicenti: It looks like this doc might also be relevant: https://v2.reactnavigation.org/docs/common-mistakes.html#explicitly-rendering-more-than-one-navigator

justsml: Don't forget to use `new TypeError` - useful stack traces depend on the constructor being called.

brentvatne:
@justsml - sorry for this, could you possibly do a rebase?

PR #3827, opened 2018-03-24
Experiment: prototype of maybe terrible idea: navigation.dangerouslyUpdateRouter

brentvatne:
# Motivation

Some people would like to update the route config dynamically at runtime, eg: to add new routes to drawer or add/remove tabs. Before pursuing this further I'd like to make sure we understand and enumerate as many use cases as possible so we can know if this covers them. But I wanted to prototype this now to get an idea of the impact it might have on codebase.

# Test plan

It's a prototype, it doesn't have tests, you can open it in NavigationPlayground though.

### DIFF
--- a/src/navigators/createNavigator.js
+++ b/src/navigators/createNavigator.js
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import NavigationActions from '../NavigationActions';
 import getChildEventSubscriber from '../getChildEventSubscriber';
 import addNavigationHelpers from '../addNavigationHelpers';
 
@@ -36,12 +37,23 @@ function createNavigator(NavigatorView, router, navigationConfig) {
       return this.props.navigation;
     };
 
+    _dangerouslyUpdateRouter = (key, newRouteConfigs) => {
+      if (router.setRoutes) {
+        this.props.navigation.dispatch(
+          NavigationActions.reinit({ key, newRouteConfigs })
+        );
+      } else {
+        console.warn('Router does not support updating routes');
+      }
+    };
+
     render() {
       const { navigation, screenProps } = this.props;
       const { dispatch, state, addListener } = navigation;
       const { routes } = state;
 
       const descriptors = {};
+
       routes.forEach(route => {
         const getComponent = () =>
           router.getComponentForRouteName(route.routeName);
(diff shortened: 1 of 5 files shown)


src/navigators/createNavigator.js
```diff
@@ -36,12 +37,23 @@ function createNavigator(NavigatorView, router, navigationConfig) {
       return this.props.navigation;
     };
 
+    _dangerouslyUpdateRouter = (key, newRouteConfigs) => {
+      if (router.setRoutes) {
```
ericvicenti: After the action creators overhaul, I don't think there would be a way to access the router from here.

brentvatne: this line is actually not needed because it's just using dispatch anyhow

brentvatne:
@ericvicenti - agreed, I think that solution is nicer. I wonder if there are other use cases where updating the router would be useful. The main use cases I have in mind right now are adding routes to drawers / tabs, and removing routes from tabs.

If anyone has other cases where you need to use Redux to update your router state like we do here, please chime in here and let us know!

jamonholmgren:
> Also, because routers are entirely stateless, I think it would be way easier to re-create the router if anything changes.

Any reason this shouldn't be the recommended way to approach this issue? I'm all for flexibility, but if it's adding complications to the implementation then it sounds like recreating the router is a reasonable approach.

kevinsperrine:
In my snooping to try to find any-way-at-all to make dynamic routes work, I stumbled on this. 

@ericvicenti 
>Would it work to put "name" in params of FriendScreen, then use a custom contentComponent in the drawer, which asynchronously loads the names?
>In other words, what is the advantage of supporting .navigate('BrentScreen') if you can already do .navigate('FriendScreen', {name: 'Brent'})?

I don't see how listing the friends is possible without replacing large portions of the essentially core behavior in the custom drawer components. I've been trying to make dynamic routes work for weeks, and just keep running into different blockers and haven't been able to make it click. I highly appreciate all the work everyone puts into this library, but there are dozens of issues asking for dynamic routes, but nothing that _really_ shows all the ins-and-outs of making them work. 

In piecing together all the comments about custom components, setting static router props, just passing params, etc. I've started this snack. Can you explain what you mean about asynchronously loading the names and passing them into a custom component? Specially, how these new items can be added to the drawer as different items without running into descriptor errors or duplicate keys. 

See a minimal repo snack here: https://snack.expo.io/rk2ZCc2Jm

Thanks again for all the hard work.

PR #3843, opened 2018-03-26
Update Flowtype to support navigation.navigate({})

gre:
according to implementation, this is legit:

```
navigation.navigate({
  routeName: "EditAccount",
  params: { account },
  key: "editaccount"
})
```

however, the flowtype don't allow that.


ericvicenti:
Yeah this looks like a good improvement. I think in some cases, a different child action might actually work other than the navigate action, but I think this is correct type for the fully-supported use case.

PR #3845, opened 2018-03-26
Fix back button in redux example

schinsue:
# Motivation

Pressing the header back button on the provided redux example doesn't work. Adding `addNavigationHelpers` solves the problem.

# Test plan
Run the ReduxExample and press back button.

### DIFF
--- a/examples/ReduxExample/src/navigators/AppNavigator.js
+++ b/examples/ReduxExample/src/navigators/AppNavigator.js
@@ -1,7 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { StackNavigator } from 'react-navigation';
+import {
+  StackNavigator,
+  addNavigationHelpers,
+} from 'react-navigation';
 
 import LoginScreen from '../components/LoginScreen';
 import MainScreen from '../components/MainScreen';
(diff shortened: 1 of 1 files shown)


padvlad:
Adding support for the Android back button may also help here. Right now it always closes the app. This would look something like this:

```
  componentDidMount() {
      BackHandler.addEventListener("hardwareBackPress", this.onBackPress);
    }
  }

  componentWillUnmount() {
      BackHandler.removeEventListener("hardwareBackPress", this.onBackPress);
    }
  }

  onBackPress = () => {
    const { dispatch, nav } = this.props;
    if (nav.index === 0) {
      return false;
    }
    dispatch(NavigationActions.back());
    return true;
  };
```

dskoneczny:
`addNavigationHelpers` was deleted in this #3619 
https://github.com/react-navigation/react-navigation/pull/3619/files#diff-4ae5155e50686f6a37f9527be1609205L11

edolix:
@padvlad with latest release it seems `hardwareBackPress` callback stop working. @dskoneczny  Can be related with deletion of `addNavigationHelpers ` ?

padvlad:
@edolix ```import { BackHandler } from 'react-native';```
The ```hardwareBackPress``` callback is not included in the ```react-navigation``` package. So this may be an unrelated issue.

edolix:
@padvlad Yes, I did not explain myself sorry. My app uses Redux with `TabNavigator` and nested `StackNavigator`.
With latest release i'm not able to handle the back button on my RootComponent using `BackHandler`, seems that  `onBackPress` never been called.

It's not the right place to discuss it, i know. Just to clarify my previous comment 👍

brentvatne:
`addNavigationHelpers` isn't needed on 2.0

Issue #3785, opened 2018-03-19
Isolate router tests


ericvicenti:
The tests within `src/routers/__tests__` often test several routers at once.

It would be awesome if somebody could help isolate each test to the router after which it is named

PR #3870, opened 2018-03-30
Refactor router tests to isolate child routers

superburrito:
Creating this PR to address https://github.com/react-navigation/react-navigation/issues/3785, not sure if I've made changes in the right direction, please advise!

### DIFF
--- a/src/routers/__tests__/Routers-test.js
+++ b/src/routers/__tests__/Routers-test.js
@@ -89,108 +89,177 @@ Object.keys(ROUTERS).forEach(routerName => {
   });
 });
 
-test('Handles no-op actions with tabs within stack router', () => {
+// Creates a StackRouter with a nested TabRouter
+const createTestRouters = stackRouterConfig => {
   const BarView = () => <div />;
   const FooTabNavigator = () => <div />;
-  FooTabNavigator.router = TabRouter({
+  const TestTabRouter = TabRouter({
     Zap: { screen: BarView },
     Zoo: { screen: BarView },
   });
-  const TestRouter = StackRouter({
-    Foo: {
-      screen: FooTabNavigator,
-    },
-    Bar: {
-      screen: BarView,
-    },
-  });
-  const state1 = TestRouter.getStateForAction({ type: NavigationActions.INIT });
-  const state2 = TestRouter.getStateForAction({
-    type: NavigationActions.NAVIGATE,
-    routeName: 'Qux',
-  });
-  expect(state1.routes[0].key).toEqual('id-0');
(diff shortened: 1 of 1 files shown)


ericvicenti:
Thanks for the contribution!

Unfortunately my issue wasn't clear enough. The main problem is that the stack router is used within the switch router test, and the switch router used from the stack router test. The reason this is done is to make sure sub-routers behave correctly. Ideally we would test the sub-router functionality without actually importing the other router.

In the case of routers-test, I think it does make sense to test several routers at once, so this file does not need to be changed. Although maybe it could be fleshed out a bit more.

As far as the changes you made here, it doesn't actually help the quality of the test. I think the test was easier to read before these changes, because your eyes don't have to jump around as much to understand what is going on. In the future, please try to make tests really really simple easy to read, even if it results in a lot of duplicated code.

I'm closing this PR, but another one would be very welcome, if it gets the stack router out of the switch router test, etc.

superburrito:
@ericvicenti thank you for taking the time to clarify! 

Would be happy to follow up with a different PR, but I'm still a bit confused about how you would go about testing sub-router behaviours without the parent router, esp for different types of routers...

Do you mean that tests involving nested routers -- in `StackRouter-test.js`, for instance -- should be strictly composed of StackRouters? For example, in:
https://github.com/react-navigation/react-navigation/blob/d57d118fda8ad46942d964f45e9303580153a9ee/src/routers/__tests__/StackRouter-test.js#L1082-L1096
we would replace `ChildNavigator.router = TabRouter({ ... })` with `ChildNavigator.router = StackRouter({ ... })`?

And since this  case concerns `setParams`, which is covered in `TabRouters-test.js`,
https://github.com/react-navigation/react-navigation/blob/d57d118fda8ad46942d964f45e9303580153a9ee/src/routers/__tests__/TabRouter-test.js#L127

we know that TabRouters can handle `setParams` in isolation, so we wouldn't need a test case that intermingles StackRouters with TabRouters / SwitchRouters?

ericvicenti:
Instead of testing a totally different router, I think each test should create a dummy/basic child router with very simple behavior, and make sure they continue to work. For the more complicated tests that really do involve situations with stack and switch router, we should move them to routers-test. 

For example we should have a test that covers the use of stack COMPLETE_TRANSITION within the switch router: https://github.com/react-navigation/react-navigation/blob/master/src/routers/SwitchRouter.js#L6

(Ideally this wouldn't be the case, and the switch router would not depend on the stack actions, but it is always good to make sure tests match reality)

superburrito:
I see, would you mind having a look at https://github.com/react-navigation/react-navigation/pull/3876 to see if I got the right idea down?

PR #3898, opened 2018-04-04
[flow] fix flow type for transitionConfig

jemmyphan:
update flowtype for transitionConfig, transitionConfig has 3 parameters since react-navigation/react-navigation@2ee8548


brentvatne:
this appears to be resolved on master now as well. if not then please re-open with conflict resolved and I'll handle it quickly

PR #3899, opened 2018-04-04
container initialization fix

ericvicenti:
This fixes an issue where the container unnecessarily renders twice on startup when persistence is not used.


brentvatne:
good catch :)

PR #3901, opened 2018-04-04
Immediate transition fix, avoid stale render

ericvicenti:
Improves scene filtering within transitioner.

The important change is to avoid ever rendering stale scenes unless a transition is happening.

Yes, this change could be made inside of NavigationScenesReducer, but that seems more painful than this approach, and the whole transitioner is already in need of an overhaul.

### DIFF
--- a/src/views/Transitioner.js
+++ b/src/views/Transitioner.js
@@ -44,11 +44,6 @@ class Transitioner extends React.Component {
     this._queuedTransition = null;
   }
 
-  componentWillMount() {
-    this._onLayout = this._onLayout.bind(this);
-    this._onTransitionEnd = this._onTransitionEnd.bind(this);
-  }
-
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/views/Transitioner.js
```diff
@@ -58,12 +53,15 @@ class Transitioner extends React.Component {
...
   componentWillReceiveProps(nextProps) {
-    const nextScenes = NavigationScenesReducer(
+    let nextScenes = NavigationScenesReducer(
       this.state.scenes,
       nextProps.navigation.state,
       this.props.navigation.state,
       nextProps.descriptors
     );
+    if (!nextProps.navigation.state.isTransitioning) {
+      nextScenes = filterStale(nextScenes);
```
ericvicenti: here's the only behavioral change. Now we will immediately get rid of stale scenes unless a transition has been explicitly requested.

PR #3903, opened 2018-04-04
no more component will mount

ericvicenti:
Aside from Transitioner, this is the only reference to cWM in the repo.

The transitioner willMount is removed here: https://github.com/react-navigation/react-navigation/pull/3901/files


GabrielScalici:
Warning "ComponentWillMount is deprecated" when I try to use Drawer

ericvicenti:
Are you using the new 2.0.0 release candidate? If you can show us a minimal example that demonstrates the warning, please file an issue for it!

PR #3914, opened 2018-04-05
fix for when a StackNavigator has an empty router state

dannycochran:
In my app I have two StackNavigators. One nested inside the other.

If navigating to the nested StackNavigator, then using the Android back button until the app exits, re-opening the app will cause a crash. The nested stack navigator is initialized with an empty stack, so it attempts to get the "key" of undefined.

I can fix this locally by initializing the nested navigator with a dummy route, but it seems like react-navigation should be able to handle nested navigators that are initialized with empty router state.

**Test plan (required)**

In my app I have two StackNavigators. One nested inside the other.

If navigating to the nested StackNavigator, then using the Android back button until the app exits, re-opening the app will cause a crash. The nested stack navigator is initialized with an empty stack, so it attempts to get the "key" of undefined.

### DIFF
--- a/src/getChildEventSubscriber.js
+++ b/src/getChildEventSubscriber.js
@@ -70,8 +70,8 @@ export default function getChildEventSubscriber(addListener, key) {
       const routes = state && state.routes;
 
       const lastFocusKey =
-        lastState && lastState.routes && lastState.routes[lastState.index].key;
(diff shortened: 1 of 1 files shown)


ericvicenti:
How does the stack navigator initialize with an empty stack? From my understanding, it should be impossible to have a navigation state with an empty routes array.

PR #3927, opened 2018-04-07
Fix header transition when mode is set to screen

janicduplessis:
This fixes a regression caused by #3821 that causes the header to translate horizontally when switching between screens with / without headers. This layout interpolation should only be added for headerMode=float.

**Test plan (required)**

Tested the change in an app where I noticed the bug.


janicduplessis:
cc @skevy @ericvicenti

PR #3939, opened 2018-04-09
Fix transition between 2 screens with no header

janicduplessis:
This fixes a broken transition between 2 screens with no header.

https://user-images.githubusercontent.com/2677334/38449859-e9efc2d0-39e2-11e8-81cf-e5832e9effe9.gif

If we detect that we are transitioning between 2 screens with no header just make sure it is hidden, this avoids adding complexity to the already complex interpolation we use.

**Test plan (required)**

Tested that the fix works in the app where I noticed the bug.


janicduplessis:
cc @skevy @ericvicenti

janicduplessis:
@ericvicenti Yea I talked with @skevy about that and the conclusion was that it might be fine since it's easier to understand that way, instead of adding even more complexity to the interpolation.

PR #3940, opened 2018-04-09
Use Header.HEIGHT instead of measuring to avoid flicker

janicduplessis:
The current technique we use to measure the header doesn't work properly because it is async and will cause the screen to jump. The main problem is that there's no way to know the height of the header synchronously for custom headers. For the default one we know the height already so it is pretty easy (although someone could use headerStyles to change the height and would break this).

This is mostly just to discuss the best way to fix this, I don't think using Header.HEIGHT is fine since it breaks custom header height but can be a good temporary patch for now.

I can see 2 possible solutions:

1. Try to read style props to figure out the height (might not be 100% reliable).
2. Add a navigationOption to customize the header height.

**Test plan (required)**

Tested that this fixes the flicker in the app where I noticed it.

### DIFF
--- a/src/navigators/__tests__/__snapshots__/StackNavigator-test.js.snap
+++ b/src/navigators/__tests__/__snapshots__/StackNavigator-test.js.snap
@@ -77,7 +77,6 @@ exports[`StackNavigator applies correct values when headerRight is present 1`] =
     </View>
     <View
       collapsable={undefined}
-      onLayout={[Function]}
       style={
         Object {
           "transform": Array [
@@ -278,7 +277,6 @@ exports[`StackNavigator renders successfully 1`] = `
     </View>
     <View
       collapsable={undefined}
-      onLayout={[Function]}
       style={
         Object {
           "transform": Array [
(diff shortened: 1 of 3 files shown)


janicduplessis:
cc @skevy @ericvicenti @brentvatne

skevy:
@janicduplessis take a look at the NavigationPlayground on master, in the "UIKit-style header transitions" section, on iPhone X. With this PR, there's a black bar at the top of the screen. When I revert it, it looks correct. I think this is probably because Header.HEIGHT is not always the same, especially where the iPhone X is concerned. Any thoughts?

janicduplessis:
@skevy Right, I forgot to test that. I guess iPhone X support is implemented using SafeAreaView. Not sure what’s the best way to go about this, the simpler solution I can think of is to make Header.HEIGHT return the proper height when on iphone X.

janicduplessis:
I'm currently using:
```js
const { width, height } = Dimensions.get('window');
const isPhoneX =
  Platform.OS === 'ios' &&
  !Platform.isPad &&
  !Platform.isTVOS &&
  (height === 812 || width === 812);
```
to detect iphoneX in JS... Far from ideal but it could work for now. We'll have to make sure to test screen rotation too.

PR #3943, opened 2018-04-10
Add component support for header / drawerIcon / drawerLabel options

timwangdev:
if `header` passed to navigation option is a class-based component. The program will fail with error message `TypeError: Cannot call a class as a function`.

Changed variable name due to react-native require component name to be capitalized.

_Screenshot_

1181523346353_ pic_hd

**Test plan (required)**

After add fix

_Screenshot_

1191523346843_ pic

**Code formatting**

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -335,7 +335,7 @@ declare module 'react-navigation' {
    */
 
   declare export type NavigationStackScreenOptions = NavigationScreenOptions & {
-    header?: ?(React$Node | (HeaderProps => React$Node)),
+    header?: ?(React$Node | React$ComponentType<HeaderProps>),
     headerTransparent?: boolean,
     headerTitle?: string | React$Node | React$ElementType,
     headerTitleStyle?: AnimatedTextStyleProp,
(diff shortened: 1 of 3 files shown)


brentvatne:
how are you using the `header` option such that this is needed? `header: ComponentName`?

timwangdev:
Hi @brentvatne ,

I think the places need to update in `navigationOptions` are:
```
drawerLabel
drawerIcon
tabBarLabel
tabBarIcon
```

Others like `tabBarComponent` should already have class support. If you can confirm the list then I could working on extend the PR to include this.

brentvatne:
@timwangdev - that looks right to me, thanks

brentvatne:
@timwangdev - I suppose in this PR you'll only want to update the drawer related options. tabs now live on https://github.com/react-navigation/react-navigation-tabs

timwangdev:
@brentvatne Done. Should we also update the website?

brentvatne:
@timwangdev - yup. oh my I'm sorry for the delay on this response :(

timwangdev:
Sorry that I decided to drop this.

PR #3950, opened 2018-04-11
Fix Easing usage in StackViewTransitionConfigs

timwangdev:
According to `react-native` source https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/Easing.js#L225-L241 `Easing.in` and `Easing.out` takes an `EasingFunction` and return an `EasingFunction`.

The original code has no effect.

**Test plan (required)**

Test via example app on android device.

The code must pass tests.

**Code formatting**


timwangdev:
Sorry, my mistake. It's a type issue. Closed.

PR #3961, opened 2018-04-11
Fix the routing error if the remained path after matched path is empty.

azurechen:
If I click a link `schema://aaa/bbb/ccc?q=123` and setup a path `aaa/bbb/ccc` in `StackNavigator`. The first run of recursive function `getActionForPathAndParams` resolve the correct matched path `aaa/bbb/ccc` from `aaa/bbb/ccc?q=123`. But next run will resolve an incorrect path `q=123` from `q=123` because the path is empty.

The easiest way to fix it is check the length of remained path after the matched path is not zero. A small change can fix this issue.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -523,11 +523,14 @@ export default (routeConfigs, stackConfig = {}) => {
       let nestedAction;
       let nestedQueryString = queryString ? '?' + queryString : '';
       if (childRouters[matchedRouteName]) {
-        nestedAction = childRouters[matchedRouteName].getActionForPathAndParams(
-          pathMatch.slice(pathMatchKeys.length).join('/') + nestedQueryString
-        );
-        if (!nestedAction) {
-          return null;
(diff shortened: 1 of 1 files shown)


brentvatne:
cc @ericvicenti

ericvicenti:
Hi @azurechen, sorry about the delayed review of this. I'd like to understand the nature of the problem a bit better. Can you write a test that confirms what behavior is fixed by this?

You say that the path you're trying to match is `aaa/bbb/ccc?q=123`, which looks like an incorrect path, because `?q=123` should already be parsed out. Instead, your path would be `aaa/bbb/ccc`, and your params would be `{q: '123'}`

If you add a test I will be able to tell for sure if this is a fixed bug or a misunderstanding.

We also may have a bug elsewhere in the library that messes up this path parsing from the full URI. Let me know if you have an example app that demonstrates any bugs here.

brentvatne:
@ericvicenti - I'm moving this to https://github.com/react-navigation/react-navigation-core/issues/12

PR #3972, opened 2018-04-12
AppNavigator.js fixed due to docs

tahaemin:
On docs, you wrote we need to use `addNavigationHelpers` but it is absent on the example code.


brentvatne:
this isn't true on master anymore, the docs aren't relevant for that branch

brentvatne:
(i appreciate the pr though!)

tahaemin:
so will you fix that branch?

Issue #3266, opened 2018-01-12
Custom close/open/toggle route name for DrawerNavigator.


novrain:
### Current Behavior
"DrawerNavigatorConfig" allow user to overwrite drawerOpenRoute/drawerCloseRoute/drawerToggleRoute (it's not mentioned in current document, but after reading the code, we find it works):
```javascript
const DrawerNavigator = (
  routeConfigs: NavigationRouteConfigMap,
  config: DrawerNavigatorConfig = {}
) => {
  const mergedConfig = { ...DefaultDrawerConfig, ...config };
  const {
    containerConfig,
    drawerWidth,
    drawerLockMode,
    contentComponent,
    contentOptions,
    drawerPosition,
    useNativeAnimations,
    drawerBackgroundColor,
    drawerOpenRoute,
    drawerCloseRoute,
    drawerToggleRoute,
    ...tabsConfig
  } = mergedConfig;
```
but in other code, such as DrawerSidebar, still using the default route:
```javascript
class DrawerSidebar extends React.PureComponent<Props> {
  props: Props;

  _getScreenOptions = (routeKey: string) => {
    const DrawerScreen = this.props.router.getComponentForRouteName(
      'DrawerClose'
    );
...
  _onItemPress = ({ route, focused }: DrawerItem) => {
    this.props.navigation.navigate('DrawerClose');
```

### Expected Behavior
We know that maybe it is't a bug, but if pass the custom  DrawerNavigatorConfig  to the DrawerSidebar through DrawerView(maybe some other codes also need to be modified in this way), we can custom the route, then we can nest multiple DrawerNavigator.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |^1.0.0-beta.25
| react-native     |0.51.0
| node             |8.1.3
| npm or yarn      |5.0.3/1.2.3

brentvatne:
in 2.0 you can call `openDrawer()` and `closeDrawer()` and use a technique i describe in https://github.com/react-navigation/react-navigation/issues/1261#issuecomment-387234411 if you want to handle multiple drawers

PR #3978, opened 2018-04-13
Fix/1.5.11 issue 3266 custom drawer route names

joewilson81:
Solves [3266](https://github.com/react-navigation/react-navigation/issues/3266) by allowing custom drawer open/close/toggle routes.

Previously, if you declared a `DrawerNavigator` with `drawerCloseRoute` other than `'DrawerClose'`, you would get an error in `react-navigation` that the `'DrawerClose'` route was not defined.

**Test plan (required)**

`npm run lint`
`npm run jest`


joewilson81:
It appears there are some TabView tests that are failing in the v1.5.11 branch.

PR #3985, opened 2018-04-14
fix for IsFocused() to return a function

slicejunk:
withNavigationFocus is still unusable after this change since 'didFocus' is not fired


brentvatne:
we should change `_isRouteFocused = route => () => {` to `_isRouteFocused = route => {` instead

brentvatne:
fixed with my suggested approach in https://github.com/react-navigation/react-navigation/commit/e5e434c9e267d1302f7ccd740f4154b48c2676db

PR #3986, opened 2018-04-14
test for isFocused props to change on tab switch

slicejunk:
this PR adds a visual feedback on the screen to check the focused props has really changed in the component wrapped with withNavigationFocus.


ericvicenti:
This is an unclear example. If you want to help expand on the navigation playground examples, can you focus on adding clear example use cases?

slicejunk:
withNavigationFocus was buggy and the example did not permit to check this

PR #4017, opened 2018-04-18
Fix header hardcoded height to accound for iPhone X and orientation changes

janicduplessis:
My previous PR did not account for iPhone X and orientation changes, this fixes it.

One main issue that remains is translucent status bar on android, the library as a whole doesn't support it well anyway at the moment so I just left a TODO. 

**Test plan (required)**

Been using this patch for a while and tested it on iPhone X, normal iPhone and Android with orientation changes.

### DIFF
--- a/src/views/StackView/StackViewLayout.js
+++ b/src/views/StackView/StackViewLayout.js
@@ -9,6 +9,7 @@ import {
   View,
   I18nManager,
   Easing,
+  Dimensions,
 } from 'react-native';
 
 import Card from './StackViewCard';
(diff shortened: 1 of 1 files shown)


janicduplessis:
This is pretty terrible but it works ¯\_(ツ)_/¯

cc @brentvatne @ericvicenti

brentvatne:
😆 thanks. we really need some more sane handling of header height but for now, if it works, it works

janicduplessis:
@brentvatne Yea that triggered the gist I sent you 😄

src/views/StackView/StackViewLayout.js
```diff
@@ -450,7 +459,19 @@ class StackViewLayout extends React.Component {
     const headerMode = this._getHeaderMode();
     let marginTop = 0;
     if (!hasHeader && headerMode === 'float') {
-      marginTop = -Header.HEIGHT;
+      const { isLandscape } = this.props;
+      let headerHeight;
+      if (Platform.OS === 'android') {
+        // TODO: Need to handle translucent status bar.
+        headerHeight = 56;
+      } else if (isLandscape && !Platform.isPad) {
```
Minishlink: Hello :) I'm migrating to v2. Was this line supposed to target only iOS? If so we might need to add a `Platform.OS === ios`. On the web it looks like this:
image

PR #4085, opened 2018-04-27
Fix missing/incorrect flow types

roballsopp:
Fixed issues:

- The typedef for [`getParam`](https://reactnavigation.org/docs/navigation-prop.html#getparam-get-a-specific-param-value-with-a-fallback) is missing.

- The return type for `withNavigation` is incorrect per the [flow documentation](https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component) for injecting props with an HOC. The injected prop should be removed from the returned component's prop types. I'm not sure why this doesn't seem to cause an issue in the example app, but flow insisted I need a `navigation` prop in my own app until I fixed this.

- `React$*` is apparently considered private, and seems to cause some issues that `React.*` does not (https://github.com/flowtype/flow-typed/issues/1302, https://github.com/flowtype/flow-typed/issues/1589). My own experience was that `React$*` causes the following cryptic flow error `object prototype is not an object.` when using the `withNavigation` HOC in some cases.

Passes flow, tested on iOS and Android

### DIFF
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -4,6 +4,8 @@
 
 import type {
   NavigationScreenProp,
+  NavigationState,
+  NavigationStateRoute,
   NavigationEventSubscription,
 } from 'react-navigation';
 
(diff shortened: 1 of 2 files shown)


ericvicenti:
Looks great, thanks for these improvements!

PR #4075, opened 2018-04-27
Introduce TypeScript to co-exist with JavaScript

jedmao:
- In response to https://github.com/react-navigation/rfcs/issues/34

### Motivation: First-Class ✈️ TypeScript ❤️ Support 💪 

The [existing TypeScript definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-navigation) have problems:
1. They are manually typed ⌨️ (very difficult to maintain).
1. Kudos to the author(s), but the missing type information is 🤦‍♂️ and not ideal. The definitions don't provide enough information about the most important features of react-navigation.
1. Even if they did, they don't come 📦 with react-navigation, so it's very likely to get definitions that are a bit out of sync.

### Strategy 🤔 

1. Introduce TypeScript into the build system to co-exist with JavaScript files.
1. Incrementally convert existing JavaScript files into TypeScript 👷, starting with the files that have no dependencies and working up. 🚧 
1. New TypeScript files are transpiled in-place with generated type definitions beside them (`.d.ts` files).
1. For now, port [`react-navigation.d.ts` from DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-navigation/index.d.ts) and publicize the typings in `package.json#types` (See f31f7b0fcf7a1c460a833f5107c79913d59195d4 and below this list).
1. At the very end 🏁 convert `src/react-navigation.js` into TypeScript, after which it will generate a 100% accurate `.d.ts` file with all the 🔔 & whistles (e.g., generics).
```json
  "types": "src/react-navigation.d.ts",
```

_Note: If this PR is merged, I'll be happy to continue the effort diligently until all source files are typed with advanced features... or my wife is 😠 at me for staying up too late 🌃._

### Tests ✅ 

The test configuration has also been updated to support TypeScript.
```
npm test
```
1. Lints (pre-task).
1. If successful, runs tests (both JS & TS).
1. If successful, transpiles Typescript into JavaScript in-place (post-task).

### Code formatting 💅 

I ensured that prettier was formatting TypeScript files on the same `precommit#lint-staged` script as well as the introduction of a simple `tslint.json` file, which we can modify as needed.

PR #4096, opened 2018-04-30
getChildEventSubscriber.js -> .ts

jedmao:
Based on #4075 

### Motivation

Providing type information via TypeScript-generated type definitions. This is the first file + test converted to TypeScript, so it sets the stage for how future changes should look regarding converting to TypeScript.

### Test plan

Preserve existing tests. Because the changes are isolated to type information and the test logic is not being changed, I see no reason to split this up any more than it is.


brentvatne:
tracking ts migration in https://github.com/react-navigation/react-navigation-core/issues/2

Issue #4097, opened 2018-04-30
(2.x) Switch Navigator mounting initial component twice


richgilbank:
### Current Behavior
- Using `createSwitchNavigator` to conditionally load screens (like in the [current Auth example](https://v2.reactnavigation.org/docs/en/auth-flow.html)), the initial component is mounted twice.

### Expected Behavior
- Any given component should only be mounted once (unless of course a new screen is explicitly pushed a second time).

### How to reproduce
- [Snack](https://snack.expo.io/@richgilbank/navigation-test)
  - I've simplified it here by making each of the navigation objects screens instead of nested navigators, but it applies in both circumstances
- Open logs panel - note "foo mounted" is logged twice

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.0-beta.10
| react-native     | Expo SDK 26 (https://github.com/expo/react-native/archive/sdk-26.0.0.tar.gz)
| node             | 8.9.4
| npm or yarn      | npm 5.6.0

richgilbank:
rel: https://github.com/react-navigation/react-navigation/issues/4033

PR #4098, opened 2018-04-30
Container state can be out of date if we have async functions between getting and using it

brentvatne:
# Motivation

Fixes https://github.com/react-navigation/react-navigation/issues/4097

# Test plan

TODO: need to come back and do this.


ericvicenti:
This makes a lot of sense! Thanks for catching the problem and clearly commenting!

richgilbank:
Nice! Thanks for the quick work @brentvatne & @ericvicenti! 
Fix confirmed in 2.0.0-rc.7.

PR #4100, opened 2018-05-01
Fix onTransitionStart and onTransitionEnd in StackView

fr1n63:
These functions weren't firing anymore as they are passed as part of navigationConfig. Not sure if this is the best way to fix, but it solves my issue.


brentvatne:
cc @ericvicenti

allmaxgit:
@fr1n63 Thank you, now I see what we can pass it in component:
```
<Navigator onTransitionStart={this.onTransitionStart} />
```
And this is better then using navigationConfig. Because we can connect component to redux and dispatch any action.

fr1n63:
I didn't make any changes to yarn.lock directly, reverted.

brentvatne:
@ericvicenti - can you review when you have a moment?

fr1n63:
Ha! we were using it in v1, and as part of upgrading to v2 this was an issue - have since refactored and don't actually need it now I'm on v2. I'll take another swing at it when I get a minute.

davelnewton:
What's the status on this? The way my app is structured now I need to use `onTransitionStart` in the config object passed to `createStackNavigation`; only `onTransitionEnd` is firing, which is too late for what I need :/

gp3gp3gp3:
Can we update the documentation to cover this use case?  The `createStackNavigator` [API documentation](https://reactnavigation.org/docs/en/stack-navigator.html#stacknavigatorconfig) still has `onTransitionStart` as a prop passed in, though it doesn't do anything passed as a config object to that function.

brentvatne:
@gp3gp3gp3 - this was fixed - https://github.com/react-navigation/react-navigation/commit/ebdd2da79f34d43e2a3a4629a3d4f15ccfce53e9 and https://github.com/react-navigation/react-navigation/commit/fa4fdb9c57b3961cf1eff55560f5d40d178c2a4e (both released in 2.7.0 but i'd update to latest, currently 2.9.1)

gp3gp3gp3:
Whoops!  Sorry, I thought I was on the latest, looks like I'm still at 2.6.1.  Thank you!

lcampanis:
@gp3gp3gp3 @brentvatne Unfortunately this is still a bug for `onTransitionStart`.
```
"react-navigation-stack": "^2.0.0-alpha.37",
"react-navigation-tabs": "^2.6.0",
```

I can confirm that on TabNavigator only `onTransitionEnd` works.
On StackNavigator, none work.

Also ended up using `onNavigationStateChange`

PR #4101, opened 2018-05-01
NavigationActions.js -> .ts

jedmao:
Based on #4075

### Motivation

Providing type information via TypeScript-generated type definitions. This is the 2nd file + test converted to TypeScript, but the first one to add documentation (referenced from [here](https://reactnavigation.org/docs/navigation-actions.html)).

### Test plan

Preserve existing tests. Notice that the tests in this PR are 100% untouched. This is ideal, of course.

image


brentvatne:
tracking ts migration in https://github.com/react-navigation/react-navigation-core/issues/2

PR #4103, opened 2018-05-01
Fix initial event in non-persisted container

ericvicenti:


ericvicenti:
Skip this, use https://github.com/react-navigation/react-navigation/pull/4104

PR #4108, opened 2018-05-02
[help wanted] Fix missing/incorrect flow types for 1.x branch

roballsopp:
These changes were merged to master already, but they also belong in 1.x

Fixed issues:

- The typedef for [`getParam`](https://reactnavigation.org/docs/navigation-prop.html#getparam-get-a-specific-param-value-with-a-fallback) is missing.

- The return type for `withNavigation` is incorrect per the [flow documentation](https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component) for injecting props with an HOC. The injected prop should be removed from the returned component's prop types. I'm not sure why this doesn't seem to cause an issue in the example app, but flow insisted I need a `navigation` prop in my own app until I fixed this.

### DIFF
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -17,7 +17,11 @@ type MyNavScreenProps = {
   banner: React.Node,
 };
 
-class MyBackButton extends React.Component<any, any> {
+type MyBackButtonProps = {
+  navigation: NavigationScreenProp<*>,
+};
+
+class MyBackButton extends React.Component<MyBackButtonProps, any> {
   render() {
     return <Button onPress={this._navigateBack} title="Custom Back" />;
   }
@@ -142,7 +146,7 @@ class MyPhotosScreen extends React.Component<MyPhotosScreenProps> {
     const { navigation } = this.props;
     return (
       <MyNavScreen
-        banner={`${navigation.state.params.name}'s Photos`}
+        banner={`${navigation.getParam('name')}'s Photos`}
         navigation={navigation}
       />
     );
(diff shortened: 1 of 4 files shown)


roballsopp:
investigating failure. i only made flow-type changes, so unsure why the snapshot is different.

roballsopp:
This test appears to fail in the base 1.x branch without my changes. Seems some code got in without a snapshot update. Should I just update the failing snapshot and push it up with this pr?

roballsopp:
I went ahead and pushed the snapshot tweaks..

roballsopp:
ok now there is a cryptic error coming from jest that appears to mean you need to update: https://github.com/facebook/jest/issues/5401. getting in a little farther than I'm comfortable. can anyone look at this?

roballsopp:
Awesome no prob! Sorry I'm not sure how merging is set up with this repo. Should I be able to merge this myself, or will someone else come along and get it in?

Ashoat:
I can merge it, but I was hoping to get some clarification on the Jest failure first. It looks like other recent PRs aren't seeing that failure for some reason, but based on Googling the failure doesn't seem to be related to the changes in the PR.

@brentvatne, any idea what's going on? [https://github.com/facebook/jest/issues/5913](This issue) seems to be the root cause, but [https://github.com/facebook/jest/issues/5401](this issue) has a bit more detail.

roballsopp:
No luck on the jiggering. Same result the second time.

Ashoat:
Looks like #4145 has the same test issue, so I think it's safe to conclude that there is something wrong with the CI/test suite and it doesn't have to do with this PR.

Thanks for the PR @roballsopp and sorry for the delay.

roballsopp:
Awesome thanks for the merge!

PR #4115, opened 2018-05-03
Fix CompletionAction trigger

ericvicenti:
The completion action should always fire when the current navigation state isTransitioning.

There were some problems surrounding this with persisted states and isTransitioning.


ericvicenti:
ping @brentvatne

luongs3:
was this issue solved? I still got it at version 2.14.

Issue #4118, opened 2018-05-04
Swipe back gesture in RTL mode is broken


spaceye:
### Current Behavior

Swipe back doesn't work in RTL mode by default and even when you set `navigationOptions.gestureDirection` to 'inverted' value to facilitate it functionality remains broken: one is able to swipe back but stack card responds to user's gesture in a jerky near-edge movement.

### Expected Behavior

Swipe back works just fine in RTL mode by default. If one needs to override it they can do it even when the RTL mode is on.

### How to reproduce

- Add a StackNavigator with some routes to your setup;
- Force app into RTL mode with `I18nManager.forceRTL(true);` call;
- Navigate from one route to another;
- Try to swipe back.

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.0-rc.1 (also on the latest release)
| react-native     | 0.54.4
| node             | 9.10.1
| npm or yarn      | 1.5.1

spaceye:
Initial functionality was added in https://github.com/react-navigation/react-navigation/pull/3077 however there is a bug in there and I'd like it to use RTL by default as everywhere else in the code we do just that (RTL is supported in all the transitions) with an ability to override it if one chooses so.

brentvatne:
thanks for the pr in #4119!

spaceye:
You're welcome!

mohamadian:
please visit #3325

PR #4119, opened 2018-05-04
Fix for broken swipe back gesture in RTL mode

spaceye:
**Motivation**

Swipe back doesn't work in RTL mode by default and even when you set navigationOptions.gestureDirection to 'inverted' value to facilitate it functionality remains broken: one is able to swipe back but stack card responds to user's gesture in a jerky near-edge movement.

**Test plan (required)**

I've tested the new code in the app we develop (however I'm not at liberty to share screenshots / video due to NDA).

Steps to reproduce original issue are given at https://github.com/react-navigation/react-navigation/issues/4118.

P.S. I also fixed the default back button's mask in RTL mode.

Thanks guys!

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -7,6 +7,7 @@ import {
   Platform,
   StyleSheet,
   View,
+  I18nManager,
   ViewPropTypes,
 } from 'react-native';
 import { MaskedViewIOS } from '../../PlatformHelpers';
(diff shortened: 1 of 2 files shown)


ericvicenti:
These changes make sense, thanks for the contribution.

I wish this stuff was easier to test in the Navigation playground, so we can check for regressions more frequently. Please send a PR for this if you have time!

PR #4124, opened 2018-05-05
Fix regression causing setting undefined value to isTransitioning

angly-cat:
PR #3393 introduced a regression: in some cases a value of `isTransitioning` property of navigation state is lost.

In that PR there's a comment:

> // Return state with new index. Change isTransitioning only if index has changed

However, if index hasn't changed, `isTransitioning` becomes `undefined`. This PR makes `isTransitioning` to be indeed unchanged if index hasn't changed.

**Test plan**

You can check [this snack](https://snack.expo.io/@angly/istransitioning-issue) for demonstration. `stateAfterSecondNavigation` should have saved `isTransitioning` value from the previous state, which was `true`, but it became `undefined` (and because of that Expo's logging just don't print it at all). With this PR, it will preserve a value of `isTransitioning`.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -327,7 +327,7 @@ export default (routeConfigs, stackConfig = {}) => {
             isTransitioning:
               state.index !== lastRouteIndex
                 ? action.immediate !== true
-                : undefined,
+                : state.isTransitioning,
             index: lastRouteIndex,
             routes,
           };


ericvicenti:
Looks great! Thanks for the repro code, clear description, and for fixing!

Could you do us a favor and write a test that verifies this behavior and prevents it from regressing again in the future?

Issue #4129, opened 2018-05-05
V2 Drawer closing when setting navigationOptions params


violabg:
### Current Behavior

I'm using a drawer as a filters panel for searching products on the main stackView. 
I'm using a contentComponent for that, so the drawer doesn't have links. When I change a filter in the drawer panel, I call an api that return products, I display the count on the navigation header on the main view, by setting the params in the navigationOptions.
This closes my drawer each time I set the params.

### Expected Behavior

I want to keep the drawer open until I click outside it and close it manually, and this was how it was working in react-navigation V1

### How to reproduce
I wrote a sample code to illustrate the problem, each time you click a button on the drawer panel it closes, if you change the code to use react-navigation V1 instead of V2, you'll see that it remains open.

https://snack.expo.io/@mediacross/test-drawernavigation

brentvatne:
cc @ericvicenti

ericvicenti:
Fix incoming, thank you for the clear bug report!

PR #4131, opened 2018-05-05
Fixes to drawer router handling of child routers

ericvicenti:
Fixes to drawer router handling of child routers

Addresses #4129, and also lets child routers swallow actions with null. Tests added


ericvicenti:
ping @brentvatne

PR #4132, opened 2018-05-05
Fix drawer stack reset-to-top behavior

ericvicenti:
“reset” is not a NavigationAction anymore.

The correct fix, for v3, to this is to navigate to the first screen inside the stack. With the less-pushy navigate behavior in v2, this will result in the first route getting selected if you specify it by routeName.


ericvicenti:
Ok, we will delete this behavior in v3.

For now this just fixes the stack action reference

PR #4143, opened 2018-05-07
Build step for and web package compatibility

ericvicenti:
use the react-native package field to point the RN packager the source version. The main version is the compiled.

Build can run with `yarn build`. With prepublishOnly flag the build will run automatically as a part of `npm publish`. Also a build is added during test run to make sure no errors happen.


ericvicenti:
I'll be doing this as a part of the universe experiment now

PR #4145, opened 2018-05-07
Added ability to set titleOffset via options

xxsnakerxx:
Header component has hardcoded value of `titleOffset`, this PR add an ability to change the value via `navigationOptions`

supported number(set both left and right) or object ({ left: number, right: number })

Before:
```
| < Back               My awesome ti...                   |
```
After:  
```js
    title: 'My awesome header',
    titleOffset: 50,
})
```
```
| < Back               My awesome title                   |
```

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -232,6 +232,19 @@ class Header extends React.PureComponent {
     const style = {};
     const { transitionPreset } = this.props;
 
+    const details = this.props.getScreenDetails(props.scene);
+    const { titleOffset } = details.options;
+
+    if (titleOffset) {
+      if (typeof titleOffset === 'number') {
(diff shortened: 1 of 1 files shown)


brentvatne:
what's the use case here?

xxsnakerxx:
The main goal of this is to give a title more space, especially on small iOS devices such as 5s, SE... when you have 320px screen width, back icon and right button "Done" long title will be truncated as I describe in the PR description... in this case title have width about 180px but it can be longer...

brentvatne:
makes sense, thanks for the info @xxsnakerxx

brentvatne:
can you submit this PR against the master branch? not adding new features to the 1.x branch anymore

xxsnakerxx:
Ok, sure

PR #4161, opened 2018-05-08
[Flow] Add missing NavigationActions.replace type definition

vpodolyan:
This PR adds type definition for `NavigationActions.replace()` method


skabbes:
@vpodolyan replace is only available on StackActions now - as you can see from [the docs](https://reactnavigation.org/docs/en/navigation-actions.html)

vpodolyan:
@skabbes oh, okay, will make PR for `1.x` branch

PR #4168, opened 2018-05-08
Fix dispatch implementation

ericvicenti:
Previously we had been emiting an action event with a null state when the router handles the action by returning null from getStateForAction.

- Fixes the case of null state getting emitted
- Renames a few things for clarity
- Refactors conditionals to read easier
- Comments to explain intent


ericvicenti:
This fixes the horrible misbehavior of isFocused and the willFocus/didFocus event that happens when you navigate to an existing screen.

PR #4172, opened 2018-05-09
fix swipe gesture

### LINKED ISSUE #4144: IOS native swipe gesture animation freezes InteractionManager.runAfterInteractions
### Current Behavior
If you navigatie from screen A to screen B and quickly swipe back to screen A, InteractionManager wont execute tasks which causes the flatlist to stop rendering new items. You can still see the previously rendered items but it wont render any new one so you just see a grey background.

Screen A contains a flatlist
Screen B detail page

The issue is caused by the PanResponder inside CardStack.js because on every render a new PanResponder is created. This is an issue since you can end up working with a old reference. In some cases like in which i describe it will call onResponderGrant on the previous PanResponder and will call onResponderEnd on the newly created PanResponder. And since PanResponder.onResponderEnd calls clearInteractionHandle which will only clear the interaction handle if there is one present it will never clear the interaction handle. So InteractionManager wont be able to run any of its scheduled tasks.

FlatList which is basically just a wrapped VirtualizedList uses react native its Batchinator to update the cells to render.
```javascript
this._updateCellsToRenderBatcher = new Batchinator(
  this._updateCellsToRender,
  this.props.updateCellsBatchingPeriod,
);
```
```javascript
class Batchinator {
  _callback: () => void;
  _delay: number;
  _taskHandle: ?{cancel: () => void};
  constructor(callback: () => void, delayMS: number) {
    this._delay = delayMS;
    this._callback = callback;
  }
  /*
   * Cleanup any pending tasks.
   *
   * By default, if there is a pending task the callback is run immediately. Set the option abort to
   * true to not call the callback if it was pending.
   */
  dispose(options: {abort: boolean} = {abort: false}) {
    if (this._taskHandle) {
      this._taskHandle.cancel();
      if (!options.abort) {
        this._callback();
      }
      this._taskHandle = null;
    }
  }
  schedule() {
    if (this._taskHandle) {
      return;
    }
    const timeoutHandle = setTimeout(() => {
      this._taskHandle = InteractionManager.runAfterInteractions(() => {
        // Note that we clear the handle before invoking the callback so that if the callback calls
        // schedule again, it will actually schedule another task.
        this._taskHandle = null;
        this._callback();
      });
    }, this._delay);
    this._taskHandle = {cancel: () => clearTimeout(timeoutHandle)};
  }
}
```
PanResponder.js
```javascript
function clearInteractionHandle(
  interactionState: {handle: ?number},
  callback: Function,
  event: Object,
  gestureState: Object
) {
  if (interactionState.handle) {
    InteractionManager.clearInteractionHandle(interactionState.handle);
    interactionState.handle = null;
  }
  if (callback) {
    callback(event, gestureState);
  }
}
```
**Example project https://github.com/wesleymooiman/react-navigation-bug-example**

### Expected Behavior

Its should clear the correct handle so InteractionManager knows its handle has been cleared and can process scheduled tasks.

### How to reproduce

Create an app with a StackRouter which contains two screens, a list screen which use a flatlist and a detail page. Navigate from screen A with the flatlist to screen B and quickly swipe back before screen B has completely loaded. Now scroll inside the FlatList and notice it wont render any new items. 

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.5.11
| react-native     | 0.53.0
| node             | 9.11.1
| npm or yarn      | 1.6.0

jk625x:
* Fixes #4144 
Create the PanResponder in "componentWillMount()", make sure that "onResponderEnd()" can be called to clear the interaction handle.


ericvicenti:
> make sure that "onResponderEnd()" can be called to clear the interaction handle.

How did you do this, exactly?

Are you sure we shouldn't do this in the constructor instead of `componentWillMount`? Its sorta deprecated now. We can just set an instance variable like `class Stack { _panResponder = ... }`

ericvicenti:
Ah, I see. We definitely do want to cache the pan responder here. I cleaned up your changes here and moved it to an instance var instead of deprecated componentWillMMount

PR #4176, opened 2018-05-09
Fix exposing undefined actions in addNavigationHelpers

skabbes:
While upgrading our codebase to react-nav 2.0, I noticed that there were some uses of NavigationActions internally within react-navigation that _appeared_ accidental.  Further reviewing the code, this does not appear to part of the public API (however, we were using it for react-native-web support).  So potentially this should be removed entirely, in favor of `router.createActionCreators`.

I'll leave it up to the maintainers to determine what the correct fix is.  This just popped up in my search when I was migrating `NavigationActions` -> `StackActions` in our internal code, and I thought I'd raise it here.

### DIFF
--- a/src/addNavigationHelpers.js
+++ b/src/addNavigationHelpers.js
@@ -51,14 +51,6 @@ export default function(navigation) {
       );
       return navigation.dispatch(NavigationActions.navigate(navigateTo));
     },
-    pop: (n, params) =>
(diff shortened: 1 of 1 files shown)


skabbes:
Closing in favor of https://github.com/react-navigation/react-navigation/pull/4179

PR #4179, opened 2018-05-09
Remove unused code addNavigationHelpers

skabbes:
While upgrading our codebase to react-nav 2.0, I noticed that there were some uses of NavigationActions internally within react-navigation that appeared accidental. Further reviewing the code, this does not appear to part of the public API and only used for tests.

Just practicing some good hygiene here and deleting.


skabbes:
@brentvatne based off our conversation in slack.

brentvatne:
thanks!

benjaminbillet:
How are we supposed to do this https://v2.reactnavigation.org/docs/en/redux-integration.html without "addNavigationHelpers"?

pilwon:
@skabbes @brentvatne Is `addNavigationHelpers` no longer supported by this library? If so, we may need to update this page as well: https://reactnavigation.org/docs/en/redux-integration.html

brentvatne:
removed! will be updated on next deploy

pilwon:
@brentvatne Thanks-

pilwon:
@brentvatne I noticed you put `/* more setup code here! this is not a runnable snippet */` in [the new doc](https://github.com/react-navigation/react-navigation.github.io/commit/9231f12d2920eb0e6b556eccc75def93cb853d40#diff-b5f1f7006bb9950937730bd3d60af781R194). What value goes in this `prop` and what more setup is necessary for redux integration? Thanks-

brentvatne:
read other parts of redux docs for that, it's the same

PR #4188, opened 2018-05-10
Add <NavigationEvents/> component

slorber:
Declarative component to register event listeners, as discussed in RFC https://github.com/react-navigation/rfcs/issues/41


slorber:
@brentvatne @satya164 @ericvicenti here is the PR as discussed in RFC.


1) There is no test. Any advice on how to test that? I'll try to mock the navigation/addListener and see existing tests related to event system

2) I noticed a weird behavior on my example. I'm not sure it's related to my code but the 2nd time we enter the NavigationEvents example, the events are not fired anymore. I suspect it's something broken in the lib because the "emit()" function is not even called so it does not look like an event listener registration problem

slorber:
Hmmmm it seems the 3) issue with events not firing has been fixed in a recent commit, after updating from upstream the problem just disappeard :)

slorber:
I've added some tests + flow declaration. 

If someone can check the flow declaration that could be nice, not exactly sure what I'm doing.

I'll do the documentation in another PR as it's out of this repository.

Should I support the "action" listener? I've seen this in the code but it's not documented.

Should I support case where the provided navigation prop change over time (ie unregister all listeners from navigation1, and re-register them on navigation2?)

src/views/NavigationEvents.js
```diff
@@ -0,0 +1,61 @@
...
+        this.addListener(eventName);
+      }
+    });
+  }
+
+  componentWillUnmount() {
+    EventNames.forEach(this.removeListener);
+  }
+
+  addListener = eventName => {
```
vonovak: as an alternative, we could consider

```
addListener = eventName => {
  this[eventName] = (...args) => {
    const listener = this.props[EventNameToPropName[eventName]];
    listener && listener(...args)
  }

  this.subscriptions[eventName] = this.props.navigation.addListener(
    eventName,
    this[eventName]
  );
};
```

Or, more verbosely (to show what I mean):

```
class NavigationEvents extends React.Component {
  willBlur = (...args) => {
    const { onWillBlur } = this.props;
    onWillBlur && onWillBlur(...args);
  };

  addListenerForWillBlur = () => {
    this.subscriptions.willBlur = this.props.navigation.addListener('willBlur', this.willBlur);
  };
}
```

This would allow to remove `componentDidUpdate` and the remove + add listener calls. Just subscribing once in `componentDidMount`, unsubscribing in `componentWillUnmount` and not caring about the changes of the props of `NavigationEvents`.

slorber: @vonovak I've proposed a similar implementation initially, using "internal listeners", but @satya164 suggested I subscribe user-provided listeners directly [here](https://github.com/react-navigation/rfcs/issues/41#issuecomment-387016663)

I'm ok for your proposal, the only drawback is that if user provides a single callback, you would have 4 listeners being subscribed, with 3 of them doing nothing. Not a big deal IMHO so if we settle on that I'll make the change.

slorber:
Hi @brentvatne @ericvicenti 

Any feedback on the PR? 
What about the implementation comment of @vonovak ?
Should I start writing documentation about the component?

Just rebased it on upstream, all tests passing on CI except the one that is broken on master: 

> NavigationContainer › warnings › detached navigators › warns when you render more than one navigator explicitly

Also this is weird but locally I get this test failing (not on CI), any idea?

image

src/views/NavigationEvents.js
```diff
@@ -0,0 +1,61 @@
...
+  willFocus: 'onWillFocus',
+  didFocus: 'onDidFocus',
+  willBlur: 'onWillBlur',
+  didBlur: 'onDidBlur',
+};
+
+const EventNames = Object.keys(EventNameToPropName);
+
+class NavigationEvents extends React.Component {
+  constructor() {
```
ericvicenti: delete this?

src/views/NavigationEvents.js
ericvicenti: Hmm, I think @satya164's suggestion makes a lot of sense and this implementation is quite good. It may be a tiny bit harder to read but it prevents creation of unnecessary upstream subscriptions

src/views/__tests__/NavigationEvents-test.js
```diff
@@ -0,0 +1,164 @@
...
+    );
+    checkAddListenerCalled(4);
+    checkAddListenerCalledWith('willBlur', onWillBlur);
+    checkAddListenerCalledWith('willFocus', onWillFocus);
+    checkAddListenerCalledWith('didBlur', onDidBlur);
+    checkAddListenerCalledWith('didFocus', onDidFocus);
+  });
+
+  it('attach all listeners with navigation context', () => {
+    const component = renderer.create(<TestComponent />);
```
ericvicenti: these tests seem to do the job, but personally I'm a fan of avoiding much work in `beforeEach` and instead making it so that each test reads more like a standalone story. The way you have it here, I need to read the whole complicated test harness to understand a single test. This results in longer test files, but I think the improved readability makes it worthwhile.

I'd say its fine for now but in the future you may want to keep the whole test isolated to within the `it(` callback

examples/NavigationPlayground/js/TabsWithNavigationEvents.js
```diff
@@ -0,0 +1,127 @@
...
+                    backgroundColor: '#e4e4e4',
+                  }}
+                >
+                  <Event event={item} />
+                </View>
+              )}
+            />
+          </View>
+
+          <NavigationEvents
```
ericvicenti: this example looks great for core developers like ourselves who want to test the library, but its a terrible example of how events might be used in practice. This is fine for now, but in the future I would recommend making an example for each purpose.

slorber: Yes I agree. Maybe I could make an example for the scenario where you want to refresh data on focus, but not if this is a focus triggered by a back navigation? (like it was my original usecase)

Do you think of other meaningful/general patterns that make sense with event?

satya164: > refresh data on focus, but not if this is a focus triggered by a back navigation

So basically on mount?

slorber: Simplified usecase: 

- I have a tab navigation with 3 tabs containing small lists (scrollable) of items each.
All items can be pressed and open a new item detail stacknav screen on top of the scrollable tab screen (like a popup)
- I want tabs data to refresh on tab focus (mount, but also swipes/tab press), but don't want to refresh in case of a back event from the items detail screen, because this would mess-up with the scroll position.

slorber:
Hi @ericvicenti thanks for the review. I made changes to the PR:

- removed the useless constructor
- refactored the test to avoid `beforeEach`. There are still some helper functions to avoid some duplication, but this should be more easy to read

The local failure persists after `rm -rf node_modules` + `yarn install`

ericvicenti:
Ah, it looks like these flow errors are legitimate. I can reproduce them locally with the following:

```
cd examples/NavigationPlayground
yarn
yarn test
```

slorber:
Thanks @ericvicenti just fixed the problem, didn't know there was another command to run playground tests

brentvatne:
@ericvicenti - let's discuss whether this is a good fit for the library itself or if it should live in a 3rd party lib

slorber:
Hi @brentvatne 

I feel like this component is of the same kind as `withNavigationFocus` HOC: it can  be built in userland yet is convenient enough to be useful for many. If it were to be put in a `react-navigation-extras` package or something, maybe it's worth keeping these 2 components togethers so that NavigationEvents does not feel too lonely ;)

slorber:
@ericvicenti @brentvatne have you decided where this component should live yet?

brentvatne:
yeah - @ericvicenti believes this would be a good fit for react-navigation itself, so let's ship it. are there any outstanding changes you would like to make first?

slorber:
great :)

nothing to add for me, except planning a new PR for some doc

also, it seems the build is now failing, not sure it's related to the code itself

slorber:
great!

about the doc, should I just create a new api reference page in .md with some frontmatter and that's all needed?

I mean copy and modify a page like this one: https://github.com/react-navigation/react-navigation.github.io/blob/45c3e0279e50e07e8701442b15f4df2f26e19828/docs/stack-actions.md

brentvatne:
ya and add to sidebars.yml (I think, afk)

slorber:
pr for doc: https://github.com/react-navigation/react-navigation.github.io/pull/211

PR #4189, opened 2018-05-10
[Flow] Add missing NavigationActions.replace type definition

vpodolyan:
This PR adds type definition for `NavigationActions.replace()` method


vpodolyan:
Sorry, just mixed up branches :grin:

PR #4195, opened 2018-05-10
Use Animated.Image for back button

R1ZZU:
**Motivation**:
Current behavior: when tintColor is an animated value (e.g. interpolated scroll position) only title color is changed, back button doesn't respect this changes

**Fix**:
Use `Animated.Image` instead of `Image` for default back button

### DIFF
--- a/src/views/Header/HeaderBackButton.js
+++ b/src/views/Header/HeaderBackButton.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import {
+  Animated,
   I18nManager,
   Image,
   Text,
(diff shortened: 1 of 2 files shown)


brentvatne:
there is extra overhead associated with using an `Animated` component and I don't think this is a common enough use case to justify it. additionally, it's not possible to animated colors with the native driver currently, and so this will not work when we introduce react-native-gesture-handler for stack gestures. I would instead render a custom back button component

brentvatne:
that all said, thanks for the pr! if you have another idea for how to do this more easily without adding overhead, feel free to open another pr!

R1ZZU:
@brentvatne Would you consider a PR with additional `<AnimatedBackButton />` component or smth like `animatedBackButton` field in `navigationOptions` in order to provide same styling as you have ATM, but with possibility to animate color?

PR #4202, opened 2018-05-11
Apply DrawerLockMode from DrawerScreen options to DrawerLayout component

### LINKED ISSUE #4201: DrawerLockMode on drawerScreen is ignored
### Current Behavior

Specifying `drawerLockMode` option on drawerScreen does not apply on `DrawerLayout` component.

### Expected Behavior

Should be apply.

### How to reproduce

```
const App = createDrawerNavigator(
  { Home: {
      screen: AppNavigator,
      navigationOptions: { drawerLockMode: 'locked-closed' }
    }
  }
);
```

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0
| react-native     | 0.54.2
| node             | latest
| npm or yarn      | latest

sebastienbarbier:
While migrating from v1 to v2, I lost the `navigationOptions through tree` feature and had to implement an alternative way to enable/disable the `DrawerLockMode` on `DrawerNavigator` .

> In previous version of React Navigation, the library used to dig through your component tree to find navigationOptions. This is no longer the case, and only navigationOptions on screen components of a particular navigator will apply to that navigator. You can read more about this in the Navigation options resolution guide.
[react-navigation v2 documentation](https://reactnavigation.org/docs/en/common-mistakes.html)

Looking at the code, seams like `DrawerLockMode` from `DrawerScreen options` is **read, initialised, but never used**. Might be a simple mistake, but this allow to be more flexible and I think would make sense to have 👍

Fix #4201

### DIFF
--- a/src/views/Drawer/DrawerView.js
+++ b/src/views/Drawer/DrawerView.js
@@ -93,6 +93,7 @@ export default class DrawerView extends React.PureComponent {
           this._drawer = c;
         }}
         drawerLockMode={
+          drawerLockMode ||
           (this.props.screenProps && this.props.screenProps.drawerLockMode) ||
           this.props.navigationConfig.drawerLockMode
         }


vladikoff:
@brentvatne @ericvicenti r? might be good for the next RC

brentvatne:
thanks!

sebastienbarbier:
🎉, my pleasure 🙏

PR #4212, opened 2018-05-12
passActiveRouteOption to support nested configs

ericvicenti:
New API to support nested configuration, such as the header title within the “StacksOverTabs” example. In v2, this nested configuration behavior is explicit.

While I’m happy with the new API and don’t see any bugs, this is a pretty invasive change to the core code, and we should review the changes to the navigation prop. The following new pieces of API have been added:

- passActiveRouteOption. See how its used in `StacksOverTabs.js`
- router.childRouters, a map of routeName to child router. Needed for static access to child router
- navigation.router refers to the router that maintains the navigation state. This should be safe
- navigation.childrenNavigation is a map of route key to child navigation objects. This might be ok, but might be a footgun
- navigation.getScreenProps, added as a convenience and I think it is safe. Maybe should rename to discourage screenProps usage altogether.
- navigation.getChildNavigation. This is how all children navigation props are accessed, and is important for the new nested configurer to get access to the child navigation prop

Lots of changed code:

- big changes to navigation prop and createNavigationContainer will break redux support
- dangerously get parent seems less dangerous now, so I use it within passActiveRouteOption
- subscriptions (addListener) are now cached inside the child navigation creator

So this does seem to address the problem fine, but I think we should be careful moving forward here. Careful code review would be appreciated!

### DIFF
--- a/examples/NavigationPlayground/js/StacksOverTabs.js
+++ b/examples/NavigationPlayground/js/StacksOverTabs.js
@@ -8,6 +8,7 @@ import {
   SafeAreaView,
   createStackNavigator,
   createBottomTabNavigator,
+  passActiveRouteOption,
 } from 'react-navigation';
 
 import Ionicons from 'react-native-vector-icons/Ionicons';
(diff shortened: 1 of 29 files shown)


ericvicenti:
Lets go with https://github.com/react-navigation/react-navigation/pull/4425 instead, its a less-breaking step in the same direction

PR #4213, opened 2018-05-12
Initial nav state is always asynchronous (for Linking.getInitialUrl)

Minishlink:
Currently, when there is an initial URL, the navigation state is (by default) constructed as such:

1. it is set to the default initial state in the NavigationContainer constructor
2. an action corresponding to the URL is played on top of it during componentDidMount, asynchronously

This has two consequences:

1. The Navigator is rendered twice at start-up, and on the first render, it mounts the component from the default initial state
2. If there is some navigation logic on this component (as it is often the case for authentication purposes), the Navigator might not successfully apply the URL because of conflict with the default initial state

What I propose is to always consider that the navigation state is hydrated asynchronously. I'm not sure why it wouldn't be the case seeing that `Linking.getInitialUrl` is.

This PR will enhance support for deep-links on native and routing on the web.

**Test plan (required)**

I'd like to have a go from a reviewer first because it's not a small change and there's plenty of tests to fix (they're not yet made to wait that componentDidMount is finished). After this, I'll be happy to add tests ! :)


Minishlink:
@ericvicenti I pushed a fix for the test of TabNavigator. I'm not sure about the approach, can you confirm I can apply the same approach for the other tests or should I take a new one?

ericvicenti:
Aha, if that change fixes the snapshot test, then it looks great! Definitely would want that for the other navigators.

Can you add a comment that explains why its needed, because the state only actually gets set on componentDidMount. Also you may want to assert that the initial render is null or the result of `renderLoadingExperimental`

Minishlink:
@ericvicenti Thanks, should be good for review 👍

brentvatne:
@ericvicenti - my concern about this is I think the vast majority of apps won't rehydrate navigation state, and I don't want to make them pay the cost (async startup) for it in that case. maybe I misunderstand though, let's catch up on slack about this

brentvatne:
I think the spirit of this is right, but probably what we should do instead of making this async is expose some native APIs that let us get the initial URL and state synchronously. I really appreciate this PR and if you're interested in pursuing the sync approach further, or making this behavior opt-in until someone gets around to that, please reopen this PR on https://github.com/react-navigation/react-navigation-native with appropriate changes!

Minishlink:
@brentvatne Will try do to it next week, thanks! I'll also need to update the appcontainer for web (though it'll be easier to do since getting the current URL is synchronous)

PR #4219, opened 2018-05-13
Fix drawer router initial state

Minishlink:
In `DrawerRouter.getStateForAction`, if the router had no initial state, the state was computed two times ([here](https://github.com/react-navigation/react-navigation/blob/master/src/routers/DrawerRouter.js#L30) and [here](https://github.com/react-navigation/react-navigation/blob/master/src/routers/DrawerRouter.js#L62)), and thus the returned state was wrong.
My fix mimics the StackRouter's logic [here](https://github.com/react-navigation/react-navigation/blob/master/src/routers/StackRouter.js#L223).

**Test plan (required)**

Added a test.


ericvicenti:
👏👏 Thank you for this fix!

Not sure why the tests are failing and I don't have much time to explore now, but this is ready to land as soon as we get things green.

Minishlink:
@ericvicenti Thanks! This should fix the CI :) https://github.com/react-navigation/react-navigation/pull/4327

brentvatne:
thanks again @Minishlink! we really appreciate all of the PRs :)

PR #4234, opened 2018-05-14
Namespaced private types that previously were exported. Added explicit module.exports.

mwalkerwells:
Motivation: Private types were being exported & previously was unable to destructure imports
Context (w/ tests): https://github.com/flowtype/flow-typed/pull/2165

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -1,1100 +1,1166 @@
 // @flow
 
-declare module 'react-navigation' {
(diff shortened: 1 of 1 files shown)


mwalkerwells:
It seems this file is out of date with... https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js

Any guidance?

Issue #3954, opened 2018-04-11
DrawerNavigator does not work, if it contains a TabNavigator containing a Stacknavigator


janrop:
### Current Behavior

https://snack.expo.io/@janrop/drawernavigator-test

Navigating with the DrawerNavigator does not work if there is a TabNavigator which itself contains a StackNavigator

- Drawer Navigator
    - TabNavigator
        - Stacknavigator
            - Screen
        - Screen
        - Screen
    - Screen

### Expected Behavior

- Navigation should work as expected

### How to reproduce

https://snack.expo.io/@janrop/drawernavigator-test

* Go to the second or third tab
* Open the Drawer
* Go To "InDrawer"
* Open the Drawer
* Try to return to "Tabs"

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.5.11
| expo sdk     | 26.0.0

kosmicbird:
+1 

I have a very similar navigation system and am experiencing the same issue

ericvicenti:
Just fixed it here: https://github.com/react-navigation/react-navigation/pull/4237

In the mean time you can make this change yourself, or use a custom drawer contentComponent that just does a normal `props.navigation.navigate()` when a link is pressed.

kosmicbird:
Oh cool! Great news to wake up to ^_^ I will implement this in my app. Thanks!

brentvatne:
we still need to remove the `StackActions.reset()` thing from drawer, issue lives on at https://github.com/react-navigation/react-navigation-drawer/issues/2

seanmorton:
https://github.com/react-navigation/react-navigation-drawer/pull/16

PR #4237, opened 2018-05-15
Fix broken tabs-in-drawer behavior

ericvicenti:
This code assumes that the child of a drawer will be a stack.

Its a very confusing behavior anyways, and the "less pushy navigate" will address the concern of the original issue

closes https://github.com/react-navigation/react-navigation/issues/3954

### DIFF
--- a/src/views/Drawer/DrawerSidebar.js
+++ b/src/views/Drawer/DrawerSidebar.js
@@ -46,23 +46,9 @@ class DrawerSidebar extends React.PureComponent {
 
   _onItemPress = ({ route, focused }) => {
     if (!focused) {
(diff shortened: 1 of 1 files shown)


brentvatne:
i'm going to extract drawer into a separate lib and then make this change there

brentvatne:
let's do this on react-navigation-drawer: https://github.com/react-navigation/react-navigation-drawer/issues/2

jaykangad:
Im still facnig this issue in latest version 6.0

Suppose if i click on any screen from drawer and then move to Tab navigator, it doesn't allow me to move to any tab. Screen is frozen it seems.

PR #4054, opened 2018-04-24
 Fix ripple on android and make ripple same as tintColor 

JasperVercammen:
This is a fix for issue #3955 
These changes target only android and should fix the ripple on the default back button

Issue #3955, opened 2018-04-11
Ripple back button Android is too wide


JasperVercammen:
### Current Behavior

- The default back arrow on android has a ripple that goes too wide.
![ripple_too_large](https://user-images.githubusercontent.com/2750271/38606731-25ead0a8-3d77-11e8-9b9a-46dfb11bb6f5.gif)

If the header ends, the ripple is cut off. 
If not, the ripple will go on till it's under the statusbar. This is visible in the default example project: 
https://expo.io/@react-navigation/NavigationPlayground

### Expected Behavior

- The ripple should have boundaries. (48 x 48 dp) 
Example from the messeges app from google: 
![ripple_as_it_should_be](https://user-images.githubusercontent.com/2750271/38606729-25d05e08-3d77-11e8-98be-4384ad810147.gif)

Can also be checked in every other app from Google. 

### How to reproduce

- Open an android project with a stacknavigator and 2 screens. Navigate to the second screen. Press (and hold) the back arrow.  (https://expo.io/@react-navigation/NavigationPlayground)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.5.3
| react-native     | 0.53

kristfal:
Nitpick: Also the ripple should inherit the tint colour of the icon. See an example at:

https://material.io/guidelines/components/buttons.html#buttons-toggle-buttons

kolyneh:
Maybe this issue is fixed then #4616 is fixed too.
The `headerLeft` component should have `marginLeft:16` to prevent from touching parent's left.

brentvatne:
fix is merged on master, to be released soon

PR #4245, opened 2018-05-15
Fix ripple on android and make ripple same as tintColor 

JasperVercammen:
This is a fix for issue #3955
These changes target only android and should fix the ripple on the default back button

Also see PR #4054 for a fix for v1.x


src/views/Header/HeaderBackButton.js
```diff
@@ -79,15 +80,19 @@ class HeaderBackButton extends React.PureComponent {
 
     const backButtonTitle = renderTruncated ? truncatedTitle : title;
 
-    return (
+    const pressColor = pressColorAndroid
+      ? pressColorAndroid
+      : `rgba(${tintColor}, .32)`;
```
brentvatne: this isn't a valid color, using the default values it returns: `rgba(#000000, .32)`

JasperVercammen: Good catch. Do you have any suggestions on how to handle this? Since `tintColor` can be any "kind" of color format (hex, rgb, colorname,..), but the alpha value should be lowered for the ripple. Is a color converter something?

brentvatne: react-native uses this function called `normalizeColor` for it: https://github.com/facebook/react-native/blob/d01ab66b47a173a62eef6261e2415f0619fefcbb/Libraries/Color/normalizeColor.js

it is based off of [tinycolor](https://github.com/bgrins/TinyColor). unfortunately `normalizeColor` isn't exposed through the public api, so we might want to pull in `tinycolor` and use it here.

JasperVercammen: And with pull in, you mean install it as a dependency, or just add the file into the utils folder?

JasperVercammen: @ricardobeat I'm a bit out of time, so if you want to adapt, be my guest :-) I've send an invitation

brentvatne: @ricardobeat - still interested?

brentvatne:
closing this for now, if anyone wants to pick up the torch it'd be very welcome!

PR #4248, opened 2018-05-15
Fixed the Header Title Center bug on Android

rbhatia46:
The Title in the header of the Stack Navigator wasn't centred by default for Android Devices and was set to 'flex-start' for any devices without iOS. 

The main motivation was when I was trying to build an Instagram UI Clone in React Native on Android. I used the title in navigation options and saw that the title wasn't centred by default for Android devices whereas it was for the iOS devices. I had a look at the Header source code and figured out that for devices other than iOS, the value should be 'centre' instead of 'flex-start' which indeed solved the issue.

Please refer to the screenshots below and after the changes:

**Before**

Before

**After**

After

Everything works fine on iOS by default but I guess this fixes the problem on Android devices.

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -576,7 +576,7 @@ const styles = StyleSheet.create({
     position: 'absolute',
     alignItems: 'center',
     flexDirection: 'row',
-    justifyContent: Platform.OS === 'ios' ? 'center' : 'flex-start',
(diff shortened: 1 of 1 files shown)


brentvatne:
hi there! this behavior is intentional. by default on android the text is aligned to the left. open issue on rfcs for this: https://github.com/react-navigation/rfcs/issues/38

if you'd like to open a rfc on that repo please do

PR #4251, opened 2018-05-16
Optimize images

charpeni:
Lossless compression algorithm saved 5 KB out of 9 KB. 54.9% per file on average (up to 67.8%).


brentvatne:
nice :)

PR #4255, opened 2018-05-16
Fix for broken transition callbacks in StackView

### LINKED ISSUE #4252: onTransitionStart and onTransitionEnd callbacks not firing
Hi, I'm currently in the process of upgrading one of our projects to version 2.0.1.

I tried to get the `onTransitionEnd` callback working on a stack, but it simply does nothing.

```
createStackNavigator({
  Dummy: { screen: Dummy }
},{
  onTransitionEnd: (props) => { Alert.alert('Just did it!'); }
})
```

Here's a snack with my full example: https://snack.expo.io/@stigi/ontransitionstart-react-navigation-2.0.1

Here's the same example working in an older version https://snack.expo.io/@stigi/ontransitionend-react-navigation-1.5.4

The [documentation](https://reactnavigation.org/docs/en/stack-navigator.html) for v2 still mentions `onTransitionStart` and `End` so I'm positive it didn't go away.

---

### Current Behavior

Using the code from above I push the same Dummy screen onto the stack.  
The `onTransitionEnd` callback doesn't get triggered and no Alert is shown.

### Expected Behavior

Using the code from above I push the same Dummy screen onto the stack.  
The `onTransitionEnd` callback gets triggered and an Alert is shown.

### How to reproduce

See my snack. There's a push button. Push it :)

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.1
| react-native | 0.54.4
| node             | 9.4.0
| yarn      | 1.6.0

stigi:
Looks like in `createNavigation` the `stackConfig` from
`createStackNavigation` is passed inside `navigationConfig` prop. This
commit pulls the values from the correct place inside `StackView`.

TestPlan:
Applied the change locally and saw the `onTransitionEnd` callback pass.
Removed the `onTransitionEnd` callback from my test code and didn't get
a red box :P

Fixes #4252

### DIFF
--- a/src/views/StackView/StackView.js
+++ b/src/views/StackView/StackView.js
@@ -24,9 +24,12 @@ class StackView extends React.Component {
         configureTransition={this._configureTransition}
         navigation={this.props.navigation}
         descriptors={this.props.descriptors}
-        onTransitionStart={this.props.onTransitionStart}
+        onTransitionStart={this.props.navigationConfig.onTransitionStart}
         onTransitionEnd={(transition, lastTransition) => {
-          const { onTransitionEnd, navigation } = this.props;
+          const {
+            navigationConfig: { onTransitionEnd },
+            navigation,
+          } = this.props;
           if (transition.navigation.state.isTransitioning) {
             navigation.dispatch(
               StackActions.completeTransition({


Bardiamist:
Dublicate of [pull4100](https://github.com/react-navigation/react-navigation/pull/4100/files)

PR #4267, opened 2018-05-17
[examples] Update ReduxExample

Ashoat:
Using two new utils:
1. `createNavigationPropConstructor`, which is now preferred over `createReduxBoundAddListener`.
2. `initializeListeners`, which is necessary to trigger the event listeners with the initial state.


brentvatne:
👌

PR #4290, opened 2018-05-22
Add enableKeyboardAvoiding to navigationConfig

yangyuan:
Add enableKeyboardAvoiding to navigationConfig, if enabled, createNavigator will wrap the NavigatorView with a KeyboardAvoidingView.

### DIFF
--- a/src/navigators/createNavigator.js
+++ b/src/navigators/createNavigator.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { KeyboardAvoidingView } from 'react-native';
 
 import getChildEventSubscriber from '../getChildEventSubscriber';
 
@@ -92,7 +93,9 @@ function createNavigator(NavigatorView, router, navigationConfig) {
         };
       });
 
-      return (
+      const { enableKeyboardAvoiding } = navigationConfig;
+
+      let navigatorView = (
         <NavigatorView
           {...this.props}
           screenProps={screenProps}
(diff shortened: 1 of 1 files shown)


yangyuan:
KeyboardAvoidingView needs keyboardVerticalOffset if the view does not start from the top of the screen.

For createStackNavigator or createMaterialTopTabNavigator there will headers and tabs.
I've seen most of the projects use magic numbers (like 64 for ios, 54+StatusBar.currentHeight for Android, and special case for iPhone X, iPad, horizontal iPhone X...) to set the keyboardVerticalOffset. Which is a bad idea.

We can extend the navigators to add a KeyboardAvoidingView. However, it will be duplicated work, especially in some UI structure like this:
```
<BottomTabNavigator>
  <KeyboardAvoidingView>
    <StackNavigator />
  </KeyboardAvoidingView>
</BottomTabNavigator>
```

I think the better idea is to let react-navigation to take care of this. So I suggest to add a new paramater in navigationConfig.
For example:
```
createStackNavigator({...}, { enableKeyboardAvoiding: true });
```

brentvatne:
hey! this is interesting. could you submit a RFC that details this more? thanks! https://github.com/react-navigation/rfcs

brentvatne:
closing until rfc

scottyantipa:
It would be really nice to get this in. I currently have to remove react-navigation headers, and make my own, to get KeyboardAvoidingView working properly. @yangyuan is this being worked on?

brentvatne:
@scottyantipa - want to open a RFC? http://github.com/react-navigation/rfcs

yangyuan:
@brentvatne I will open a RFC today.

yangyuan:
Hi @brentvatne I've created the RFC. https://github.com/react-navigation/rfcs/pull/58

Issue #4250, opened 2018-05-16
JS Errors get stuck in createNavigationContainer so cannot be caught and handled in Error Boundaries


owinter86:
### Current Behavior

- Errors get stuck in the `componentDidCatch` of the `createNavigationContainer.js` component and currently not possible to create custom error boundaries to catch and handle JS errors

### Expected Behavior

- JS errors in screens not associated with `_reactNavigationIsHydratingState` should cascade down the chain and have the possibility to be handled in custom Error Boundaries 

### How to reproduce
- `componentDidCatch` lifecycle methods in navigations screens are never called during JS errors.

### Proposed solution

``` javascript
// createNavigationContainer.js
...

componentDidCatch(e, errorInfo) {
  if (_reactNavigationIsHydratingState) {
    //..
  } else {
    // Throw error if not caught in above conditional
    // Will ensure errors cascade to any custom errorBoundaries
    throw new Error(e);
  }
}
```

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.1
| react-native     |  0.55.4
| node             |  8.4.0
| yarn      | 1.6.0

brentvatne:
@ericvicenti - this seems pretty reasonable

brentvatne:
resolved as of 2.4.1 https://github.com/react-navigation/react-navigation/releases/tag/2.4.1

PR #4298, opened 2018-05-22
fix(create-nav-container): pass up error on catch

AndreiCalazans:
Errors are getting stuck inside the `createNavigationContainer`, so implementing `componentDidCatch` above any Navigator isn't working. This fixes according to this issue #4250


brentvatne:
thanks!

Issue #3787, opened 2018-03-19
Params are not passed between two StackNavigators wrapped in SwitchNavigator


animaonline:
I know this has been reported earlier (https://github.com/react-navigation/react-navigation/issues/143) but there's no workaround and this is clearly a bug.

E.g.

```
const RegisterStack = StackNavigator(
  {
    RegisterStep1: RegisterStep1,
    RegisterStep2: RegisterStep2,
    RegisterStep3: RegisterStep3,
    RegisterStep4: RegisterStep4,
    RegisterStep4: RegisterStep4,
    RegisterStepFailed: RegisterStepFailed
  },
  {
    headerMode: "screen"
  }
);

const UserInfoStack = StackNavigator(
  {
    UserInfo,
  },
  {
    headerMode: "screen"
  }
);

const RootSwitchNavigator = SwitchNavigator(
  {
    Register: RegisterStack,
    UserInfo: UserInfoStack
  },
  {
    headerMode: "screen",
    initialRouteName: "Register"
  }
);
```

When I navigate to UserInfo from RegisterStep1 like so:
`this.props.navigation.navigate("UserInfo", {hello: "world });`

The params are not passed. 

Here's a snack taken from your documentation that reproduces this problem: https://snack.expo.io/r1Ih9W6tG

Notice the alert(JSON.stringify(this.props.navigation.state.params)) in HomeScreen's componentDidMount

brentvatne:
you have two routes named UserInfo 🤔 so it's switching to that stack and passing the param in to the stack route but not to the actual screen. rename the one at top level to `UserInfoStack` or something and it shold resolve the issue

animaonline:
I'm sorry there's a typo in my code, but the error is still reproducible, have a look at the snack, it uses the code from your documentation.

brentvatne:
@animaonline - it's still not the expected behavior based on what is in the snack. if you want to pass the params to the route inside of a stack, navigate to that route directly.

so in the snack it navigates to `App`, but `Home` is the route in `App` that you want to receive the params. so instead you navigate to `Home`

brentvatne:
here's an example: https://snack.expo.io/H1ZGLKZ5G

notice that I have to use `key` here so it doesn't push another route on top. this is pretty inconvenient and resolved in 2.0 via https://github.com/react-navigation/rfcs/blob/master/text/0004-less-pushy-navigate.md

animaonline:
[...]

PR #4306, opened 2018-05-23
Fix bug: params not be passed to navigator inside SwitchNavigator

liuqiang1357:
As @sahil290791 described in [ issue #3787](https://github.com/react-navigation/react-navigation/issues/3787), I have implemented my navigator and use some logic depend on params passed from last screen, like choosing some page to show, providing context for all the screen. So I think it is necessary for child navigator to receive params like a ordinary screen component.

### DIFF
--- a/src/routers/SwitchRouter.js
+++ b/src/routers/SwitchRouter.js
@@ -189,7 +189,7 @@ export default (routeConfigs, config = {}) => {
             newChildState = childRouter
               ? childRouter.getStateForAction(action.action, childState)
               : null;
-          } else if (!action.action && !childRouter && action.params) {
+          } else if (!action.action && action.params) {
             newChildState = {
               ...childState,
               params: {


brentvatne:
I think this is an open topic for discussion: https://github.com/react-navigation/rfcs/issues/43

My question for you is: why not navigate directly to the switch screen that you are interested in and pass the params to that?

liuqiang1357:
@brentvatne  In my occasion, I get a authorization information in AuthStack from server, and pass the information to HomeStack. So all the screens in HomeStack need the information, not only the first screen.

I don't use redux, and I don't think put something into global is a good practice. So I wrap the HomeNavigator and provide the information by contex. This is my code for HomeStack:
`
const HomeNavigator = createTabNavigator({
  Screen1: { screen: Screen1 },
  Screen2: { screen: Screen2 },
  Screen3: { screen: Screen3 },
})

export default class HomeStack extends Component {
  static router = HomeNavigator.router

  render() {
    const auth = this.props.navigation.getParam('auth')
    return (
      <Provider auth={this.auth}>
        <HomeNavigator navigation={this.props.navigation}/>
      </Provider>
    )
  }
}
`

fbartho:
I needed to Use a StackNavigator with 1 item to get a custom Header, and then a TopTabNavigator nested inside in order to build our UserProfile Screen correctly. Unfortunately, to Pass the UserId that the Profile needs, we need to use NavParams. This unfortunately is blocked by the absence of this feature! Is there any further support for this need, that I can provide a more through description of our stack if that helps @brentvatne

brentvatne:
@ericvicenti ping me about this when you’re around :)

mistenkt:
Can we expect a merge of this?

PR #4311, opened 2018-05-23
Only set state when component is mounted

kimihiro64:

Fixes a warning that debugger catches on for setState no-op when navigation container is no longer mounted.

**Test plan (required)**

Not much to test here, it was a no-op before, it's a no-op after, just no warning now.

The code must pass tests.

**Code formatting**

### DIFF
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -344,11 +344,15 @@ export default function createNavigationContainer(Component) {
       if (navState !== lastNavState) {
         // Cache updates to state.nav during the tick to ensure that subsequent calls will not discard this change
         this._navState = navState;
-        this.setState({ nav: navState }, () => {
-          this._onNavigationStateChange(lastNavState, navState, action);
-          dispatchActionEvents();
-          this._persistNavigationState(navState);
-        });
+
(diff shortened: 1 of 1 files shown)


kimihiro64:
@brentvatne  Perhaps it is a bug on my side, but I'm not sure what the best way to handle it is. I have multiple network calls going off at the same time, and the handler will navigate to login if a 403 is received. I assume when one triggers it, the rest trigger while the original navigator container is no longer mounted, and that's why I get the warning. The additional navigates don't do anything, which is the expected behavior. My top level navigator is a SwitchNavigator, which has 'Login' and 'App', so possibly navigating away from 'App' back to 'Login' is killing the container. I don't really /want/ to write a bunch of code just to track that and stop a warning when functionally it works.

brentvatne:
can't you cancel responding to those network requests when the component is unmounted? or is this being handled outside of the component tree?

kimihiro64:
I guess I can pass in the component to check that, because yeah, the handler is outside the components.

brentvatne:
going to close this because i'm not sure it's a good fit. i do appreciate the pr nonetheless

PR #4319, opened 2018-05-24
Fix Codecov link -> react-community vs react-navigation

slorber:
OLD (link is broken and code coverage is coverage before renaming)
[codecov](https://codecov.io/gh/react-community/react-navigation)

NEW:
[codecov](https://codecov.io/gh/react-navigation/react-navigation)


brentvatne:
I prefer the old number 😅

Issue #1728, opened 2017-05-30
Cannot Properly Center Title on Android


StevePotter:
The default for title alignment on Android is left but in our app we center it.   

To center the title, I specified a style in `headerTitleStyle` that has `alignSelf: 'center'`.  That moved the title to the middle, but it was offset by 40px, as shown here:
<img width="377" alt="screenshot 2017-05-29 16 27 41" src="https://cloud.githubusercontent.com/assets/271103/26582058/8d04106e-450d-11e7-8460-1450789bf38d.png">

I found the cause, which is in 
[this line from Header.js](https://github.com/react-community/react-navigation/blob/30ca3c1f80617abc28fdd71f0b37d4f62b3e388a/src/views/Header.js#L163).  On Android, this essentially cancels out the offset from `TITLE_OFFSET`.  Since there is not 'right' component but there is a 'left' back button, the style on the title's container is `{ left: 40, right: 0 }`, which causes the problem.  It would also be nice to have access to magic constants like `TITLE_OFFSET` as well.  Maybe there could be a config setting for those as well.

It took a lot of work to get the title to align to the center.  I suggest adding a property like `headerTitleAlign`.  That could allow someone like me to override the default behavior.  Would you be open to something like that?

By the way, I have fixed the problem in our app by subclassing Header.js and adding:
```
  _renderTitle(props: SceneProps, options: *): ?React.Element<*> {
    const view = super._renderTitle(props, options)
    // see https://github.com/react-community/react-navigation/blob/30ca3c1f80617abc28fdd71f0b37d4f62b3e388a/src/views/Header.js#L162
    // by setting the right to 0 and not the left, the title is off center
    // style is an array and we check for left/right being set and if so, match 
    let left, right
    for (const style of view.props.style) {
      if (style.left !== undefined) {
        left = style.left
      }
      if (style.right !== undefined) {
        right = style.right
      }
    }
    if (left !== undefined || right !== undefined) {
      if (left === undefined) {
        left = right
      } else if (right === undefined) {
        right = left
      } else {
        left = right = Math.max(left, right)
      }
      view.props.style.push({ right, left })
    }

    return view
  }
```
This logic works for centering only.

hecvasro:
@StevePotter Adding an empty view as 'headerRight' works too.

ShenghuGong:
@hecvasro   empty view means headerRight=null or headerRight=<View  /> ?

hecvasro:
[...]

PR #4331, opened 2018-05-24
added header container styles to be customized

rghorbani:
This push adds a request to be able to center title on android ( issue #1728 ).

No specific test required to this feature.

Kept the original code formatting.


brentvatne:
this is reasonable but it gets into an api design discussion. if we add this prop, what happens when someone wants to customize the `top` or `left`, etc?

I think it's reasonable to make the following changes:

`headerLayoutPreset: 'center' | 'left'` - would default to `center` on ios and `left` on android. on rtl, `left` would become `right`, in the same way `headerLeft` and `headerRight` are reversed on rtl
`headerTitleContainerStyle` - override anything on the `title` style in Header.js
`headerLeftContainerStyle`- override anything on the `left` style in Header.js
`headerRightContainerStyle` - override anything on the `right` style in Header.js

brentvatne:
related: https://github.com/react-navigation/react-navigation/pull/4187

cc @xxsnakerxx  in case you're interested in implementing the above

rghorbani:
@brentvatne Requested changes applied.

src/views/Header/Header.js
```diff
@@ -211,6 +211,11 @@ class Header extends React.PureComponent {
 
     const { transitionPreset } = this.props;
 
+    let { style } = props;
+    if (options.headerLeftContainerStyle) {
+      style = [style, headerLeftContainerStyle];
```
brentvatne: should be `options.headerLeftContainerStyle` here? same for others

rghorbani: fixed.

brentvatne:
@stantoncbradley - I'm going to finish https://github.com/react-navigation/react-navigation/pull/4588 before landing this, I'll try to get to that this weak. until then you can cherrypick this into a fork and use it if you need it!

brentvatne:
changed my mind and landed this to keep working on related functionality in #4588

PR #4353, opened 2018-05-27
Make headers of non-standard heights work correctly in StackNavigator when header is null (Fixes #4208)

### LINKED ISSUE #4208: MaterialTopTabNavigator with null header in StackNavigator cause empty space between tabs and top of the phone screen
### Current Behavior

I have `MaterialTopTabNavigator` with `header: null` in `StackNavigator`.  Setting `height` more than 44 for a header causes a gap between screen top and the header top. I've tested on iOS less than iPhone X.

screen shot 2018-05-12 at 02 21 43

Setting negative `margin` for the stack header fixes appearance but breaks screen change animation on iOS.

### Expected Behavior

Header should start from the top of the screen.

screen shot 2018-05-12 at 02 22 40

### How to reproduce

Put `StackNavigator` inside `MaterialTopTabNavigator`. Set `header: null` for MaterialTopTabNavigator screen.  Set `tabBarOptions.style` for first and  `navigationOptions.headerStyle` for second to `{ height: 64 }`.

https://snack.expo.io/SyMs6vXCf

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.1
| react-native     | 0.54.2
| node             | 10.1.0
| yarn      | 1.6.0

vshab:
Hello!

Current implementation uses constants for negative margin when Card's header is null. This causes problems when header has non-standard height(which is allowed through navigationOptions.headerStyle option). See  #4208 for example. This PR wraps header in View when in floating mode and measures its height using onLayout callback.

As @brentvatne said in #4208 this is probably a temporary fix.

Thanks.


brentvatne:
can you add an example to navigation playground that shows what this does?

vshab:
@brentvatne done!

vshab:
@brentvatne could you take a look please? :) we are kind of blocked by this issue.

brentvatne:
hi @vshab - can you change this to only use this new logic when a custom header height is used, and continue to execute the old codepath otherwise?

brentvatne:
if it's set with `headerStyle` yeah.. but if this is a custom header component then I don't believe so. so I guess there are a couple of cases to consider here. I'll merge this and play with it today but if I run into any issues I'll revert and let you know.

brentvatne:
you can use this in 2.4.0-beta.0

vshab:
Thank you!

PR #4392, opened 2018-05-31
Fix race condition in DrawerView

ericvicenti:
Without maintaining this state, an infinite loop happens when rapidly opening and closing the drawer. Better fix would be inside the drawer component itself, but... ¯\_(ツ)_/¯


tankers746:
Fixed issue for me

PR #4394, opened 2018-05-31
Improve DrawerRouter logic to close drawer

ericvicenti:
The drawer-closing logic was too aggressive. This change makes sure that the drawer will close when any navigation happens, but not when params of child routes change

Fixes https://github.com/react-navigation/react-navigation/issues/4342

### DIFF
--- a/src/routers/DrawerRouter.js
+++ b/src/routers/DrawerRouter.js
@@ -6,6 +6,13 @@ import withDefaultValue from '../utils/withDefaultValue';
 
 import DrawerActions from './DrawerActions';
 
+const getActiveRouteKey = route => {
+  if (route.routes && route.routes[route.index]) {
+    return getActiveRouteKey(route.routes[route.index]);
+  }
+  return route.key;
+};
+
 export default (routeConfigs, config = {}) => {
   config = { ...config };
   config = withDefaultValue(config, 'resetOnBlur', false);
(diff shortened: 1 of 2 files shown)


fanlion:
> The drawer-closing logic was too aggressive. This change makes sure that the drawer will close when any navigation happens, but not when params of child routes change
> 
> Fixes #4342

why not give us a choice?

brentvatne:
you have a choice, you can call closeDrawer after setParams :)

micabe:
@brentvatne What do you mean, can we not set a boolean in the drawerConfig object such as it will not close on route change ?

src/routers/DrawerRouter.js
```diff
@@ -6,6 +6,13 @@ import withDefaultValue from '../utils/withDefaultValue';
 
 import DrawerActions from './DrawerActions';
 
+const getActiveRouteKey = route => {
+  if (route.routes && route.routes[route.index]) {
+    return getActiveRouteKey(route.routes[route.index]);
+  }
+  return route.key;
```
shashankpandit: @ericvicenti Just a quick note here. This assumes nested navigators have globally unique route names, which may not be the case. Would it be better to use a route key that composes all the routes along the way?

For instance, my navigation stack is:
Drawer > Stack > Tab

Two different routes in the drawer can end up having routes down in the tab navigator with the same key. Tabs are just a way to filter a given page into separate areas, so they don't declare globally unique routes.

I can of course easily work around this, but thought I'd point it out in case it's causing issues for other users.

satya164: Key is always unique for each route

shashankpandit: @satya164 I see. I am not using any keys anywhere, but two of my route names in nested low level navigators collide (which is not a problem for my application, cuz I never directly navigate to those routes). And this results in the following check failing: https://github.com/react-navigation/react-navigation/pull/4394/files#diff-be823f6faddbc25a62b32d94cef3e4f5R81

Either way, this is not a big deal. I can easily make those routes unique and stuff works. Feel free to ignore.

satya164: keys are created internally

PR #4404, opened 2018-06-01
Transitioner: Fix instantaneous transitions

### LINKED ISSUE #4178: Reset action causes animation "glitches" on iOS
### Current Behavior

In React Native, when using using a reset action, the animation in `2.0` is like this... It’s also not often the same, black headers, flickers...

### Expected Behavior

In `1.5` the animation (with the same code, minus the replacement of `StackActions`) the animation is as expected.

### How to reproduce

This can be reproduced by simply firing a reset action.

```js
const resetAction = StackActions.reset({
  index: 0,
  actions: [NavigationActions.navigate({ routeName: "Intro" })],
});

<Button onPress={() => navigation.dispatch(resetAction)}>Continue</Button>
```

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.0.0
| react-native     | 0.55.3
| node             | 9.3.0
| npm or yarn      | 1.3.2

ericvicenti:
The transitioner should not always perform an animation on state change. Instead, we should check if the navigation state is requesting a transition with isTransitioning. If not, we follow a similar codepath to transitioning, without any animation. This will still call the transition hooks in the props, so its unlikely to break things.

fixes #4178

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,18 +7,18 @@
     "prettier/react"
   ],
   "parser": "babel-eslint",
-  "plugins": [
-    "react",
-    "prettier"
-  ],
+  "plugins": ["react", "prettier"],
   "env": {
     "jasmine": true
   },
   "rules": {
-    "prettier/prettier": ["error", {
-      "trailingComma": "es5",
-      "singleQuote": true
-    }],
+    "prettier/prettier": [
+      "error",
+      {
+        "trailingComma": "es5",
+        "singleQuote": true
+      }
+    ],
 
     "no-underscore-dangle": "off",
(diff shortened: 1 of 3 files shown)


cranberyxl:
@ericvicenti This is causing transitions not to fire when using `navigation.navigate` to go to a screen in a child navigator. Should a transition fire in this case? Can you suggest a workaround otherwise?

ericvicenti:
@cranberyxl is there a minimum code example of the bug? What do you mean exactly when you say transitions do not fire? Please file a new issue with a mcve: https://stackoverflow.com/help/mcve

PR #4425, opened 2018-06-04
Refactor, move child navigation to navigation prop

ericvicenti:
This fixes our descriptor caching issue, and unblocks explicit nested navigation options.

As a side effect, the following APIs are introduced to the navigation prop:

- navigation.getChildNavigation(routeKey) , which is useful for explicitly getting children config info
- navigation.router, access to the static router

And routers have a new way to statically expose children behavior. This is particularly important for getChildNavigation to access custom action creators on the child router.

- router.childRouters[routeName] , an optional way to access the children routers directly. If childRouters are not provided in a router, we will fall back on getComponentForRouteName(routeName).router, which is the previous external API for this (although it may be slower because it will require the whole screen component).

IMO, these are all safe and sensible APIs to include in the library

### DIFF
--- a/src/getNavigation.js
+++ b/src/getNavigation.js
@@ -9,7 +9,10 @@ export default function getNavigation(
   getScreenProps,
   getCurrentNavigation
 ) {
+  const actions = router.getActionCreators(state, null);
+
   const navigation = {
+    actions,
     router,
     state,
     dispatch,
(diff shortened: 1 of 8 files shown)


src/getNavigation.js
```diff
@@ -0,0 +1,49 @@
...
+  const navigation = {
+    router,
+    state,
+    dispatch,
+    getScreenProps,
+    getChildNavigation: childKey =>
+      getChildNavigation(navigation, childKey, getCurrentNavigation),
+    isFocused: childKey => {
+      const { routes, index } = getCurrentNavigation().state;
+      console.log('Top level isFocused!', childKey, index, routes[index].key);
```
brentvatne: 👋 comment

ericvicenti:
Cool. Should we tag this change 2.2 and announce the new APIs?

Issue #2019, opened 2017-06-28
PropTypes and React.createClass are deprecated


BerndWessels:
`react-navigation` seems to still be using `PropTypes` and `React.createClass` which is deprecated and throws annoying warnings now.

Is anybody already fixing this or do you want PRs to fix it?

matthamil:
What version of react navigation are you using? Can you provide links to the source files that are causing these warnings? The `PropTypes` warnings should have been resolved with https://github.com/react-community/react-navigation/pull/1928 and https://github.com/react-community/react-navigation/pull/1968.

szhigunov:
I have updated examples code, and haven't checked `react-navigation` sources yet. So PRs are welcome.

jsteiner:
@matthamil it looks like the latest release (v1.0.0-beta.11) was released before some of those were merged. Maybe time for a new release?

sanjayholla94:
@kelset I think this has been resolved with the latest release

kelset:
it is :) Thanks for letting me know!

ekimlinger:
Still seeing these issues on the latest version (^1.0.0-beta.19), am I missing something? 

I could not find any instances in this particular node_modules folder.

StackTrace:

```
2:58:47 PM
▼Warning: PropTypes has been moved to a separate package. Accessing React.PropTypes is no longer supported and will be removed completely in React 16. Use the prop-types package on npm instead. (https://fb.me/migrating-from-react-proptypes)
- node_modules\expo\src\Logs.js:187:24 in newConsoleFunc
- node_modules\fbjs\lib\warning.js:33:20 in printWarning
- node_modules\fbjs\lib\warning.js:57:25 in warning
- node_modules\react\cjs\react.development.js:2133:18 in get
- node_modules\react-native\Libraries\polyfills\babelHelpers.js:162:78 in interopRequireWildcard
- node_modules\metro-bundler\src\Resolver\polyfills\require.js:174:12 in loadModuleImplementation
- node_modules\react-navigation\src\navigators\StackNavigator.js:5:0 in <unknown>
- node_modules\metro-bundler\src\Resolver\polyfills\require.js:174:12 in loadModuleImplementation
- node_modules\react-navigation\src\react-navigation.js:27:19 in StackNavigator
* navigation\rootNavigator.js:8:15 in <unknown>
- node_modules\metro-bundler\src\Resolver\polyfills\require.js:174:12 in loadModuleImplementation
* App.js:15:0 in <unknown>
[...]

PR #4428, opened 2018-06-05
Change 'React' import writing, Prevent warnings in old version React

cccRaim:
When use `createStackNavigator`, old version React will warning, look image below  
### BUG PERFORMANCE
**In debugger**  
image  
**In Simulator**  
image  
It's uncomfortable.

### THE REASON
Look at this comment  
https://github.com/react-navigation/react-navigation/issues/2019#issuecomment-347547140  
Because the import writing is
```javascript
import * as React from 'react'
```
In this writing, babel will traverse all property in package(react etc.), so the getter in react will be call, and then, the deprecated warning will be executed.  
image
search the keywords in **react-navigation** project.
image  
Only four files contains this writing, and these four files is created 3 months ago.   
So I think it's a personal writing, but this writing cause the warnings.

### DIFF
--- a/src/navigators/createStackNavigator.js
+++ b/src/navigators/createStackNavigator.js
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React from 'react';
 import createNavigationContainer from '../createNavigationContainer';
 import createKeyboardAwareNavigator from './createKeyboardAwareNavigator';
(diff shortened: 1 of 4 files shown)


cccRaim:
@ericvicenti

brentvatne:
thanks @cccRaim!

PR #4435, opened 2018-06-06
Move all drawer code into react-navigation-drawer

brentvatne:
# Motivation

I want us to be able to iterate on this separately. In particular, I want to be able to use react-native-gesture-handler as per https://github.com/react-navigation/react-navigation/pull/3751

# Test plan

Tests pass, NavigationPlayground works, example on react-navigation-drawer works and tests pass there.


package.json
```diff
@@ -38,6 +38,7 @@
     "react-native-drawer-layout-polyfill": "^1.3.2",
```
ericvicenti: remove?

PR #2982, opened 2017-11-15
Fix deep link not working when nesting have more than two levels

moriyoshi:
This is a proposed fix for #1950.

PR #4451, opened 2018-06-07
Add missing tests about DeepLinking 

Arlevoy:
Hello! I've seen that there were a small test missing and I took the opportunity to do a first contribution to this repo. See https://github.com/react-navigation/react-navigation/pull/2982#issuecomment-392223790


brentvatne:
thank you!

PR #4453, opened 2018-06-07
Refactor redo, child navigation to navigation prop

ericvicenti:
Re-doing #4425 without isFocused bug on child navigation


brentvatne:
rebase plis 👼

PR #4455, opened 2018-06-08
throw readable error when wrong navigation prop is passed

Tommos0:
rebase of https://github.com/react-navigation/react-navigation/pull/3823


brentvatne:
thanks!

brentvatne:
in 2.2.5

Issue #4458, opened 2018-06-08
onTransitionEnd is not triggered


ensecoz:
onTransitionEnd is not triggered properly. I'll create PR to fix this. The problem is that it extract wrong property name from the props.

```
createStackNavigator(
  {
    HomeScreen: { screen: HomeScreen },
    DetailScreen: { screen: DetailScreen }
  },
  {
    onTransitionEnd: () => console.log('ended')
  }
);
```

brentvatne:
merged your pr #4459 which fixed this and released in react-navigation@2.2.5

PR #4459, opened 2018-06-08
Extract correct property for onTransitionEnd

ensecoz:
Fix bug https://github.com/react-navigation/react-navigation/issues/4458


brentvatne:
in 2.2.5

PR #4489, opened 2018-06-12
redux example remove screenProps

ericvicenti:
This example was changed in the PR for redux-helpers v2, to kill screenProps from redux integration


ericvicenti:
https://github.com/react-navigation/react-navigation-redux-helpers/pull/37#issuecomment-396689873

PR #4493, opened 2018-06-13
@ericvicenti/universe

ericvicenti:
This cleanup is necessary for the universe hacks. It makes core modules more 'importable' to the other repo where I'm working on full-stack navigation


ericvicenti:
all of these are surface-level changes

PR #4529, opened 2018-06-18
Add possibility to define default params for a single route

vcebotari:
**Motivation**: i want to be able to set default param on a single route, not for entire stack of routes,
for now i can only set 'initialRouteParams' and need this workaround to solve my problem.

**Problem**: i have some categorized list of articles what are in different tabs, want to set header title from route params what i set in default route(tab) params.

**Now it's look like**:
```
Tab1: {
    index: 0,
    params: { title: 'Title 1', url: '/' },
    screen: props => (
      <NewsScreen {...props} />
    ),
 },
Tab2: {
    index: 1,
    params: { title: 'Title 2', url: '/' },
    screen: props => (
      <NewsScreen {...props} />
    ),
 },

```

### DIFF
--- a/src/routers/SwitchRouter.js
+++ b/src/routers/SwitchRouter.js
@@ -56,7 +56,7 @@ export default (routeConfigs, config = {}) => {
 
   function resetChildRoute(routeName) {
     const params =
-      routeName === initialRouteName ? initialRouteParams : undefined;
+      routeName === initialRouteName ? initialRouteParams : routeConfigs.params ? routeConfigs.params : undefined;
     const childRouter = childRouters[routeName];
     if (childRouter) {
       const childAction = NavigationActions.init();
--- a/src/routers/SwitchRouter.js
+++ b/src/routers/SwitchRouter.js
@@ -56,7 +56,11 @@ export default (routeConfigs, config = {}) => {
 
   function resetChildRoute(routeName) {
     const params =
-      routeName === initialRouteName ? initialRouteParams : routeConfigs.params ? routeConfigs.params : undefined;
+      routeName === initialRouteName
+        ? initialRouteParams
+        : routeConfigs[routeName].params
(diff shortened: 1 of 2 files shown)


vcebotari:
can we refuse to initialRouteParam, and use only params from the route ?

brentvatne:
heya @vcebotari - could we call this `initialParams` instead?

ericvicenti:
imo, `defaultParams` would be a more accurate name for this

src/routers/SwitchRouter.js
```diff
@@ -56,7 +56,11 @@ export default (routeConfigs, config = {}) => {
 
   function resetChildRoute(routeName) {
     const params =
-      routeName === initialRouteName ? initialRouteParams : undefined;
+      routeName === initialRouteName
+        ? initialRouteParams
```
ericvicenti: shouldn't this be `{...initalRouteParams, ...routeConfigs[routeName].params }`

brentvatne:
yeah let's do `defaultParams` good call @ericvicenti

src/routers/SwitchRouter.js
vcebotari: yeah, that's better )

vcebotari: @ericvicenti , should we assign initialRouteParams to every child route? or ,
let "initalRouteParams" - for initial route, and simply add defaultParams to child route ?

vcebotari:
@brentvatne , @ericvicenti , if we will use something like: `{...initalRouteParams, ...routeConfigs[routeName].defaultParams }`, maybe we should rename `initialRouteParams` to `defaultStackParams`...

brentvatne:
closing this in favour of: https://github.com/react-navigation/react-navigation-core/pull/11

thanks for getting the ball rolling here!

PR #4550, opened 2018-06-21
Adds SwitchNavigator to react-navigation-web

jpaas:
Adds the SwitchNavigator for web since it does not have a view, it is portable.


ericvicenti:
Looks great, thanks for the fix!

PR #4551, opened 2018-06-22
Adds getNavigation to web exports

jpaas:
getNavigation is a core export. Hard to make web work without it.


ericvicenti:
Thanks!

PR #4559, opened 2018-06-22
Improve changelog format

rodrigobdz:
Bug fixes should be listed in the "Fixed" section instead of "Changed" according to [keepachangelog](https://keepachangelog.com/en/1.0.0/#how).


brentvatne:
thanks for being on top of this!

PR #4561, opened 2018-06-23
Fix changelog format

rodrigobdz:
Newest version 2.5.1 was not shown as a link because declaration at the bottom was missing.


brentvatne:
fixed this already in another commit, oops :)

PR #4586, opened 2018-06-27
Clean up PlatformHelpers

ericvicenti:
This hack had previously been required for old versions of react native and react-native-web

Now it just causes troubles because new versions of react-native-web do not export `BackAndroid` at all

### DIFF
--- a/src/PlatformHelpers.native.js
+++ b/src/PlatformHelpers.native.js
@@ -1,9 +1,0 @@
-import {
-  BackAndroid as DeprecatedBackAndroid,
-  BackHandler as ModernBackHandler,
-  MaskedViewIOS,
-} from 'react-native';
-
-const BackHandler = ModernBackHandler || DeprecatedBackAndroid;
-
-export { BackHandler, MaskedViewIOS };
--- a/src/PlatformHelpers.web.js
+++ b/src/PlatformHelpers.web.js
@@ -1,6 +1,0 @@
-import React from 'react';
-import { BackHandler, View } from 'react-native';
-
-const MaskedViewIOS = () => <View>{this.props.children}</View>;
-
-export { BackHandler, MaskedViewIOS };
--- a/src/createNavigationContainer.js
+++ b/src/createNavigationContainer.js
@@ -1,8 +1,7 @@
 import React from 'react';
-import { AsyncStorage, Linking, Platform } from 'react-native';
+import { AsyncStorage, Linking, Platform, BackHandler } from 'react-native';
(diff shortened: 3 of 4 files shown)


brentvatne:
I think we need to keep this file for things like MaskedViewIOS, no?

Kif11:
@ericvicenti I don't think it is true. MaskedViewIOS is still  `const MaskedViewIOS = UnimplementedView;` and there is no component in [exports](https://github.com/necolas/react-native-web/tree/b195f2b1f507a211f03beef4fa9931c5428ce0ac/packages/react-native-web/src/exports) for it.

Can we put this fix back?

 "react-native-web": "^0.8.7",

remohammadi:
```
ERROR in ./node_modules/react-navigation/src/views/Header/Header.js
Module not found: Error: Can't resolve 'react-native-web/dist/exports/MaskedViewIOS'
```

```
react-navigation@^2.6.2:
  version "2.6.2"
  resolved "https://registry.yarnpkg.com/react-navigation/-/react-navigation-2.6.2.tgz#a8fdb69c5876698f7e58c5a3962a78ba24bb2058"

react-native-web@^0.8.8:
  version "0.8.8"
  resolved "https://registry.yarnpkg.com/react-native-web/-/react-native-web-0.8.8.tgz#65e7fd866d26eacbeebe6de53fa106df2b11a2de"
```

I don't get the error with `2.5.5`.

brentvatne:
https://github.com/necolas/react-native-web/blob/b195f2b1f507a211f03beef4fa9931c5428ce0ac/packages/react-native-web/src/index.js#L84

it looks like it's there?

remohammadi:
Yes, but I don't know why it's trying to find it among `exports/*`. Maybe it's related to the way `react-native` is aliased to `react-native-web` in webpack config. 🤔

I should note that I initialized my project a few hours ago with this command:
```
create-react-native-app my_project --with-web-support
```

And upgraded the `react-native-web` package when I saw the error while calling `yarn web`, after I had added the `createStackNavigator` import. Upgrading didn't helped. I removed the `node_modules` and `yarn install`ed to be sure that there is nothing left from the old `react-native-web`, and it also didn't help.

jstacoder:
@brentvatne i am getting the same error as @remohammadi when using the latest version of react-navigation: `^2.12.0`

jstacoder:
@brentvatne @remohammadi  i was able to get it working by installing `babel-plugin-react-native-web` and adding this to my `.babelrc`: 

```json
"plugins": [
    ["react-native-web", {commonjs: true}]
  ],
```

Issue #4185, opened 2018-05-10
Nested Navigator self setParams not working when a subScreen is above its stack


kevinscience:
### Current Behavior

Suppose I have:

TabNavigator: Main
- Screen: A
- Screen:B

StackNavigator:Root
- Screen:Main
- Screen:C

In Screen A, I have 
```
setTitle(){
        console.log('setTitle called');
        this.props.navigation.setParams({title:'Went to C'})
    }
```
will be pass to Screen C and will be called in Screen C as a callback function.
But, The fact is that, when Screen C is on, setParams in Screen A is not working.

### Expected Behavior

`this.props.navigation.setParams({title:'Went to C'})` is working in any circumstances.

### How to reproduce

I have made a sample project to reproduce the bug
https://snack.expo.io/HkoCblM0z

When one press the button on Screen A, it navigate to Screen C, there is a callback function called in Screen C's constructor. But it's not working. Nothing happened.
please see the log on console. 

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 1.5.12
| react-native     | 0.55.2
| node             | 8.11.1
| npm or yarn      | 5.6.0

brentvatne:
repro'd this issue on 2.0 as well: https://snack.expo.io/rkSZTDXJm

fredriksebek:
I initially thought it was due to API changes, but didn't find anything that stood out. Happy I found this post or else I would have gone crazy. Happy you are able to reproduce it @brentvatne!

normdoow:
I am having a similar issue to this. I have a StackNavigator nested inside a DrawerNavigator. Stacks set up like this...
```
const innerDrawerStack = StackNavigator({
    Home: { screen: Menu },
    CardsView: { screen: CardsView },
    Feedback: { screen: FeedbackView }
}, {
    navigationOptions: ({navigation}) => ({
        headerLeft: <HamburgerButton navigation={ navigation } />,
        headerStyle: {
            backgroundColor: colors.primaryColor,
        },
        headerTitleStyle: {
            fontWeight: 'bold',
            fontSize: 16,
            color: 'white'
        },
        gesturesEnabled: false
    }),
    initialRouteName: 'Home'
})

const DrawerStack = DrawerNavigator({
    InnerStack: { screen: innerDrawerStack } //inner drawer stack adds the header since drawers don't have a header
},{ contentComponent: CustomDrawerView })
```

[...]

PR #4587, opened 2018-06-27
Delegate to child routers for more than just the top screen in the stack

brentvatne:
## Motivation

This fixes https://github.com/react-navigation/react-navigation/issues/4185. It seems like it was an oversight to only try to, uh, bubble down actions to the child router of the active screen in the stack.

## Test plan

I ran the example from https://github.com/react-navigation/react-navigation/issues/4185 against this and it fixed the issue.

### DIFF
--- a/src/StateUtils.js
+++ b/src/StateUtils.js
@@ -133,10 +133,15 @@ const StateUtils = {
    * Replace a route by a key.
    * Note that this moves the index to the position to where the new route in the
    * stack is at. Does not prune the routes.
+   * If preserveIndex is true then replacing the route does not cause the index
+   * to change to the index of that route.
    */
-  replaceAt(state, key, route) {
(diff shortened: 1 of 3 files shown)


src/StateUtils.js
```diff
@@ -133,10 +133,15 @@ const StateUtils = {
...
    * stack is at. Does not prune the routes.
+   * If preserveIndex is true then replacing the route does not cause the index
+   * to change to the index of that route.
    */
-  replaceAt(state, key, route) {
+  replaceAt(state, key, route, preserveIndex = false) {
     const index = StateUtils.indexOf(state, key);
-    return StateUtils.replaceAtIndex(state, index, route);
+    const nextIndex = preserveIndex ? state.index : index;
+    let nextState = StateUtils.replaceAtIndex(state, index, route);
```
ericvicenti: wouldn't this be a const?

Issue #4092, opened 2018-04-30
Option to force the header view in iOS to be rendered like Android (left-aligned) and vice versa


afshin-hoseini:
Is it possible to force the stack navigator's header on iOS to be rendered as like as what it appears on Android?

I applied this headerTitleStyle:

```
{
    fontFamily : "fontName",
    textAlign : 'left',
    position : 'absolute',
    top : 0, left : 20 , right:20, bottom:0,
    backgroundColor : 'red'
}
```

And everything went well except the headerBack min width.

**The result without back button**
![withoutbackbtn](https://user-images.githubusercontent.com/15011928/39420379-91688f6c-4c79-11e8-839b-49f3498324ab.png)

**The result width back button**
![withoutbackbtn](https://user-images.githubusercontent.com/15011928/39420445-dcaf6dd8-4c79-11e8-970f-781198e04e83.png)

My problem is the min width of back button. Is there any solution to change this min width?

pzatorski:
Hey @afshin-hoseini I am not sure but I think the only way that you can change the 'back button' it that you can create your own one. You can do it by adding component to `headerLeft` key in `navigationOptions`. If you set the `headerLeft: null` then it wouldn't return the 'back button' at all.

brentvatne:
hi there! I think this is an important feature for us to add - it should be easy to set the header mode to use the android style of header on ios.

brentvatne:
related: https://github.com/react-navigation/react-navigation/issues/3677

brentvatne:
have a WIP PR open for this, will hopefully land it soon

brentvatne:
done in #4588, check out react-navigation@2.9.1

pzatorski:
Great work @brentvatne !

PR #4588, opened 2018-06-28
Add headerLayoutPreset option to stack

brentvatne:
## Motivation

https://github.com/react-navigation/react-navigation/issues/4092

- `headerLayoutPreset: 'center' | 'left'` to provide an easy solution for [questions like this](https://github.com/react-navigation/react-navigation/issues/4615).
- `headerBackTitleEnabled` - this configuration option for stack navigator allows you to force back button titles to either be rendered or not (if you disagree with defaults for your platform and layout preset).
- Also fixes https://github.com/react-navigation/react-navigation/issues/3955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `headerLayoutPreset: 'center' | 'left'` to provide an easy solution for [questions like this](https://github.com/react-navigation/react-navigation/issues/4615).
+- `headerBackTitleEnabled` - this configuration option for stack navigator allows you to force back button titles to either be rendered or not (if you disagree with defaults for your platform and layout preset).
+
+### Fixed
+- Android back button ripple is now appropriately sized (fixes [#3955](https://github.com/react-navigation/react-navigation/issues/3955)).
+
 
 ## [2.8.0] - [2018-07-19](https://github.com/react-navigation/react-navigation/releases/tag/2.8.0)
 ### Added
--- a/examples/NavigationPlayground/js/SimpleStack.js
+++ b/examples/NavigationPlayground/js/SimpleStack.js
@@ -10,7 +10,7 @@ import type {
 } from 'react-navigation';
 
 import * as React from 'react';
-import { ScrollView, StatusBar } from 'react-native';
+import { Platform, ScrollView, StatusBar } from 'react-native';
 import {
   createStackNavigator,
   SafeAreaView,
@@ -23,6 +23,8 @@ import invariant from 'invariant';
 import SampleText from './SampleText';
 import { Button } from './commonComponents/ButtonWithMargin';
 import { HeaderButtons } from './commonComponents/HeaderButtons';
+
+const DEBUG = false;
 
 type MyNavScreenProps = {
   navigation: NavigationScreenProp<NavigationState>,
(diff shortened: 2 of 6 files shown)


ahmetabdi:
@brentvatne Is this in 2.9.1? I have it in my stack navigator but it is not positioned to the left.

```
navigationOptions: () => ({
      headerTransparent: false,
      title: 'Settings',
      headerLeft: SettingsIcon,
      headerLayoutPreset: 'left',
}),
```

brentvatne:
it is not part of `navigationOptions`, you need to use it in the stack config

ahmetabdi:
Oh thanks

giantss:
@brentvatne  I always feel that the `headerLayoutPreset` configuration item should belong to `navigationOptions`.

brentvatne:
feel free to open a rfc to explain why you think that is the case: https://github.com/react-navigation/rfcs

arsenalol69:
Can you updata Typescript definition ? Because TSLint can't find your new param :/

brentvatne:
@arsenalol69 https://github.com/react-navigation/react-navigation/pull/5276

jcbertolucci:
> it is not part of `navigationOptions`, you need to use it in the stack config

that saved my life

danielsotopino:
> @brentvatne I always feel that the `headerLayoutPreset` configuration item should belong to `navigationOptions`.

I feel the same

danielsotopino:
I wouldn't be able to tell you how because i haven't read the code of react-navigation but the why is quite simple since many of the header styling can be done on navigationOptions.

I'll check the code later if i can, would be interesting to help on this.

eleallegue:
I am dealing with the same problem and I feel same way. I think that in my case for example, I just want to have one header in the whole app and I want to center the name and so I have no header configuration anywhere. The only config of stack I have is the bottom navigation, so now I just cannot center the title unless I use a dirt solution that is leaving blank the headerLeft with a View component.

esr360:
Don't know what an rfc is, but from a Developer Experience point of view, at least to me (as an inexperienced newbie), I agree it seems to make more sense for `headerLayoutPreset` to be part of `navigationOptions`.

esr360:
For me it makes sense because I am calling `navigationOptions` on each screen, and have control of the navigation where I am using it. Setting the alignment of the navigation title seems like an option of the navigation rather than an option of the stack. I think it's feasible that different screens could potentially have different layout requirements.

As I'm a newbie, the only transitioning I've actually seen of screens is slide/fade animations so in that scenario at least it wouldn't have an impact.

enderandpeter:
Would someone mind taking a look at this again? From what I am seeing the `headerLayoutPreset: 'center'` stackConfig setting is not centering the header title on Android. I'm having to use [a workaround](https://github.com/react-navigation/react-navigation/issues/4615#issuecomment-402012158) instead.

andywu0408:
const MainStack = createStackNavigator({
    ......
    ......
},
{
  headerLayoutPreset: 'center'
});


Adding it into stackConfig works for me on both iOS and Android.

andywu0408:
const StackNavigator = createStackNavigator({
    screen: MyScreen,
navigationOptions: {
    title: 'MyTitle'
}
},
{
    headerLayoutPreset: 'center'
});

Adding it into stackConfig works for me on both iOS and Android.

elie222:
Thanks @andywu0408. I ran into this issue myself recently. The issue is that it's not part of `navigationOptions` or `defaultNavigationOptions`. It works as in the Andy's example.

radupm:
Since React navigation 5 use [headerTitleAlign](https://reactnavigation.org/docs/stack-navigator/#headertitlealign):

`  defaultNavigationOptions: {
      title: 'Centered',
      headerTitleAlign: 'center'
  }`

Defaults to `center` on iOS and `left` on Android.

hadpro24:
You can go this link : https://reactnavigation.org/docs/stack-navigator/#headertitlealign

marcvangend:
Documentation for version 6: https://reactnavigation.org/docs/elements/#headertitlealign

PR #4590, opened 2018-06-28
Routers: Deep Linking Overhaul

ericvicenti:
This fixes a number of deep linking issues. Primarily:

- Stack router no longer attempts to parse query params within path handling
- Switch router now has exact same param treatment for URLs as stack router does


brentvatne:
nice work!

PR #4632, opened 2018-07-03
Added withNavigationDebounce and added NavigationUtility which will support with functionalities which are required when redux integration is removed

sinhashubham95:
**NavigationUtility**
The intention behind adding this utility is the requirements that arise when the redux integration has been removed like dispatching actions and getting the navigation state at any component despite the fact that it has the navigation prop. Other requirements that are fulfilled through this utility are getting the current route or key, searching for any route or key and getting the key to the route next to the route name specified (which is used in a go back).

**withNavigationDebounce**
This is a HOC which has the same navigation methods debounced with the wait time passed as a parameter. This requirement could have been fulfilled through the redux integration using which we had a chance to change the navigation state at our end manually, but with the removal of redux integration, this can be used as an alternative.

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -751,15 +751,10 @@ declare module 'react-navigation' {
   };
 
   declare export var NavigationUtils: {
-    setNavigatorReference: (navigator?: {
-      dispatch?: Function,
-      state?: { nav?: Object },
-    }),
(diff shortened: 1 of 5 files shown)


brentvatne:
I don't think we want to add this to core, but you could release it as a standalone library on npm just fine for anyone who would like this!

PR #4636, opened 2018-07-03
Feat/allow keyless replace

lfkwtz:
## Motivation

Per [my conversation](https://github.com/react-navigation/react-navigation/pull/3440#issuecomment-401928084) with @ericvicenti, the StackNavigator replace action shouldn't need to require a key. If we leave key as undefined, we can replace current screen by pulling the last index.

## Test plan

Added a unit test to show it is working as expected


ericvicenti:
Looking good, thanks for the contribution!

Issue #4643, opened 2018-07-04
[Android] 'absolute' position hides header - happened sometime between 2.0.4 & 2.5.5


noambonnie:
### Current Behavior
Using Stack Navigator, trying to set header position to absolute (with the intention of animating it). In 2.0.4 works as expected. In 2.5.5 the header is hidden. I tried modifying `zIndex` to no avail.

When removing `position: 'absolute'` from the headerStyle in `navigationOptions`, the header shows as expected.

I compared the diff between the two versions. I found a diff in src/views/Header/Header.js. In 2.5.5 line 503 of Header.js: 
```
      <Animated.View
        style={[
          this.props.layoutInterpolator(this.props),
          { backgroundColor: DEFAULT_BACKGROUND_COLOR },
        ]}
      >
```
The "backgroundColor" part is new. Oddly enough, if I remove that line I do see the header. This works:
```
      <Animated.View
        style={[
          this.props.layoutInterpolator(this.props),
        ]}
      >
```

I can't find a good explanation as to why this works but I've tested it a bunch of times.

### Expected Behavior
Absolute positioning of header should not hide the header. 

### How to reproduce

Snack: https://snack.expo.io/BJskvP9Mm

Since it's my first time using snack I'll also post my `App.js` here just in case - it's short:

```
import React from 'react';
import { View, Text } from 'react-native';
import { createStackNavigator } from 'react-navigation';

class PostScreen extends React.Component {
    render() {
        return (
            <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
                <Text>Test Screen</Text>
            </View>
        );
    }
}

const HeaderStack = createStackNavigator(
    {
        Post: {
            screen: PostScreen,
            navigationOptions: {
                title: "Test"
            }
        },
    },
    {
        navigationOptions: {
            headerStyle: {
                position: 'absolute', // <==== Comment this out and the header would show
                top: 0,
                left: 0,
                right: 0,
                bottom: 0,
                // zIndex: 1,
                // What I'm really trying to do is animate, but once the absolute issue is resolved I think the rest will be too.
                // transform: [{ 
                //     translateY: navbarTranslate
                // }],
            },
        },
    }
);

export default class App extends React.Component {
    render() {
        return (
            <HeaderStack/>
        )
    }
}

```

Simply run the above code and header would not show. Comment out the `position: 'absolute'` line and header appears.

### Your Environment

| software         | version
| ---------------- | -------
[...]

Issue #4607, opened 2018-06-30
Upgrade to 2.5.5 and removed header shadow


AlirezaAkbarix:
Just upgrade package from `2.3.1` to the latest version (`2.5.5`) and the shadow below header is gone!

`2.3.1`:
![image](https://user-images.githubusercontent.com/15062651/42121364-e18845de-7c42-11e8-8ace-e1c09b7e859f.png)

`2.5.5`:
![image](https://user-images.githubusercontent.com/15062651/42121175-50158c90-7c3f-11e8-995e-15be570250a7.png)

Please don't notice issue is not respecting the issue template as it's a clear issue.

Environment:
  OS: macOS High Sierra 10.13
  Node: 8.9.4
  Yarn: 1.6.0
  npm: 6.0.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.1 AI-173.4720617

Packages: (wanted => installed)
  react: ^16.4.1 => 16.4.1
  react-native: 0.55.4 => 0.55.4

brentvatne:
can you please post a minimal example that reproduces this?

ramsanath:
I am facing the same issue 

Steps to reproduce

Create a StackNavigator with a basic screen

>  
    import {createStackNavigator} from 'react-navigation';
    import React, {Component} from 'react'
    import {View} from 'react-native'
    
    class Demo extends Component {
        render() {
            return (
                <View
                    style={{
                        flex: 1,
                        backgroundColor: 'powderblue'
                    }}
                />
            )
        }
    }
    
    const RootNavigator = createStackNavigator({
        Home: Demo,
    });
    
    export default RootNavigator;

What is expected:

![expected](https://user-images.githubusercontent.com/25882182/42167993-63be3432-7e2d-11e8-811e-379990c6d359.png)

Actual output:

![screenshot_1530539481](https://user-images.githubusercontent.com/25882182/42168021-78c38bfc-7e2d-11e8-933b-e4a139c908ee.png)

**Info**
platform: OSX
react-navigation: ^2.5.5
react-native: 0.55.4

kolyneh:
Have the same issue on 2.5.5.

kidmysoul:
Have the same issue on 2.5.5.

xanderberkein:
Snack: https://snack.expo.io/rkGrH3dMX

[...]

PR #4645, opened 2018-07-05
fix(header): fixing bug when header disappear when absolute

Jekiwijaya:
## Motivation
There's a bug in Android header after version >= 2.5.4
Fixing:
- missing header when position absolute #4643 
- missing shadow in android. #4607


brentvatne:
hi! I ended up doing this: https://github.com/react-navigation/react-navigation/commit/fe95bdeee6b70201e74ff40d7b87e0617ebbc3e1

thanks so much for opening the pr!

PR #4652, opened 2018-07-06
Makes "easy" to define left and right Header buttons

Piropa:
The idea behind this is to ease the creation of left and right buttons to the header bar preserving the default styling or defining a new one.

## Motivation

I've found quite unpleasant to add custom buttons to the header bar, since the properties "headerLeft" and "headerRight" only allow me to define a custom element or custom component without any of the predefined platforms margins and colors already established by react-navigation, forcing me to rebuild the whole component and clone styling. Also found that in order to define a new action for the left button, then I have to create a whole new custom object, which I don't like.

Then, by implementing this changes, I can create Right Button which will respect all default styling, I can also easily change the Left or Right button icons or actions without having to create a whole component.

## Changes

I have added the following variables to the navigation options:
```
options: {
   // Default Title for right button (iOS only)
   headerRightTitle: string

   // Default Truncated Title for right button (iOS only)
   headerTruncatedRightTitle: string

   // Style for the right button Title (iOS only)
   headerRightTitleStyle: TextStyle

   // Default action for left button
   headerLeftOnPress: () => void

   // Default action for right button
   headerRightOnPress: () => void

   // Default image for left button
   headerBackImage:
      React.Element<> |
      React.Component<{tintColor, title}> |
      {
         source:
            React.Element<> |
            React.Component<{style<ViewStyle> , tintColor, title}> |
            ImageSourcePropType
      }

   // Default image for right button
   headerRightImage:
      React.Element<> |
      React.Component<{tintColor, title}> |
      {
         source:
            React.Element<> |
            React.Component<{style<ViewStyle> , tintColor, title}> |
            ImageSourcePropType
      }

   // Left button (all definitions)
   headerLeft:
      React.Element<> |
      React.Component<{onPress, pressColorAndroid, tintColor, backImage, 
                       title, truncatedTitle, titleStyle, width}> |
      {
         source:
            React.Element<> |
            React.Component<{style<ViewStyle> , tintColor, title}> |
            ImageSourcePropType,
         onPress: () => void,
      }

   // Right button (all definitions)
   headerRight:
      React.Element<> |
      React.Component<{onPress, pressColorAndroid, tintColor, rightImage, 
                       title, truncatedTitle, titleStyle, width}> |
      {
         source:
            React.Element<> |
            React.Component<{style<ViewStyle> , tintColor, title}> |
            ImageSourcePropType,
         onPress: () => void,
         title: string,
         truncatedTitle: string,
      }
}
```
This fork is based on [@rghorbani's fork](https://github.com/react-navigation/react-navigation/pull/4331/commits/29189a6e517e7287d47ffd58b6089bf0d3da3e3e) for react-navigation which adds Custom Container Styles for the header Title and Buttons within the navigator options:
```
options {
    headerLeftContainerStyle: ViewStyle,
    headerRightContainerStyle: ViewStyle,
    headerTitleContainerStyle: ViewStyle
}
```

## Usage Examples:
1. Add a Right default button (the default icon is the ios-more vertical three dots)
```
options = {
    headerRight: {}
}
```

2. Use a Custom Action for Right Button:
```
options = {
    headerRight: { onPress: myCustomActionRight }
}
```
OR
```
options = {
    headerRightOnPress: myCustomActionRight
}
```

3. Use a custom Icon for Right Button:
```
options = {
    headerRightImage: { source: myImageURISourceOrReqSource }
}
```
OR
```
options = {
    headerRight: { source: myImageURISourceOrReqSource }
}
```

4. Use a Custom Action for Left Button:
```
options = {
    headerLeft: { onPress: myCustomActionLeft }
}
```
OR
```
options = {
    headerLeftOnPress: myCustomActionLeft 
}
```

5. Use a custom Icon for Left Button:
```
options = {
    headerBackImage: { source: myImageURISourceOrReqSource }
}
```
OR
```
options = {
    headerLeft: { source: myImageURISourceOrReqSource }
}
```

The customized actions will respect the default styles, tintcolor and spacing calculated by react-navigation or customized via style customizing options


hegelstad:
It already exists, https://github.com/vonovak/react-navigation-header-buttons. The docs are probably lacking tho..

Piropa:
Hello @hegelstad,

I have already checked react-navigation-header-buttons before I wrote this approach. So, whilst you may say overlook both approaches as similar, indeed they are not.

react-navigation-header-buttons deals with the creation and rendering of complex React Element configurations to replace the headerLeft or headerRight option variables, and, as the author says, "it tries to mimic" the react-navigation appearance of native navbar buttons.

This approach is different since it does not deal with the creation of complex button configurations nor tries to mimic native appearance. What it does it does is to extend and customize the **already present react-navigation left button code and styling** and mirrors it for the right button, hence giving you the chance to work inside the navigator options to get advantage of its already present and well defined behavior, like tintcolor, android underlay, margins, sizes, etc. and allowing you to customize the most basic ones like "onPress" or the icon "Image".

Then while you can use both approaches to accomplish similar results (from the outside), both solutions are different, this one straighthens the coding for most common scenarios, like left "menu" button (instead of back) or right  "more" button. Whilst the approach you mentioned can deal with more complex and not as striaghtforward configurations, like, i.e., two right buttons in the same header.

I hope this clearifies the motivation to create this extension and the differences with the one you have mentioned.

So no, it does not "already exists". It's a completely different solution.

Piropa:
Hi @hegelstad,

Well, I can number some advantages, however I wouldn't point them as "better", since "better" depends on personal preferences. I rather say this approach is more convenient for the simplest and more common use case, that is, Setting up a right button with same styling as left button, customizing button actions or customizing buttons icons, without having to deal with styling.

So, given the use case, some of the advantages would be:

1.- It saves the developer the need to dig into react-navigation code to duplicate the styles and measures this module already calculates based on platform and device.
2.- It makes straightforward the process of customizing icons and actions for header buttons
3.- It reduces the pile stock of code needed for putting a simple header button ready to work
4.- It takes advantage of react-navigation code and validations

You are welcome to try.

hegelstad:
1. 2. and 4. are as said ameliorated by react-navigation-header-buttons.

Point 3. is hardly an issue. You import the buttons and set them in the navigationConfig or in the navigationOptions of any screen. Customize if needed, or just drop them in ready to work.

The developers already stated that they have decided to modularize this library, so I am suggesting that you put less effort into going against the stream.

Piropa:
That's your point of view.... Mine is different. I think this is a more suitable solution for the most common case scenario.

mvanroon:
Thanks for your @Piropa I would love to see this merged.

My use case: I am looking for a way for the default react-navigation's headerLeft component to go back to the previous tab instead of going back in the stack navigator. I'm left now with having to recreate the default angle on ios and arrow on android and then providing a custom onPress function. This MR would get rid of 90% of the custom code needed to accomplish this.

vonovak:
> I'm left now with having to recreate the default angle on ios and arrow on android and then providing a custom onPress function

@mvanroon let me just point out this is not accurate. You can import the default `HeaderBackButton`, wrap it, and override the `onPress` handler, like this:

```js

export const SelectionAwareBackButton = (
  ({ onPress, ...other }) => {
    const onPressOverride = someLogicHere() ? someCustomOnPress : onPress;
    return <HeaderBackButton {...other} onPress={onPressOverride} />;
  }
);
```

then you may use it like this:

```
  static navigationOptions = () => ({
    headerLeft: SelectionAwareBackButton
  });
```


Also note that the header views have a [new home](https://github.com/react-navigation/react-navigation-stack/tree/master/src/views/Header)

mvanroon:
@vonovak excellent! No idea this existed. My apologies. Is this in the documentation somewhere? If not I’d be happy to add it.

vonovak:
@mvanroon this probably isn't explicitly documented anywhere; if you submit a short guide to the "Assorted Guides" in https://reactnavigation.org/docs/en/getting-started.html that would be appreciated!!

brentvatne:
@Piropa - thank you for this PR! I think the main issue that this points out is that we need some container component for the headers, so if you want to respect the standard platform margins it can be done for you out of the box. I'm hesitant to add any abstractions beyond that right now because it will be a maintenance burden and I think would provide limited value over just having a container. if someone would like to open a RFC (https://github.com/react-navigation/rfcs) to propose this (cc @vonovak as we discussed previously, maybe you're interested) it would be appreciated!

Piropa:
> @Piropa - thank you for this PR! I think the main issue that this points out is that we need some container component for the headers, so if you want to respect the standard platform margins it can be done for you out of the box. I'm hesitant to add any abstractions beyond that right now because it will be a maintenance burden and I think would provide limited value over just having a container. if someone would like to open a RFC (https://github.com/react-navigation/rfcs) to propose this (cc @vonovak as we discussed previously, maybe you're interested) it would be appreciated!

Hello @bretvane!... please tell me more about this you are proposing. It happens that right now I am currently rolling over the real burden of upgrading this I did for 2.6.2 to the late latests 2.18.1, by using react-navigation-stack.... and then it comes 3.0.0.!!... So, how can I help?

brentvatne:
hey @Piropa - what I'm saying is that I think the biggest issue people have is to add the proper amount of padding when they add a button (whether text-only or an icon) to the header left or header right, so that it's not just touching the edge of the screen

PR #4658, opened 2018-07-08
Error when building with haul: ref to pathToRegexp.compile

0xfurai:

## Motivation
Error in source code that cause problem on react native using haul https://puu.sh/AT1uZ/158623d5a4.png

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,27 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Error when building with haul: ref to pathToRegexp.compile(#4658).
 
+## [2.9.1] - [2018-07-24](https://github.com/react-navigation/react-navigation/releases/tag/2.9.1)
+### Fixed
+- Incorrect parameters passed to title offset calculation led to bug in header layout when no right component (https://github.com/react-navigation/react-navigation/issues/4754)
+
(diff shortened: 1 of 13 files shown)


ericvicenti:
Erm this is strange.. we're using normal documented behavior of pathToRegexp here.. https://github.com/pillarjs/path-to-regexp

I'm not sure whats going on with haul

cc @satya164

satya164:
cc @thymikee

thymikee:
I think it's a quirk in Babel acutally. And haul's failure indicates correct behavior (imho). Wouldn't a module like that be imported like this?:
```diff
-import pathToRegexp from 'path-to-regexp';
+import * as pathToRegexp from 'path-to-regexp';
```
The documentation uses require, so I'd assume you either do the above or:
```js 
import {compile} from 'path-to-regexp';
compile(...)
```

ericvicenti:
Oh, very strange that it somehow works in metro.

I'm not a fan of `import * ...` because I've seen it go wrong under various environments..

@Piotrovskyi, can you change this PR to import compile with `import {compile} from 'path-to-regexp';` ? Thanks!

brentvatne:
@ericvicenti - look good to you now?

rodrigobdz:
@brentvatne About the changelog entry, now that I think about it… It should be add under the _Unreleased_ section, right?

PR #4667, opened 2018-07-10
[flow] Make NavigationRoute types exact

Ashoat:
React Native 0.56 introduces Flow 0.75, which makes it impossible to refine `NavigationRoute` based on the presence of `index` or `routes` properties.

This PR turns `NavigationStateRoute` and `NavigationLeafRoute` into exact types, which addresses this issue.


ericvicenti:
This makes sense. Thanks for the improvement!

PR #4679, opened 2018-07-12
Fix #4608 - remove header left component only when a headerLeft optio…

### LINKED ISSUE #4608: Can't specify a headerLeft component when at the starting index of a StackNavigator
### Current Behavior
Just upgraded to v2. I found that I can no longer specify a `headerLeft` component in the StackNavigator when the scene index is at zero due to [this line](https://github.com/react-navigation/react-navigation/blob/3d06d19d6add53e8951e8f0bc793bf2bb8946940/src/views/Header/Header.js#L144).

### Expected Behavior
To maintain consistency with the previous version of my app, I need to be able to specify this component on the main screen, before navigation occurs. I am not sure if with v2 the API changed and there is now a preferred other way of achieving this.

### How to reproduce
1. Create a new StackNavigator
2. Assign it a `headerLeft` component in `navigationOptions`
3. Navigate between scenes, and the component should not render on the initial scene

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 2.4.1
| react-native     | 0.55.4
| node             | 8.11.3
| npm or yarn      | 1.7.0

Is there a known workaround to this issue?

papasmurf17:
## Motivation

Fix #4608 - remove header left component only when a headerLeft option is not specified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - StackNavigator.replace method no longer requires a key param. If the key is left undefined, the last screen in the stack will be replaced.
 
 ### Fixed
-- Remove header left component only when the headerLeft option is not specified (#4608).
+- Support headerLeft component for the first screen in a stack (#4608).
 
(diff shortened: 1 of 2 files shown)


ericvicenti:
Thanks for fixing this!

PR #4699, opened 2018-07-16
Fix issue undefined is not an object evaluating r.default

ebaynaud:
Fix https://github.com/aksonov/react-native-router-flux/issues/3124

## Motivation

Release builds fails without these changes

## Test plan

No more errors

## Code formatting

Simple letter put from uppercase to lowercase

## Changelog

Not required I guess

### DIFF
--- a/src/navigators/DrawerNavigator.js
+++ b/src/navigators/DrawerNavigator.js
@@ -4,7 +4,7 @@ import SafeAreaView from 'react-native-safe-area-view';
 
 import createNavigator from './createNavigator';
 import createNavigationContainer from '../createNavigationContainer';
-import TabRouter from '../routers/TabRouter';
+import tabRouter from '../routers/TabRouter';
 import DrawerScreen from '../views/Drawer/DrawerScreen';
 import DrawerView from '../views/Drawer/DrawerView';
 import DrawerItems from '../views/Drawer/DrawerNavigatorItems';
@@ -63,8 +63,8 @@ const DrawerNavigator = (routeConfigs, config = {}) => {
     ...tabsConfig
   } = mergedConfig;
 
-  const contentRouter = TabRouter(routeConfigs, tabsConfig);
-  const drawerRouter = TabRouter(
+  const contentRouter = tabRouter(routeConfigs, tabsConfig);
+  const drawerRouter = tabRouter(
     {
       [drawerCloseRoute]: {
         screen: createNavigator(contentRouter, routeConfigs, config)(props => (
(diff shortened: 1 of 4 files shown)


ericvicenti:
I'm confused why this would fix anything.. what bundler are you using?

ebaynaud:
Standard metro bundler... I do not know what really happens as you can see in https://github.com/aksonov/react-native-router-flux/issues/3124 but what I'm sure of is that it fixes the encountered issue for me and other people!

brentvatne:
thanks for submitting a pr! although this seems to fix your problem, i'm hesitant to land something that works without any reason that anyone can understand. it seems more like a bug in another tool.

ebaynaud:
I understand @brentvatne but it's a really minor change that could avoid me and others folks still using react-native-router-flux to fork react-navigation. That's the only workaround found yet...

brentvatne:
sorry unless someone can explain why this fixes the problem we cannot land it. the diff on its own makes no sense

TwilightOwl:
@brentvatne  https://babeljs.io/docs/en/babel-plugin-transform-class-constructor-call
This babel plugin is removed from babel 7, therefore your module doesn't work in React Native 0.56 in some cases. Now creating instance without 'new' keyword is impossible. So use 'tabRouter' instead 'TabRouter' or use new with 'TabRouter'

martincarrera:
Please @brentvatne could you reopen this PR? Thanks.

brentvatne:
thanks! yes

rishiankush:
I updated my react-navigation package with this above fix in all the files:
src/navigators/DrawerNavigator.js
src/navigators/StackNavigator.js
src/navigators/TabNavigator.js

but still getting the same issue in release mode:
`Unhandled JS Exception: TypeError: undefined is not an object (evaluating 'r.default')`

brentvatne:
cc @ebaynaud

warlock1024127:
I solved the same issue by: https://github.com/facebook/react-native/issues/20150, Workaround 2

abartolo:
@ebaynaud Thanks for the fix. I am now able to compile release build successfully.

@rishiankush Did you update all files? I see 4 files are changed but you only mentioned 3.

cc: @brentvatne

brentvatne:
if someone can confirm that this works as expected i can land it and publish an update to 1.x

abartolo:
@brentvatne I can confirm this is working. I have been building release builds for IOS and Android using the following "git+https://github.com/ebaynaud/react-navigation#1.x" in my package.json for about 3 weeks now.

brentvatne:
please create a new issue and follow the template @luco

Issue #4695, opened 2018-07-15
Set `headerTransparent` will still leave a headerline in iOS device


IronLu233:
### Current Behavior
I created a new project by `react-native init` command. And installed the react-navigation. So there is no extra packages in my project.
Then I change the code and the code is below:
```
/** Javascript
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 * @flow
 */

import React, { Component } from "react";
import { Platform, StyleSheet, Text, View } from "react-native";
import { createStackNavigator } from "react-navigation";

class Home extends React.Component {
  static navigationOptions = {
    headerTransparent: true
  };
  render() {
    return <View style={{ backgroundColor: "#E6E6FA" }} flex={1} />;
  }
}
const RootStack = createStackNavigator({
  Home: {
    screen: Home
  }
});

export default class App extends Component {
  render() {
    return <RootStack />;
  }
}

```
I run this code in iOS and Android simulator.
### Expected Behavior
![image](https://user-images.githubusercontent.com/20639676/42730062-ba540430-881d-11e8-8ee4-f3a8fcbf4107.png)
As we say, in Android simulator, header set to be transparent correctly. But in iOS, there leave an 
0.5 width headerline with color `#A7A7AA` in the headers bottom.
I find the code [here](https://github.com/react-navigation/react-navigation/blob/1fe11c100e72a632258e749d76bf542d425584b7/src/views/Header/Header.js#L561)may not consider  the transparent header situation.

### How to reproduce
I upload my repo to 
https://github.com/IronLu233/bug-report-4-react-navigation/blob/master/README.md

### Your Environment

| software         | version
| ---------------- | -------
| react-navigation |2.6.2
| react-native     |0.56
| node             |16.4.1
| yarn      |1.5.1

kirakik:
Hi @IronLu233, this is the intended behaviour of `headerTransparent` [according to the documentation](https://reactnavigation.org/docs/en/stack-navigator.html#headertransparent). You can override `headerStyle` to remove the bottom border. 

Is your intention to make the header completely transparent, or to hide the header entirely? By making it transparent, it still takes its full size, but just doesn't have a background color. If you want to hide it entirely, you should be setting the `header` value to `null` (see [here](https://reactnavigation.org/docs/en/stack-navigator.html#header))

kirakik:
good point @brentvatne, the naming is indeed confusing if the border style and elevation remain, I can open a PR for this later today

[...]

PR #4701, opened 2018-07-16
Set borderBottom to transparent and of size 0 if headerTransparent is true

kirakik:
## Motivation

As explained in issue #4695, setting `headerTransparent` still leaves a borderBottom of width 0.5 and color of `#A7A7AA`. This PR aims to fix that issue.

## Test plan

Before: 
img_1185

After:
img_1184


brentvatne:
thanks a lot! this looks good!

DarisCalinor:
Much appreciated! Looks great!

PR #4709, opened 2018-07-17
pass tintColor prop to HeaderRight

fondoger:
Just a few lines. Pass `tintColor` props to custom `headerRight` component, which is useful though not necessary.

And this is my first time to make a pr. So please tell me if I did something wrong.

Thanks for your time.

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -213,15 +213,15 @@ class Header extends React.PureComponent {
   _renderRightComponent = props => {
     const { options } = props.scene.descriptor;
     const headerRight = options.headerRight;
-    if (headerRight) {
-      return (
-        <headerRight
-          pressColorAndroid={options.headerPressColorAndroid}
-          tintColor={options.headerTintColor}
-        />
-      );
+    if (React.isValidElement(headerRight) || headerRight === null) {
+      return headerRight;
     }
-    return null;
+    return (
+      <headerRight
+        pressColorAndroid={options.headerPressColorAndroid}
+        tintColor={options.headerTintColor}
+      />
+    );
   };
 
   _renderLeft(props) {
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -217,8 +217,9 @@ class Header extends React.PureComponent {
       return headerRight;
     }
     if (!headerRight) return null;
+    const RenderedRightComponent = headerRight;
     return (
-      <headerRight
+      <RenderedRightComponent
         pressColorAndroid={options.headerPressColorAndroid}
         tintColor={options.headerTintColor}
       />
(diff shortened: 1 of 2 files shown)


fondoger:
And we can write code more conveniently, such as:
```
static navigationOptions = {
    headerRight: SomeHeaderRightButton,
};
```
instead of:
```
static navigationOptions = {
    headerRight: <SomeHeaderRightButton />,
};
```

fondoger:
OK, I will do it tomorrow when I can use my computer.

fondoger:
I have updated the changelog file. It's it ok now?

src/views/Header/Header.js
```diff
@@ -211,8 +211,18 @@ class Header extends React.PureComponent {
...
-    return headerRight || null;
+    const { options } = props.scene.descriptor;
+    const headerRight = options.headerRight;
+    if (React.isValidElement(headerRight) || headerRight === null) {
+      return headerRight;
+    }
+    if (!headerRight) return null;
+    return (
+      <headerRight
+        pressColorAndroid={options.headerPressColorAndroid}
```
rodrigobdz: Is `options.headerPressColorAndroid` always defined?

fondoger: I don't know. But there is a `headerPressColorAndroid` in the `_renderLeftComponent()` function, which is used to define the press ripple color on Android. So I copied it to the `_renderRightComponent()` function. 
They both receive props from `props.scene.descriptor`.

src/views/Header/Header.js
```diff
@@ -211,8 +211,18 @@ class Header extends React.PureComponent {
...
-    const { headerRight } = props.scene.descriptor.options;
-    return headerRight || null;
+    const { options } = props.scene.descriptor;
+    const headerRight = options.headerRight;
+    if (React.isValidElement(headerRight) || headerRight === null) {
+      return headerRight;
+    }
+    if (!headerRight) return null;
+    return (
+      <headerRight
```
ericvicenti: This won't work, components need to be capitalized. Plus the headerRight is an element  (it is already rendered as JSX). and is not a component.

fondoger: I want to make headerRight similar to headerLeft.

As `headerLeft` option can receive both rendered JSX element and react component, so `headerRight` might also can do this.
If the provided headerRight is an element, there is no difference to previous code. Here is the code:
```
if (React.isValidElement(headerRight) || headerRight === null) {
    return headerRight;
}
```
If the provided headerRight is a React component, then we do the render work and renturn it.

Similar to `headerLeft`, both two ways of providing `headerRight` component below are supported:
```
static navigationOptions = {
    headerRight: <SomeHeaderRightButton />,
};
```
```
static navigationOptions = {
    headerRight: SomeHeaderRightButton,
};
```

The new way allows headerRight access `tintColor` prop passed down from upper.

I will add a line so that the rendered component become capitalized.

Hope you can understand me.

fondoger:
The code of master has changed so much. I can't find `Header.js` files now.

This PR should be **CLOSE** now.

brentvatne:
@fondoger - yeah the source for stack lives in https://github.com/react-navigation/react-navigation-stack now

PR #4735, opened 2018-07-20
fix double backgroundColor iOS (causes problems with transparent/transluscent colors)

JulianKingman:
## Motivation

With a recent change, the header background no longer works if it's transluscent or transparent (in iOS). This change modifies make the background color apply in only one place.

## Test plan

I tested it on iOS and Android, with and without a background, and it appears to work fine. I also tried Android with and without elevation, and it also appears to work (the problem code involved a fix for android shadows)

## Code formatting

Done

## Changelog

Done


JulianKingman:
I can't tell by reading them, what are the failed tests about?

brentvatne:
fixed in 2.9.0

brentvatne:
fixed in 2.9.0

PR #4757, opened 2018-07-24
Fix typo in header transition preset check (uitkit -> uikit)

simonedavico:
## Motivation

The header transition preset check in `StackViewLayout` contains a typo, checking for `uitkit` instead of `uikit`.


brentvatne:
yikes thank you

PR #4787, opened 2018-07-30
Animate headerBackground between scene transitions

erickreutz:
Currently, headerBackground is sort of just plopped in place between scene changes which leads to it looking pretty jarring between transitions. What I'm proposing here is that we keep `headerBackground` rendered absolutely between scenes much like we do for the left, right and center. The default background interpolator transforms the x-axis moving it to the left and right as the scene transitions. This allows one to pull off the below in a much smoother fashion. 

background

As far as a complete implementation goes I'm not positive what I may be missing as I hacked this together generally quickly and have not tested it in places like android. Although the additions are pretty minimal and the current `headerBackground` option doesn't seem to be used outside of just placing it in the header container. If there is interest around getting this merged I'd be happy to see it to the finish line. What do you guys think?

### DIFF
--- a/src/views/Header/Header.js
+++ b/src/views/Header/Header.js
@@ -80,6 +80,7 @@ class Header extends React.PureComponent {
     titleFromLeftInterpolator: HeaderStyleInterpolator.forCenterFromLeft,
     titleInterpolator: HeaderStyleInterpolator.forCenter,
     rightInterpolator: HeaderStyleInterpolator.forRight,
+    backgroundInterpolator: HeaderStyleInterpolator.forBackground,
   };
 
   static get HEIGHT() {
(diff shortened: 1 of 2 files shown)


erickreutz:
I'm also confused as to why these tests are failing as they seem to be passing on my local machine.

ericvicenti:
Ooh, this looks fantastic! The header can be pretty finicky so hopefully we can have people test this a bit before we land it.

I suspect the test failures are unrelated. I'll rebase to re-run them

erickreutz:
Awesome! I think it’s a great addition too. Still looks like the tests are failing. Do you know why?

erickreutz:
@brentvatne https://github.com/react-navigation/react-navigation-stack/pull/3 doen

PR #4789, opened 2018-07-30
reactotron hello world

paulwilson123:

## Motivation

## Test plan

The code must pass tests.

## Code formatting

## Changelog

Add an entry under the "Unreleased" heading in [CHANGELOG.md](https://github.com/react-navigation/react-navigation/blob/master/CHANGELOG.md#unreleased) which explains your change.


ericvicenti:
Cool, I was playing around with something similar. We should definitely also log the navigation state change from createNavigationContainer's `dispatch` method, because that will show people what actions are happening and what state changes

PR #4791, opened 2018-07-31
Fixes bug where `null` doesn't work in routerOptions `paths` object for deeplinking

jparkrr:
## Motivation

Fixes the deeplink bug where, if you specified `null` in the routerOptions `paths`, the logic would actually fall back to the `path` specified on the route. This was happening because the logic was `option1 || option2`, and the value we wanted to define with option1 was falsey.

This was especially confusing for me at first because the fallback option is likely to be undefined, and then the path would be set as the default of the `routeName`.

By the way, this is related to https://github.com/react-navigation/react-navigation/pull/4671 and https://github.com/react-navigation/react-navigation/pull/1811, where it became possible to pass `null` as the path for the child router to handle the whole path. Thanks so much for this deeplinking work! It was hard to find, though, so what is the best place to document this?

## Test plan

Added a test

## Changelog

I will add this later if you think this solution is good enough. Thanks!

### DIFF
--- a/src/routers/__tests__/PathHandling-test.js
+++ b/src/routers/__tests__/PathHandling-test.js
@@ -547,6 +547,24 @@ const performRouterTest = createTestRouter => {
     expect(action.type).toEqual(NavigationActions.NAVIGATE);
     expect(action.routeName).toEqual('baz');
   });
+
+  test('paths option set as null on router overrides path from route config', () => {
+    const router = createTestRouter(
+      {
+        main: {
+          screen: MainNavigator,
+        },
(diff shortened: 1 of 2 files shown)


ericvicenti:
Thanks for the fix Justin, this looks great!

As for documentation, we should definitely add to this doc, https://github.com/react-navigation/react-navigation.github.io/blob/source/docs/deep-linking.md . I've been meaning to write a new doc on path handling, but have been distracted. If you want to draft one up, I would be happy to review and build upon it!

PR #4797, opened 2018-08-01
Updated NavigationEvents to support children

piet-v:
## Motivation

The current implementation requires the user to add a useless parent class if your own code is a neighbour of NavigationEvents.

```
import React from 'react';
import { View } from 'react-native';
import { NavigationEvents } from 'react-navigation';

const MyScreen = () => (
  <View>
    <NavigationEvents
      onWillFocus={payload => console.log('will focus',payload)}
      onDidFocus={payload => console.log('did focus',payload)}
      onWillBlur={payload => console.log('will blur',payload)}
      onDidBlur={payload => console.log('did blur',payload)}
    />
    {/* 
      Your view code
    */}
  </View>
);

export default MyScreen;
```
The new implementation would allow the user to get rid of this by adding the configuration without interfering with the layout of the component.

```
import React from 'react';
import { View } from 'react-native';
import { NavigationEvents } from 'react-navigation';

const MyScreen = () => (
  <NavigationEvents
    onWillFocus={payload => console.log('will focus',payload)}
    onDidFocus={payload => console.log('did focus',payload)}
    onWillBlur={payload => console.log('will blur',payload)}
    onDidBlur={payload => console.log('did blur',payload)}
  >
    {/* 
      Your view code
    */}
  </NavigationEvents>
);

export default MyScreen;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Updated `NavigationEvents` to support children.
 
 ## [2.9.3] - [2018-07-26](https://github.com/react-navigation/react-navigation/releases/tag/2.9.3)
 ### Added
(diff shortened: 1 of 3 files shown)


brentvatne:
why not use `React.fragment` or the shorthand `<>` for this?

piet-v:
Actually, I wasn't aware of ´React.fragment´. 

The RN project I'm currently working on is still on 0.51 while ´React.fragment´ is introduced in 0.52. Guess I'll make work of upgrading my version instead.

Thanks for the info! I'll close this ticket as it makes no sense to pollute the lib.

PR #4809, opened 2018-08-03
Extract stack to react-navigation-stack

brentvatne:
## Motivation

Make it easier to work on / for people to test an alternative stack built on top of screen components while staying up to date with any upstream changes.

## Test plan

Run navigation playground


src/__tests__/NavigationContainer-test.js
```diff
@@ -4,7 +4,9 @@ import { StyleSheet, View } from 'react-native';
 import renderer from 'react-test-renderer';
 
 import NavigationActions from '../NavigationActions';
-import createStackNavigator from '../navigators/createStackNavigator';
+
+// TODO: we should create a dummy navigator here
```
ericvicenti: What is a dummy navigator?

brentvatne: discussed on slack

PR #4824, opened 2018-08-07
Add disableRouteNamePaths option to router configs

charpeni:
## Motivation

Since `react-navigation@2.7.0`, the router lost a bit of the magic and to be able to deep link to a nested path we have to append each `routeName` we navigate through or make it explicit that paths are allowed to flow through with a `null` path, because routeName is used as default path.

So, I add `disableRouteNamePaths` to the router configs which is affecting the behavior here:
https://github.com/react-navigation/react-navigation/blob/73c76f1e4bbb015581eaae79a22e0076615ef392/src/routers/pathUtils.js#L86-L89

I find it more flexible than:

```javascript
const tabs = {
  ScreenA: screenGenerator('Screen A'),
  ScreenB: StackNavigator,
};

const TabNavigator = createTabNavigator(tabs, {
  paths: Object.keys(tabs).reduce((paths, path) => ({ ...paths, [path]: null }), {}),
});
```

## Test plan

I've added tests with Stack, Switch, and Tab router.

## Changelog

Add disableRouteNamePaths option to the router configs to disable the default behavior of using routeName as the default path.

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -107,7 +107,7 @@ export default (routeConfigs, stackConfig = {}) => {
   const {
     getPathAndParamsForRoute,
     getActionForPathAndParams,
-  } = createPathParser(childRouters, routeConfigs, stackConfig.paths);
+  } = createPathParser(childRouters, routeConfigs, stackConfig);
 
   return {
     childRouters,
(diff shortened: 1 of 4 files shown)


brentvatne:
this is an @ericvicenti pr

brentvatne:
@ericvicenti - when you get a chance, this seems pretty reasonable to me

charpeni:
@ericvicenti? 😺

ericvicenti:
This looks great, apologies for the delayed review!

brentvatne:
👌 @charpeni - can you resolve the merge conflict and update the PR? can land it and release when that's ready

charpeni:
Branch is up to date, eslint is fixed. 

Should be ready to go!

PR #4917, opened 2018-08-29
[flow] Fix missing headerLayoutPreset option in StackNavigatorConfig

MortalKastor:
As specified in the [official docs](https://reactnavigation.org/docs/en/stack-navigator.html#stacknavigatorconfig), `StackNavigatorConfig` allows one to optionally set `headerLayoutPreset` to either `left` or `center`, but the current flow libdef doesn't reflect that so `const stack = createStackNavigator({}, { headerLayoutPreset: "center" });` raises a flow error.

## Test plan

`const stack = createStackNavigator({}, { headerLayoutPreset: "center" });` should no longer raise error.


brentvatne:
thanks!

PR #4950, opened 2018-09-04
Pin create-react-context at v0.2.2

### LINKED ISSUE #4934: create-react-context is no longer Open Source in 0.2.3
As per https://github.com/jamiebuilds/create-react-context/commit/392ef2b40a4a6239ff63131c03797b7607708040, `create-react-context` is no longer Open Source. This was released as 0.2.3, which is inappropriately not a major version release because this is a breaking change.

`react-navigation` is currently depending on `create-react-context` `^0.2.1`, which will automatically install `0.2.3`

An easy solution would be to pin the dependency at either version `0.2.1` or `0.2.2`, without a `^` or `~`.

React Navigation's license is currently BSD, an Open Source license, but because of its dependency on `create-react-context`, installing it today necessarily means installing non-Open Source code.

ckknight:
Because `create-react-context` updated its license from MIT (read: Open
Source) to a more-restrictive, non-Open Source license without
appropriately bumping its major version in violation of semver, the
`create-react-context` version within `react-navigation` necessarily
needs to have its version pinned to a version of `create-react-context`
before the license change.

v0.2.2 is the most recent version of `create-react-context` that still
retains an Open Source license.

Note: Another solution to this would be to require `react` `16.3.x` or
higher, which includes the official `React.createContext` API without
need to rely on a third-party library.

Closes #4934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Pinned `create-react-context` dependency to `0.2.2` (https://github.com/react-navigation/react-navigation/issues/4934)
+
+
 ## [2.12.1] - [2018-08-23](https://github.com/react-navigation/react-navigation/releases/tag/2.12.1)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


brentvatne:
thank you!

PR #5067, opened 2018-10-05
[flow] Update StackViewConfig to match recent changes

Ashoat:
Two of these are recent property additions, and the other is a recent change to the signature for the `transitionConfig` function.


ericvicenti:
Looks good. `isModal` makes me cry a little inside, but thanks for flowifying the current state of the world!

PR #5074, opened 2018-10-07
[flow] typecheck getParam to actually check that key and fallback are valid

Klathmon:
## Motivation

I noticed that the type definition for the navigation prop's `getParam` was very loose and always returned `any`. This means if you are using flow and make a typo in the `key` argument of `getParam`, flow won't know there is a problem! This showed up in our codebase here:

image

(note `singup` instead of `signup`)

After implementing this fix, the error is caught right away!

image

Additionally, the return type of `getParam` is now inferred automatically meaning you have to spend less time defining types to get better flow coverage:

image

And it also typechecks the `fallback` argument to ensure types aren't mixed incorrectly:

image

### Details of typedef:

Just in case anyone needs it, a quick explanation of the typedef.

`getParam` is defined as a generic now, where the first parameter is `ParamName` which must match one of the keys defined in the `state.params` object. `ParamName` is inferred from the first argument of the function (`paramName`), and the `fallback` argument is now required to be of the same type as the `paramName` in the `state.params` object (in other words, `typeof fallback === state.params[paramName]`). `fallback` is still optional.

The return value of the function is also the same as the `fallback` type check, in that it mirrors the type definition in `state.params[paramName]`.

## Changelog

Added this entry:

- Update `getParam` flow typings to check `key` and `fallback` arguments, as well as return the correct type automatically.

Let me know if there are any changes or clarifications anyone needs or wants!


Klathmon:
1. Thanks! Forgot to change that... (we have the type from react-navigation wrapped in our app that applies these changes, and I used `State` instead of `S`!)

2. So i was worried about making any changes to the NavigationState as it would have pretty big effects on a lot of code, and to be honest I was having trouble following where all of the parts of it were defined. I figured this would end up breaking some apps, but I didn't realize that a lot of people were defining `NavigationScreenProp` without `params`...

To really solve the problem and keep backwards compatibility, i think we would need to basically disable this PR if `params` aren't defined, and enable it if they are. I have a hunch there's a way to do that, but I'm not 100% sure. It also means that users that don't define props won't ever know that they can have them typechecked, but I think that's an issue that can be solved through documentation.

I probably won't have time today to look at this, but I should early next week. And thanks for the in-depth checking, I appreciate it! I'm hoping that if this goes well, I can dive into other areas of react-navigation's types and see what can be improved.

Ashoat:
Unfortunately my `+params?: void` solution above doesn't quite work. It gets `$PropertyType<S, 'params'>` to resolve to `undefined` instead of erroring out, but `$Keys` still can't be applied to `undefined`. I think you're right that we need to figure out a way to conditionally apply this change only in the case where `params` is defined.

Looking forward to future improvements!

Klathmon:
@Ashoat I think I cracked it, but it's not pretty, and I think there's a lot that can be improved about it:

```
  getParam: <ParamName>(
    paramName: ParamName,
    fallback?: $ElementType<$PropertyType<{|
      ...{| params: {| [ParamName]: any |} |},
      ...$Exact<S>,
    |}, 'params'>, ParamName>,
  ) => $ElementType<$PropertyType<{|
    ...{| params: {| [ParamName]: any |} |},
    ...$Exact<S>,
  |}, 'params'>, ParamName>,
```

It's only changing the `getParam` function, no other type changes necessary. It's a bit verbose, and I'm not 100% sure it's working across flow versions, but it does seem to work correctly for me (flow 0.82).

If you don't have any props defined, it will return `any` type. If you have props defined, it returns the type of the prop by looking up the ParamName. It will also warn if you pass a `ParamName` in that isn't defined in the `params` object (but only if you have a `params` object defined at all). And it will warn if your fallback isn't the same type as what is defined (but again, only if you have a `params` object defined at all)

It works and typechecks if you define a `params` object, it works and doesn't typecheck if you don't, and it even works if you pass `any` or `*` as the `S` type.

I think at this point it's just testing, and validating that it works in all cases, and then making it a bit prettier and possibly simplifying some of the spreads out.

I don't have time to do any of that right now, but I will next week. But I figured I'd throw it here for you if you wanted to take a look before then.

Klathmon:
@Ashoat 

I updated the PR to have my latest (which is unchanged from my last comment).

It's not perfect (like you pointed out, it doesn't handle the maybe relationship with the fallback correctly), but It's more than good enough for now.

I noticed that your issue you opened is marked as closed in 0.83. That's still a bit too new to rely on IMO, but if you are cool with it I can update it to that.

Otherwise if this PR is working for you, I think it's ready to merge!

Ashoat:
I do still think it would be better `void` instead of `any` in the `{| params: {| [ParamName]: any |} |}` part. I'm actually surprised that misspelling a param name generates a type error at all when using `any`, since `any` usually hides type errors.

Klathmon:
But void is analogous to `undefined` isn't it? We want it to return `any` when there is no param structure defined in the`NavigationScreenProp`, because the value could really be `any`.

`void` would imply that it is returning `undefined`, which would throw errors for everyone that is currently using something like `const signUp = this.navigation.getParam('signUp')` without defining the shape of the params in `NavigationScreenProp`.

And the reason misspelling a param name generates a type error when you define `params` in `NavigationScreenProp` is because of the way the spreading is happening. If you don't define `params` in `NavigationScreenProp` then params is defined as `{| [ParamName]: any |}`. But if you do define ANY param name at all, it will override the entire `params` prop with what you have, meaning the `{| [ParamName]: any |}` isn't ever used at all. That's why it's safe here to use `any` because it won't ever apply if you define any `params` object (even `{}` will trigger this).

Ashoat:
`void` is the type of the value `undefined`. I don’t think it’s possible to avoid typing `NavigationScreenProp`’s type parameter. You can use the existential type (`*`), but then Flow should just infer the shape based on usage.

That said, the most generic type parameters people should be using are `NavigationStateRoute` or `NavigationLeafRoute`, which would return `mixed` rather than `any`, which avoids “crippling” the typechecker. Of course it’s better to type the precise shape of the parameters (like I did [https://github.com/Ashoat/squadcal/commit/5fb267826342c80c351104cfe4f331440ccdbad8](here), for instance), but it’s not a bad fallback.

Klathmon:
You have to type`NavigationScreenProp`, but you don't have to put a `params` key in the prop, which is what this workaround takes advantage of.

But I was under the assumption that using `void` would make flow mark the return value from `getParam` as `undefined`, but it seems like it doesn't do that strangely... So I guess I still have some learning to do about `void`!

I can switch to using void there and i'll commit that soon. Thanks again for all the help on this

Ashoat:
Awesome, thanks again for making this PR - it's a huge improvement to `getParam`. Looking forward to any future PRs from you!

@brentvatne - I don't seem to be able to merge PRs, so I'll leave it to you.

brentvatne:
@Ashoat - strange! you're a collaborator on the project. maybe it's the reviewer block

Issue #1527, opened 2017-05-15
How to make deep linking for nested navigators?


diegorodriguesvieira:
I have this code below, I am trying to do deep linking with nested navigators, but it is not working...

I'm using the URL: test://test/main/mainscreentwo

I'm trying to reproduce this example
https://github.com/react-community/react-navigation/issues/687

```
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Text,
  View,
  Platform,
} from 'react-native';

import { StackNavigator, TabNavigator } from 'react-navigation';

class MainScreenOne extends Component {
  render() {
    return (
      <View>
        <Text>
          MainScreenOne
        </Text>
      </View>
    );
  }
}

class MainScreenTwo extends Component {
  render() {
    return (
      <View>
        <Text>
          MainScreenTwo
        </Text>
      </View>
    );
  }
}

class MainScreenThree extends Component {
  render() {
    return (
      <View>
        <Text>
          MainScreenThree
        </Text>
      </View>
    );
  }
}

class SecondScreen extends Component {
  render() {
    return (
      <View>
        <Text>
          SecondScreen
        </Text>
      </View>
    );
  }
}

const App = StackNavigator({
  Main: {
    screen: TabNavigator({
      MainScreenOne: {
        screen: MainScreenOne,
        path: 'mainscreenone',
      },
      MainScreenTwo: {
        screen: MainScreenTwo,
        path: 'mainscreentwo',
      },
      MainScreenThree: {
        screen: MainScreenThree,
        path: 'mainscreenthree',
      },
    }),
    path: 'main',
  },
  Second: {
    screen: SecondScreen,
    path: 'second',
  },
})

const prefix = Platform.OS == 'android' ? 'test://test/' : 'test://';
const MainApp = () => <App uriPrefix={prefix} />;
AppRegistry.registerComponent('AppMuffato', () => MainApp);
```

filippofilip95:
I have the same question :)

rickardinho:
Me too

aaronbuchanan:
@diegorodriguesvieira @filippofilip95 @rickardinho any of you make progress on this in the last 27 days :)?

aroth:
Anything?

filippofilip95:
Unfortunately no :(

twairball:
This works if you pass `prefix` to the child navigator. 

````
# Root.js
const prefix = Platform.OS == 'android' ? 'test://test/' : 'test://';
[...]

PR #5120, opened 2018-10-11
add recursive function to return deeply nested actions

Taylor123:
## Motivation
Adds out of the box support for #1527

Deep linking does not currently support navigating within nested navigators.

Example Navigator architecture:

```
StackNavigator {
  View1
  TabNavigator: {
    Stack1: {
      View2
      View3
    }
    Stack2
  }
}
```

Calling a deep link within `StackNav -> TabNav -> Stack1 -> View2` to move to `view3` does not dispatch the right action. If I were to call open the deep link from a view within `Stack2`, then it works.

## Test plan

I've tested on both `iOS` and `android`. Happy to have others test it within their use cases!

Existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+- Updated `_handleOpenUrl` to get deeply nested actions
+
 ## [2.17.0] - [2018-09-25](https://github.com/react-navigation/react-navigation/releases/tag/2.17.0)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


ericvicenti:
I don't think this is a good idea, won't it result in unexpected behavior most of the time?

I think it should be fine to provide a sub-action, and ensure the router produces the correct top-level action. If this isn't working correctly for you on the latest versions of react-navigation, can you create a fresh issue with a simple example of the bug?

Taylor123:
@ericvicenti I didn't see it creating any unexpected behavior in my application, but you know more about how the actions are created than I.

Do you have any examples of how I could accomplish this? In the question I linked to, the only solutions that worked for me were directly changing the action that is dispatched since the router isn't generating the right top level action.

PR #5173, opened 2018-10-25
[flow] Fix type of getCurrentNavigation

Ashoat:
As of 2.18 we seem to be okay returning `null` here.

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -535,14 +535,26 @@ declare module 'react-navigation' {
     ) => NavigationEventSubscription,
     getParam: <ParamName>(
       paramName: ParamName,
-      fallback?: $ElementType<$PropertyType<{|
-        ...{| params: {| [ParamName]: void |} |},
-        ...$Exact<S>,
-      |}, 'params'>, ParamName>,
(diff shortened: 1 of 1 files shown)


Ashoat:
My force pushing is confusing GitHub...

Ashoat:
Actually that's just Prettier's reformatting, I guess the last Flow commit somehow skipped it.

brentvatne:
i don't think anything changed in 2.18.0, so i don't think anything changed in 2.18.0 but in 2.18.1 we call `getCurrentNavigation()` and expect it to either be undefined or the navigation object.

```js
  let getNavigation = obj => {
    console.log({ inner: obj.getCurrentNavigation() });
    return 1;
  };
  this._navigation = getNavigation({
    getCurrentNavigation: () => this._navigation,
  });
  console.log({ outer: this._navigation });
```

```js
{inner: undefined}
{outer: 1}
```

so in the first pass through `getNavigation` we pass the result of `getCurrentNavigation` into `getChildrenNavigationCache` (introduced in 2.18.1) and if it is falsey then we just return an empty object. we could return some other result from here if that is helpful for redux.

regardless, this pr is correct, thanks

Issue #3031, opened 2017-11-22
Cannot reset nested navigator when resetting root one


JCMais:
Imagine that we have the following structure
```
const ChildNavigatorA = StackNavigator({
  ScreenA: {
    screen: ScreenA,
  },
  ScreenB: {
    screen: ScreenB,
  },
});
const ChildNavigatorB = StackNavigator({
  ScreenC: {
    screen: ScreenC,
  },
  ScreenD: {
    screen: ScreenD,
  },
});
const AppNavigator = StackNavigator({
  Home: {
    screen: RandomHomeComponent,
  },
  ScreenA: {
    screen: ChildNavigatorA,
  },
  ScreenC: {
    screen: ChildNavigatorB,
  },
});
```

Let's say that I'm on `ScreenA` and navigate to `ScreenC`.

The navigation state will look like this:
```
  navigation: {
    index: 2,
    routes: [
      {
        routeName: 'Home',
        key: 'Init-id-1511364125347-0',
      }
      {
        index: 0,
        routes: [
          {
            routeName: 'ScreenA',
            key: 'Init-id-1511364125347-1',
          },
        ],
        key: 'id-1511364125347-2',
        routeName: 'ScreenA'
      },
      {
        index: 0,
        routes: [
          {
            routeName: 'ScreenC',
            key: 'Init-id-1511364125347-3',
          }
        ],
        key: 'id-1511364125347-4',
        routeName: 'ScreenC'
      }
    ],
    allOptions: {
      navigationOptions: {
        header: null
      }
    }
  },
```

Now, the user did something on `ScreenC`, and I would like to reset their navigation state, so it was like they were going from `Home` => `ScreenA` => `ScreenB`, so I call:

```
navigation.dispatch(
  NavigationActions.reset({
    key: null,
    index: 1,
    routeName:  'Home',
    actions: [
      NavigationActions.navigate({
        routeName: 'Home',
      }),
      NavigationActions.navigate({
        routeName: 'ScreenA',
        action: NavigationActions.reset({
          index: 1,
          routeName: 'ScreenA',
          actions: [
            NavigationActions.navigate({
              routeName: 'ScreenA',
            }),
            NavigationActions.navigate({
              routeName: 'ScreenB',
            }),
          ],
        }),
      }),
    ],
  }),
);
```

### Current Behavior
[...]

PR #5175, opened 2018-10-26
StackRouter.getStateForAction(reset()) should work without passing state 2nd arg

slorber:
I just migrated from 1.0-beta to 2.18.1 and the behavior of StackRouter.getStateForAction changed regarding reset action.

Let's consider this navigator:

```
const AppNavigation = createStackNavigator(
  {
    HomeScreen: { screen: HomeScreen },
    SecondScreen: { screen: SecondScreen },
  },
);
```

Before, one could do:

```js
const SECOND_SCREEN_STATE = getStateForAction(
  StackActions.reset({
    index: 0,
    actions: [NavigationActions.navigate({ routeName: 'SecondScreen' })],
  }),
);
```

now, the 2nd arg is mandatory for the reset action to behave correctly. If you don't pass the initial state arg, getStateForAction always reset to the initial route name ("HomeScreen")

```js
const INITIAL_STATE = getStateForAction(
  NavigationActions.navigate({ routeName: 'HomeScreen' }),
);

const SECOND_SCREEN_STATE = getStateForAction(
  StackActions.reset({
    index: 0,
    actions: [NavigationActions.navigate({ routeName: 'SecondScreen' })],
  }),
  INITIAL_STATE, // => this is now needed
);
```

I think this behavior is confusing for the users that would expect reset to work without providing any state. For example: https://github.com/react-navigation/react-navigation/issues/3031#issuecomment-385751703

I don't know when this behavior changed, but as far as I know I don't remember seing it in breaking changes of 2.x.

About the PR: it is probably not a very elegant implementation. Maybe I should handle this in "getInitialState(action)" instead, not sure how however. Also there's the "replace" action to consider? If this change is approved, please tell me how to polish my PR

### DIFF
--- a/src/routers/StackRouter.js
+++ b/src/routers/StackRouter.js
@@ -109,7 +109,7 @@ export default (routeConfigs, stackConfig = {}) => {
     getActionForPathAndParams,
   } = createPathParser(childRouters, routeConfigs, stackConfig);
 
-  return {
+  const router = {
     childRouters,
 
     getComponentForState(state) {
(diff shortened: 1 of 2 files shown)


satya164:
@ericvicenti can you take a look? :)

satya164:
I'm closing this since we're already at 4.x and it doesn't make much sense to change the behaviour in 2.x.

PR #5266, opened 2018-11-22
./src/createNavigationContainer -> @react-navigation/native/src/createAppContainer on NavigationTestUtils.js

watanabeyu:
## Motivation

For App-test.js on expo.
I updated react-navigation to v3, and want to test navigation, but error was occured.  
So replace it.

## Test plan

The code must pass tests.

## Code formatting

Lint-Staged handled

## Changelog

- [x] Add an entry under the "Unreleased" heading in [CHANGELOG.md](https://github.com/react-navigation/react-navigation/blob/master/CHANGELOG.md#unreleased) which explains your change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


brentvatne:
thank you!

Issue #5293, opened 2018-11-27
Cannot run NavigationPlayground ("Tried to register two views with the same name RNGestureHandlerButton")


koheku:
### Current Behavior
I currently get a redbox when trying to run the `NavigationPlayground` sample app.

![img_3474](https://user-images.githubusercontent.com/1695036/49094242-148a6400-f266-11e8-9b15-d6813f2a23b3.PNG)

### How to reproduce

Following the docs, run:
```
yarn install
cd examples/NavigationPlayground
yarn install
yarn start
```

### Resolving it
Similar [to this commit](https://github.com/koheku/react-navigation/commit/a393818875ab3c8c47849fa656c7c084d3a9d4f3), deleting the `react-native-gesture-handler` post-install from `NavigationPlayground/node_modules` solves it in my environment. I will try and submit a PR :)

brentvatne:
🙏

PR #5294, opened 2018-11-27
Fix for NavigationPlayground red box (duplicate native module)

koheku:
Solves https://github.com/react-navigation/react-navigation/issues/5293


brentvatne:
thanks!

PR #5622, opened 2019-02-25
remove unused import and unused destructure on CustomTransitioner.js

TitusEfferian:
## Motivation

when i was looking for reference on how to make custom transitioner, i saw this example code and found some unused import and unused destructure on `_renderScene` method (line of code 64).


matthamil:
Thanks!

PR #5627, opened 2019-02-25
fixed typo in Custom Tab UI example

FibreFoX:
## Motivation

While browsing the examples, especially around the custom tab bar options, one obvious typo seems to have sneaked itself into the code. This pull-request fixes that wrong spelling.

## Test plan

No test-plan, it's some already existing example.

## Code formatting

Did that, nothing bad was found.

## Changelog

No changelog required


brentvatne:
thx

PR #5653, opened 2019-03-05
On action before swipe

hoangpham95:

## Motivation

I want to have custom action before the swipe animation took place.

## Test plan

Video demonstration will be provided on both Android and iOS.


brentvatne:
we don't support react-navigaton 1.x anymore, you're welcome to maintain this on your own fork though! :)

PR #5655, opened 2019-03-05
On action before swipe

hoangpham95:
## Motivation
Custom behavior for action before the swiping back animation.

## Test plan
Video UI will be uploaded on both Android and iOS


brentvatne:
hi! same comment applies as https://github.com/react-navigation/react-navigation/pull/5653#issuecomment-469498986

maybe you opened the pr to the public repo by accident, rather than on your own fork?

hoangpham95:
My mistake. It was supposed to be the update for the other PR.

PR #5664, opened 2019-03-07
CustomTabs – consolidate all imports from react-navigation on the sam…

StevenBlack:
…e line.

## Motivation

We had two `import` statements from the same resource (`react-navigation`).

If merged, this simple PR consolidates all `import ... from "react-navigation"` in the same statement.

## Test plan

Not applicable 

## Code formatting

👍 

## Changelog

Not required


eriveltonelias:
thanks!

PR #5276, opened 2018-11-23
Add TypeScript declaration file.

janhesters:
## Motivation
As discussed in [this issue](https://github.com/react-navigation/react-navigation-core/issues/2) we should manage the TypeScript typings from within react-navigation.

This file extends the old DefinitelyTyped file:

What I added (and needs to be checked / reviewed): `unmountInactiveRoutes`, `drawerType` and missing `contentOptions` for `DrawerItems`.
Added `defaultNavigationOptions`.
Added default params inside route definitions.
Added `headerBackgroundTransitionPreset`.
Added `cardOverlayEnabled` and `cardShadowEnabled`.
Added `createAppContainer` (thanks to @BrendonSled).
Added `Context` and types in `package.json` (thanks to @bySabi).
 
What I couldn't add:
The support for React Hooks (I didn't find the code in V3).
~~Type definitions for `createAppContainer` (I didn't have time to look into it and understand the code).~~

**TODO:**
1. Changelog
2. React Hooks
3. See [this comment](https://github.com/react-navigation/react-navigation/pull/5276#issuecomment-444816485) below for how you can help.

## Changelog

**TODO:** Add an entry under the "Unreleased" heading in [CHANGELOG.md](https://github.com/react-navigation/react-navigation/blob/master/CHANGELOG.md#unreleased) which explains your change.

PR #5681, opened 2019-03-13
Add TypeScript declaration file and convert NavigationPlayground to TypeScript

brentvatne:
I pulled #5276, made some changes to it and rebased


brentvatne:
https://github.com/react-navigation/react-navigation/commit/eac5feb64e2170f370a815f10af77b471e146648

more work to do but we have some initial thing in there now that we can work off of

solkaz:
@brentvatne which version of `react-navigation` would ship the TS definitions? After that's published, it'd be good to have the typings [removed from DT](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package).

brentvatne:
3.4.0 will include, i've released a rc for it - 3.4.0-rc.0 - feel free to try it in your app and let me know how it goes

PR #5715, opened 2019-03-20
[TypeScript] Export DrawerIconProps and DrawerLabelProps

itsjgf:
## Motivation

The props used in `drawerIcon` and `drawerLabel` are not exported, this PR fixes it.


brentvatne:
thanks!

PR #5727, opened 2019-03-22
unhandled promise rejection failed to delete storage directory

oxbits:
with:

await AsyncStorage.clear();

I'm getting this in ios:

unhandled promise rejection failed to delete storage directory

found this:

https://stackoverflow.com/questions/46736268/react-native-asyncstorage-clear-is-failing-on-ios

switched to use this instead:

await AsyncStorage.getAllKeys().then(AsyncStorage.multiRemove)

but that causes unhandled promise rejection in android.

this is working:

Platform.OS === 'ios' ? await AsyncStorage.getAllKeys().then(AsyncStorage.multiRemove) : await AsyncStorage.clear()

## Motivation

In: "Switch between routes" item in example app

Example does not sign out:

unhandled promise rejection failed to delete storage directory

## Test plan

I tested the app in both android and ios

## Code formatting

I am happy to reformat per any suggestions.

### DIFF
--- a/examples/NavigationPlayground/app/SwitchWithStacks.tsx
+++ b/examples/NavigationPlayground/app/SwitchWithStacks.tsx
@@ -5,6 +5,7 @@ import {
   StatusBar,
   StyleSheet,
   View,
+  Platform,
 } from 'react-native';
 import { createStackNavigator, createSwitchNavigator } from 'react-navigation';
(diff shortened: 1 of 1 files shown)


brentvatne:
yikes this is a really silly upstream react-native bug..

PR #5796, opened 2019-04-09
Improved Type Definition for Navigator navigationOptions.

guptaamol:
`NavigationTabRouterConfigBase` as a type  made no sense for `navigationOtions` updated it  appropriately for the two tabNavigators.

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -556,9 +556,7 @@ declare module 'react-navigation' {
     defaultNavigationOptions?: NavigationScreenConfig<
       NavigationBottomTabScreenOptions
     >;
-    navigationOptions?: NavigationScreenConfig<
-      any
-    >;
+    navigationOptions?: NavigationScreenConfig<any>;
   }
   export interface TabScene {
     route: NavigationRoute;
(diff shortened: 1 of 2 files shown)


typescript/react-navigation.d.ts
```diff
@@ -549,14 +549,16 @@ declare module 'react-navigation' {
...
+    navigationOptions?: NavigationScreenConfig<NavigationScreenOptions>;
   }
   export interface NavigationBottomTabRouterConfig
     extends NavigationTabRouterConfigBase {
     defaultNavigationOptions?: NavigationScreenConfig<
       NavigationBottomTabScreenOptions
     >;
-    navigationOptions?: NavigationTabRouterConfigBase;
+    navigationOptions?: NavigationScreenConfig<
+      NavigationBottomTabScreenOptions
```
brentvatne: we can't actually type this very well because the type depends on the parent navigator :\ 

```js
let Tabs = createBottomTabNavigator({ Home }, {
  defaultNavigationOptions: { /* this applies to Home route to configure bottom tabs */ },
  navigationOptions: { /* this applies to Tabs route on stack to configure stack */ },
});
let Stack = createStackNavigator({ Tabs });
```

might have to do `NavigationScreenConfig<any>` or `NavigationScreenConfig< NavigationScreenOptions>` for now. of course this isn't great either because `NavigationScreenOptions` is only a union of build in navigation option types.. so probably `any` is the most accurate

guptaamol: Hmmm, makes sense i guess it is the best option.

PR #5806, opened 2019-04-11
[flow] Update for Flow 0.92

Ashoat:
Changes:

1. We can now type React components with static members as `React.ComponentType` (instead of requiring the `StatelessFunctionalComponent`) hack.
2. We can now type-parameterize React components on all of their props, instead of just the ones we care about. Not sure why this wasn't previously. Note: this is a breaking change, in the sense that people's parameterization of `NavigationScreenComponent`, `NavigationNavigator`, and `NavigationContainer` may need to change.
3. The order of object-type-spread in the `getParam` definition has been switched. It was giving errors with the old order.
4. Avoid recursive type inference `NavigationScreenProp`. This was yielding an "*** Recursion limit exceeded ***" error, which was rather difficult to debug. We can just use a `NavigationState` and let the user cast.
5. Fix `onTransitionStart`/`onTransitionEnd` types in `NavigationStackViewConfig`.
6. Add `navigationConfig` property to `NavigationView`.

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -353,12 +353,8 @@ declare module 'react-navigation' {
   declare export type NavigationScreenComponent<
     Route: NavigationRoute,
     Options: {},
-    Props: {}
-  > = React$ComponentType<{
-    ...Props,
-    ...NavigationNavigatorProps<Options, Route>,
-  }> &
-    withOptionalNavigationOptions<Options>;
+    Props: NavigationNavigatorProps<Options, Route>
+  > = React$ComponentType<Props> & withOptionalNavigationOptions<Options>;
 
   declare interface withRouter<State, Options> {
     router: NavigationRouter<State, Options>;
@@ -367,11 +363,8 @@ declare module 'react-navigation' {
   declare export type NavigationNavigator<
     State: NavigationState,
     Options: {},
-    Props: {}
-  > = React$StatelessFunctionalComponent<{
-    ...Props,
-    ...NavigationNavigatorProps<Options, State>,
-  }> &
+    Props: NavigationNavigatorProps<Options, State>
+  > = React$ComponentType<Props> &
     withRouter<State, Options> &
     withOptionalNavigationOptions<Options>;
 
@@ -463,8 +456,14 @@ declare module 'react-navigation' {
       prevTransitionProps: ?NavigationTransitionProps,
       isModal: boolean
     ) => TransitionConfig,
-    onTransitionStart?: () => void,
-    onTransitionEnd?: () => void,
+    onTransitionStart?: (
+      transitionProps: NavigationTransitionProps,
+      prevTransitionProps: ?NavigationTransitionProps
+    ) => void,
+    onTransitionEnd?: (
+      transitionProps: NavigationTransitionProps,
+      prevTransitionProps: ?NavigationTransitionProps
+    ) => void,
     transparentCard?: boolean,
     disableKeyboardHandling?: boolean,
   |};
@@ -589,8 +588,8 @@ declare module 'react-navigation' {
       fallback?: $ElementType<
         $PropertyType<
           {|
-            ...{| params: {| [ParamName]: void |} |},
             ...$Exact<S>,
+            ...{| params: {| [ParamName]: void |} |},
           |},
           'params'
         >,
@@ -599,14 +598,14 @@ declare module 'react-navigation' {
     ) => $ElementType<
       $PropertyType<
         {|
-          ...{| params: {| [ParamName]: void |} |},
           ...$Exact<S>,
+          ...{| params: {| [ParamName]: void |} |},
         |},
         'params'
       >,
       ParamName
     >,
-    dangerouslyGetParent: () => NavigationScreenProp<*>,
+    dangerouslyGetParent: () => ?NavigationScreenProp<NavigationState>,
     isFocused: () => boolean,
     // Shared action creators that exist for all routers
     goBack: (routeKey?: ?string) => boolean,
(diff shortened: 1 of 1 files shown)


wschurman:
> The order of object-type-spread in the getParam definition has been switched. It was giving errors with the old order.

Arrived here by trying your solution at https://github.com/flow-typed/flow-typed/issues/2959. What errors you were seeing with the old order? With the new ordering, the value type for `getParam` always resolves to `void` (undefined) since the fallback spread (`...{| params: {| [ParamName]: void |} |}`) always clobbers the params key supplied in the first spread (type `$Exact<S>`).

```
$ElementType<
  $PropertyType<
    {|
      ...$Exact<S>,
      ...{| params: {| [ParamName]: void |} |},
    |},
    'params'
  >,
  ParamName
>
```

Ashoat:
Ah, interesting. `flow type-at-pos` reports that you're right - the type of `getParam`'s return is now `empty` for me, although it seems to have been `any`-ified at some point, as Flow reports no errors when it is downcast. That must be why switching the orders didn't yield any errors.

I'm seeing an error purportedly about a mix-up between the `mixed` return of `NavigationParams` and the `void` default return. Here's the results I'm seeing on Flow 0.92 when running `flow` and `flow --show-all-branches`:

```
ashoat@ashoatmbp2018 [~/Dropbox/src/squadcal/native]$ flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ app.react.js:100:54

Cannot call createReduxContainer with RootNavigator bound to Navigator because undefined [1] is incompatible with
mixed [2] in property getParam of the first argument of property router.getScreenOptions.

     app.react.js
      97│ }
      98│
      99│ const msInDay = 24 * 60 * 60 * 1000;
     100│ const ReduxifiedRootNavigator = createReduxContainer(RootNavigator);
     101│
     102│ type NativeDispatch = Dispatch & ((action: NavigationAction) => boolean);
     103│

     flow-typed/npm/@react-navigation/core_v3.x.x.js
 [2] 100│     [key: string]: mixed,

     flow-typed/npm/react-navigation_v3.x.x.js
 [1] 595│             ...{| params: {| [ParamName]: void |} |},



Found 1 error

Only showing the most relevant union/intersection branches.
To see all branches, re-run Flow with --show-all-branches
ashoat@ashoatmbp2018 [~/Dropbox/src/squadcal/native]$ flow --show-all-branches
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ app.react.js:100:54

Cannot call createReduxContainer with RootNavigator bound to Navigator because:
 • Either property router is missing in React.ComponentType [1] but exists in withRouter [2].
 • Or undefined [3] is incompatible with mixed [4] in property getParam of the first argument of property
   router.getScreenOptions.
 • Or property router is missing in withOptionalNavigationOptions [5] but exists in withRouter [2].

     app.react.js
      97│ }
      98│
      99│ const msInDay = 24 * 60 * 60 * 1000;
     100│ const ReduxifiedRootNavigator = createReduxContainer(RootNavigator);
     101│
     102│ type NativeDispatch = Dispatch & ((action: NavigationAction) => boolean);
     103│

     flow-typed/npm/@react-navigation/core_v3.x.x.js
 [4] 100│     [key: string]: mixed,
        :
 [2] 350│     withRouter<State, Options> &

     flow-typed/npm/react-navigation_v3.x.x.js
 [1] 371│   > = React$ComponentType<Props> &
     372│     withRouter<State, Options> &
 [5] 373│     withOptionalNavigationOptions<Options>;
        :
 [3] 595│             ...{| params: {| [ParamName]: void |} |},



Found 1 error
```

wschurman:
I can repro and get a similar error, though not that exact one. Thinking about it more, the ordering more depends on how people prefer to type their codebases. I also saw that there was some discussion about this on https://github.com/react-navigation/react-navigation/pull/5074.

In code without explicit param type specification in `NavigationScreenProp` (i.e. just `NavigationScreenProp<NavigationRoute>`), I see the two following behaviors:
- By ordering them the way it is now (after this commit), the error is in the at the callsite and is due to the return type being undefined.
- Reverting the ordering change seems to have the same effect for me except the error shows in two places (the mixed and undefined), which is more confusing.
I argue the current ordering is better for this case, but only slightly as they're both still errors which can be alleviated by users by forcing the type as `any`.

In code with explicit navigation params type specification (i.e. `NavigationScreenProp<{| params: {...} |}>`), I see the two following behaviors:
- By ordering them the way it is now the type information is lost and the return type is always void.
- Reverting the ordering produces the correct type error: `Cannot call this.props.navigation.getParam with 'blah' bound to paramName because property
blah is missing in NavigationParams [1].`

I'm almost more tempted to make the types more strict by requiring typing the navigation params for components that use `getParam`. That's more up to the active contributors (you) though as it's more of a design question. The type could then be something simpler like:
```
$ElementType<
  $PropertyType<
    $Exact<S>,
    'params'
  >,
  ParamName
>
```
The downside is more cryptic errors like `an indexer property is missing in undefined [1].` when no params typing is specified (just using `NavigationScreenProp<NavigationRoute>` which is common practice).

Ashoat:
Just using `$PropertyType<$Exact<S>, 'params'>` won't work because `S` can be `NavigationState` (for the root navigator), which doesn't have a `params` property.

After some work I figured out a working solution. It actually looks really clean. The trick is to use an unsealed object instead of a sealed object.

```js
     ) => $ElementType<
       $PropertyType<
         {|
+          ...{| params: { } |},
           ...$Exact<S>,
-          ...{| params: {| [ParamName]: void |} |},
         |},
         'params'
       >,
```

I can put up some PRs for this later, or if you'd like to please feel free to preempt me!

PR #5809, opened 2019-04-13
chore: remove createTabNavigator out of ts definition

jarvisluong:
## Motivation

Remove `createTabNavigator` out of ts definition in v3 since it is removed in the source


brentvatne:
thanks @jarvisluong!

PR #5820, opened 2019-04-17
Add missing type definitions for DrawerNavigatorConfig

edmundojr:
Noticed that some definitions were missing when using `drawerType`: https://reactnavigation.org/docs/en/drawer-navigator.html#drawernavigatorconfig

First PR here, let me know if I need to change anything :)


brentvatne:
lgtm, if you want to follow up with another pr to add some annotations about default values that could be helpful!

PR #5828, opened 2019-04-18
Removes duplicate NavigationInjectedProps export

mlodato517:
## Motivation

I just noticed that `NavigationInjectedProps` is defined the same way twice so I think we can remove one of them.

## Changelog

Didn't know if this warranted an entry in the changelog but let me know if it does and I'll gladly add it!


brentvatne:
thanks!

PR #5856, opened 2019-04-29
[Typescript] `TabBarTop` is `MaterialTopTabBar` now

IsaacClem:
The export of type `TabBarTop` does not match the actual export of `MaterialTopTabBar`.
This fixes the issue where <TabBarTop {...props} /> would fail as it wasn't being exported.

(It is basically #5800 for typescript)


brentvatne:
thanks!

PR #5872, opened 2019-05-02
remove old exports and add export for SceneView and SwitchRouter

jarvisluong:
## Motivation

I removed some old export such as `StackNavigator`, `TabNavigator`, `SwitchNavigator`. Also, this typedef updates will help me to fulfill this issue: https://github.com/react-navigation/animated-switch/issues/4 since I will need the typedefs for `SceneView` and `SwitchRouter`

## Test plan

Tried to export those newly added in my own personal project to see if it is compatible

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -1460,7 +1460,7 @@ declare module 'react-navigation' {
   export interface SceneViewProps {
     component: React.ComponentType;
     screenProps: ScreenProps;
-    navigation: NavigationScreenProp;
+    navigation: NavigationScreenProp<NavigationRoute>;
   }
 
   export class SceneView extends React.Component {}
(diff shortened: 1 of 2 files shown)


typescript/react-navigation.d.ts
```diff
@@ -1467,4 +1452,16 @@ declare module 'react-navigation' {
...
   export const DrawerGestureContext: React.Context<React.Ref<PanGestureHandler>>;
+  
+  /**
+   * SceneView
+   */
+
+  export interface SceneViewProps {
+    component: React.ComponentType;
+    screenProps: ScreenProps;
+    navigation: NavigationScreenProp;
```
brentvatne: looks like this is causing an error on CI: https://circleci.com/gh/react-navigation/react-navigation/4289#tests/containers/1

jarvisluong: It should be resolved by now

PR #5893, opened 2019-05-08
Add type on BottomTabNavigatorConfig

carlosroberto555:
Add missed type

## Motivation

I did added this type on DefinitelyTyped, but this module override types from there.


brentvatne:
thanks! i merged the original pr. much appreciated nonetheless

PR #5942, opened 2019-05-28
fix: add reset to navigation screen prop TypeScript def

danielholmes:
## Motivation

Stop TypeScript compiler errors for usage of `reset` which is described at:

<https://reactnavigation.org/docs/en/navigation-prop.html#reset>

## Test plan

I've been using this in a real world project and it stopped the tsc errors when adding this change.


satya164:
Thank you for the PR

PR #5961, opened 2019-06-02
Add generics to NavigationRouteConfigMap

nartc:
## Motivation

`navigationOptions` in `routeConfigMap` does not have intellisense. For example:

```javascript
const mainTabNavigator = createBottomTabNavigator(
   {
      Home: {
          screen: HomeScreen,
          navigationOptions: <-- this does not have any intellisense as to what's available to be modify
      }
   }
);
```
## Changelog
- Improve `navigationOptions` typings under `NavigationRouteConfigMap` to provide proper intellisense for `navigationOptions` for each Navigator type

Add an entry under the "Unreleased" heading in [CHANGELOG.md](https://github.com/react-navigation/react-navigation/blob/master/CHANGELOG.md#unreleased) which explains your change.


yarn.lock
```diff
@@ -1,10203 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
osdnk: What happened here?

satya164:
Closing since this should be fixed in 4.x.

PR #5993, opened 2019-06-11
Integrated deprecated createTabNavigator in react-navigation@3.x

dev-0x1:
## Motivation
BottomTabNavigator has crash issues in order to avoid inconvenience this provides an alternate way of using tabNavigator.


satya164:
You should use that package directly, or upgrade to the actively maintained `react-navigation-tabs` package.

satya164:
Closing since we don't want this.

Issue #6002, opened 2019-06-14
getActiveChildNavigationOptions && flow


Preeternal:
I got the flow error:
"Cannot import `getActiveChildNavigationOptions` because there is no `getActiveChildNavigationOptions` export in `react-navigation`.Flow(InferError)"

Ashoat:
The [libdef](https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js) doesn't currently cover `getActiveChildNavigationOptions`. Could you make a PR adding it? Here are the [docs](https://reactnavigation.org/docs/en/navigation-options-resolution.html#getactivechildnavigationoptions) and [an example use](https://github.com/react-navigation/react-navigation/blob/eac5feb64e2170f370a815f10af77b471e146648/examples/NavigationPlayground/app/StacksOverTabs.tsx#L123-L134).

Preeternal:
> cover `getActiveChildNavigationOptions`

I tried to do it. But the issue is still present. Could you please review the PR?

PR #6018, opened 2019-06-19
flow cover for getActiveChildNavigationOptions

Preeternal:
## Motivation

Try to solve the issue - https://github.com/react-navigation/react-navigation/issues/6002#issuecomment-502908696

### DIFF
--- a/flow/react-navigation.js
+++ b/flow/react-navigation.js
@@ -585,11 +585,6 @@ declare module 'react-navigation' {
     remove: () => void,
   };
 
-  declare export function getActiveChildNavigationOptions<S>(
-    navigation: NavigationProp<S>,
-    screenProps?: {}
-  ): NavigationParams;
-
   declare export type NavigationScreenProp<+S> = {
     +state: S,
     dispatch: NavigationDispatch,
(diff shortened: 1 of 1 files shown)


Ashoat:
Check the example usage in #6002 - you should be using `NavigationScreenProp`, not `NavigationProp`

Preeternal:
Thanks for review. Seems it works as expected. Could you please check it too?

Ashoat:
The sounds wrong, `NavigationParams` is not the same as "navigation options"

Preeternal:
typescript defs from the RNN sources:

`  // @todo - any..
  export function getActiveChildNavigationOptions<S>(
    navigation: NavigationProp<S>,
    screenProps?: ScreenProps
  ): NavigationParams;`

Ashoat:
Awesome, thanks! Next step is to get these changes onto `flow-typed`. You'll need to add your new code to the [libdef for `react-navigation`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-navigation_v3.x.x/flow_v0.92.x-/react-navigation_v3.x.x.js) as well as [`@react-navigation/core`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/%40react-navigation/core_v3.x.x/flow_v0.92.x-/core_v3.x.x.js).

Preeternal:
Thanks. Could you please take a look https://github.com/flow-typed/flow-typed/pull/3448

PR #6019, opened 2019-06-19
fix: add missing isFirstRouteInParent type in typescript and flow

Shchepotin:
## Motivation

Using method isFirstRouteInParent in typescript or flow.


satya164:
Thank you for the PR!

PR #6034, opened 2019-06-24
fix: SceneView type to set the PropTypes

giuseppelt:
Updated SceneView Props Type.
ScreenProps need to be optional (?) as in createNavigator function typing.


satya164:
I think it doesn't need to be optional. Closing since the PR is out of date with master.

PR #6159, opened 2019-08-07
Fix DrawerNavigatorConfig.contentOptions TypeScript definition

mbfisher:
## Motivation

The definition of the `DrawerNavigatorConfig` interface does not appear to match the documentation [here](https://reactnavigation.org/docs/en/drawer-navigator.html#contentoptions-for-draweritems). Specifically, the `contentOptions` property contains a subset of the documented options.

My specific instance of this bug is adding `itemsContainerStyle` to `contentOptions` e.g.

```ts

createDrawerNavigator({
  {...},
  {
    contentOptions: {
      itemsContainerStyle: {...}
    }
})
```

## Implementation

I've made a small refactor to the structure of the interfaces related to `DrawerNavigatorConfig`; obviously I need a maintainer to help decide if it's the best approach 🙂

## Test plan

I haven't provided a repro of this issue because it's a bug with the TypeScript definitions, so should hopefully be demonstrable without one.

I've tested my updated config in my project and it fixes the TypeScript build error I was getting.

I've run `yarn test` in this repository and all is well, though I don't think there is anything to cover this change.


satya164:

Drawer navigator has moved to https://github.com/react-navigation/drawer and is now written in TypeScript. If you think this is still an issue, please send a PR there.

Issue #6047, opened 2019-07-01
Missing navigationOptions in SwitchNavigatorConfig typing


sinxwal:
### Current Behavior
SwitchNavigatorConfig interface is missing navigationOptions and defaultNavigationOptions options, however [docs](https://reactnavigation.org/docs/en/switch-navigator.html#switchnavigatorconfig) page says they are allowed here.

So when I use them, I get typings error:
Argument of type '{ navigationOptions: { tabBarLabel: any; }; }' is not assignable to parameter of type 'SwitchNavigatorConfig'.
  Object literal may only specify known properties, and 'navigationOptions' does not exist in type 'SwitchNavigatorConfig'.ts(2345)

### Expected Behavior
No typings error.

### How to reproduce
Code example:
`const AppStack = createSwitchNavigator(
  {},
  {
    navigationOptions: {
      title: 'My awesome title'
    }
  }
);`
### Your Environment

| software         | version
| ---------------- | -------
| react-navigation | 3.11.0
| react-native     | 0.59.9
| node             | 8.11.2
| yarn      | 1.10.1

alexmcode:
Hi. I would like to help with this issue. Is anyone working on it?

sinxwal:
@alexmcode The issue is really simple, so you can create pr if you have time.

alexmcode:
Sure. I'll do that

alexmcode:
@sinxwal the solution I found is to just add the two properties to the SwitchNavigatorConfig interface. Is this the right approach?

satya164:
Should be fixed now.

PR #6160, opened 2019-08-07
Add navigationOptions and defaultNavigationOptions for SwitchNavigatorConfig

alexmcode:
This PR aims to fix issue #6047 by adding navigationOptions and defaultNavigationOptions to SwitchNavigatorConfig interface. I don't know if this is the right solution of if the solution I found is applied correctly.

My first question is:
1. Do I need to follow the example from [StackNavigatorConfig](https://github.com/react-navigation/react-navigation/blob/master/typescript/react-navigation.d.ts#L963) and redo the whole SwitchNavigatorConfig interface object or the approach I have is sufficient?

2. The solution from this PR adds the two properties in to the interface object with type `object`. Is this correct.

Please advice on further directions. Thanks

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -973,6 +973,8 @@ declare module 'react-navigation' {
 
   export interface SwitchNavigatorConfig {
     initialRouteName: string;
+    navigationOptions?: object;
+    defaultNavigationOptions?: object;
     resetOnBlur?: boolean;
     paths?: NavigationPathsConfig;
     backBehavior?: 'none' | 'initialRoute';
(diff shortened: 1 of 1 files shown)


wilhelmeek:
Hey @alexmcode, thanks for taking the time to do this. `object` is very loosely defined and won't provide as much value as a more concretely defined type.

You might be able to get away with extending `NavigationSwitchRouterConfig` similar to how it's done with `StackNavigatorConfig`.

alexmcode:
Hi @wilhelmeek thanks for you response and feedback. I did the changes according to your advice. Please let me know if it's all good. Thanks

satya164:

The type definitions have changed a bit. If you think this is still an issue, please send an updated PR.

alexmcode:
@satya164 from what I see the problem described in #6047 still exists (on master branch). @sinxwal can you please confirm this?

PR #6164, opened 2019-08-07
Typescript: Mark "params" in NavigationState as possibly undefined

maximeaubaret:
## Motivation
Unless I'm declaring the type of my Props wrong (I'm still new to Typescript), this change is required to prevent issues where no errors are thrown when accessing any properties on the "params" field in the NavigationState.

```ts
function MyScene(props: {
  navigation: NavigationScreenProp<NavigationState, { userId: number }>
}) {

	// Should warn about params being possibly undefined.
	const { userId } = props.navigation.state.params;

	// ...
}
```


brentvatne:
this is reasonable, thanks

PR #6186, opened 2019-08-14
Add generic to navigation screen config

extrememito:
## Motivation

To be able to have strongly typed NavigationParams inside NavigationScreenConfig function
while dynamically setting NavigationScreenOptions based on NavigationParams

## Test plan

image

Update:
image

## Changelog

Add generic for NavigationScreenConfig to be able to pass custom NavigationParams


satya164:

The type definitions have changed a bit. If you think this is still an issue, please send an updated PR.

PR #6444, opened 2019-11-08
feat: allow navigation params to be typed in navigate method

alburkerk:
## Motivation

We can type navigation params in component by writing something like this in component props declaration : 

``` ts
export interface SomeNavigationParams {
  foo: string;
  bar: number;
}

export interface Props {
  navigation: NavigationScreenProp<any, SomeNavigationParams>;
}
``` 

But when we are navigating to this screen, we only call navigate method without any type checking.

``` ts
// Next line doesn't trigger ts error
this.props.navigation.navigate('MyRouteName', { foo: "foo", bar: "bar"})
``` 

By making navigate a generic method, we can type those params correctly

``` ts
import { SomeNavigationParams } from '@Screens/....../SomeScreen.types.ts'

......
// Next line triggers an error
this.props.navigation.navigate<SomeNavigationParams>('MyRouteName', { foo: "foo", bar: "bar"});
......
``` 

## Test plan

No .js code so it's can't be breaking anything. Plus, can't also break typing as changes are seamless for react-navigation API.

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -564,22 +564,23 @@ declare module 'react-navigation' {
     dispatch: NavigationDispatch;
     goBack: (routeKey?: string | null) => boolean;
     dismiss: () => boolean;
-    navigate(options: {
+    navigate<T extends NavigationParams>(options: {
       routeName:
         | string
         | {
             routeName: string;
-            params?: NavigationParams;
+            params?: T;
             action?: NavigationNavigateAction;
             key?: string;
           };
-      params?: NavigationParams;
+      params?: T;
       action?: NavigationAction;
       key?: string;
     }): boolean;
-    navigate(
+
+    navigate<T extends NavigationParams>(
       routeNameOrOptions: string,
-      params?: NavigationParams,
+      params?: T,
       action?: NavigationAction
     ): boolean;
     openDrawer: () => any;


jtschoonhoven:
Alternative implementation here (I didn't see this until after I'd submitted my own): https://github.com/react-navigation/react-navigation/pull/6515

avlonder:
Hi guys,
in react-navigation v4.4.3, I needed to apply the following patch to get this working:
```
diff --git a/node_modules/react-navigation/typescript/react-navigation.d.ts b/node_modules/react-navigation/typescript/react-navigation.d.ts
index 12bff27..174619a 100644
--- a/node_modules/react-navigation/typescript/react-navigation.d.ts
+++ b/node_modules/react-navigation/typescript/react-navigation.d.ts
@@ -569,7 +569,7 @@ export interface NavigationScreenProp<S, P = NavigationParams> {
   dispatch: NavigationDispatch;
   goBack: (routeKey?: string | null) => boolean;
   dismiss: () => boolean;
-  navigate<T extends NavigationParams>(options: {
+  navigate<T extends P>(options: {
     routeName:
       | string
       | {
@@ -583,7 +583,7 @@ export interface NavigationScreenProp<S, P = NavigationParams> {
     key?: string;
   }): boolean;
 
-  navigate<T extends NavigationParams>(
+  navigate<T extends P>(
     routeNameOrOptions: string,
     params?: T,
     action?: NavigationAction
```

PR #6486, opened 2019-12-04

mtt87:
## Motivation

This code is actually throwing a type error because of a wrong type declaration with the action expecting to be only a `NavigationNavigateAction` while in reality any action can be dispatched not only the `navigate`.
```jsx
      StackActions.replace({
        routeName: 'MyRoute1',
        // this action is throwing a type error
        action: StackActions.push({
          routeName: 'MyRoute2',
        }),
      }),
```

## Code formatting

It's just a simple type fix, I edited it in the GitHub UI, I didn't run any test locally

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -291,7 +291,7 @@ declare module 'react-navigation' {
     newKey?: string;
     routeName: string;
     params?: NavigationParams;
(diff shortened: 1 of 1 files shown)


liangjianyu303:
easy!!! I use "reset" simulate "replace" to solve

satya164:
Thanks for the PR

PR #6503, opened 2019-12-11
Add NavigationContainerComponent state types

soh335:

## Motivation

Type `createAppContainer` return is `NavigationContainer` and [get via ref in render](https://reactnavigation.org/docs/en/navigating-without-navigation-prop.html) is `NavigationContainerComponent`.

Current type definition, `NavigationContainer` has state but `NavigationContainerComponent` dont have. So cant access like `NavigationService.state.nav...` .

## Test plan

The code must pass tests.

## Code formatting


baudev:
This definition is still missing into the `3.13.0` version.

avlonder:
> Current type definition, `NavigationContainer` has state but `NavigationContainerComponent` dont have. So cant access like `NavigationService.state.nav...` .

I'm using react-navigation 4.4.3. Does somebody have a patch for this?

PR #6525, opened 2019-12-20

taoqf:

## Motivation

I find method `replace` is missing in types, but this can be used in js.

## Test plan

won't need to, I just past the types from v3.11.1

## Code formatting

I run `yarn format` with error: `Command "format" not found`

### DIFF
--- a/typescript/react-navigation.d.ts
+++ b/typescript/react-navigation.d.ts
@@ -596,6 +596,19 @@ declare module 'react-navigation' {
       eventName: string,
       callback: NavigationEventCallback
     ) => NavigationEventSubscription;
+    push: (
+      routeName: string,
+      params?: NavigationParams,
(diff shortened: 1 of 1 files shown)


satya164:
Replace comes from stack navigator. Please refer to the docs on TypeScript https://reactnavigation.org/docs/en/typescript.html

strdr4605:
@satya164, another question is about drawer functions. As they are in https://github.com/react-navigation/drawer/blob/master/src/types.tsx#L28
maybe they should also be removed from https://github.com/react-navigation/react-navigation/blob/ace2cd0c6c6faa8957ff3fe633d843ee7524db4a/typescript/react-navigation.d.ts#L585

taoqf:
How can I reopen this? @satya164

PR #6557, opened 2020-01-09

taoqf:

## Motivation

I find method `replace` is missing in types, but this can be used in js.

## Test plan

won't need to, I just past the types from v3.11.1

## Code formatting

I run `yarn format` with error: `Command "format" not found`

see https://github.com/react-navigation/react-navigation/pull/6525


satya164:
Thanks for the PR, but as I replied in the previous PR, `replace` comes from stack navigator (`NavigationStackProp`). Please refer to the docs on TypeScript https://reactnavigation.org/docs/en/typescript.html

PR #6794, opened 2020-02-08
feat(history) sync browser history with navigation state

hosseinmd:
I have implemented few functions for web history support and they work well in react-native-web.


satya164:
Thank you for the PR. We already have basic history integration here: https://github.com/react-navigation/react-navigation/blob/master/packages/native/src/useLinking.native.tsx

PR #6796, opened 2020-02-08
feat: useScrollToTop accepts an optional offset param for scrollables where the "top" is not at 0

christianjuth:
I have a FlatList component in my app that extends the full height of the device to achieve the iOS "frosted glass" effect. I had to do a bunch of tricks to get all the content to align correctly. One of those tricks was passing contentOffset to the FlatList to have it initialize to where I want the top of my list to be. Currently, there is no way to pass an offset value to useScrollToTop hook in React Navigation. I added a second parameter (after ref) that accepts an offset.

christianjuth/react-navigation-1.git#react-navigation-native-v5.0.0-gitpkg

```tsx
let contentInsets = {
  top: TopTabBar.useHeight({ safe: true }),
  bottom: Footer.useHeight({ safe: true })
};

let scrollIndicatorInsets = {
  top: TopTabBar.useHeight({ safe: false }),
  bottom: Footer.useHeight({ safe: false })
};

// tap tab to scroll to top
let ref = useRef();
useScrollToTop(ref, -contentInsets.top);

return(
  <FlatList
    contentContainerStyle={{
      paddingTop: 6,
      paddingBottom: 6
    }}
    contentInset={contentInsets}
    contentOffset={{
      y: -contentInsets.top,
      x: 0
    }}
    automaticallyAdjustContentInsets={false}
    scrollIndicatorInsets={scrollIndicatorInsets}
    contentInsetAdjustmentBehavior="never"
  />
);
```

Usage Example

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useNavigation, useRoute, EventArg } from '@react-navigation/core';
 
-type ScrollOptions = { y?: number; animated?: boolean };
+type ScrollOptions = { offset?: number; animated?: boolean };
 
 type ScrollableView =
   | { scrollToTop(): void }
@@ -43,7 +43,8 @@ function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
 }
 
 export default function useScrollToTop(
-  ref: React.RefObject<ScrollableWrapper>
+  ref: React.RefObject<ScrollableWrapper>,
+  offset: number = 0
 ) {
   const navigation = useNavigation();
   const route = useRoute();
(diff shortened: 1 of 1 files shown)


packages/native/src/useScrollToTop.tsx
```diff
@@ -82,20 +83,23 @@ export default function useScrollToTop(
           const scrollable = getScrollableNode(ref);
 
           if (isFocused && isFirst && scrollable && !e.defaultPrevented) {
-            if ('scrollToTop' in scrollable) {
```
satya164: I'm not sure about this, since the hook is called `useScrollToTop`, it should prioritize the `scrollToTop` method if available.

packages/native/src/useScrollToTop.tsx
```diff
@@ -82,20 +83,23 @@ export default function useScrollToTop(
...
-              scrollable.scrollToTop();
-            } else if ('scrollTo' in scrollable) {
-              scrollable.scrollTo({ y: 0, animated: true });
+            if ('scrollTo' in scrollable) {
+              scrollable.scrollTo({ offset, animated: true });
```
satya164: `scrollTo` doesn't take `offset`, it takes `y`

packages/native/src/useScrollToTop.tsx
```diff
@@ -82,20 +83,23 @@ export default function useScrollToTop(
...
             } else if ('scrollToOffset' in scrollable) {
-              scrollable.scrollToOffset({ y: 0, animated: true });
+              scrollable.scrollToOffset({ offset, animated: true });
             } else if ('scrollResponderScrollTo' in scrollable) {
-              scrollable.scrollResponderScrollTo({ y: 0, animated: true });
+              scrollable.scrollResponderScrollTo({
+                offset,
```
satya164: `scrollResponderScrollTo` takes `y`

packages/native/src/useScrollToTop.tsx
```diff
@@ -43,7 +43,8 @@ function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
 }
 
 export default function useScrollToTop(
-  ref: React.RefObject<ScrollableWrapper>
+  ref: React.RefObject<ScrollableWrapper>,
+  offset: number = 0
```
satya164: It should be in an options object for future flexibility

christianjuth:
The issue with using `paddingTop: 6 + headeheight` is it doesn't position the activity indicator correctly. Using content inset does move the activity indicator. I will try and implement supporting a callback function as you suggested.

Freddy03h:
Hi! I've got the same problem than you @christianjuth and I tried a different solution: #6872

christianjuth:
@Freddy03h I decided to go with @satya164's solution instead of my own

```javascript
const scrollRef = React.useRef();

useScrollToTop(React.useRef({
  scrollToTop: () => ref.current?.scrollToOffset({ offset: -contentInsets.top }),
}));
```

PR #6815, opened 2020-02-10
feat: add headerLeft prop to native-stack

dsznajder:
I've added `headerLeft` prop for `native-stack` package. However, it will not work correctly on Android because of [this code](https://github.com/kmagiera/react-native-screens/blob/fd7fcfbc5bd67875bbbf29d270a8579894fd6e1d/android/src/main/java/com/swmansion/rnscreens/ScreenStackHeaderConfig.java#L216) inside `react-native-screens`, but for iOS it works correctly as far as I've tested, but maybe I'm missing something important here.

EDIT: 

There is possibility to actually show title "properly" sticked to left side with something like this:
```js
      {headerLeft !== undefined ? (
        <ScreenStackHeaderLeftView>
          {headerLeft()}
          <Text style={...}>{
            headerTitle !== undefined
              ? headerTitle
              : title !== undefined
                ? title
                : route.name
          }</Text>
        </ScreenStackHeaderLeftView>
      ) : null}
```

I'm not experienced enough to try do something inside java but, it seems like some workaround. Also, there should be some condition for `ScreenStackHeaderCenterView` if someone would like to render centred title. I can add this, but I would like to know what you guys think about this approach.

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -107,6 +107,12 @@ export type NativeStackNavigationOptions = {
    */
   headerLargeTitle?: boolean;
   /**
+   * Function which returns a React Element to display on the left side of the header.
+   *
+   * @platform ios
+   */
+  headerLeft?: () => React.ReactNode;
+  /**
    * Function which returns a React Element to display on the right side of the header.
    */
   headerRight?: () => React.ReactNode;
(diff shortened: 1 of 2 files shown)


dsznajder:
@satya164 This can be actually handled same on iOS and Android but with some additional code inside `HeaderConfig`. For now it would be some workaround. 
Also, I've tried to do something with this in `ScreenStackHeaderConfig` in java but it seems like `Toolbar` limitation, but still not sure about that. 

@kmagiera please take a look in free time.

satya164:
Hey, native-stack has been moved to [software-mansion/react-native-screens](https://github.com/software-mansion/react-native-screens). Please send the pull request there. Extremely sorry for the trouble.

PR #6825, opened 2020-02-11
fix: static container memo check

abhiin1947:
### What were the changes?
- Memo check compared elements of `prevProps` to `nextProps` but failed to take new props into account.
- Fixed the logic and added a new test.

### Repro steps
- Memoize a component and add a new prop. Memo returns previous result instead of re-rendering the component.

PS: I couldn't find a PR template, please let me know if I missed it. I'll update my description as per the template.

### DIFF
--- a/packages/core/src/StaticContainer.tsx
+++ b/packages/core/src/StaticContainer.tsx
@@ -8,12 +8,19 @@ function StaticContainer(props: any) {
 }
 
 export default React.memo(StaticContainer, (prevProps: any, nextProps: any) => {
-  for (const prop in prevProps) {
-    if (prop === 'children') {
+  const prevPropKeys = Object.keys(prevProps);
+  const nextPropKeys = Object.keys(nextProps);
+
+  if (prevPropKeys.length !== nextPropKeys.length) {
+    return false;
+  }
+
+  for (const key of prevPropKeys) {
+    if (key === 'children') {
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/src/StaticContainer.tsx
```diff
@@ -8,6 +8,10 @@ function StaticContainer(props: any) {
 }
 
 export default React.memo(StaticContainer, (prevProps: any, nextProps: any) => {
+  if (Object.keys(prevProps).length !== Object.keys(nextProps).length) {
+    return false;
+  }
+
```
rkstar: if we're going through the trouble of creating arrays here, we could ditch the code below in favour of a reduce method since we're looping anyway:
```
const prevKeys = Object.keys(prevProps);
const nextKeys = Object.keys(nextProps);
return prevKeys.length === nextKeys.length
  ? prevKeys.reduce((returnValue, key) => returnValue ? (prevProps[key] === nextProps[key]) : false, true)
  : false;
```

satya164:
Thanks for the PR. I spilled water on my MacBook so I'll merge when I have access to a working laptop.

packages/core/src/StaticContainer.tsx
abhiin1947: I'm not sure reduce is a good idea this case as it would extend the iterations beyond the required number of them. But I did reuse the keys extracted from props instead of getting them again.

PR #6859, opened 2020-02-14
feat: add split mode

jaulz:
When you use a transparent background for cards you sometimes expect the items in the background to be touchable as well. This fix keeps the current behaviour but if you want you can make the background touchable by passing `pointerEvents="none"` to the container. I added an example in "Transparent Stack" where you can press on "Show Sheet" but you are still able to press the components in the background .


jaulz:
@satya164 what do you think in general? should this be the default behavior or shall I add a flag?

jaulz:
@satya164 I have something like a master detail layout where the master screen is on the left and the detail view is on the right hand side:
image

The list should still be interactable while the details screen is visible on tablets. For some other reasons I cannot simply place the list on top of the details screen.

Could we for example add a property `cardContainerPointerEvents` (which is `auto` by default)? Let me know what you think and I'm happy to make the adjustments.

satya164:
@jaulz I think it'd be better to support such layout in core. Would you like to send a PR for that? Then we won't need hacks like this.

Also there's this if you just want a static content on the side: https://github.com/react-navigation/react-navigation/pull/6948

jaulz:
@satya164 you mean a new package like `@react-navigation/master-details`? Though, even then it would be wise to use the stack navigator including the changes of this PR. Otherwise we would copy large parts of the stack navigator.

satya164:
not a different package, but a new `mode="split"` or something on stack navigator which will apply the correct styling.

jaulz:
Sure, let me check it and adjust the PR 👍

jaulz:
@satya164 I just pushed the new changes. Could you please have a look and tell me if this goes into the right direction you had in mind? You can test the changes in the example app where I added a new menu item "Split Stack". When you test it in the browser and resize the app you should be able to see the behaviour. Unfortunately, I was not yet able to run the example app on iOS as I am facing some issues with the project but I will try to set it up in the meantime.

satya164:
I'll check soon.

satya164:
Ok, I just checked, and I think this layout is already possible using drawer navigator and `drawerType="permanent"`. Do you think you can use drawer instead of stack here for your use case?

jaulz:
Hm, how would you then render the content on top of the Drawer (see small screen example)? Also, we would miss all the nice other things from `Stack` (e.g., cardStyleInterpolator etc.), don't we? Correct me if I'm wrong as I probably don't know all the functionalities of the library yet 😉

satya164:
> how would you then render the content on top of the Drawer

That looks like `drawerType="back"` with drawer width set to screen width. The only thing that seems to be missing is to keep the drawer open by default instead of closed. We can add a prop for that.

> Also, we would miss all the nice other things from Stack (e.g., cardStyleInterpolator etc.)

I don't see any animations when switching the screens from sidebar. What are the animations you need in this setup?

jaulz:
The example is of course just a simplified version of my use case and I'm very grateful for your inputs! I just tried to implement it with a `Drawer` but I hit a point that I can't solve with this approach I think. In my case the details component must be a direct sibling of the master component due to some layout constraints (i.e. zIndex) so I would really appreciate if we could solve it somehow via `Stack`. There could also be other use cases when the `Drawer` would not solve everything e.g. when you want the master view to be horizontal above the details view.

satya164:
> In my case the details component must be a direct sibling of the master component due to some layout constraints (i.e. zIndex) so I would really appreciate if we could solve it somehow via Stack

That sounds like a very specific case. Whether they are direct siblings or not is an implementation detail and not something you should rely on whether it's drawer or stack. There is no guarantee that it'll continue to be the case. Can you give an example of why you'd need that?

> There could also be other use cases when the Drawer would not solve everything e.g. when you want the master view to be horizontal above the details view. 

I want avoid adding features for potential use cases unless I am confident that users will need it. Personally I have never seen such a UI. Anyway, if there's demand for such a UI, it's always possible to build custom navigators to accommodate special cases.

I'm closing this PR now because I just published a new version of drawer with a new `openByDefault` option as well as some bug fixes which make it possible to use it as a master-detail view. You can see an example here: https://github.com/react-navigation/react-navigation/blob/master/example/src/Screens/MasterDetail.tsx

ezgif-5-6513b1a6ce1f

If you face any issues with it, feel free to open new issues.

exentrich:
The last piece I miss in Drawer is to be able to use standard header bar in `drawerContent`.
In example project there a custom header bar from `react-native-paper`...

PR #6872, opened 2020-02-17
fix: useScrollToTop and offset

Freddy03h:
Hello, two things on this PR:

- `scrollToOffset` should use `offset` and not `y` in options https://facebook.github.io/react-native/docs/virtualizedlist#scrolltooffset
- get the `contentOffset.y` from the scrollable props instead of using `0`. The definition of `contentOffset` is 

> Used to manually set the starting scroll offset

So I think it's ok to use it to prevent unwanted behavior if we use `contentOffset`/`contentInset` on iOS to achieve some effects, and avoiding the need to give the offset as a param for `useScrollToTop `

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -2,11 +2,12 @@ import * as React from 'react';
 import { useNavigation, useRoute, EventArg } from '@react-navigation/core';
 
 type ScrollOptions = { y?: number; animated?: boolean };
+type ScrollOffsetOptions = { offset?: number; animated?: boolean };
 
 type ScrollableView =
   | { scrollToTop(): void }
   | { scrollTo(options: ScrollOptions): void }
-  | { scrollToOffset(options: ScrollOptions): void }
+  | { scrollToOffset(options: ScrollOffsetOptions): void }
   | { scrollResponderScrollTo(options: ScrollOptions): void };
 
 type ScrollableWrapper =
@@ -87,7 +88,7 @@ export default function useScrollToTop(
             } else if ('scrollTo' in scrollable) {
               scrollable.scrollTo({ y: 0, animated: true });
             } else if ('scrollToOffset' in scrollable) {
-              scrollable.scrollToOffset({ y: 0, animated: true });
+              scrollable.scrollToOffset({ offset: 0, animated: true });
             } else if ('scrollResponderScrollTo' in scrollable) {
               scrollable.scrollResponderScrollTo({ y: 0, animated: true });
             }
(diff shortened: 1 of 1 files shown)


packages/native/src/useScrollToTop.tsx
```diff
@@ -82,14 +83,16 @@ export default function useScrollToTop(
           const scrollable = getScrollableNode(ref);
 
           if (isFocused && isFirst && scrollable && !e.defaultPrevented) {
+            const contentOffsetY = scrollable.props.contentOffset?.y || 0;
```
satya164: Thanks for the PR! I think this isn't the right solution since this relies an implementation detail. Since the scrollables are usually class components, it's possible to access their props via `ref.props` (since `ref` is the instance of the class), but say if someone refactprs them to a function component in future and uses `useImperativeHandle` to set the ref instead, this code will break. imo the approach I suggested in #6796 will be the most flexible and future proof.

Freddy03h: Thanks for you fast answer and review! You're right it's not ok to rely on `ref.props`.
I tried this approach because your is more flexible indeed, but need to write much implementation detail already written in the hook implementation (like the `getScrollableNode` function) and it also need to knowing the implementation that `scrollToTop` is prioritized and will not change in the future.

For example, for a FlatList wrapped in Animated:
```js
useScrollToTop(useRef({
  scrollToTop: () => scrollRef.current?.getNode().scrollToOffset({ offset: contentOffset.y }),
}));
```
(and i'm not a fan of the polymorphic `ref | () => void` of the simpler example, but it's subjective).

I think the only thing we want to change in the `useScrollToTop` hook is the hardcoded `0` that assume doing a `scrollToTop` and so that why I understand @christianjuth proposal to give it as a param.

But more I think about it, more I think it's not the job of react-navigation and a `scrollToTop`/`scrollToStart` should exist for `<ScrollView>` directly in react-native. The function will take into consideration the `contentOffset` prop or, I don't know the implementation detail, called the native fonction 'scrollToTop' equivalent. The 'scrollToTop' work out of the box with `contentOffset` and `contentInset` if we touch the StatusBar so I think something like this already exist in the native part or in the react-native implementation.

Maybe you're already know something about it on the react-native part ? That's why `scrollToTop` is prioritized ?

For now I'm closing this issue.

Freddy03h: Ok, I found a new use case where we need more flexibility so your design is better.
If you set an `initialScrollIndex` you want to use `scrollToIndex()` for `useScrollToTop()`.

Issue #6877, opened 2020-02-18
Modals can't be transparent without changing ThemeContext


Jarred-Sumner:
**Current Behavior**
Say you're on iOS and you want to show a modal with a transparent background.
 
All views have a background color set by default, so instead of the modal being transparent, it shows the system background color.

https://github.com/react-navigation/react-navigation/blob/1a28c299b5e3f0805eb6e9ea3cf5e9cc90c7a280/packages/native-stack/src/views/NativeStackView.tsx#L73-L79

Currently, to make it transparent, you'd need to set a custom `ThemeProvider` above the stack, which would override `colors.background`.

Example:

```jsx
const TransparentThemeProvider = ({ children }) => {
  const currentTheme = useTheme();
  const newTheme = React.useMemo(() => {
    const theme = cloneDeep(currentTheme);
    theme.colors.background = "transparent";
    return theme;
  }, [currentTheme]);
  return <ThemeProvider value={newTheme}>{children}</ThemeProvider>;
};
```

**Expected Behavior**

It'd be great if there was a way to ensure it is transparent when it should be without overriding the ThemeContext. For example, if `stackPresentation` is `transparentModal`, then the background should be null or `transparent`.

This functionality works on `@react-navigation/stack` via the following props:
- `cardStyle`
- `cardOverlayEnabled`
- `cardShadowEnabled`

**Your Environment**
```
├─ @react-navigation/core@5.1.2
│  ├─ @react-navigation/routers@^5.0.1
├─ @react-navigation/native-stack@5.0.3
├─ @react-navigation/native@5.0.3
│  └─ @react-navigation/core@^5.1.2
├─ @react-navigation/routers@5.0.1
├─ @react-navigation/stack@5.0.3
├─ react-navigation-animated-switch@0.4.1
├─ react-navigation-shared-element@0.5.0
```

satya164:
We can remove the background if `stackPresentation` is `transparentModal`. Do you want to send a PR?

Joaojuby:
Was having an issue with a modal stack in a partial migration from v4 to v5 and I would never find out to fix it wasn't it for this.

Also, we don't even use the ThemeContext in our app (well we do now) so a clarification in the docs or migration guide would also be useful aside from a PR with the proposed change

satya164:
[...]

PR #6888, opened 2020-02-18
chore: remove background when stackPresentation is transparentModal

kaioduarte:
## Summary
Fixes https://github.com/react-navigation/react-navigation/issues/6877

## Before / After
<div style="display: flex;">

</div>

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.tsx
+++ b/packages/native-stack/src/views/NativeStackView.tsx
@@ -73,7 +73,12 @@ export default function NativeStackView({
             <View
               style={[
                 styles.container,
-                { backgroundColor: colors.background },
+                {
+                  backgroundColor:
+                    stackPresentation !== 'transparentModal'
+                      ? colors.background
+                      : undefined,
+                },
                 contentStyle,
               ]}
             >


packages/native-stack/src/views/NativeStackView.tsx
```diff
@@ -73,7 +73,12 @@ export default function NativeStackView({
             <View
               style={[
                 styles.container,
-                { backgroundColor: colors.background },
+                {
+                  backgroundColor:
+                    stackPresentation !== 'transparentModal'
```
bang88: How about the old configuration `transparentCard:true`. Does this settings works too

kaioduarte: This option is not related with the changed package (native-stack) and this option was [removed in v4](https://github.com/react-navigation/stack/blob/48b1651e99e3feb01c3caa5e0cab87b24b90ccf0/src/utils/validateDeprecatedConfig.tsx#L16).

satya164:
Hey, thanks for the PR. We're moving the native-stack to screens repo: https://github.com/software-mansion/react-native-screens/pull/345

Would be great if you can send the PR there. Thanks and extremely sorry for the trouble.

kaioduarte:
No problem, I'll manage to send there when your PR is merged. Thanks anyway!

Issue #6909, opened 2020-02-20
Keyboard or D Pad navigation on stack navigator Focusing on previous screen Touchables


mohamedshuaau:
Sorry for my poor english. So I have a stack navigator with a Home screen page which has a mapped array of "Menu Items". They are aligned horizontally. My problem is that when I navigate to any of the page (using keyboard or dpad on android) and use my keyboard to navigate on elements of that screen, if I press down key, it focuses on the home screen menu items. But those items are in home screen and this is another screen. I know it focused because I loose focus on my current screens focused elements and If I press return, it opens the screen from which it is focused on in the main menu.

App.js:

```
<Stack.Navigator initialRouteName='Home' headerMode='none'>
                        <Stack.Screen name="Home" component={HomeScreen} />
                        <Stack.Screen name="TV" component={TVScreen} />
                        <Stack.Screen name="Movie" component={MovieScreen} />
                        <Stack.Screen name="Messages" component={MessageScreen} />
                        <Stack.Screen name="HotelInfo" component={HotelInformationScreen} />
                        <Stack.Screen name="Daily" component={DailyUpdateScreen} />
                        <Stack.Screen name="IVD" component={IVDScreen} />
                        <Stack.Screen name="Shopping" component={ShoppingScreen} />
                        <Stack.Screen name="Weather" component={WeatherScreen} />
                        <Stack.Screen name="Language" component={LanguageSelectionScreen} />
                        <Stack.Screen name="Survey" component={SurveyScreen} />
                        <Stack.Screen name="TVPlayer" component={TVPlayer} />
                        <Stack.Screen name="MoviePlayer" component={MoviePlayer} />
                        <Stack.Screen name="MovieDetail" component={MovieDetailScreen} />

                        <Stack.Screen name="Registration" component={Registration} />
</Stack.Navigator>
```

HomeScreen.js:

```

navigations = debounce((screen_name, background_image) => {
        this.props.navigation.push(screen_name, {background: background_image});
    }, 200)

    navigate(screen_name, background_image){
        this.navigations(screen_name, background_image);
    }

<View style={styles.thumbnailContainer}>
                    {
                        this.state.mainMenu.map((e, index) => {
                            return (
                                <CustomIcon
                                    key={index}
                                    source={{uri: baseURL.baseImageURL + '/main-menu/icon/' + e.icon}}
                                    imageStyle={styles.thumbnail}
                                    onPress={()=>this.navigate(e.screen_name, e.background_image)}
[...]

PR #6934, opened 2020-02-22
refactor: enable screens only for last screen

satya164:
This will avoid issues such as #6909


osdnk:
Tested and looks good 🤷‍♂

PR #6948, opened 2020-02-24
feat: add sidebar drawer type

NoemiRozpara:
moved from https://github.com/react-navigation/navigation-ex/pull/301


pincheira:
@NoemiRozpara is this compiling for you? I tried it out and it's got some double declaration error:

```
 web  Failed to compile.
./packages/drawer/src/views/DrawerView.tsx
SyntaxError: /Users/jp/Work/spare/permanent-sidebar/react-navigation/packages/drawer/src/views/DrawerView.tsx: Identifier 'handleDrawerOpen' has already been declared (158:8)

  156 |   }
  157 |
> 158 |   const handleDrawerOpen = () => {
      |         ^
  159 |     if (bigScreenSidebar) {
  160 |       return;
  161 |     }
```

Thank you 🙏

NoemiRozpara:
Can you re-run the checks?

osdnk:
rerunning

packages/drawer/src/views/Drawer.tsx
```diff
@@ -92,6 +92,7 @@ type Props = {
   renderDrawerContent: Renderer;
   renderSceneContent: Renderer;
   gestureHandlerProps?: React.ComponentProps<typeof PanGestureHandler>;
+  isBigScreen?: boolean;
```
satya164: We don't need a `isBigScreen` prop. We can pass something like `drawerType={screenWidth > 800 ? 'sidebar' : 'front'}`

packages/drawer/src/views/Drawer.tsx
```diff
@@ -601,35 +610,59 @@ export default class DrawerView extends React.PureComponent<Props> {
               importantForAccessibility={open ? 'no-hide-descendants' : 'auto'}
               style={styles.content}
             >
-              {renderSceneContent({ progress: this.progress })}
+              {renderSceneContent({
+                // make 'sidebar' initially visible on the big screen to avoid annoying animation
+                progress: this.bigScreenSidebar
+                  ? new Animated.Value(1)
```
satya164: nit: lets move this to top of the scope to avoid creating this animated value on render

packages/drawer/src/views/Drawer.tsx
```diff
@@ -195,6 +197,9 @@ export default class DrawerView extends React.PureComponent<Props> {
     }
   };
 
+  private bigScreenSidebar =
```
satya164: we need to move it to render so that it respects prop changes

packages/drawer/src/views/Drawer.tsx
```diff
@@ -214,11 +219,12 @@ export default class DrawerView extends React.PureComponent<Props> {
...
   private gestureX = new Value<number>(0);
   private offsetX = new Value<number>(0);
-  private position = new Value<number>(0);
+  private position = new Value<number>(this.bigScreenSidebar ? 1 : 0);
 
   private containerWidth = new Value<number>(0);
   private drawerWidth = new Value<number>(0);
-  private drawerOpacity = new Value<number>(0);
+  // make drawer initially visible on the big screen to avoid annoying animation
+  private drawerOpacity = new Value<number>(this.bigScreenSidebar ? 1 : 0);
```
satya164: I think we can just to `opacity: drawerType === 'sidebar' : 1 : this.drawerOpacity` below

packages/drawer/src/views/Drawer.tsx
```diff
@@ -214,11 +219,12 @@ export default class DrawerView extends React.PureComponent<Props> {
   private velocityX = new Value<number>(0);
   private gestureX = new Value<number>(0);
   private offsetX = new Value<number>(0);
-  private position = new Value<number>(0);
+  private position = new Value<number>(this.bigScreenSidebar ? 1 : 0);
```
satya164: Probably don't need to adjust `position`. We could probably hardcode the value in render if we need it

packages/drawer/src/views/Drawer.tsx
```diff
@@ -601,35 +610,59 @@ export default class DrawerView extends React.PureComponent<Props> {
...
           <Animated.View
             accessibilityViewIsModal={open}
             removeClippedSubviews={Platform.OS !== 'ios'}
             onLayout={this.handleDrawerLayout}
             style={[
               styles.container,
-              isRight ? { right: offset } : { left: offset },
-              {
-                transform: [{ translateX: drawerTranslateX }],
+              !this.bigScreenSidebar && {
```
satya164: Let's keep the style in a style object in `StyleSheet.create` and use it here

packages/drawer/src/views/DrawerView.tsx
```diff
@@ -88,8 +88,12 @@ export default function DrawerView({
   sceneContainerStyle,
 }: Props) {
   const [loaded, setLoaded] = React.useState([state.index]);
+
+  const { width, height } = Dimensions.get('window');
+  const isBigScreen = Math.min(width, height) >= 768;
```
satya164: I think maybe we can leave the automatic switching for now since now it's stable and this can be considered a breaking change. Let the user pass the `drawerType` explicitly

packages/drawer/src/types.tsx
```diff
@@ -27,7 +27,7 @@ export type DrawerNavigationConfig<T = DrawerContentOptions> = {
    * - `back`: The drawer is revealed behind the screen on swipe.
    * - `slide`: Both the screen and the drawer slide on swipe to reveal the drawer.
    */
-  drawerType?: 'front' | 'back' | 'slide';
+  drawerType?: 'front' | 'back' | 'slide' | 'sidebar';
```
satya164: Let's change to `'sidebar'` to `'permanent'` according to @EvanBacon's suggestion

```suggestion
  drawerType?: 'front' | 'back' | 'slide' | 'permanent';
```

packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
```diff
@@ -1,4 +1,5 @@
 import { Animated } from 'react-native';
+import { isIphoneX } from 'react-native-iphone-x-helper';
```
satya164: This change seems unrelated

satya164:
Merged in https://github.com/react-navigation/react-navigation/commit/6a5d0a035afae60d91aef78401ec8826295746fe

PR #7666, opened 2020-02-25
feat: add ability add listeners with listeners prop

### LINKED ISSUE #6756: No way to listen to `tabPress` event outside of Tab.Navigator.
**Current Behavior**

There is no way to set `tabPress` listener in a component that is not a Tab.Screen or its child.

**Expected Behavior**

There should be a way to listen to `tabPress` event globally and not only in TabNavigator's child components.

**Use case**

I want to render a FAB button that is visible in all Tab.Navigator Tabs and I want to change its appearance depending on currently active tab.

To make FAB component visible in all tabs I render it as a Tab.Navigator sibling. Since I can only set `tabPress` listener inside Tab.Navigator screen currently, I have to pass a callback to each Tab.Navigator screen to know if it was pressed. It would be handy to be able to set a `tabPress` listener outside of Tab.Navigator that would tell me which tab was just pressed.

**Your Environment**

| software                      | version |
| ----------------------------- | ------- |
| @react-navigation/native      | 5.0.0-alpha.27
| @react-navigation/bottom-tabs |5.0.0-alpha.34

satya164:
This adds ability to listen to events from the component where the navigator is defined, even if the screen is not rendered.

```js
<Tabs.Screen
  name="Chat"
  component={Chat}
  options={{ title: 'Chat' }}
  listeners={{
    tabPress: e => console.log('Tab press', e.target),
  }}
/>
```

Closes #6756

### DIFF
--- a/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx
+++ b/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx
@@ -48,6 +48,8 @@ function BottomTabNavigator({
 }
 
 export default createNavigatorFactory<
+  TabNavigationState,
   BottomTabNavigationOptions,
+  BottomTabNavigationEventMap,
   typeof BottomTabNavigator
 >(BottomTabNavigator);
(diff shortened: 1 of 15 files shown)


chakrihacker:
Awesome, thanks Satya! much needed feature that I am missing in V5

PR #7689, opened 2020-02-28
fix: export all types from bottom-tabs

emanuelmd:
Noticed that some types were not exported form `types.tsx`, I think it makes more sense to handle export control to that file instead of `index.tsx`

I also renamed the 'index' & 'types' files extension to `.ts` since they don't contain react code.


emanuelmd:
I disagree

PR #7800, opened 2020-03-12
feat: emit gesture navigation events from stack view

martintreurnicht:
Allows you to subscribe to gesture navigation events, we have a custom keyboard that we want to hide and show when gesture is being used to navigate (same as native keyboard)

### DIFF
--- a/example/src/index.tsx
+++ b/example/src/index.tsx
@@ -51,7 +51,6 @@ import MaterialBottomTabs from './Screens/MaterialBottomTabs';
 import DynamicTabs from './Screens/DynamicTabs';
 import AuthFlow from './Screens/AuthFlow';
 import CompatAPI from './Screens/CompatAPI';
-import LinkComponent from './Screens/LinkComponent';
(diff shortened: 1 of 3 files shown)


martintreurnicht:
> Hey! Thanks for the PR. These callbacks are used for convenience internally and I don't know if it makes sense to expose them as a public API.
> 
> We already expose `transitionStart`, `focus` and `blur` events which can help with this. We're missing `gestureStart`, `gestureEnd` and `gestureCancel` events. Can you update the PR to emit them instead?

Hi @satya164, I've made the changes you requested. I'm not sure what you wanted me to change exactly, i assume it was the names. Please let me know if there is anything else

martintreurnicht:
@satya164 Any news on this PR? Do you want me to change something, or are you happy with it the way that it is?

satya164:

Regarding the PR, as I requested, please add specific events such as `gestureStart`, `gestureEnd` etc which fire on gesture start, gesture end etc, not rename `pageChangeStart` to `gestureStart` because they are not the same things.

PR #7805, opened 2020-03-13
feat: add Link component as useLinkTo hook for navigating to links

satya164:
The `Link` component can be used to navigate to URLs. On web, it'll use an `a` tag for proper accessibility. On React Native, it'll use a `Text`.

Example:

```js
<Link to="/feed/hot">Go to 🔥</Link>
```

Sometimes we might want more complex styling and more control over the behaviour, or navigate to a URL programmatically. The `useLinkTo` hook can be used for that.

Example:

```js
function LinkButton({ to, ...rest }) {
  const linkTo = useLinkTo();

  return (
    <Button
      {...rest}
      href={to}
      onPress={(e) => {
        e.preventDefault();
        linkTo(to);
      }}
    />
  );
}
```

### DIFF
--- a/example/src/Screens/LinkComponent.tsx
+++ b/example/src/Screens/LinkComponent.tsx
@@ -1,0 +1,140 @@
+import * as React from 'react';
+import { View, StyleSheet, ScrollView } from 'react-native';
+import { Button } from 'react-native-paper';
+import {
+  Link,
+  RouteProp,
+  ParamListBase,
+  useLinkTo,
+} from '@react-navigation/native';
+import {
+  createStackNavigator,
+  StackNavigationProp,
+} from '@react-navigation/stack';
+import Article from '../Shared/Article';
+import Albums from '../Shared/Albums';
+
+type SimpleStackParams = {
+  Article: { author: string };
+  Album: undefined;
+};
+
+type SimpleStackNavigation = StackNavigationProp<SimpleStackParams>;
+
+const LinkButton = ({
+  to,
+  ...rest
+}: React.ComponentProps<typeof Button> & { to: string }) => {
+  const linkTo = useLinkTo();
+
+  return <Button onPress={() => linkTo(to)} {...rest} />;
+};
+
+const ArticleScreen = ({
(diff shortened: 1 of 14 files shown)


packages/native/src/useLinking.tsx
```diff
@@ -92,10 +104,10 @@ export default function useLinking(
   const numberOfIndicesAhead = React.useRef(0);
 
   React.useEffect(() => {
-    window.addEventListener('popstate', () => {
+    const onPopState = () => {
```
EvanBacon, line 107: Maybe it would make sense to split this into a new hook `usePopState` to keep the logic divided more clearly 🤔

example/src/Screens/LinkComponent.tsx
```diff
@@ -0,0 +1,140 @@
...
+  Album: undefined;
+};
+
+type SimpleStackNavigation = StackNavigationProp<SimpleStackParams>;
+
+const LinkButton = ({
+  to,
+  ...rest
+}: React.ComponentProps<typeof Button> & { to: string }) => {
+  const linkTo = useLinkTo();
```
EvanBacon, line 28: Maybe it would make sense to make this return an array like `const [linkTo, props] = useLinkTo(to)` so people can utilize the complex `onPress` method created if they want. Also returning `props` could be an elegant way to strip `href` on non-web targets.

packages/native/src/useLinking.tsx
satya164: I'll refactor after adding some tests so I don't accidentally break anything 😅

example/src/Screens/LinkComponent.tsx
satya164: I think maybe a separate hook will be good for it instead of overloading this one. I'm also thinking maybe `Link` can accept an `as` prop instead (ala styled components) which will serve the same purpose.

Lemme think a bit.

satya164:
Going to merge for now and will make more changes later. Thanks for the review.

PR #7809, opened 2020-03-14
feat: add an option to change use a custom card overlay

osdnk:
I find it sometimes useful to define overlay renderer on my own. Eg. I needed to replace the background with BlurView and with this API I find it quite easy

### DIFF
--- a/packages/stack/src/types.tsx
+++ b/packages/stack/src/types.tsx
@@ -314,9 +314,9 @@ export type StackNavigationOptions = StackHeaderOptions &
       left?: number;
     };
     /**
-     * Function that given `style` returns a React Element to display as a overlay.
+     * Function that` returns a React Element to display as a overlay.
      */
-    overlay?: (style: StyleProp<ViewStyle>) => React.ReactNode;
+    overlay?: (props: { style: StyleProp<ViewStyle> }) => React.ReactNode;
   };
 
 export type StackNavigationConfig = {
--- a/packages/stack/src/types.tsx
+++ b/packages/stack/src/types.tsx
@@ -314,9 +314,9 @@ export type StackNavigationOptions = StackHeaderOptions &
       left?: number;
     };
     /**
-     * todo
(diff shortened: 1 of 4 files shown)


packages/stack/src/types.tsx
```diff
@@ -313,6 +313,10 @@ export type StackNavigationOptions = StackHeaderOptions &
       bottom?: number;
       left?: number;
     };
+    /**
+     * Function that given `style` returns a React Element to display as a overlay.
+     */
+    overlay?: (style: StyleProp<ViewStyle>) => React.ReactNode;
```
satya164: can we change it to an object with `style` property?
```suggestion
    overlay?: (style: StyleProp<ViewStyle>) => React.ReactNode;
```

osdnk: yes, it will be more intuitive

packages/stack/src/types.tsx
```diff
@@ -313,6 +313,10 @@ export type StackNavigationOptions = StackHeaderOptions &
       bottom?: number;
       left?: number;
     };
+    /**
+     * Function that` returns a React Element to display as a overlay.
```
satya164:
```suggestion
     * Function that returns a React Element to display as a overlay.
```

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -471,12 +475,9 @@ export default class Card extends React.Component<Props> {
...
     return (
       <View pointerEvents="box-none" {...rest}>
-        {overlayEnabled && overlayStyle ? (
-          <Animated.View
-            pointerEvents="none"
-            style={[styles.overlay, overlayStyle]}
-          />
-        ) : null}
+        {overlayEnabled && overlayStyle
+          ? this.props.overlay(overlayStyle)
```
satya164: The code doesn't match the type definitions (should be `({ style: overlayStyle })`)

PR #7824, opened 2020-03-18
fix: fix closing stack using inverted gesture.

AlexeyVlasenko:
Allow you to close Stack that has inverted gesture config using correct closing bool calculation.

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -276,7 +276,7 @@ export default class Card extends React.Component<Props> {
           gestureDirection === 'vertical-inverted'
         ) {
           translation = Math.abs(translation);
-          velocity = Math.abs(translation);
+          velocity = Math.abs(velocity);
         }
 
         const closing =
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -275,8 +275,8 @@ export default class Card extends React.Component<Props> {
           gestureDirection === 'horizontal-inverted' ||
           gestureDirection === 'vertical-inverted'
         ) {
-          translation = Math.abs(translation);
-          velocity = Math.abs(velocity);
+          translation *= -1;
+          velocity *= -1;
         }
 
         const closing =
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -271,6 +271,14 @@ export default class Card extends React.Component<Props> {
           velocity = nativeEvent.velocityX;
         }
 
+        if (
+          gestureDirection === 'horizontal-inverted' ||
+          gestureDirection === 'vertical-inverted'
+        ) {
+          translation = Math.abs(translation);
+          velocity = Math.abs(translation);
+        }
+
         const closing =
           translation + velocity * gestureVelocityImpact > distance / 2
             ? velocity !== 0 || translation !== 0


AlexeyVlasenko:
> Thanks for the PR. Is `Math.abs` the correct solution or we should multiply with `-1` instead?
> 
> cc @tienph6m who removed `Math.abs` here #7700

I think both Math.abs and multiply with -1 will produce equal performance, so Math.abs look better there

satya164:
> I think both Math.abs and multiply with -1 will produce equal performance

It's not about performance. The linked PR fixed a bug by removing `Math.abs`, using it for inverted directions will likely re-introduce that bug for inverted directions.

AlexeyVlasenko:
> > I think both Math.abs and multiply with -1 will produce equal performance
> 
> It's not about performance. The linked PR fixed a bug by removing `Math.abs`, using it for inverted directions will likely re-introduce that bug for inverted directions.

No, it wouldn't, because in my fix i use Math.abs only for  INVERTED config.

satya164:
> No, it wouldn't, because in my fix i use Math.abs only for INVERTED config.

So the bug will only happen only for INVERTED config like I said.

tienph6m:
Hi, sorry I wasn't aware of the gestureDirection option when making that PR.
I think using Math.abs in this case would re-introduce the bug like @satya164 said, we should multiple the values with `-1`.

AlexeyVlasenko:
> Hi, sorry I wasn't aware of the gestureDirection option when making that PR.
> I think using Math.abs in this case would re-introduce the bug like @satya164 said, we should multiple `translation + velocity * gestureVelocityImpact` with `-1` when the gesture direction is inverted, probably don't modify `velocity` here because we still need to use the original `velocity` to perform the closing animation.

I can't reproduce bug mentioned in [#7700](https://github.com/react-navigation/react-navigation/pull/7700) using this fix, can you check it, please?

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -271,6 +271,14 @@ export default class Card extends React.Component<Props> {
           velocity = nativeEvent.velocityX;
         }
 
+        if (
+          gestureDirection === 'horizontal-inverted' ||
+          gestureDirection === 'vertical-inverted'
+        ) {
+          translation = Math.abs(translation);
+          velocity = Math.abs(translation);
```
tienph6m: I think you meant `velocity = Math.abs(velocity);`?

AlexeyVlasenko: Yes, my bad, sorry.

tienph6m:
Mar-19-2020 16-13-49
I was able to reproduce the bug pretty easily, the bug occurs when you swipe back really fast.

I think the directions of `translation` and `velocity` matter and stripping them away with `Math.abs` will cause unexpected behaviors. You can try multiplying `translation + velocity * gestureVelocityImpact` with `-1` when the gesture is inverted and see that it will behave correctly.

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -271,6 +271,14 @@ export default class Card extends React.Component<Props> {
           velocity = nativeEvent.velocityX;
         }
 
+        if (
+          gestureDirection === 'horizontal-inverted' ||
+          gestureDirection === 'vertical-inverted'
+        ) {
+          translation = Math.abs(translation);
```
satya164:
```suggestion
          translation *= -1;
```

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -271,6 +271,14 @@ export default class Card extends React.Component<Props> {
...
+          velocity = Math.abs(velocity);
```
satya164:
```suggestion
          velocity *= -1;
```

Issue #6475, opened 2019-11-27
Drawer is not closing when clicking outside of drawer


kranthijulakantiwork:
I have a navigation drawer which can be opened by clicking a button and closed by swiping. But I can't close the drawer by clicking outside of it. I am opening drawer navigator using this.props.navigation.openDrawer().

Here's my navigator: 

```
const navigationOptions = {
  defaultNavigationOptions: {
    gesturesEnabled: false,
    header: null
  },
  transitionConfig
};

const DrawerNavigator = createDrawerNavigator(
  {
    Home: { screen: Tabs }
  },
  {
    initialRouteName: 'Home',
    contentComponent: DrawerScreen,
    drawerWidth: (2 * width) / 3
  }
);

const AppNavigator = createStackNavigator(
  {
    Splash: { screen: Splash },
    DrawerNavigator: { screen: DrawerNavigator },
  },
  navigationOptions
);
```

### **MainActivity.js**

```
import android.content.Intent;
import android.os.Bundle;
import com.facebook.react.ReactActivity;
import org.devio.rn.splashscreen.SplashScreen;
import com.facebook.react.ReactActivityDelegate;
import com.facebook.react.ReactRootView;
import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;

public class MainActivity extends ReactActivity {

    /**
     * Returns the name of the main component registered from JavaScript.
     * This is used to schedule rendering of the component.
     */
    @Override
    protected String getMainComponentName() {
        return "mobile_app";
    }

    @Override
    protected ReactActivityDelegate createReactActivityDelegate() {
        return new ReactActivityDelegate(this, getMainComponentName()) {
            @Override
            protected ReactRootView createRootView() {
                return new RNGestureHandlerEnabledRootView(MainActivity.this);
            }
        };
    }

    @Override
    public void onActivityResult(int requestCode, int resultCode, Intent data) {
        super.onActivityResult(requestCode, resultCode, data);
        MainApplication.getCallbackManager().onActivityResult(requestCode, resultCode, data);
    }

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        SplashScreen.show(this);
        super.onCreate(savedInstanceState);
    }

}
```

### **Expected Behavior**

    What do you expect should be happening?

In the previous version we used (3.0.9) before upgrading to react-navigation 4.0.10, the drawer was closing fine if clicking outside the drawer container area on android.
[...]

PR #7834, opened 2020-03-19
feat: add swipeEnabled option to disable swipe gesture in drawer

raajnadar:
Commit to fix this issue https://github.com/react-navigation/react-navigation/issues/6475#issuecomment-587539763

I added separate prop `closeDrawerOnTap` which is `true` by default to close drawer by tap even when `enableGesture` is `false`.

### DIFF
--- a/packages/drawer/src/types.tsx
+++ b/packages/drawer/src/types.tsx
@@ -115,6 +115,11 @@ export type DrawerNavigationOptions = {
    */
   gestureEnabled?: boolean;
   /**
+   * Whether you can close the drawer with tap gesture.
+   * Defaults to `true`
+   */
+  closeDrawerOnTap?: boolean;
+  /**
    * Whether this screen should be unmounted when navigating away from it.
    * Defaults to `false`.
    */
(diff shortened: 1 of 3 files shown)


satya164:
Thanks for the PR. Can you add a new option called `swipeEnabled` instead which only enables/disables the swipe?

raajnadar:
@satya164 Is this the changes you want?

satya164:
@raajnadar it needs to enable/disable swipe (so put it on `PanGestureHandler`, the code you have will disable taps instead coz it's on `TapGestureHandler`.

Basically, in every place `gestureEnabled` is used, make it `gestureEnabled && swipeEnabled`, except the backdrop where tap gesture is used, leave that as is. It'll preserve `gestureEnabled`'s behaviour and add new prop `swipeEnabled` to control only swipes.

raajnadar:
If `gestureEnabled` is `false` & `swipeEnabled` is `true` both Pan and tap gesture are disabled.

If `gestureEnabled` is `true` & `swipeEnabled` is `true` both Pan and tap gesture are enabled.

If `gestureEnabled` is `true` & `swipeEnabled` is `false` Pan is disabled and tap gesture is enabled.

PR #7853, opened 2020-03-22
feat: add more info to useLinking error

bell-steven:
I had this error thrown, and I only declared useLinking once.  I was only getting the error on Android.
I realized that I had declared `android:launchMode="singleTask"` in the `application` section of the Android Manifest file,  instead of the `activity` section.

This should hopefully give users a little more direction if they didn't declare it in the manifest or implemented it incorrectly.


satya164:
Merged in https://github.com/react-navigation/react-navigation/commit/d94e43c3c8625b209a5c883b8cb560496d07fda7

PR #7871, opened 2020-03-26
fix: add pointerEvents=box-none to overlay View

wojtus7:
This `pointerEvents='none'` was present in the View before the introduction of overlayRender.
Without it, all interactions are blocked for 1-2 seconds after closing the screen.

It can be found in diff here: https://github.com/react-navigation/react-navigation/pull/7809/files

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -480,7 +480,10 @@ export default class Card extends React.Component<Props> {
       <CardAnimationContext.Provider value={animationContext}>
         <View pointerEvents="box-none" {...rest}>
           {overlayEnabled ? (
-            <View style={StyleSheet.absoluteFill}>
+            <View
+              pointerEvents="none"
+              style={StyleSheet.absoluteFill}
+            >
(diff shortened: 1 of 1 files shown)


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -480,7 +480,10 @@ export default class Card extends React.Component<Props> {
       <CardAnimationContext.Provider value={animationContext}>
         <View pointerEvents="box-none" {...rest}>
           {overlayEnabled ? (
-            <View style={StyleSheet.absoluteFill}>
+            <View
+              pointerEvents="none"
```
satya164: Thanks! Can you change it to `box-none` so the user can customize it by specifying it in the overlay itself?

```suggestion
              pointerEvents="box-none"
```

wojtus7: Cool by me - done

satya164:
Thanks! The CI is failing. Can you fix it so I can merge?

wojtus7:
My bad 😶- fixed

PR #7894, opened 2020-03-29
fix: undefined pattern allows path building (#7893)

### LINKED ISSUE #7893: Navigators without path result in pattern being undefined
**Current Behavior**

If no path is included on a navigator in the config provided to `useLinking`, navigating to another screen in another navigator fails for web. I have no issues on iOS.

This is the config I'm providing to `useLinking`

```
{
        [HOME]: {
            initialRouteName: GROUP_LIST,
            screens: {
                [GROUP_NAVIGATOR]: {
                    initialRouteName: GROUP_LIST,
                    screens: {
                        [GROUP_LIST]: 'groups',
                        [GROUP_DETAIL]: { path: 'groups/:id', parse: { id: String } },
                    },
                },
            },
        },
        [LANDING]: 'landing',
        [LOGIN]: 'login',
    };
```

If no path is provided to the `GROUP_NAVIGATOR` (which is  a StackNavigator) then routing to other paths complains about `pattern` being undefined.

Note, I intentionally omitted a path on the navigator as it made the URLs for web really weird. Whatever was listed in the navigator showed in the address bar. I want the URL to be `domain.com/groups/id`, including a path on the navigator resulted in `domain.com/navigatorpath/grops/:id`.

**Expected Behavior**

I would expect navigation to work even with no paths set on the navigators.

**How to reproduce**

I [built this snack](https://snack.expo.io/@anthony_crosscomm/pattner-undefined-bug). Make sure you are in Web view and tap 'ScreenSelector'. You'll get `F is undefined`.

I traced this back to the function [getPathFromState](https://github.com/react-navigation/react-navigation/blob/179b6312fedc02dff23346e6b4849d6f06eb28af/packages/core/src/getPathFromState.tsx); it does a check to see if `pattern` is an empty string. I've found `pattern` can be `undefined` with my above config, ultimately resulting in this error. This seems to happen only on web, I found no issue on iOS.

The solution is to change `if(pattern !== '')` to `if(!!pattern)`. I tested this locally and all worked properly.

I'm working on submitting a pull request to have this change effected.

The main things I would like another pair of eyes on are:

1. My config is correct
2. There isn't a specific 

**Your Environment**

| iOS or Android                 | iOS 13.3 simulator
| @react-navigation/native       | 5.1.3
| react-native                   | 36.0.0
| expo                           | 36.0.0
| node                           | 12.12

callmetwan:
Resolves #7893

Expand conditional check of `pattern` from checking for an empty string to checking for a falsey value. This solves use cases where `pattern` may be undefined. See #7893 for more details.

### DIFF
--- a/packages/core/src/getPathFromState.tsx
+++ b/packages/core/src/getPathFromState.tsx
@@ -105,7 +105,7 @@ export default function getPathFromState(
     }
 
     // we don't add empty path strings to path
-    if (pattern !== '') {
+    if (pattern) {
       const config =
         currentOptions[route.name] !== undefined
           ? (currentOptions[route.name] as { stringify?: StringifyConfig })


WoLewicki:
Thanks for your contribution! The change you made here is not enough since we need to put the routes we encountered into the path string, otherwise, they would be cut while resolving the state from the path. It is done on #7911, so I will close this PR.

Issue #6918, opened 2020-02-21
[ios] TextInput immediately gets unfocused after navigating back


MaxBolotnyi:
**Current Behavior**
After going back from any screen, if you click on the text input, it will focus and unfocus immediately 

I am attaching 2 snack's playgrounds. It's present both on react-navigation 4 and 5 versions.

**Expected Behavior**
After going back from any screen, if you click on the text input, it should focus and stay focused

**How to reproduce**
1. Go to the second screen.
2. Go back.
3. Right after transition ends click on the textInput component.

**SNACK** 
React-navigation 4 - https://snack.expo.io/@shonsy/navigationandfocus
React-navigation 5 - https://snack.expo.io/@shonsy/newnav

**Your Environment**
    "react-navigation": "^4.0.10",
    "react-native-screens": "^2.0.0-alpha.20",
    "react-navigation-stack": "2.1.1",
    "react-native-gesture-handler": "^1.5.0",
    "react-native-safe-area-context": "^0.6.0",
    "@react-native-community/masked-view": "^0.1.1"

**OR**
   "react-native-screens": "^2.0.0-alpha.20",
    "@react-navigation/stack": "^5.0.0",
    "@react-navigation/native": "^5.0.0",
    "react-native-gesture-handler": "^1.5.0",
    "react-native-safe-area-context": "^0.6.0",
    "@react-native-community/masked-view": "^0.1.5"

Rafal-JAR:
Having the same issue. On first screen load input focuses normally after navigating to another screen and back, for a short time the input will blur after focus.
```
"react-navigation": "^4.0.10",
"react-navigation-animated-switch": "^0.1.2",
"react-navigation-hooks": "^1.1.0",
"react-navigation-stack": "2.1.1",
"react-navigation-tabs": "^2.5.3",
"react-navigation-transitions": "^1.0.12",
```

Btw @MaxBolotnyi it looks like your snacks are incomplete 
`React-navigation 4` - example does not have navigation action on button
`React-navigation 5` - has only one screen

MaxBolotnyi:
@Rafal-JAR 
Thanks for pointing that out, I replaced the links so they are now complete.

thetminko:
+1

yuriybarnovich:
Reproducing as well.

declanelcocks:
[...]

PR #7907, opened 2020-03-29
fix: dismiss keyboard on screen change for android (#7875)

### LINKED ISSUE #7875: [android] Keyboard does not hide on screen change.
**Current Behavior**
Demo project to test (Expo):
https://github.com/MaxBolotnyi/android-keyboard-issue

When you open the keyboard and then switch the screen - the keyboard remains to be open.

**Expected Behavior**

The keyboard should hide when the screen changes.

**How to reproduce**

https://github.com/MaxBolotnyi/android-keyboard-issue

1. Clone the project.
2. Run `expo install`
3. Run `react-native run-android`
4. Click on the input field.
5. Click on the 'goSecond' button.

**Your Environment**

  "@react-native-community/masked-view": "^0.1.7",
    "expo": "~36.0.0",
    "react": "~16.9.0",
    "react-dom": "~16.9.0",
    "react-native": "~0.61.4",
    "react-native-gesture-handler": "^1.5.6",
    "react-native-reanimated": "^1.4.0",
    "react-native-safe-area-context": "^0.7.3",
    "react-native-screens": "^2.0.0-alpha.12",
    "react-native-unimodules": "~0.7.0",
    "react-native-web": "~0.11.7",
    "react-navigation": "^4.3.3",
    "react-navigation-stack": "^2.3.6"

MaxBolotnyi:
Resolves #7875

Should dismiss the keyboard when the screen change is confirmed on android ( already does it on ios ).
Checked both platforms, works fine both on android & ios.

### DIFF
--- a/packages/stack/src/views/KeyboardManager.tsx
+++ b/packages/stack/src/views/KeyboardManager.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { TextInput } from 'react-native';
+import { TextInput, Platform, Keyboard } from 'react-native';
 
 type Props = {
   enabled: boolean;
@@ -58,6 +58,10 @@ export default class KeyboardManager extends React.Component<Props> {
 
     if (input) {
       TextInput.State.blurTextInput(input);
+    }
+
+    if (Platform.OS === 'android') {
+      Keyboard.dismiss();
     }
 
     // Cleanup the ID on successful page change


MaxBolotnyi:
@satya164 previously there was only `Keyboard.dismiss()` line for both platforms. To solve the #6918 there was a PR replacing that line with blurring only previously focused textInput, while still dismissing keyboard on ios. However, this same code broke android behavior. 

My PR has the #6918 issue in mind and works fine on both platforms, I checked this issue too before submitting the PR itself :)

MaxBolotnyi:
@satya164 
The same code does not work for android as it works for ios due to the platform-specific features, and because in fact, these are two different issues, one was happening during the back navigation, while the new one is specific for android and exists anytime you need to change the screen

satya164:
I get that you brought back the old code for Android only and these are 2 different issues, but what I'm curious about is why does the same code produce different behaviour on Android than iOS? Like why doesn't it produce the bug on Android since `Keyboard.dismiss()` should close dismiss any opened keyboard, thereby closing input in newly opened screen as well. Is it because of timing? 

> The same code does not work for android as it works for ios due to the platform-specific features
Seems like `blurTextInput(input)` is also supported on Android looking at this code: https://github.com/facebook/react-native/blob/bbd91446dd7a71e785b0405df68bc87fa1c9f3aa/Libraries/Components/TextInput/TextInputState.js#L132. So I'm curious about what's the difference between the 2 platforms.

MaxBolotnyi:
@satya164 

I cannot say for sure where exactly this happens, but the ios Keyboard gets being hidden somewhere outside of the react-navigation code. When it comes to android, the keyboard stays open until you hide it explicitly, at least with the current react-navigation realization.

At the moment you want to navigate to a new screen, not back to the previous one, you would have this check 
`const input = this.previouslyFocusedTextInput;

    if (input) {`
 go through the false scenario, which will not hide anything for android

MaxBolotnyi:
@satya164  the`blurTextInput` method itself works for both platforms but is not called on android when it should be

packages/stack/src/views/KeyboardManager.tsx
```diff
@@ -60,6 +60,10 @@ export default class KeyboardManager extends React.Component<Props> {
       TextInput.State.blurTextInput(input);
     }
 
+    if (Platform.OS === 'android') {
+      Keyboard.dismiss();
+    }
```
satya164:
```suggestion
    if (Platform.OS === 'android') {
      Keyboard.dismiss();
    } else if (input) {
      TextInput.State.blurTextInput(input);
    }
```

satya164:
Merged in 8432e5ab25f041af8538ea7fb35e97cfcf1f983e

I'll investigate why it happens later

PR #7931, opened 2020-04-02
fix: android/ios focus on navigating back and keyboard management (#8094)

MaxBolotnyi:
This commit 8432e5a brought the issue  #6918  back for Android

I've come up with the idea solving both  #6918 and #7875 (#7907)

### DIFF
--- a/packages/stack/src/vendor/views/KeyboardManager.tsx
+++ b/packages/stack/src/vendor/views/KeyboardManager.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react';
-import { TextInput, Platform, Keyboard } from 'react-native';
+import { TextInput, Keyboard } from 'react-native';
 
 type Props = {
   enabled: boolean;
   children: (props: {
     onPageChangeStart: () => void;
-    onPageChangeConfirm: () => void;
+    onPageChangeConfirm: (closing: Boolean) => void;
     onPageChangeCancel: () => void;
   }) => React.ReactNode;
 };
@@ -47,7 +47,7 @@ export default class KeyboardManager extends React.Component<Props> {
     this.startTimestamp = Date.now();
   };
 
-  private handlePageChangeConfirm = () => {
+  private handlePageChangeConfirm = (closing: Boolean) => {
     if (!this.props.enabled) {
       return;
     }
@@ -56,10 +56,10 @@ export default class KeyboardManager extends React.Component<Props> {
 
     const input = this.previouslyFocusedTextInput;
 
-    if (Platform.OS === 'android') {
-      Keyboard.dismiss();
-    } else if (input) {
+    if (closing && input) {
       TextInput.State.blurTextInput(input);
+    } else {
+      Keyboard.dismiss();
     }
 
     // Cleanup the ID on successful page change
--- a/packages/stack/src/vendor/views/Stack/CardContainer.tsx
+++ b/packages/stack/src/vendor/views/Stack/CardContainer.tsx
@@ -43,7 +43,7 @@ type Props = TransitionPreset & {
   ) => void;
   onTransitionEnd?: (props: { route: Route<string> }, closing: boolean) => void;
   onPageChangeStart?: () => void;
-  onPageChangeConfirm?: () => void;
+  onPageChangeConfirm?: (closing: Boolean) => void;
   onPageChangeCancel?: () => void;
   gestureEnabled?: boolean;
   gestureResponseDistance?: {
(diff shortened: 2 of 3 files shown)


MaxBolotnyi:
@satya164  Hi, please review when have time

packages/stack/src/vendor/views/KeyboardManager.tsx
```diff
@@ -56,10 +56,10 @@ export default class KeyboardManager extends React.Component<Props> {
 
     const input = this.previouslyFocusedTextInput;
 
-    if (Platform.OS === 'android') {
-      Keyboard.dismiss();
-    } else if (input) {
+    if (closing && input) {
       TextInput.State.blurTextInput(input);
```
satya164, line 60: Hey, thanks for the PR, but it seems confusing as to why there's need for separate logic for `closing`. When a screen is closing, it's going to unmount and that should remove focus from the input anyway and hide the keyboard without having to blur it manually.

So I'm not really sure if the issue here is with the keyboard handling. Maybe your view is not unmounting somehow and it's a different issue? Can you share a repro for this?

MaxBolotnyi: @satya164 
it's there to resolve this issue - #6918
Which happens when the screen closes, when you close the screen and right away focus on the input on the previous screen, you'll get the input blur and focus back

satya164: I just want to understand the logic behind all the changes. Without understanding the logic, it's easy to break it in future updates and impossible to make any kind of changes. It may fix the issue, but how? Can you add some more description of the behaviour you observe so I can better understand what exactly is happening?

MaxBolotnyi: @satya164 
Sure, the current behavior is the following: 
Preconditions: you have 2 screens available. On the 1st screen, you have a TextInput, on the second one you don't need anything at all.
1. You open the second screen
2. You click on go back btn
3. You click on input right away while the transition is still in progress
4  The input get focused, then it blurs right away.
If you focus the input second time - it will work just fine, also if you focus after the transition it will also work fine

To resolve this issue, you added the following code: 
`  if (input) {
      TextInput.State.blurTextInput(input);...
`
But the thing is it broke the behavior for the opening of the screen, which caused this issue #7875 

So keeping all this in mind we have to:
1. Dismiss the keyboard on the transition to the new screen.
2. Only unfocus the previously focused inputs if we going back from the screen to allow people immediately focus when they see the previous screen.

The last PR we made for both of these issues solved it only partially, while this one is supposed to address it on both platforms in the right way

MaxBolotnyi: @satya164  did I miss something important for you here, or this is enough? :)

hamzaouhssain1993:
I confirm that this problem is back for Android devices. Going back and clicking right away on the text input makes the keyboard dismiss. Please if you can check and approve the PR would be great. Thanks

satya164:
Can you share a repro and mention more details about Android version and device model?

Trancever:
@hamzaouhssain1993 I've just closed #7875 because it seems to be resolved in the latest version of the library (both v4 and v5). Please create a new issue and provide a small repro. Make sure to include information about react-navigation version that you use.

MaxBolotnyi:
@Trancever Hi Dawid, [ here ](https://github.com/react-navigation/react-navigation/issues/8094) is the link to the issue with the snack repro and the video of the issue.

The issue is located in the stack, not in the react-navigation itself, as you can see from the PR.

packages/stack/src/vendor/views/KeyboardManager.tsx
thekingrenz23: > @satya164
> Sure, the current behavior is the following:
> Preconditions: you have 2 screens available. On the 1st screen, you have a TextInput, on the second one you don't need anything at all.
> 
> 1. You open the second screen
> 2. You click on go back btn
> 3. You click on input right away while the transition is still in progress
>    4  The input get focused, then it blurs right away.
>    If you focus the input second time - it will work just fine, also if you focus after the transition it will also work fine
> 
> To resolve this issue, you added the following code:
> ` if (input) { TextInput.State.blurTextInput(input);...`
> But the thing is it broke the behavior for the opening of the screen, which caused this issue #7875
> 
> So keeping all this in mind we have to:
> 
> 1. Dismiss the keyboard on the transition to the new screen.
> 2. Only unfocus the previously focused inputs if we going back from the screen to allow people immediately focus when they see the previous screen.
> 
> The last PR we made for both of these issues solved it only partially, while this one is supposed to address it on both platforms in the right way

i have this issue did you fixed it ?

PR #7943, opened 2020-04-03
fix: ios presentation modal cuts the topOffset on the bottom

### LINKED ISSUE #7856: TransitionPresets.ModalPresentationIOS cause card to overflow the device screen
**Current Behavior**

Please look at the orange border which is set on the `cardStyle` of a screen.

Standard Modal Transition:
Simulator Screen Shot - iPhone 8 - 2020-03-24 at 15 57 13

When using `TransitionPresets.ModalPresentationIOS`, note the bottom of the orange border is not inside the device's screen:

Simulator Screen Shot - iPhone 8 - 2020-03-24 at 15 57 07

Note: The `borderWidth` I set is 11, it looks like the exceeding height is 10.

**Expected Behavior**

The screen should shrink in height to be fully displayed when `TransitionPresets.ModalPresentationIOS` is used.

**How to reproduce**

**Your Environment**

| iOS or Android                 | ios |
|    "@react-navigation/bottom-tabs" | "^5.2.2", |
|    "@react-navigation/native" | "^5.1.0",
|   "@react-navigation/stack" | "^5.2.1",
| react-native                   | 0.61.4 |
| expo                           | ^36.0.0 | 
| node                           | v12.0.0 | 
| npm or yarn                    | yarn |

Ferossgp:
Because of the translateY moving the screen out to the bottom of view by 10 pt, these 10pt are hidden under the screen, or steal this size from the safe area. To avoid cutting elements, the size of the screen could be decreased by the `topOffset` using padding on the bottom. Fixes #7856

### DIFF
--- a/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
+++ b/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
@@ -164,6 +164,7 @@ export function forModalPresentationIOS({
       borderTopLeftRadius: borderRadius,
       borderTopRightRadius: borderRadius,
       marginTop: index === 0 ? 0 : statusBarHeight,
+      paddingBottom: topOffset,
       transform: [{ translateY }, { scale }],
     },
     overlayStyle: { opacity: overlayOpacity },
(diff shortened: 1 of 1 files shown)


packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
```diff
@@ -164,6 +164,7 @@ export function forModalPresentationIOS({
       borderTopLeftRadius: borderRadius,
       borderTopRightRadius: borderRadius,
       marginTop: index === 0 ? 0 : statusBarHeight,
+      paddingBottom: topOffset,
```
SDSLeon, line 167: index === 0 ? 0 : topOffset

we don't need padding for the first screen

satya164:
Thanks for the PR!

PR #7957, opened 2020-04-06
fix: use 1 as default in compat pop action

### LINKED ISSUE #7956: Compatability Package: Calling pop without an argument using withNavigation errors
**Current Behavior**

When using `withNavigation` on a component, and then calling `navigation.pop()` on the given navigation prop gives an error:

> undefined is not an object 

**Expected Behavior**

It should use the default of 1 as a number, instead of erroring.

pmk1c:
resolves #7956 

Not sure if this is the best fix, but I think this should resolve the problem. Maybe the better way is to give `{}` to `StackActions.pop` instead of `n`. But I'm not sure if people are already putting objects into `pop` (which they shouldn't, I guess).

### DIFF
--- a/packages/compat/src/helpers.tsx
+++ b/packages/compat/src/helpers.tsx
@@ -57,7 +57,7 @@ export function push(routeName: string, params?: object, action?: never) {
   });
 }
 
-export function pop(n: number) {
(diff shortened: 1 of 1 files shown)


satya164:
thanks for the pr

Issue #7962, opened 2020-04-06
How to filter DrawerItems


ROODAY:
**Current Behavior**
Our app has a Home screen, Login Screen, and Article Screen. Article is a reusable screen that changes depending on the required params passed to it, so we don't want a user clicking the Article screen in the Drawer and navigating to it (instead we navigate using the API when an article button is clicked). I want to hide the DrawerItem for Article, and I did so by filtering the routes (I did extensive googling and all the results were a couple years old so this is the best I could find).
```
const Drawer = createDrawerNavigator();
const visibleRoutes = ['Home', 'Login'];
const Sidebar = props => {
    const clonedProps = {...props};
    clonedProps.state.routes = props.state.routes.filter(route => visibleRoutes.includes(route.name));
    return (
        <DrawerContentScrollView {...clonedProps}>
            <DrawerItemList {...clonedProps} />
        </DrawerContentScrollView>
    );
}

...
<Drawer.Navigator
initialRouteName="Home"
drawerContent={Sidebar} >
...
```

**Expected Behavior**
This works and removes the unwanted DrawerItems from the Sidebar. However, when clicking an Article button, I get this error:
```
The action 'NAVIGATE' with payload {"name":"Article","params":{"id":1,"title":"Democratic Debate","subtitle":"Bernie dominates debate, but democrats seem to rally behind Biden still.","dateString":"April 06, 2020"}} was not handled by any navigator.

Do you have a screen named 'Article'?

If you're trying to navigate to a screen in a nested navigator, see https://reactnavigation.org/docs/nesting-navigators#navigating-to-a-screen-in-a-nested-navigator.
```

I do have a screen named article, defined like so:
```
<Drawer.Screen
    name="Article"
    component={ArticleScreen}
    listeners={{
        focus: e => StatusBar.setHidden(true),
        blur: e => StatusBar.setHidden(false)
    }}
/>
```
Yet I get this error whenever I click on the button. It would seem that the filtered routes I send to the DrawerItems are the same ones that the entire navigation uses. What's the point of a custom component then?

**How to reproduce**
[Full Code](https://github.com/Veri-News/Veri-Mobile/blob/master/components/AppNavigator.js)

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | Android
| @react-navigation/native       |  5.1.4
| @react-navigation/drawer       | 5.4.0
| react-native-reanimated        | 1.4.0
[...]

PR #7964, opened 2020-04-06
feat: allow hiding of routes in DrawerItemList

ROODAY:
I made this based on my own issue #7962 ([plus it seems like a lot of people have wanted this for a while](https://www.google.com/search?q=react+navigation+hide+drawer+item)).

What this proposes is that now users can define a list of routes they want to be visible when defining a custom `drawerContent` component:
```
function CustomDrawerContent(props) {
  return (
    <DrawerContentScrollView {...props}>
      <DrawerItemList {...props} visibleRoutes={["Home", "Login", "Settings"]} />
    </DrawerContentScrollView>
  );
}
```

This would then filter out any routes not in the `visibleRoutes` prop. If `visibleRoutes` is not set, it will fallback to using `state.routeNames` and display all the routes.

### DIFF
--- a/packages/drawer/src/views/DrawerItemList.tsx
+++ b/packages/drawer/src/views/DrawerItemList.tsx
@@ -30,11 +30,11 @@ export default function DrawerItemList({
   inactiveBackgroundColor,
   itemStyle,
   labelStyle,
-  visibleRoutes
+  hiddenRoutes
 }: Props) {
-  if (!visibleRoutes) visibleRoutes = state.routeNames;
+  const hidden = hiddenRoutes ? hiddenRoutes : [];
   
-  return (state.routes.filter(route => visibleRoutes.includes(route.name)).map((route, i) => {
+  return (state.routes.filter(route => !hidden.includes(route.name)).map((route, i) => {
     const focused = i === state.index;
     const { title, drawerLabel, drawerIcon } = descriptors[route.key].options;
 
--- a/packages/drawer/src/views/DrawerItemList.tsx
+++ b/packages/drawer/src/views/DrawerItemList.tsx
@@ -30,8 +30,11 @@ export default function DrawerItemList({
   inactiveBackgroundColor,
   itemStyle,
   labelStyle,
+  visibleRoutes
 }: Props) {
-  return (state.routes.map((route, i) => {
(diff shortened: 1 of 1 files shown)


ROODAY:
I've never used TypeScript before, so I think one thing I missed was defining a type for this new prop I added. Could someone point me to some relevant docs/tutorials so I can add checking that `visibleRoutes` is either an array of `Route<String>` or `undefined`?

ROODAY:
Is that Circle CI lint failure going to be enforced? If I understand it correctly it wants me to put the filter call on a newline, but then shouldn't the map call also be put on a newline? I was just trying to follow the style inside the file, could a maintainer advise?

ROODAY:
Hi @satya164, could I get a review when you get a chance?

xstable:
Good approach, but wouldn't it be better to have an parameter working in the opposite way? Means to hiddenRoutes={["Secret", "Login", "Imprint"]}. 
This would minimize the size of data to add to this property, because generally there will be most of the time more Items to show as to hide.

ROODAY:
@xstable Sounds good, just made an update!

satya164:
Thanks for the PR.

Why not put the screen in a parent/child stack navigator if you don't want it to be displayed in the drawer sidebar? It could be confusing to the user have a screen rendered, but not have anything selected in the drawer.

ROODAY:
@satya164 That's fair, at the time I was developing a class project and we just needed to get it out by the deadline and hiding the drawer item was the fastest solution (the specific screen was an article screen after a user clicked an article card, we didn't want users to navigate to the article screen through the drawer, only when they clicked the corresponding article card). If I were to go back and refactor it, I guess the way you described would be the most idiomatic (placing the article screen inside a stack navigator for the home screen).

samipshah100:
This fits my use case where by I don't want to nest navigators because: 
1. I am navigating conditionaly, between routes hence I don't always know if the navigator is nested. 
2. Using Deep links so need the navigation routes state to be as flat as possible

satya164:
I'm closing the issue. We recommend nesting a stack inside if you don't want to show a screen in drawer/tabs.

@samijaber not sure why deep links would require flat state.

cabhara:
How can drawerItems be hidden by nesting a stack?

Issue #7981, opened 2020-04-09
Build error after updating to 5.2.10


panda0603:
**Current Behavior**

![image](https://user-images.githubusercontent.com/34116285/78845827-b3bb9000-7a44-11ea-8a72-4b3d19b4a6cd.png)

updated @react-navigation/bottom-tabs, native, stack to latest version just now. 
Getting above error. 

Already tried deleting the cache, but no luck 

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | both
| @react-navigation/native       |5.1.5
| @react-navigation/stack        |5.2.10
| react-native-gesture-handler   |1.6.1
| react-native-safe-area-context |0.7.3
| react-native-screens           |2.4.0
| react-native                   |0.61.5
| npm or yarn                    | npm

gregoryfm:
Same issue here

raajnadar:
Did you try upgrading typescript to latest version?

raajnadar:
Follow this steps

1) Delete `node_modules` & lock file (yarn.lock or package.lock)
2) Upgrade typescript to latest i.e. 3.8.3
3) Run `yarn` or `npm install`
4) Run `yarn react-native run-android`

If it still did not work disable hermes.

I had to disable hermes IDK the exact reason but it caused insta crash in RN 0.62

gregoryfm:
Hi @raajnadar 

I'm not using either typescript or hermes. 

I already tried removed node_modules, yarn.lock, cleared yarn cache and installed using npm/yarn
But still the same error.

raajnadar:
Try doing this for android.

`cd android` and run `gradlew clean`

gregoryfm:
I tried that as well, same error..

raajnadar:
This is the last solution that I can give delete the apk that is installed by adb on you phone. Make sure you delete the app for all the users.

After that `gradlew clean` and re run now.

raajnadar:
I am adding all the possible solutions in this comment https://github.com/react-navigation/react-navigation/issues/7981#issuecomment-611377210 so keep an eye on the edits.

nickyleach:
I wonder if something went wrong published the packages? I tried downgrading to `5.2.8` but it resolves to `5.2.10`. From my `yarn.lock`:


```"@react-navigation/stack@^5.2.8":
  version "5.2.10"
[...]

PR #7993, opened 2020-04-09
fix: Unexpected Token Error

Savinvadim1312:
The last version was crashing with the `Unexpected Token Error`. This commit fixes it.

```
SyntaxError: /Users/vadim/Projects/Fitenium/fitenium-mobile/node_modules/@react-navigation/stack/src/index.tsx: Unexpected token (51:12)

  52 |   StackNavigationOptions,
  49 |  * Types
  53 |   StackNavigationProp,
  50 |  */
> 51 | export type {
```

### DIFF
--- a/packages/stack/src/index.tsx
+++ b/packages/stack/src/index.tsx
@@ -48,7 +48,7 @@ export { default as useGestureHandlerRef } from './utils/useGestureHandlerRef';
 /**
  * Types
  */
-export type {
+export {
   StackNavigationOptions,
   StackNavigationProp,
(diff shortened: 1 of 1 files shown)


raajnadar:
The solution is here https://github.com/react-navigation/react-navigation/issues/7981#issuecomment-611377210 please don't send unnecessary PR

raajnadar:
This is not a solution. The solution is to upgrade `@babel/core`, `@babel/runtime` & `typescript` (typescript if you use) to latest.

PR #8018, opened 2020-04-13
feat: use anchor elements on web for a11y features

EvanBacon:
Using `<a />` elements for the drawer items, this adds the popup on hover and right-click actions. The actual href isn't used for linking because it would cause the page to completely re-render.

There is a problem in the example where `Root` is used for the route name but the actual path is `/` which means the hover preview is partially incorrect. Open to feedback on how to get the path for a route.

Screen Shot 2020-04-12 at 5 44 14 PM


satya164:
Closing since drawer and tabs now include proper links.

PR #8088, opened 2020-04-20
fix: 'unknown token type' error

meet-panchal:


raajnadar:
This is the solution. https://github.com/react-navigation/react-navigation/issues/7981#issuecomment-611377210

Issue #8090, opened 2020-04-21
createMaterialTopTabNavigator upperCaseLabel seems to have been deleted, What's the substitution for 5x？


MiaoMiaoai:
**Current Behavior**

- What code are you running and what is happening?
- Include a screenshot or video if it makes sense.

**Expected Behavior**

- What do you expect should be happening?
- Include a screenshot or video if it makes sense.

**How to reproduce**

- You must provide a way to reproduce the problem. If you are having an issue with your machine or build tools, the issue belongs on another repository as that is outside of the scope of React Navigation.
- Either re-create the bug on [Snack](https://snack.expo.io) or link to a GitHub repository with code that reproduces the bug.
- Explain how to run the example app and any steps that we need to take to reproduce the issue from the example app.
- Keep the repro code as simple as possible, with the minimum amount of code required to repro the issue.
- Before reporting an issue, make sure you are on latest version of the package.

**Your Environment**

| software                            | version | 
| ----------------------------------- | ------- |
| iOS or Android                      | iOS 13.3
| @react-navigation/native            | v5.1.6
| @react-navigation/material-top-tabs | v5.1.9
| react-native-tab-view               |v2.14.0
| react-native                        |v0.62.2
| npm or yarn                         | yarn

upperCaseLabel seems to have been deleted, What's the substitution for 5x？

raajnadar:
This is the solution

> It can be changed by specifying a different `textTransform` in `labelStyle`.

https://github.com/react-navigation/react-navigation/pull/8091#issuecomment-617836054

PR #8091, opened 2020-04-21
feat: ability to change text case for material top tabs

raajnadar:
Fix https://github.com/react-navigation/react-navigation/issues/8090

Added the `upperCaseLabel` prop for material top tabbar which was available in version 4.


satya164:
Thanks for the PR.

There's no need for a separate prop. It can be changed by specifying a different `textTransform` in `labelStyle`. So I'm closing the PR.

Issue #7317, opened 2020-02-24
backBehavior: 'history' doesn't work


firofame:
### Current behavior

when I call goBack() nothing happens.

### Expected behavior

when I call goBack() it should go to the last visited tab.

### Code sample

```
import React from 'react';
import {View, Button} from 'react-native';
import {createAppContainer} from 'react-navigation';
import {createMaterialTopTabNavigator} from 'react-navigation-tabs';

const Screen = ({navigation: {goBack}}) => {
  return (
    <View>
      <Button onPress={() => goBack()} title="Go Back" />
    </View>
  );
};

const TabNavigator = createMaterialTopTabNavigator(
  {
    Screen1: Screen,
    Screen2: Screen,
    Screen3: Screen,
  },
  {
    backBehavior: 'history',
  },
);

export default createAppContainer(TabNavigator);
```

### Your Environment

ios & Android

"react-native": "0.60.5",
"react-navigation": "^4.0.0",
"react-navigation-tabs": "^2.3.0",

aliazizi:
I have same issue

ainar:
It looks like someone mixed the "initialRoute" and first route:
https://github.com/react-navigation/react-navigation/blob/9a8fea8f2c1bdabfc5dd87e5c3ff4e7b97aef47d/packages/routers/src/TabRouter.tsx#L68

2xSamurai:
Is there any fix for this other than downgrading. My packages are as follows

"@react-navigation/bottom-tabs": "^5.6.1",
    "@react-navigation/material-top-tabs": "^5.2.12",
    "@react-navigation/native": "^5.6.1",
    "@react-navigation/stack": "^5.6.2",

backBehavior="history" does nothing

Someone please help. Thank you.

ronastlelobo:
Is there any fix for this ? I am also facing this issue.

aliazizi:
downgrade to old version

> 
> 
> Is there any fix for this ? I am also facing this issue.

shivanitp:
is there any solution get in this ?
i also facing same issue 
 "react-navigation-tabs": "^2.5.5",
 "react-navigation": "^4.0.9",

chris-hud:
Running into same issue.
`    <Tab.Navigator
      backBehavior="history"
      screenOptions={{
        headerShown: true
      }}
    >`
    
For all navigations that originate on the second tab, back button still returns to first tab. 

This is a real problem for me.

PR #8110, opened 2020-04-23
fix: backBehavior with initialRoute

ainar:
I corrected the behavior when we push the back button and we defined an initialRouteName. (https://github.com/react-navigation/react-navigation/issues/7317) Before, back button always got back to the first route, even if the initial route was not the first.

I added relevant tests to check the correct behavior.

### DIFF
--- a/packages/routers/src/TabRouter.tsx
+++ b/packages/routers/src/TabRouter.tsx
@@ -64,15 +64,13 @@ const getRouteHistory = (
   const history = [{ type: TYPE_ROUTE, key: routes[index].key }];
 
   switch (backBehavior) {
-    case 'initialRoute':
-      if (index !== 0) {
-        history.unshift({ type: TYPE_ROUTE, key: routes[0].key });
-      }
-      break;
     case 'order':
       for (let i = index; i > 0; i--) {
         history.unshift({ type: TYPE_ROUTE, key: routes[i - 1].key });
       }
+      break;
+    case 'initialRoute':
+      // The history always begin with the initial route (see changeIndex)
       break;
     case 'history':
       // The history will fill up on navigation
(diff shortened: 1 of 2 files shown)


packages/routers/src/TabRouter.tsx
```diff
@@ -64,16 +64,14 @@ const getRouteHistory = (
...
-        history.unshift({ type: TYPE_ROUTE, key: routes[0].key });
-      }
-      break;
     case 'order':
       for (let i = index; i > 0; i--) {
         history.unshift({ type: TYPE_ROUTE, key: routes[i - 1].key });
       }
       break;
+    case 'initialRoute':
+      // The history always begin with the initial route (see changeIndex)
```
satya164, line 73: This is not guaranteed during rehydration. It uses the `getHistory` function to populate the history, so this function shouldn't return incorrect history.

packages/routers/src/TabRouter.tsx
```diff
@@ -95,6 +93,13 @@ const changeIndex = (
     history = state.history
       .filter((it) => (it.type === 'route' ? it.key !== currentKey : false))
       .concat({ type: TYPE_ROUTE, key: currentKey });
+  } else if (backBehavior === 'initialRoute') {
+    const currentKey = state.routes[index].key;
+
+    history = [state.history[0]];
```
satya164, line 99: History should be populated inside the `getHistory` function for it to work correctly. There are changes to history in this function when `backBehavior` is `history`, because in that case, `history` is coupled to the `index` change.

ainar:
@satya164 I fixed rehydration. (I messed the word in my commit message)
Otherwise, I did not find any other way than refactoring to get this initial route.

ainar:
@satya164 I Think it is better now.

ainar:
Any news? I am using the fix in my app and it seems to work like a charm.

ashishfeels:
@ainar for me its not working the bottomTabs history is still messed up , it exits the app instead of going to previous tab.

ronastlelobo:
@ashishfeels  same here. Any luck in figuring out?

PR #8117, opened 2020-04-24
feat: Update README.md

cmcodes1:
Added installation steps in detail so that it will be easier of a new user to install this package in their project.


cmcodes1:
@brentvatne Kindly review the same.

brentvatne:
i think we should just link to https://reactnavigation.org/docs/getting-started rather than duplicating the installation instructions in the README

satya164:
We already have a link to the documentation website, both in the repository description and in the README. So I'm going to close the PR since like @brentvatne said, we don't want to duplicate the instructions here.

PR #8124, opened 2020-04-25
feat: support to deeplink url of universal language

FuzzyFade:
I recently found that deeplink url of universal language isn't supported at present. I have slightly modified it.


FuzzyFade:
hello？

PR #8140, opened 2020-04-27
feat: pass deep linking url to getStateFromPath

willcaul:
# Overview
I would like to pass the full URL to the `getStateFromPath()` function so that my custom `useLinking()` `getStateFromPath` param can use the`url` when determining the state. 

## Background
I want my app to support deep links from multiple TLDs and I want to show different screens for certain TLDs.  As a result, I cannot rely on only the `path` when calculating the state (i.e. the path can be the same, but the TLDs are different).

Please let me know if this is acceptable or if there is a preferred way to achieve supporting different states based on different TLDs.

### DIFF
--- a/packages/core/src/__tests__/getStateFromPath.test.tsx
+++ b/packages/core/src/__tests__/getStateFromPath.test.tsx
@@ -88,6 +88,61 @@ it('converts path string to initial state with config', () => {
   );
 });
 
+it('converts path string to initial state with config and url', () => {
+  const path = '/foo/bar/sweet/apple/baz/jane?count=10&answer=42&valid=true';
+  const config = {
+    Foo: 'foo',
+    Bar: 'bar/:type/:fruit',
+    Baz: {
+      path: 'baz/:author',
+      parse: {
+        author: (author: string) =>
+          author.replace(/^\w/, (c) => c.toUpperCase()),
+        count: Number,
+        valid: Boolean,
+      },
+      stringify: {
+        author: (author: string) => author.toLowerCase(),
+      },
+    },
+  };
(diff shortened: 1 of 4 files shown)


satya164:
Thanks for the PR. Maybe you can specify an empty prefix instead so you receive the full URL for `path`?

```js
prefixes: ['']
```

willcaul:
> Thanks for the PR. Maybe you can specify an empty prefix instead so you receive the full URL for `path`?
> 
> ```js
> prefixes: ['']
> ```

Thanks for the suggestion @satya164... I will try that. Is there a way for me to route to the same screen for different paths/urls using the `useLinking()` `config` param? for example, is there a way to do something like: 
```
config: {
   ThingScreen: [
      'https://www.example.com/path1/thing', 
      'https://www.example.com/path2/thing', 
      'https://www.example.co.uk/path1/thing', 
      'https://sub.example.com/path2/thing'
   ]
}
```

satya164:
If you specify your own `getStateFromPath`, then you can normalize these URLs to a single format and call `getStateFromPath` from the library.

```js
(path, config) =>
  // normalize the URLs to specific path and strip the prefixes in `normalize(...)`
  getStateFromPath(normalize(path), config)
```

willcaul:
> If you specify your own `getStateFromPath`, then you can normalize these URLs to a single format and call `getStateFromPath` from the library.
> 
> ```js
> (path, config) =>
>   // normalize the URLs to specific path and strip the prefixes in `normalize(...)`
>   getStateFromPath(normalize(path), config)
> ```

Thank you @satya164, that is very helpful! I will give that a try!

PR #8155, opened 2020-04-29
support screen options reset tabBarOptions

malacca:
when i make bottom tab bar like this

tabbar

1. switch first page, tab bar is transparent
2. switch other page, tab bar background is black

i found it's difficult to do with current api

so, i add code, support screen options reset tabBarOptions, and it's very easy to do this, it's not  break change

--------------------
another change is add  insets.bottom to custom style.height auto.

this may not be compatible with the original, but it is more logical, because `style.padding` is add insets.bottom auto

if you think it is inappropriate, i can remove this change, this change is not the point

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -96,6 +96,11 @@ export type BottomTabNavigationOptions = {
    * Defaults to `false`.
    */
   unmountOnBlur?: boolean;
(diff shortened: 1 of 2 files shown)


malacca:
I researched it in depth and found that what I actually need is to combine `tabBarOptions` props to `options`, but this will cause a lot of changes, so close PR

PR #8158, opened 2020-04-30
[v4.x] fix: change old docUrl to new docUrl due v5

linzera:
This PRs fixes the docs URLs shown on some warns to guide us to the right path on docs, following it you should get stuck here:

Page Not Found | React Navigation 2020-04-30 04-01-49

With this PR, we can reach the URL for the v4 documentation, since v5 is out.


satya164:
Thanks!

PR #8174, opened 2020-05-02
document title

osdnk:


packages/core/src/useDescriptors.tsx
```diff
@@ -116,43 +121,47 @@ export default function useDescriptors<
...
           return (
             <NavigationBuilderContext.Provider key={route.key} value={context}>
-              <SceneView
-                navigation={navigation}
-                route={route}
-                screen={screen}
-                getState={getState}
-                setState={setState}
-              />
+              <NavigationContext.Provider value={navigation}>
```
satya164, line 151: The `NavigationContext` is already there in `SceneView` so changes to this file shouldn't be needed.

packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,75 @@ it("doesn't throw if children is null", () => {
...
+            </TestNavigator>
+          )}
+        </Screen>
+      </TestNavigator>
+    </BaseNavigationContainer>
+  );
+
+  render(element).update(element);
+
+  await sleep(200);
```
satya164, line 1543: Instead of `sleep`, you can do `jest.useFakeTimers()` at the top, and then tests, `jest.runAllTimers()`

packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,75 @@ it("doesn't throw if children is null", () => {
...
+  const TestNavigator = (props: any) => {
+    const { state, descriptors } = useNavigationBuilder(MockRouter, props);
+
+    return descriptors[state.routes[state.index].key].render();
+  };
+
+  let navigate: () => void = () => {};
+
+  const TestScreen = (to: string) => (props: any) => {
+    navigate = () =>
```
satya164, line 1504: I think the test might be more clear if the `to` parameter is passed to `navigate`

packages/core/src/useTitleContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { SharedScreenNavigationOptions, useIsFocused } from './index';
+import NavigationDocumentTitleContext from './NavigationDocumentTitleContext';
+
+let timeout: NodeJS.Timeout | number | undefined;
+
+let nonNavigationTitle: string | undefined;
+
+let recentSetter: string | undefined;
+
+function debounceSetTitle(title: string | undefined, key: string) {
```
satya164, line 11: Is debounce needed? I guess it might screw up SEO stuff.

packages/core/src/useDescriptors.tsx
osdnk: I moved it from scene view here to have it accessible in ScreeView component

packages/core/src/useTitleContext.tsx
osdnk: I've just tested and seems to be not needed

packages/core/src/NavigationDocumentTitleContext.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+/**
+ * Context which holds data for proper setting title of document on web
+ */
+const NavigationDocumentTitleContainer = React.createContext<
```
WoLewicki, line 6: Shouldn't it be `NavigationDocumentTitleContext`?

PR #8182, opened 2020-05-03
feat: update react-native-safe-area-context to 1.0.0

janicduplessis:
I made sure 1.0 is backwards compatible with react-navigation, which means using rn-safe-area-context@1+ with older versions of react-navigation will still work.


janicduplessis:
@satya164 Updated, 1.0.0 has no breaking changes that affect react-nav, only soft deprecations for now (no warning). I also made sure it is binary compatible, which means using v1 js works with v0.7.0 native code which is included in expo. Tested that the example app runs fine on expo.

PR #8188, opened 2020-05-04
feat: add iconStyle prop to bottom tab bar options

gfpacheco:
Currently, if I make the bottom bar higher it'll expand the icon height and keep the label at the bottom and there's [no way to customize that](https://github.com/gfpacheco/react-navigation/blob/master/packages/bottom-tabs/src/views/BottomTabItem.tsx#L233) aside from replacing the whole tabBar component because BottomTabItem doesn't take an `iconStyle` prop.

This PR addresses that by creating a `iconStyle` option. That way I can set `flex: 0` and have both icon and label vertically centered in my tall bar.

Other options like adding bottom margin or negative top to the labels don't seem right to me.

Please let me know if that's something you'd want on your library or if I need to make any changes.

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabItem.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabItem.tsx
@@ -105,7 +105,7 @@ type Props = {
   /**
    * Style object for the icon element.
    */
-  iconStyle?: StyleProp<TextStyle>;
+  iconStyle?: StyleProp<ViewStyle>;
   /**
    * Style object for the wrapper element.
    */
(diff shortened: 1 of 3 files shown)


gfpacheco:
@satya164 pinging because newer PRs are getting reviewed while mine still wasn't. Is there anything I can do to improve it?

packages/bottom-tabs/src/views/BottomTabItem.tsx
```diff
@@ -102,6 +102,10 @@ type Props = {
    * Style object for the label element.
    */
   labelStyle?: StyleProp<TextStyle>;
+  /**
+   * Style object for the icon element.
+   */
+  iconStyle?: StyleProp<TextStyle>;
```
satya164, line 108:
```suggestion
  iconStyle?: StyleProp<ViewStyle>;
```

PR #8210, opened 2020-05-07
fix: return a promise-like from getInitialState

LinusU:
Before this change, the typings for `getInitialState` got very... _interesting_ 😁 

```ts
export default function useLinking(ref: React.RefObject<NavigationContainerRef>, { enabled, config, getStateFromPath, getPathFromState, }: LinkingOptions): {
    getInitialState: () => {
        then: (callback: (state: (Partial<Pick<NavigationState, "index" | "history">> & {
            stale?: true | undefined;
            type?: string | undefined;
            routes: (Pick<import("@react-navigation/core").Route<string>, "name" | "params"> & {
                key?: string | undefined;
                state?: import("@react-navigation/core").InitialState | undefined;
            })[];
        } & {
            state?: (Partial<Pick<NavigationState, "index" | "history">> & {
                stale?: true | undefined;
                type?: string | undefined;
                routes: (Pick<import("@react-navigation/core").Route<string>, "name" | "params"> & {
                    key?: string | undefined;
                    state?: import("@react-navigation/core").InitialState | undefined;
                })[];
            } & any) | undefined;
        }) | undefined) => void) => Promise<void>;
        catch(): any;
    };
};
```

This led to a lot of errors when trying to use it as a promise:

Screenshot 2020-05-07 at 18 19 55

Screenshot 2020-05-07 at 18 21 07

-----

This patch changes the return to simply return `Promise<ResultState | undefined>`, instead of a custom thenable.

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -95,18 +95,17 @@ export default function useLinking(
       }
     }
 
-    const then = (callback: (state: ResultState | undefined) => void) =>
-      Promise.resolve(callback(value));
-
     // Make it a thenable to keep consistent with the native impl
     const thenable = {
-      then,
+      then(onfulfilled?: (state: ResultState | undefined) => void) {
+        return Promise.resolve(onfulfilled ? onfulfilled(value) : value);
+      },
       catch() {
         return thenable;
       },
     };
 
-    return thenable;
+    return thenable as PromiseLike<ResultState | undefined>;
   }, []);
 
   const previousStateLengthRef = React.useRef<number | undefined>(undefined);
--- a/packages/native/src/useThenable.tsx
+++ b/packages/native/src/useThenable.tsx
@@ -1,8 +1,6 @@
 import * as React from 'react';
 
-type Thenable<T> = { then(cb: (result: T) => void): void };
-
-export default function useThenable<T>(create: () => Thenable<T>) {
+export default function useThenable<T>(create: () => PromiseLike<T>) {
   const [promise] = React.useState(create);
 
   // Check if our thenable is synchronous


LinusU:
Okay, it just occurred to me that you might to this intentionally to avoid scheduling an extra micro task (and thus [releasing Zalgo](https://blog.izs.me/2013/08/designing-apis-for-asynchrony)).

However, I'm not really sure how I would use that as an end user, since the `useEffect` callback will be scheduled to run after the initial mounting of the component anyways.

```js
const [initialState, setInitialState] = useState()
const { getInitialState } = useLinking(...)

useEffect(() => {
  getInitialState().then(setInitialState)
}, [getInitialState])
```

This could be avoided with `useMemo`, but then it won't be compatible with the native path anymore 🤔 

```js
const { getInitialState } = useLinking(...)

const initialState = useMemo(() => {
  let state
  getInitialState().then(s => { state = s })
  return state
}, [getInitialState])
```

The only way I can think of to make it work with both, and render the correct state on the initial render in the web version, would be something like this:

```js
const [nativeState, setNativeState] = useState()
const { getInitialState } = useLinking(...)

const webState = useMemo(() => {
  let state
  let zalgo = true
  getInitialState().then(s => {
    if (zalgo) {
      state = s
    } else {
      setNativeState(s)
    }
  })
  zalgo = false
  return state
}, [getInitialState])

const initialState = webState || nativeState
```

**Am I missing something here?** If I'm not, maybe we could provide a `useInitialState` hook that can be used something like this:

```
const { loading, initialState } = useInitialState(...)
// Possible values:
// { loading: false, initialState: undefined } -> there was no initial state
// { loading: false, initialState: ... } -> there was initial state
// { loading: true, initialState: undefined } -> we are querying the native platform for initial state
```

Would appreciate some guidance here 😅

satya164:
Thanks for the PR. Yea we use the synchronous value if possible https://github.com/react-navigation/react-navigation/blob/master/packages/native/src/useThenable.tsx

Can you update the PR to fix the types instead? Something like this should work I think:

```js
const then = <T extends any>(callback: (state: ResultState | undefined) => T) =>
  Promise.resolve(callback(value));

// Make it a thenable to keep consistent with the native impl
const thenable = {
  then,
  catch(_: any) {
    return thenable;
  },
};
```

LinusU:
@satya164 took another approach before seeing your comment, what do you think of this?

This also makes `getInitialState().then().then(console.log)` work, which didn't before.

Since `PromiseLike` is part of the standard TypeScript types it's easy to work with ☺️

satya164:
Looks good. Thanks!

PR #8222, opened 2020-05-09
fix: focus/blur events emitted with slight delay

IjzerenHein:
## Why

The lifecycle ‘focus’ and ‘blur’ events are called with a slight delay and in the next animation-frame on react-native; and after the DOM is updated on web.

This causes problems with applications which want to use the underlying animated-value upon receiving focus of a screen, such as shared-element transitions.

## How

This fix solves that by running the event-emitter code directly after the mutations have been applied using `useLayoutEffect`.

## Test plan

```jsx
class TestScreen extends React.Component {
  componentDidMount() {
    const { navigation } = this.props;
    console.log('componentDidMount');
    navigation.addListener("transitionStart", () => console.log('navigation.transitionStart'));
    navigation.addListener("focus", () => console.log('navigation.focus'));
  }

  ...
}
```

### Before

image

### After

image

## Shared element transitions

This fixes intial jank witnessed when performing a shared-element transition using react-navigation 5

### Before

> Sometimes a jank is witnessed because the navigator slide animation has already progressed, but before the focus/blur events have been emitted.

useEffect

### After

> No more janks

useLayoutEffect

### DIFF
--- a/packages/core/src/useFocusEvents.tsx
+++ b/packages/core/src/useFocusEvents.tsx
@@ -37,7 +37,7 @@ export default function useFocusEvents({ state, emitter }: Options) {
     [currentFocusedKey, emitter, navigation]
   );
 
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     const lastFocusedKey = lastFocusedKeyRef.current;
 
     lastFocusedKeyRef.current = currentFocusedKey;


IjzerenHein:
Unit tests are failing, will check them out 👍

IjzerenHein:
Hey man, thanks for the feedback and the elaborate explanation, they are very helpful.
I'm going to think about how to address this in a different manner 👍

PR #8233, opened 2020-05-11
feat: Export `StackNavigator`, `StackNavigatorProps` and `TypedStackNavigator` type defs

jtomaszewski:
Thanks to this, you can easily import and construct a type of a value returned by `createStackNavigator<T>()` expression.

We would use this because our app is quite complex, and so we extract constructing routes/params/stacks to separate files. We could use this like that:

```tsx
import React from "react";
import { TypedStackNavigator } from "@react-navigation/stack";
import { RootStackNavigatorParamList, Screens } from "../navigation";
import { ChatScreen, ChatListScreen } from "./chat";

export const getChatStackScreens = (
  Stack: TypedStackNavigator<RootStackNavigatorParamList>
): React.ReactElement => {
  return (
    <>
      <Stack.Screen
        name={Screens.ChatList}
        component={ChatListScreen}
      />
      <Stack.Screen
        name={Screens.Chat}
        component={ChatScreen}
      />
    </>
  );
};

// and then in `Stack.Navigator` just use it like `{getMoreStackScreens(Stack)}`
```

### DIFF
--- a/packages/stack/src/index.tsx
+++ b/packages/stack/src/index.tsx
@@ -6,7 +6,11 @@ import * as TransitionPresets from './TransitionConfigs/TransitionPresets';
 /**
  * Navigators
  */
-export { default as createStackNavigator } from './navigators/createStackNavigator';
+export {
+  default as createStackNavigator,
+  StackNavigator,
+  Props as StackNavigatorProps
+} from './navigators/createStackNavigator';
 
 export const Assets = [
   // eslint-disable-next-line import/no-commonjs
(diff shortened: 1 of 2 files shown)


satya164:

Regarding the PR, any types that's exported need to be exported for all navigators, not just stack. Although I'm not sure about exposing a type called `TypedStackNavigator`

jtomaszewski:
Why are you not sure about usability of exposing a type `TypedStackNavigator` ? I gave you the actual use case above ;) (What exactly do you think is wrong with it?)

satya164:
> What exactly do you think is wrong with it

The name. "typed" stack navigator doesn't sound right. Something like `StackNavigatorType` seems more intuitive.

PR #8252, opened 2020-05-13
Work around bug in getInitialURL on Expo SDK <= 37

LinusU:
This adds a workaround for the current version of Expo, that sends the wrong scheme when opening a universal link.

Issue on Expo: https://github.com/expo/expo/issues/6609
Merged PR w/ fix: https://github.com/expo/expo/pull/7508

`Linking.addEventListener('url', ...)` is not affected by this bug.

I don't know what the "policy" is on working around bugs in others code is here 😅 But it would much appreciated as it's quite inconvenient to work around in your own app, since you cannot use `<NavigationContainer linking={...} />` if you want to get between the call to `getInitialUrl` and the dispatch navigation action.

### DIFF
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -71,7 +71,7 @@ export default function useLinking(
       return undefined;
     }
 
-    const url = await (Promise.race([
+    let url = await (Promise.race([
       Linking.getInitialURL(),
       new Promise((resolve) =>
         // Timeout in 150ms if `getInitialState` doesn't resolve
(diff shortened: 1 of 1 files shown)


satya164:
Why not specify `exps://` in `linking.prefixes`? I'm not sure it makes sense to add in the library and could break unrelated apps which use `exps://` scheme.

LinusU:
True, I could work around it by something like this:

```js
prefixes: prefixes.flatMap(prefix => [prefix.replace(/^https:\/\//, 'exps://'), prefix]),
```

Maybe that is better than working around it here. But on the other hand, I guess that anyone doing universal linking on Expo will run into this. Thus I figured that it would better to solve it where anyone could benefit from it.

I guess that it could be possible that someone is using the `exps` scheme in their own app 🤔 

I don't really have a strong opinion either way, just wanted to help out the next person running into this as I just spent a few hours trying to get universal linking working in my Expo app ☺️

satya164:
Thanks! Maybe adding a note in the docs would be good enough.

LinusU:
Sure thing! 

https://github.com/react-navigation/react-navigation.github.io/pull/777

ClassyD:
Hey just wanted to find out if anyone got the workaround to work with react nav 3. or is this only going to work with react nav 5. 

Currently, my universal links open the app but does not deep-link to the actually content.
works fine on Android.

PR #8253, opened 2020-05-13
feat: merge path patterns for nested screens

satya164:
Currently, when we define `path` patterns in the linking config, they ignore the parent screen's `path` when parsing, for example, say we have this config:

```js
{
  Home: {
    path: 'home',
    screens: {
      Profile: 'u/:id',
    },
  },
}
```

If we parse the URL `/home`, we'll get this state:

```js
{
  routes: [{ name: 'Home' }],
}
```

If we parse the URL `/home/u/cal`, we'll get this state:

```js
{
  routes: [
    {
      name: 'Home',
      state: {
        routes: [
          {
            name: 'Home',
            state: {
              routes: [
                {
                  name: 'Profile',
                  params: { id: 'cal' },
                },
              ],
            },
          },
        ],
      },
    },
  ],
}
```

Note how we have 2 `Home` screens nested inside each other. This is not something people usually expect and it seems to trip up a lot of people. Something like following will be more intuitive:

```js
{
  routes: [
    {
      name: 'Home',
      state: {
        routes: [
          {
            name: 'Profile',
            params: { id: 'cal' },
          },
        ],
      },
    },
  ],
}
```

Essentially, we're treating patterns as relative to their parent rather than matching the `exact` segments. This PR changes the behavior of parsing links to treat configs like so. I hope it'll be easier to understand for people.

There is also a new option, `exact`, which brings back the olde behavior of matching the exact pattern:

```js
{
  Home: {
    path: 'home',
    screens: {
      Profile: {
        path: 'u/:id',
        exact: true,
      },
    },
  },
}
```

Which will be useful if `home` is not in the URL, e.g. `/u/cal`.

This change only affects situations where both parent and child screen configuration have a pattern defined. If the parent didn't have a pattern defined, nothing changes from previous behavior:

```js
{
  Home: {
    screens: {
      Profile: {
        path: 'u/:id',
      },
    },
  },
}
```

### DIFF
--- a/packages/core/src/getStateFromPath.tsx
+++ b/packages/core/src/getStateFromPath.tsx
@@ -13,6 +13,7 @@ type Options = {
     | string
     | {
         path?: string;
+        exact?: boolean;
         parse?: ParseConfig;
         screens?: Options;
         initialRouteName?: string;
@@ -21,10 +22,11 @@ type Options = {
 
 type RouteConfig = {
   screen: string;
-  match: RegExp | null;
+  regex?: RegExp;
+  path: string;
   pattern: string;
   routeNames: string[];
-  parse: ParseConfig | undefined;
+  parse?: ParseConfig;
 };
 
 type InitialRouteConfig = {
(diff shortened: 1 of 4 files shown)


packages/core/src/getStateFromPath.tsx
```diff
@@ -159,7 +157,46 @@ export default function getStateFromPath(
       remaining = segments.join('/');
     }
 
-    const state = createNestedStateObject(routeNames, initialRoutes, params);
+    const state = createNestedStateObject(
+      routeNames.map((name) => {
```
WoLewicki, line 161: Maybe extract this sometime. Kind of big block of code that is a param of a function.

PR #8254, opened 2020-05-13
feat: add getCurrentRoute

osdnk:
I'll add docs when we'll add getCurrentOptions

### DIFF
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -200,7 +200,10 @@ const BaseNavigationContainer = React.forwardRef(
     }, [getStateForRoute]);
 
     const getCurrentRoute = React.useCallback(() => {
-      let state: NavigationState = getRootState() as NavigationState;
+      let state = getRootState();
+      if (state === undefined) {
+        return undefined;
+      }
       while (state.routes[state.index].state !== undefined) {
         state = state.routes[state.index].state as NavigationState;
       }
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -199,6 +199,14 @@ const BaseNavigationContainer = React.forwardRef(
       return getStateForRoute('root');
     }, [getStateForRoute]);
 
+    const getCurrentRoute = React.useCallback(() => {
+      let state: NavigationState = getRootState() as NavigationState;
+      while (state.routes[state.index].state !== undefined) {
+        state = state.routes[state.index].state as NavigationState;
+      }
+      return state.routes[state.index];
+    }, [getRootState]);
+
     const emitter = useEventEmitter();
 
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,38 @@ it("doesn't throw if children is null", () => {
 
   expect(() => render(element).update(element)).not.toThrowError();
 });
+
+it('handles getCurrentOptions', () => {
```
satya164, line 1490:
```suggestion
it('returns currently focused route with getCurrentRoute', () => {
```

packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,38 @@ it("doesn't throw if children is null", () => {
...
+  const TestNavigator = (props: any): any => {
+    const { state, descriptors } = useNavigationBuilder(MockRouter, props);
+
+    return descriptors[state.routes[state.index].key].render();
+  };
+
+  const navigation = React.createRef<NavigationContainerRef>();
+
+  const second = (
+    <BaseNavigationContainer ref={navigation} onStateChange={jest.fn()}>
```
satya164, line 1500: Don't need `onStateChange`:

```suggestion
    <BaseNavigationContainer ref={navigation}>
```

packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,38 @@ it("doesn't throw if children is null", () => {
...
+      <TestNavigator>
+        <Screen name="bar" options={{ a: 'b' }}>
```
satya164, line 1502: Maybe we can add one more screen? With one screen, it will always be focused anyway

packages/core/src/__tests__/index.test.tsx
```diff
@@ -1486,3 +1486,38 @@ it("doesn't throw if children is null", () => {
...
+          {() => (
+            <TestNavigator initialRouteName="bar-a">
+              <Screen
+                name="bar-a"
+                component={() => null}
```
satya164, line 1507: I think this we print a warning in a console with this. Can you change to `children` prop?

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -199,6 +199,14 @@ const BaseNavigationContainer = React.forwardRef(
       return getStateForRoute('root');
     }, [getStateForRoute]);
 
+    const getCurrentRoute = React.useCallback(() => {
+      let state: NavigationState = getRootState() as NavigationState;
```
satya164, line 203: The state could be `undefined` if there's no navigator rendered. We should probably add a check to avoid `undefined is not an object` error.

SaeedZhiany:
> I'll add docs when we'll add getCurrentOptions

@osdnk add the link of the doc in this thread when you add it.  Thanks

PR #8281, opened 2020-05-18
refactor: memoize initializedState in useNavigationBuilder

Ashoat:
The current code can lead to an infinite render loop if a custom navigator implemented as a hook calls `setState` directly in its body.

This owes to a rather quirky edge case of how hooks work. The [recommended pattern](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops) for implementing derived data for a transitioner component implemented as a hook is to call `setState` directly within the body. When you do this, React will immediately rerender the component, *without calling any effects*.

This means that if a custom navigator calls `setState` within its body, `previousStateRef` will not get updated between the subsequent renders. This forces a new call to `router.getRehydratedState` every time, delivering a new state object. The custom navigator sees the new state and calls `setState` again to update it internal derived data, leading to the infinite loop.

As a more general comment - in the case of a `prev` ref initialized to do a comparison, it's safest to update it immediately after the comparison occurs.

### DIFF
--- a/packages/core/src/useNavigationBuilder.tsx
+++ b/packages/core/src/useNavigationBuilder.tsx
@@ -264,38 +264,36 @@ export default function useNavigationBuilder<
     getKey,
   } = React.useContext(NavigationStateContext);
 
-  const previousStateRef = React.useRef<
-    NavigationState | PartialState<NavigationState> | undefined
-  >();
-  const initializedStateRef = React.useRef<State>();
-
-  let isFirstStateInitialization = false;
-
-  if (
-    initializedStateRef.current === undefined ||
-    currentState !== previousStateRef.current
-  ) {
+  const [initializedState, isFirstStateInitialization] = React.useMemo(() => {
     // If the current state isn't initialized on first render, we initialize it
     // We also need to re-initialize it if the state passed from parent was changed (maybe due to reset)
     // Otherwise assume that the state was provided as initial state
     // So we need to rehydrate it to make it usable
     if (currentState === undefined || !isStateValid(currentState)) {
-      isFirstStateInitialization = true;
-      initializedStateRef.current = router.getInitialState({
-        routeNames,
-        routeParamList,
-      });
-    } else {
-      initializedStateRef.current = router.getRehydratedState(
-        currentState as PartialState<State>,
-        {
+      return [
+        router.getInitialState({
           routeNames,
           routeParamList,
-        }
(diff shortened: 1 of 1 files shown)


packages/core/src/useNavigationBuilder.tsx
```diff
@@ -264,49 +264,44 @@ export default function useNavigationBuilder<
...
-  React.useEffect(() => {
-    previousStateRef.current = currentState;
-  }, [currentState]);
+    // We explicitly don't include routeNames/routeParamList in the dep list
+    // below. We want to avoid forcing a new state to be calculated in cases
+    // where routeConfigs change without affecting routeNames/routeParamList.
+    // Instead, we handle changes to these in the nextState code below. Note
+    // that some changes to routeConfigs are explicitly ignored, such as changes
+    // to initialParams
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Ashoat, line 295: This is admittedly a little ugly.

As an alternative, I tried memoizing `routeNames`/`routeParamList` and including them in the dep list, but found that this lead to unnecessary recalculation of `initializedState` in the case of no-ops, for instance when `initialParams` change.

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -374,6 +369,12 @@ export default function useNavigationBuilder<
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  // We initialize this ref here to avoid a new getState getting initialized
+  // whenever initializedState changes. We want getState to have access to the
+  // latest initializedState, but don't need it to change when that happens
+  const initializedStateRef = React.useRef<State>();
+  initializedStateRef.current = initializedState;
```
Ashoat, line 376: We're admittedly setting `initializedStateRef.current` in the body here. However, this shouldn't cause problems with Concurrent Mode because:

1. Setting `initializedStateRef` is an idempotent operation. In contrast with something like a ref being incremented, we aren't worried about it happening multiple times, so long as we have the correct value at the end
2. The hook itself only ever sets `initializedStateRef`. Since we never read from it, we don't have to worry about one concurrent render interfering with another

Also worth noting that this PR doesn't introduce this ref, it just moves its declaration down. In `master` we currently have the same ref being set during render, but it's done in a concurrent-unsafe way (relying on the current value for comparison)

PR #8285, opened 2020-05-19
fix: relatively position float Header if !headerTransparent

Ashoat:
## Motivation

Right now `headerMode: float` renders an absolutely-positioned header. To offset the content appropriately, it then measures the height of the header and compensates with a margin. This approach unfortunately doesn't work well for animations.

Before             |  After
:-------------------------:|:-------------------------:
  |  

## Approach

When rendering the header absolutely we want to render it above (after, in sibling order) the content. But when rendering it relatively we want to render it first (before, in sibling order).

The margin compensation code is no longer necessary so I removed it.

## Test plan

I used the `StackHeaderCustomization` example to make sure transitions between `headerTransparent` and `!headerTransparent` looked good. I added a custom (taller) header to test if height transitions looked good, and toggled `headerShown` to make sure that transitioned well too.

Would be open to any other suggestions of things to test!

### DIFF
--- a/packages/stack/src/views/Stack/CardContainer.tsx
+++ b/packages/stack/src/views/Stack/CardContainer.tsx
@@ -55,6 +55,7 @@ type Props = TransitionPreset & {
   headerMode: StackHeaderMode;
   headerShown?: boolean;
   headerTransparent?: boolean;
+  isFloatHeaderAbsolute: boolean;
   headerHeight: number;
   onHeaderHeightChange: (props: {
     route: Route<string>;
@@ -85,6 +86,7 @@ function CardContainer({
   headerShown,
   headerStyleInterpolator,
   headerTransparent,
+  isFloatHeaderAbsolute,
   headerHeight,
   onHeaderHeightChange,
   index,
@@ -188,7 +190,7 @@ function CardContainer({
       pointerEvents={active ? 'box-none' : pointerEvents}
       pageOverflowEnabled={headerMode === 'screen' && mode === 'card'}
       containerStyle={
-        headerMode === 'float' && !headerTransparent && headerShown !== false
+        isFloatHeaderAbsolute && !headerTransparent && headerShown !== false
           ? { marginTop: headerHeight }
           : null
       }
--- a/packages/stack/src/views/Stack/CardStack.tsx
+++ b/packages/stack/src/views/Stack/CardStack.tsx
@@ -335,6 +335,21 @@ export default class CardStack extends React.Component<Props, State> {
     return state.routes[state.index];
   };
 
+  private getSomeFloatHeaderNeedsAbsolutePositioning = () => {
+    if (this.props.headerMode !== 'float') {
+      return false;
+    }
+    return this.state.scenes.slice(-2).some((scene) => {
+      const { descriptor } = scene;
+      const options = descriptor ? descriptor.options : {};
+      const { headerTransparent, headerShown } = options;
+      if (headerTransparent || headerShown === false) {
+        return true;
+      }
+      return false;
+    });
+  };
+
   render() {
     const {
       mode,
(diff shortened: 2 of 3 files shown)


packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -299,7 +300,7 @@ export default class CardStack extends React.Component<Props, State> {
           props.insets,
           state.descriptors,
           layout,
-          {}
+          state.headerHeights
```
Ashoat, line 303: I'm not sure why this was previously an empty object. It had the effect of reseting `headerHeights` to be based on `getDefaultHeaderHeight` every time the scene container calls `onLayout`, which led to jittery behavior. Would be helpful to know the background for this decision, curious if there's an edge case I missed in testing

packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -334,6 +335,55 @@ export default class CardStack extends React.Component<Props, State> {
     return state.routes[state.index];
   };
 
+  private getSomeFloatHeaderNeedsAbsolutePositioning() {
```
satya164, line 338: Change to a class property to keep consistent with other methods

packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -334,6 +335,55 @@ export default class CardStack extends React.Component<Props, State> {
...
+        const prevRoute = this.props.getPreviousRoute({ route });
+        if (prevRoute) {
+          visibleKeys.add(prevRoute.key);
+        }
+      }
+    }
+
+    for (let scene of this.state.scenes) {
+      const { route, descriptor } = scene;
+      if (!visibleKeys.has(route.key)) {
```
satya164, line 375: Maybe it's simpler to just check if the route is one of the last 2 scenes since the last 2 are always the screens that can be visible at a time. Or check last 2 scenes when closingRouteKeys/openingRouteKeys are non empty, otherwise only the last one. Though the simpler the better. Currently this method looks a bit complex.

packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -383,8 +434,31 @@ export default class CardStack extends React.Component<Props, State> {
     // For modals, usually we want the screen underneath to be visible, so also disable it there
     const isScreensEnabled = Platform.OS !== 'ios' && mode !== 'modal';
 
+    let floatingHeader;
+    if (headerMode === 'float') {
+      floatingHeader = renderHeader({
```
satya164, line 439: I think this needs to be wrapped in a `React.Fragment` with `key` to preserve local state when being re-arranged.

packages/stack/src/views/Stack/CardStack.tsx
satya164: Looks like a mistake

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -55,6 +55,7 @@ type Props = TransitionPreset & {
   headerMode: StackHeaderMode;
   headerShown?: boolean;
   headerTransparent?: boolean;
+  absoluteFloatHeader: boolean;
```
satya164, line 58: Maybe something like `isFloatHeaderAbsolute`. `absoluteFloatHeader` feels like it contains the absolute float header.

Ashoat:
Sorry for the delay here, have some pressing stuff I need to get to this week but will hopefully be able to update this PR later in the week!

packages/stack/src/views/Stack/CardStack.tsx
Ashoat: Good catch!

packages/stack/src/views/Stack/CardStack.tsx
Ashoat: You're right, this is way more complex than it needs to be. I did some testing and the last two scenes should be all we need to check.

PR #8286, opened 2020-05-19
feat: animate changes to tabBarVisible in BottomTabBar

Ashoat:
## Motivation

Some designs call for custom keyboard inputs, or other bottom-aligned views meant overlap over the keyboard. Right now the best option on Android for this case is to set `tabBarVisible`. However changes to `tabBarVisible` doesn't get animated currently, which makes the custom-keyboard-open experience a bit more jarring than the native-keyboard-open one.

## Approach

I basically cribbed the `Animated.Value` we were using for `keyboardHidesTabBar` and made it depend on both. Note that the offset height depends on which of the two uses cases we're dealing with, which is explained in the code.

## Test plan

I played around with the `BottomTabs` example, setting certain screens to `tabBarVisible: true` and making sure it animated.

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -53,51 +53,64 @@ export default function BottomTabBar({
   const { colors } = useTheme();
   const buildLink = useLinkBuilder();
 
-  const [dimensions, setDimensions] = React.useState(() => {
-    const { height = 0, width = 0 } = Dimensions.get('window');
+  const focusedRoute = state.routes[state.index];
+  const focusedDescriptor = descriptors[focusedRoute.key];
+  const focusedOptions = focusedDescriptor.options;
 
-    return { height, width };
-  });
-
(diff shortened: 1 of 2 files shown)


packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -53,51 +53,64 @@ export default function BottomTabBar({
...
+  const prevShouldShowTabBarRef = React.useRef<boolean | undefined>();
 
-  const { routes } = state;
+  const [tabBarPartiallyHidden, setTabBarPartiallyHidden] = React.useState(
+    !shouldShowTabBar
+  );
+
+  const visibleRef = React.useRef<Animated.Value | undefined>();
+  if (!visibleRef.current) {
+    visibleRef.current = new Animated.Value(shouldShowTabBar ? 1 : 0);
```
Ashoat, line 73: This was previously initialized via state. I should probably change it back, since I don't have a good understanding of how `Animated.Value` behaves in concurrent mode

PR #8297, opened 2020-05-20
feat: add a `ServerContainer` component for SSR

satya164:
When doing SSR, the app needs to be aware of request URL to render correct navigation state.
The `ServerContainer` component lets us pass the `location` object to use for SSR.
The shape of the `location` object matches the `location` object in the browser.

Usage:

```js
ReactDOM.renderToString(
  <ServerContainer location={{ pathname: req.path, search: req.search }}>
    <App />
  </ServerContainer>
);
```

Updated example: https://github.com/react-navigation/react-navigation/pull/8298

### DIFF
--- a/packages/native/src/ServerContainer.tsx
+++ b/packages/native/src/ServerContainer.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import ServerContext, { ServerContextType } from './ServerContext';
+
+type Props = ServerContextType & {
+  children: React.ReactNode;
+};
+
+/**
+ * Container component for server rendering.
+ *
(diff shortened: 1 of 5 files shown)


ericvicenti:
This ServerContainer API is great, super easy to understand. 

Other features that should be considered when designing this API: the title and other page metadata (description, photo) is set via navigation options, so the container will need an interface to access the options of the selected route. Also we should find a way to support redirects, in the case your app wants to send a http 302, as well as 404 "not found" errors.

Implementation makes sense, but I wonder if there is a way to warn people against using this on the client.

PR #8312, opened 2020-05-22
fix: wrong use of `type` keyword for exported stack types

tomsotte:
When exporting types in TypeScript the keyword `type` should not be used.


tomsotte:
I didn't know TypeScript 3.8 brought flow like export of types only. Sorry for the bother, I was still using 3.7.

PR #8314, opened 2020-05-23
fix: allow HeaderBackground's subViews to be touchable

benevbright:
When you're using the following options on `Stack`, the touch event goes to the behind of the Header and we can't really solve this problem in given `headerBackground` component level.
```
 options={{
          headerTransparent: true,
          headerBackground: 
```

### Problem
May-23-2020 09-31-01


satya164:
Reverted since this broke header on Android
image

benevbright:
Oh... thanks. let me investigate.

PR #8317, opened 2020-05-23
fix: allow HeaderBackground's subViews to be touchable

benevbright:
This was merged https://github.com/react-navigation/react-navigation/pull/8314 but reverted because it made Android header disappear. 

After some tests, the problem was `elevation: 4` on `HeaderBackground` component, which is likely making a problem when you use `absolute + elevation + pointerEvent: everything else except "none"`. If I remove `elevation: 4`, the header is ok. 
It seems to be an old bug of RN Android. I found several similar issues.
- https://github.com/facebook/react-native/issues/19894
- https://github.com/facebook/react-native/issues/27548 
- https://react-native.canny.io/feature-requests/p/zindex--positionabsolute-breaks-with-dynamic-components-on-android
- https://github.com/react-native-community/react-native-linear-gradient/issues/370

#### Solution
And adding `zIndex: 0` without deleting `elevation: 4` did a trick.

### Original issue
When you're using the following options on `Stack`, the touch event goes to the behind of the Header and we can't really solve this problem in given `headerBackground` component level.
```
 options={{
          headerTransparent: true,
          headerBackground: 
```

### Before
before

May-23-2020 09-31-01

### After
after

### DIFF
--- a/packages/stack/src/views/Header/HeaderSegment.tsx
+++ b/packages/stack/src/views/Header/HeaderSegment.tsx
@@ -313,7 +313,7 @@ export default class HeaderSegment extends React.Component<Props, State> {
       <React.Fragment>
         <Animated.View
           pointerEvents="box-none"
-          style={[StyleSheet.absoluteFill, backgroundStyle, { zIndex: 0 }]}
+          style={[StyleSheet.absoluteFill, { zIndex: 0 }, backgroundStyle]}
         >
           {headerBackground ? (
             headerBackground({ style: safeStyles })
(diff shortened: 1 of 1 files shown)


packages/stack/src/views/Header/HeaderSegment.tsx
```diff
@@ -312,8 +312,8 @@ export default class HeaderSegment extends React.Component<Props, State> {
     return (
       <React.Fragment>
         <Animated.View
-          pointerEvents="none"
-          style={[StyleSheet.absoluteFill, backgroundStyle]}
+          pointerEvents="box-none"
+          style={[StyleSheet.absoluteFill, backgroundStyle, { zIndex: 0 }]}
```
satya164, line 316:
```suggestion
          style={[StyleSheet.absoluteFill, { zIndex: 0 }, backgroundStyle]}
```

PR #8329, opened 2020-05-25
fix(react-native-web): body scrolling if you are using <ScrollView>

milesibastos:
Current behaviour:
Screen Recording 2020-05-25 at 18 58 27

[Expected behaviour](https://youtu.be/3YJUATCgBXE)
Screen Recording 2020-05-25 at 18 58 27 2


satya164:
This is intentional. Vertical space is important on mobile devices. By not hiding the header and address bar, it takes a lot of space with less space for content. You can manually set height in your apps if you don't want it.

PR #8357, opened 2020-05-29
fix: typo on drawerPosition default props

diegolmello:
Fix a minor typo from `drawerPostion` to `drawerPosition` :)

Fix https://github.com/react-navigation/react-navigation/issues/8358

### DIFF
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -238,7 +238,6 @@ export default function DrawerView({
               renderDrawerContent={renderNavigationView}
               renderSceneContent={renderContent}
               keyboardDismissMode={keyboardDismissMode}
-              drawerPostion={drawerPosition}
               dimensions={dimensions}
             />
           </DrawerOpenContext.Provider>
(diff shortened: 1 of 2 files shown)


tarbundiyahitesh21:
> Fix a minor typo from `drawerPostion` to `drawerPosition` :)

thanks, now it's working

packages/drawer/src/views/DrawerView.tsx
```diff
@@ -238,7 +238,6 @@ export default function DrawerView({
               renderDrawerContent={renderNavigationView}
               renderSceneContent={renderContent}
               keyboardDismissMode={keyboardDismissMode}
-              drawerPostion={drawerPosition}
```
satya164, line 241: Why did you remove it from here?

diegolmello: @satya164 Because it's also a typo and it's duplicated.
[Here (correct)](https://github.com/react-navigation/react-navigation/blob/master/packages/drawer/src/views/DrawerView.tsx#L211) and [here (incorrect and removed)](https://github.com/react-navigation/react-navigation/blob/master/packages/drawer/src/views/DrawerView.tsx#L241)

PR #8365, opened 2020-06-01
fix: don't use deprecated currentlyFocusedField

rkdqudtjs1:
currentlyFocusedField => currentlyFocusedInput

### DIFF
--- a/packages/stack/src/views/KeyboardManager.tsx
+++ b/packages/stack/src/views/KeyboardManager.tsx
@@ -35,7 +35,7 @@ export default class KeyboardManager extends React.Component<Props> {
 
     this.clearKeyboardTimeout();
 
-    // currentlyFocusedField => currentlyFocusedInput
+
     const input = TextInput.State.currentlyFocusedInput(); 
 
     // When a page change begins, blur the currently focused input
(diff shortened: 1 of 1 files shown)


packages/stack/src/views/KeyboardManager.tsx
```diff
@@ -35,7 +35,8 @@ export default class KeyboardManager extends React.Component<Props> {
 
     this.clearKeyboardTimeout();
 
-    const input = TextInput.State.currentlyFocusedField();
+    // currentlyFocusedField => currentlyFocusedInput
```
satya164, line 38: Can you remove this comment?

rkdqudtjs1: I removed this comment.

packages/stack/src/views/KeyboardManager.tsx
```diff
@@ -35,7 +35,8 @@ export default class KeyboardManager extends React.Component<Props> {
 
     this.clearKeyboardTimeout();
 
-    const input = TextInput.State.currentlyFocusedField();
+
+    const input = TextInput.State.currentlyFocusedInput();
```
satya164, line 39:
```suggestion
    const input = TextInput.State.currentlyFocusedInput
      ? TextInput.State.currentlyFocusedInput()
      : TextInput.State.currentlyFocusedField();
```

ahilles107:
It will be backported to 4.x branch?

PR #8367, opened 2020-06-01
refactor: rework history stack integration

### LINKED ISSUE #8230: URL not being updated correctly when navigating through
**Current Behavior**

I am using a drawer navigator. Each drawer item uses its own stack navigator.
```
const Drawer = createDrawerNavigator(); 
const Stack1 = createStackNavigator(); 
const Stack2 = createStackNavigator(); 
const Stack3 = createStackNavigator(); 
```

This is the navigation container:

```
<NavigationContainer linking={LinkingConfiguration}>
  <Drawer.Navigator initialRouteName="Home">
    <Drawer.Screen name="Home" component={
      () => <Stack1.Navigator>
        <Stack1.Screen name="Home" component={HomeScreen}/>
      </Stack1.Navigator>
    }/>
    <Drawer.Screen name="Links" component={
      () => <Stack2.Navigator>
        <Stack2.Screen name="Links" component={LinksScreen}/>
      </Stack2.Navigator>
    }/>
    <Drawer.Screen name="New" component={
      () => <Stack3.Navigator>
        <Stack3.Screen name="New" component={NewScreen}/>
      </Stack3.Navigator>
    }/>
  </Drawer.Navigator>
</NavigationContainer>
```

Each of the 3 screens is looking like this:
```
export default function HomeScreen({navigation}) {
  return (
    <ScrollView style={styles.container} contentContainerStyle={styles.contentContainer}>
      <Text>HOME SCREEN</Text>
      <Button onPress={() => navigation.openDrawer()} title="open drawer" />
    </ScrollView>
  );
}
```

```
export default function LinksScreen({navigation}) {
  return (
    <ScrollView style={styles.container} contentContainerStyle={styles.contentContainer}>
      <Text>LINKS SCREEN</Text>
      <Button onPress={() => navigation.openDrawer()} title="open drawer" />
    </ScrollView>
  );
}
```

LinkConfiguration.js
```
import * as Linking from 'expo-linking';

export default {
  prefixes: [Linking.makeUrl('/')],
};
```

I am opening the drawer, and the URL is updating correctly.
When I am navigating to a screen that I already visited, the URL is not getting updated correctly.

**Expected Behavior**

When navigating to a DrawerItem, the URL should update accordingly.

**How to reproduce**

1. open app (home screen). The URL should be /Home/Home
2. programmatically open drawer (navigation.openDrawer())
3. click another drawer item. (Links)The URL is now /Links/Links
4. programmatically open drawer (navigation.openDrawer())
5. click Home drawer item. **The URL is now /Links/Links** instead of /Home/Home

1-nav
2-nav

Note that when I open the drawer, the URL gets updated correctly and the Home item is selected inside the drawer. However if I toggle out the drawer, the URL is wrong again.

3-nav

**Your Environment**

| iOS or Android                 | web
| @react-navigation/native       |^5.2.1
| @react-navigation/drawer       |^5.7.2
| react-native-reanimated        |^1.8.0
| react-native-gesture-handler   |~1.6.0
| react-native-safe-area-context |0.7.3
| react-native-screens           |~2.2.0
| react-native                   |https://github.com/expo/react-native/archive/sdk-37.0.1.tar.gz
| expo                           |~37.0.9
| node                           |12.0.6
| npm or yarn                    |npm

### LINKED ISSUE #8284: web url update wrong when navigate between tab and stack
**Current Behavior**

- when navigate between tab.screen and statck.screen with linking, the web url goes wrong
-  image

**Expected Behavior**

- Detail page's url should be 'detail' not to be cart
- I make a repo at [https://snack.expo.io/@bestv2/linking-web](https://snack.expo.io/@bestv2/linking-web)

**How to reproduce**
1. click the bottom Cart, url update to 'cart'
2. click the bottom Me, url update to 'me'
3. click the bottom Home, url update to 'home'
4. click the 'navigate to stack detail page' btn, url update to 'cart', it should be 'detail'

**Your Environment**

| software                      | version |
| ----------------------------- | ------- |
| iOS or Android                |web
| @react-navigation/native      |5.4.0
| @react-navigation/bottom-tabs |5.4.5
| @react-navigation/stack |5.3.7
| react-native-screens          |2.7.0
| react-native                  |0.62

### LINKED ISSUE #8344: Web, URL differs from what we expect
Hey :wave:

**Current Behavior**

We are using react navigation on the web with the Link component and sometimes our URL differs from what we expect.
We have a bottom bar with three screens ("Home, Profile, Contact") and a stack navigator who contains our bottom bar and a screen call "QRScreen".
We made some tests by navigating on each screen:
 - Home -> QRScreen => my URL is correct (/qr-screen)
 - Home -> Profile -> QRScreen => my URL is correct (/qr-screen)
 - Home -> Profile -> Contact -> QRScreen => my URL isn't correct (/profile)

**Expected Behavior**

- Home -> Profile -> Contact -> QRScreen
By navigating on these screens our url should be /qr-screen

**How to reproduce**

I create an example in this [repository](https://github.com/M4gie/nestedNavigators)

1. Navigate on these screens: Home -> Profile -> Contact -> QRScreen
2. Check the URL 

**Your Environment**

| iOS or Android                 |
| @react-navigation/native       |5.4.1
| @react-navigation/bottom-tabs       |5.4.6
| @react-navigation/stack       |5.3.8
| react-native                   |
| expo                           |37.0.3

**Not sure if it's a bug or if we missed something in your documentation**

satya164:
The PR reworks history integration to better integrate with browser's history stack and supports nested navigation more reliably:

- On each navigation, save the navigation in memory and use it to reset the state when user presses back/forward
- Improve heuristic to determine if we should do a push, replace or back

This closes #8230, closes #8284 and closes #8344

### DIFF
--- a/packages/native/src/__mocks__/window.tsx
+++ b/packages/native/src/__mocks__/window.tsx
@@ -1,0 +1,69 @@
+const location = new URL('', 'http://example.com');
+
+let listeners: (() => void)[] = [];
+let entries = [{ state: null, href: location.href }];
+let index = 0;
+
+let currentState: any = null;
+
+const history = {
+  get state() {
+    return currentState;
+  },
+
+  pushState(state: any, _: string, path: string) {
+    Object.assign(location, new URL(path, location.origin));
+
+    currentState = state;
+    entries = entries.slice(0, index + 1);
+    entries.push({ state, href: location.href });
+    index = entries.length - 1;
+  },
+
+  replaceState(state: any, _: string, path: string) {
+    Object.assign(location, new URL(path, location.origin));
+
+    currentState = state;
+    entries[index] = { state, href: location.href };
+  },
+
+  go(n: number) {
+    setTimeout(() => {
+      if (
+        (n > 0 && n < entries.length - index) ||
+        (n < 0 && Math.abs(n) <= index)
+      ) {
+        index += n;
+        Object.assign(location, new URL(entries[index].href));
+        listeners.forEach((cb) => cb);
+      }
+    }, 0);
+  },
+
+  back() {
+    this.go(-1);
+  },
+
+  forward() {
+    this.go(1);
+  },
+};
+
+const addEventListener = (type: 'popstate', listener: () => void) => {
+  if (type === 'popstate') {
+    listeners.push(listener);
+  }
+};
+
+const removeEventListener = (type: 'popstate', listener: () => void) => {
+  if (type === 'popstate') {
+    listeners = listeners.filter((cb) => cb !== listener);
+  }
+};
+
+export default {
+  location,
+  history,
+  addEventListener,
+  removeEventListener,
+};
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -23,6 +23,25 @@ const createMemoryHistory = () => {
 
   const items: HistoryRecord[] = [];
 
+  const getDelta = (count: number) => {
+    const times = window.history.state?.times ?? 1;
+
+    let delta = count - window.history.state?.times ?? 1;
+
+    if (count > times) {
+      let previous = window.history.state?.previous;
+
+      while (previous) {
+        delta -= previous.times ?? 1;
+        previous = previous.previous;
+      }
+    }
+
+    // We need to keep at least one item in the history
+    // Otherwise we'll exit the page
+    return Math.min(delta + 1, index - 1);
+  };
+
   return {
     get latest() {
       return index;
@@ -40,44 +59,62 @@ const createMemoryHistory = () => {
       return items[index].state;
     },
 
-    push(path: string, key: string, state: NavigationState) {
+    find({ path, key }: { path: string; key: string }) {
+      return items.findIndex((item) => item.path === path && item.key === key);
+    },
+
+    push({
+      path,
+      key,
+      state,
+      times = 1,
+    }: {
+      path: string;
+      key: string;
+      state: NavigationState;
+      times?: number;
+    }) {
       items.splice(index + 1);
       items.push({ path, state, key });
       index = items.length - 1;
 
-      window.history.pushState({ index }, '', path);
-    },
-
-    replace(path: string, key: string, state: NavigationState) {
-      items[index] = { path, state, key };
-      window.history.replaceState({ index }, '', path);
-    },
-
-    back(n: number) {
-      index -= n;
-      window.history.go(-n);
-    },
-
-    go(path: string, key: string) {
-      const nextIndex = items.findIndex(
-        (item) => item.path === path && item.key === key
+      window.history.pushState(
+        { index, times, previous: window.history.state },
+        '',
+        path
       );
+    },
 
-      if (nextIndex === -1) {
-        return false;
-      }
+    replace({
+      path,
+      key,
+      state,
+    }: {
+      path: string;
+      key: string;
+      state: NavigationState;
+    }) {
+      items[index] = { path, state, key };
+      window.history.replaceState(
+        {
+          index,
+          times: window.history.state?.times ?? 1,
+          previous: window.history.state?.previous,
+        },
+        '',
+        path
+      );
+    },
 
-      const delta = nextIndex - (window.history.state?.index ?? 0);
+    go(n: number) {
+      const delta = n < 0 ? getDelta(-n) : n;
 
       if (delta === 0) {
-        return true;
+        return;
       }
 
-      index = nextIndex;
-
-      window.history.go(delta);
-
-      return true;
+      index -= delta;
+      window.history.go(-delta);
     },
   };
 };
@@ -192,19 +229,6 @@ export default function useLinking(
 
   const previousStateLengthRef = React.useRef<number | undefined>(undefined);
 
-  const pendingIndexChangeRef = React.useRef<number | undefined>();
-  const pendingStateUpdateRef = React.useRef<boolean>(false);
-  const pendingStateMultiUpdateRef = React.useRef<boolean>(false);
-
-  // If we're navigating ahead >1, we're not restoring whole state,
-  // but just navigate to the selected route not caring about previous routes
-  // therefore if we need to go back, we need to pop screen and navigate to the new one
-  // Possibly, we will need to reuse the same mechanism.
-  // E.g. if we went ahead+4 (numberOfIndicesAhead = 3), and back-2,
-  // actually we need to pop the screen we navigated
-  // and navigate again, setting numberOfIndicesAhead to 1.
-  const numberOfIndicesAhead = React.useRef(0);
-
   React.useEffect(() => {
     const onPopState = () => {
       const navigation = ref.current;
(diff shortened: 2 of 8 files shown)


packages/native/src/__mocks__/window.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  replaceState(state: any, _: string, path: string) {
+    Object.assign(location, new URL(path, location.origin));
+
+    currentState = state;
+    entries[index] = { state, href: location.href };
+  },
+
+  go(n: number) {
+    setTimeout(() => {
```
osdnk, line 31: Maybe setImmediate?

packages/native/src/useLinking.tsx
```diff
@@ -6,37 +6,225 @@ import {
...
+        if (item.path === path) {
+          return i;
+        }
+      }
+
+      return -1;
+    },
+
+    push({ path, state }: { path: string; state: NavigationState }) {
+      const id = nanoid();
```
osdnk, line 64: can't we just use `counter++.toString()`

packages/native/src/useLinking.tsx
```diff
@@ -6,37 +6,225 @@ import {
...
+
+      // We pass empty string for title because it's ignored in all browsers except safari
+      // We don't store state object in history.state because:
+      // - browsers have limits on how big it can be, and we don't control the size
+      // - while not recommended, there could be non-serializable data in state
+      window.history.pushState({ id }, '', path);
+    },
+
+    replace({ path, state }: { path: string; state: NavigationState }) {
+      const id = window.history.state?.id ?? nanoid();
```
osdnk, line 81: how can it happen that `window.history.state` is `undefined` and call `replace`?

packages/native/src/useLinking.tsx
```diff
@@ -6,37 +6,225 @@ import {
...
-  pushState(state: HistoryState, title: string, url: string): void;
-  replaceState(state: HistoryState, title: string, url: string): void;
+      return new Promise((resolve) => {
+        pending = true;
+
+        const done = () => {
+          pending = false;
+
+          window.removeEventListener('popstate', done);
+          resolve();
```
osdnk, line 122: can we move it out somewhere? It's hacky .-.

packages/native/src/useLinking.tsx
```diff
@@ -116,203 +306,143 @@ export default function useLinking(
...
+        state
+      );
 
       if (
-        pendingStateUpdateRef.current &&
-        location.pathname + location.search === path
+        previousFocusedState &&
+        focusedState &&
+        // We should only handle push/pop if path changed from what was in last `popstate`
+        // Otherwise it's likely a change triggered by `popstate`
```
osdnk, line 398: can it happen?

packages/native/src/useLinking.tsx
```diff
@@ -6,37 +6,225 @@ import {
...
+): [T | undefined, T | undefined] => {
+  if (a === undefined || b === undefined || a.key !== b.key) {
+    return [undefined, undefined];
+  }
 
-  if (state.history) {
-    length = state.history.length;
-  } else {
-    length = state.index + 1;
+  const aHistoryLength = a.history ? a.history.length : a.routes.length;
```
osdnk, line 168: Can you add some comments here? Why it's like this?

packages/native/src/useLinking.tsx
```diff
@@ -116,203 +306,143 @@ export default function useLinking(
...
+          // If history length is decreased, i.e. entries were removed, we want to go back
+
+          const nextIndex = history.backIndex({ path });
+          const currentIndex = history.index;
+
+          if (nextIndex !== -1 && nextIndex < currentIndex) {
+            // An existing entry for this path exists and it's less than current index, go back to that
+            await history.go(nextIndex - currentIndex);
+          } else {
+            // We couldn't find an existing entry to go back to, so we'll go back by the delta
```
osdnk, line 423: can you explain why it can happen? Maybe we can just `go` in this case?

packages/native/src/useLinking.tsx
```diff
@@ -70,6 +258,8 @@ export default function useLinking(
     };
   });
 
+  const [history] = React.useState(createMemoryHistory);
```
osdnk, line 261: You mean sth like `useMemoOne` here? Why not `x = useRef(); if x.current === null` etc.?

packages/native/src/__mocks__/window.tsx
satya164: Not sure if the operation happens at the end of event loop so `setImmediate` might be incorrect. It's not in browsers so I can't test :<

Anyway, it's just for mocking so shouldn't matter.

packages/native/src/useLinking.tsx
satya164: A counter wouldn't work because our in-memory value would reset after reloading/restarting browser etc. but the value in state object would refer to a positive counter which would incorrectly match with existing entry.

packages/native/src/useLinking.tsx
satya164: The `state` won't exist when the user initially visits the page. It could also get replaced if some other user code is calling `history.replaceState`.  So we always need to assume that it can be `null`.

packages/native/src/useLinking.tsx
satya164: where would you move it? it's hacky but not sure what's a better alternative

packages/native/src/useLinking.tsx
satya164: `popstate` calls `resetRoot` (or dispatches navigate action), so every `popstate` will trigger a state change event

packages/native/src/useLinking.tsx
satya164: Tab and drawer will have `history`, but stack will have `routes`. Will add a comment.

packages/native/src/useLinking.tsx
satya164: The user could change the screen configuration or linking configuration which would result in different path, or user has specified some initial screens with existing history so we haven't actually visited this `path` etc. I think going by delta would be more correct here.

packages/native/src/useLinking.tsx
satya164: Not really a memo since I only want initialize this once. Not using a ref because it's just extra code.

packages/native/src/useLinking.tsx
martindido: Hey, sorry to jump in here completely out of the blue, but I've been following this PR since the very beginning because I was actually tracking down a bug I experienced when using linking with nested navigators on web that originates exactly around this part of the code. You guys might already know about it, but I figured there's no harm on pointing it out, so here it is.

Lets say you have a nested navigation tree that looks sort of like this (where N=navigator and S=screen):

N1
|- N2
|     |- S3
|     |- S4
|     |- S5
|     |- S6
| 
|- S1
|- S2

If, by using linking, you directly access any screen on N2, let's say S3 directly by url. Then you navigate through a couple of screens on the same level, let's say first to S4 and then to S6 (at which point your history length becomes at least 2). If you then try to navigate to S1, this piece of the code was actually comparing the history length (which could potentially be any number) with the length of the newly created state as returned from "getStateFromPath" for S1 via "getStateLength". Since the state for S1 was new, it had no history, so its length was always 1 (the depth of its nesting). Therefore, your previous history length was always greater than your current state length, causing this to always trigger a "history.go(-delta);" which made the url point to a previously visited url (S4 following my example), instead of S1's url as it should.

Since the navigation itself is independent of this piece of core, the site continues to work properly, except for the fat that urls from that point onwards are completely miss-matched with the screens being rendered.

Sorry for the long message but I wanted to make it as clear as possible. If you where already aware of this issue, feel free to completely disregard this message! Or if you need more info on how to reproduce it or whatever, don't hesitate to ask me. It caused me more than one headache last week, and I was very happy to see work being done on this file.

Thanks!

satya164: @martindido sounds like the linked issues

Issue #7944, opened 2020-04-03
Drawer not opening on swipe after once tapped outside to close.


adiwon9555:
**Current Behavior**

In React navigation 5

Drawer properly opens and closes on event openDrawer() and closeDrawer() events, but when facing issues while swipping to open drawer.

Steps to Reproduce:-
1)Open Drawer once
2)Tap outside Drawer and close.
3)Swipe to open drawer. ( Facing problem here, not able to open)

bijayalaxmi19:
dreawer is not close after Touching outside of drawer.

raajnadar:
Give the versions! A repro! Read the docs.

https://reactnavigation.org/docs/drawer-navigator#swipeenabled

https://reactnavigation.org/docs/drawer-navigator#gestureenabled

Nehal-29:
I'm also facing the same issue
I open the drawer then close it by tapping outside but DrawerActions.closeDrawer() is not dispatched automatically
since the state isDrawerOpen is not updated hence on tapping the icon again to open the drawer doesn't work
@brentvatne @ericvicenti

raajnadar:
**@Nehal-29 please don't tag the maintainers.**

> I'm also facing the same issue

We don't have a repro to test the issue so how can we help to solve this issue? If you can create a snack or repo with the issue we can find what is going wrong.

Nehal-29:
Hey @adiwon9555 
were you able to resolve this ?

adiwon9555:
I guess, there is some gesture issue with RN5 drawer. In my project i just removed the functionality of opening drawer via gesture.

satya164:
Closing since you didn't provide any repro. We need a minimal repro to be able to debug issues. Feel free to open a new issue if you have a repro.

rgoldiez:
@adiwon9555 @satya164 - I just submitted https://github.com/react-navigation/react-navigation/pull/8369 to fix this issue. Hopefully it gets reviewed soon!

PR #8369, opened 2020-06-02
fix: tapping overlay should update drawer state properly

rgoldiez:
There is an issue when tapping on the drawer overlay that is reported in https://github.com/react-navigation/react-navigation/issues/7944 and we've experience in our app.  

To reproduce it, you can use any non-web app that uses the drawer.  The issue is that tapping on the overlay (outside of the drawer) will indeed close the drawer.  But subsequent swipes to _open_ the drawer will immediately close the drawer.  I believe this is because the internal drawer state is not being updated properly.  

This PR simply calls `toggleDrawer(false)` to use already available methods to close the drawer.  When using the method in this PR, internal state is properly updated and the user can swipe/tap/toggle the drawer and experience the correct behavior.

### DIFF
--- a/packages/drawer/src/views/Drawer.tsx
+++ b/packages/drawer/src/views/Drawer.tsx
@@ -516,7 +516,7 @@ export default class DrawerView extends React.Component<Props> {
         oldState: (s: Animated.Value<number>) =>
           cond(
             eq(s, GestureState.ACTIVE),
-            set(this.manuallyTriggerSpring, TRUE)
+            call([], () => this.toggleDrawer(false))
           ),
       },
     },


packages/drawer/src/views/Drawer.tsx
```diff
@@ -516,7 +516,7 @@ export default class DrawerView extends React.Component<Props> {
         oldState: (s: Animated.Value<number>) =>
           cond(
             eq(s, GestureState.ACTIVE),
-            set(this.manuallyTriggerSpring, TRUE)
+            call([], () => this.toggleDrawer(false))
```
satya164, line 519: This will trigger callback to JS which can make the drawer lag if JS thread is busy. We need to figure out a different solution.

rgoldiez: A lag when tapping?  I'm not sure a lag when tapping would be noticeable (vs a lag when releasing a drag gesture).

rgoldiez:
Sure, I can put together a minimal example.

packages/drawer/src/views/Drawer.tsx
satya164: it might be noticeable on lower end devices, similar to how touchables don't feel very responsive

theZulqarnain:
facing the same problem, any update on it

rgoldiez:
@theZulqarnain - I haven't had the time to put together a demo, but I have applied this PR to our internal app with patch package to get around this issue.

satya164:

PR #8378, opened 2020-06-03
Fix type of headerBackTitle in stack navigation

eduard-malakhov:
As per documentation, `headerBackTitle` treats `null` as a special value to disable the title, however `null` is not assignable to `headerBackTitle` due to its type.

### DIFF
--- a/packages/stack/src/types.tsx
+++ b/packages/stack/src/types.tsx
@@ -135,7 +135,7 @@ export type StackHeaderOptions = {
   /**
    * Title string used by the back button on iOS, or `null` to disable label. Defaults to the previous scene's `headerTitle`.
    */
-  headerBackTitle?: string;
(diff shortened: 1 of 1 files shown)


satya164:
> As per documentation, headerBackTitle treats null

Not sure where you found it in documentation, but that's not correct. Back title should be disabled with `headerBackTitleVisible: false`

eduard-malakhov:
@satya164, the comment just above the lines that I've proposed to change says "Title string used by the back button on iOS, or `null` to disable label. Defaults to the previous scene's `headerTitle`." If that is not true, then I think we need to fix the comment. I realized that the docs I was referring to were for an older version, in 5.x docs there is indeed no mention about null.

satya164:
Yes, the comment is incorrect

Issue #7936, opened 2020-04-02
navigation.navigate() not enforcing params in TypeScript


oli-laban:
**Current Behavior**

I've followed the guide [here](https://reactnavigation.org/docs/typescript#type-checking-screens) to set up type checking, but `navigation.navigate()` isn't enforcing that the params are passed to the screen.

```
// rootStackNavigator.tsx

export type RootStackParamList = {
  Login: undefined;
  Leads: undefined;
  WebView: { url: string };
};

...
```

```
// screens/Login/index.ts

interface Props {
  navigation: StackNavigationProp<RootStackParamList, 'Login'>;
}

class Login extends Component<Props> {
  ...
}
```

Calling `navigation.navigate('WebView')` in the above component does not error.

**Expected Behavior**

`navigation.navigate('WebView')` should error if the param `{ url: 'something' }` is not passed.

Typing for the navigate method:
```
navigate<RouteName extends keyof ParamList>(...args: ParamList[RouteName] extends undefined | any ? [RouteName] | [RouteName, ParamList[RouteName]] : [RouteName, ParamList[RouteName]]): void;
```

The `ParamList[RouteName] extends undefined | any` conditional check in the typing is obviously returning `true` and the type that is being enforced on the call is:
```
navigate<"WebView">(...args: ["WebView"] | ["WebView", {
    url: string;
}]): void
```

I would expect the resulting type to be:
```
navigate<"WebView">(...args: ["WebView", {
    url: string;
}]): void
```

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | iOS 13.3
| @react-navigation/native       | 5.1.3
| @react-navigation/stack        | 5.2.8
| react-native-gesture-handler   | 1.6.1
| react-native-safe-area-context | 0.7.3
| react-native-screens           | 2.4.0
| react-native                   | 0.61.5
| node                           | 12.12.0
| npm or yarn                    | yarn 1.19.1
| typescript | 3.8.3

Miyou:
I had the same issue. Solved it by changing the typing for the navigate method to:
`navigate<RouteName extends keyof ParamList>(...args: ParamList[RouteName] & undefined extends never ? [RouteName, ParamList[RouteName]] : [RouteName] | [RouteName, ParamList[RouteName]]): void;`

I'm very new to typescript and I would have thought that simply removing the `| any` in the original typing would have worked, but it seems that for example `string | undefined` does not extend undefined for some reason.

[...]

PR #8389, opened 2020-06-05
fix: catch missing params when they are required in navigate

jeanregisser:
There is a problem with the enforcement of required params in `navigation.navigate(...)` in TypeScript as described in https://github.com/react-navigation/react-navigation/issues/7936

@Miyou found a fix 🥳. All credits go to him.

I needed this so went ahead and submitted a PR and added a test to avoid it from breaking in the future.

However note that until this project switches to TypeScript 3.9 with support for [`@ts-expect-error`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/#ts-expect-error-comments), it can still break silently.

Before the change, this is how the test looks:
Screenshot 2020-06-05 at 22 51 00
As you can see it doesn't catch the missing params which are required on line 28.

After the change, all expected errors are raised:
Screenshot 2020-06-05 at 22 51 59

Let me know what you think.


jeanregisser:
Yes done 👍

Miyou:
Thanks @jeanregisser for submitting the PR! :)

PR #8406, opened 2020-06-10
Allow bottomTabsBar height override

benhowes:
In my app the tab bar is slightly taller than the default. (`65` points)

So, setting the height works normally, unless there is a safe area inset, e.g. on more recent iOS devices, at which point my nav bar content is squashed.

I'm proposing to pluck the height from styles, falling back to the default and add the detected inset to that height.

Not sure if this is a good pattern, happy to alter.

Thanks for a great library - 5.0 took a day to add, but has been well worth it in terms of the performance/feature gains! :horse_racing:

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -181,7 +181,10 @@ export default function BottomTabBar({
   };
 
   const defaultInsets = useSafeArea();
-
+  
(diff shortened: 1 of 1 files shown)


satya164:
You can pass a height including safe area:

```js
import { useSafeArea } from 'react-native-safe-area-context';

// ...

```js
const insets = useSafeArea();
const tabBarHeight = whatever + insets.top;
```

benhowes:
Yep, that's what I did. Thanks all the same.

PR #8436, opened 2020-06-13
feat: add devtools package

satya164:
The `devtools` package extracts the redux devtools extension integration to a separate package. In future we can add more tools such as flipper integration to this package.

Usage:

```js
import { useReduxDevToolsExtension } from '@react-navigation/devtools';

  const navigationRef = React.useRef();

  useReduxDevToolsExtension(navigationRef);

  return (
    <NavigationContainer ref={navigationRef}>{/* ... */}</NavigationContainer>
  );
}
```


packages/devtools/src/useReduxDevToolsExtension.tsx
```diff
@@ -0,0 +1,106 @@
...
+    devToolsRef.current === undefined &&
+    typeof __REDUX_DEVTOOLS_EXTENSION__ !== 'undefined'
+  ) {
+    devToolsRef.current = __REDUX_DEVTOOLS_EXTENSION__.connect({
+      name: '@react-navigation/devtools',
+    });
+  }
+
+  const lastStateRef = React.useRef<NavigationState | undefined>();
+  const lastActionRef = React.useRef<NavigationAction | undefined>(undefined);
```
Ashoat, line 39: Nit - why specify `undefined` explicitly here?

packages/devtools/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "prepare": "bob build",
+    "clean": "del lib"
+  },
+  "dependencies": {
+    "@react-navigation/core": "^5.10.0",
```
Ashoat, line 35: I guess you considered `peerDependencies` and it probably didn't work given how it's transitive through `native` in most cases... a little worried about two versions of `@react-navigation/core` getting installed if somebody forgets to update `@react-navigation/devtools` when upgrading other packages, but can't think of a better solution

packages/devtools/src/useReduxDevToolsExtension.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    if (!navigation || !devTools) {
+      return;
+    }
+
+    if (lastStateRef.current === undefined) {
+      const state = navigation.getRootState();
+
+      devTools.init(state);
+      lastStateRef.current = state;
```
Ashoat, line 66: I feel like I'm missing something, but does `lastStateRef` only get set the first time, when it's `undefined`? Shouldn't it get updated every time?

packages/devtools/src/useReduxDevToolsExtension.tsx
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+    );
+
+    const unsubscribeState = navigation.addListener('state', (e) => {
+      if (lastStateRef.current === e.data.state) {
+        return;
+      }
+
+      const state = navigation.getRootState();
```
Ashoat, line 88: I'm a little confused about the difference between `e.data.state` and `navigation.getRootState`, and why you do a shallow comparison against the former and a deep comparison against the latter

packages/devtools/src/useReduxDevToolsExtension.tsx
satya164: Probably autocomplete added it by mistake xD

packages/devtools/package.json
satya164: We just import types from this package so it would be fine.

packages/devtools/src/useReduxDevToolsExtension.tsx
satya164: Yea, missed it setting below

packages/devtools/src/useReduxDevToolsExtension.tsx
satya164: The first check is to skip it if we did the `resetRoot(state)` ourselves, in which case, this state object would be same as that one. Deep equality would work, but a simple referential equality is less expensive.

For using `getRootState()`, it returns fully hydrated state object, so that's what we should be using most of the time anyway.

PR #8460, opened 2020-06-17
feat: Export `StackNavigator`, `StackNavigatorProps` and `TypedStackNavigator` type defs

jtomaszewski:
(Same as https://github.com/react-navigation/react-navigation/pull/8233 but rebased from master to main)

Thanks to this, you can easily import and construct a type of a value returned by `createStackNavigator<T>()` expression.

We would use this because our app is quite complex, and so we extract constructing routes/params/stacks to separate files. We could use this like that:

```tsx
import React from "react";
import { TypedStackNavigator } from "@react-navigation/stack";
import { RootStackNavigatorParamList, Screens } from "../navigation";
import { ChatScreen, ChatListScreen } from "./chat";

export const getChatStackScreens = (
  Stack: TypedStackNavigator<RootStackNavigatorParamList>
): React.ReactElement => {
  return (
    <>
      <Stack.Screen
        name={Screens.ChatList}
        component={ChatListScreen}
      />
      <Stack.Screen
        name={Screens.Chat}
        component={ChatScreen}
      />
    </>
  );
};

// and then in `Stack.Navigator` just use it like `{getMoreStackScreens(Stack)}`
```

### DIFF
--- a/packages/stack/src/index.tsx
+++ b/packages/stack/src/index.tsx
@@ -6,7 +6,11 @@ import * as TransitionPresets from './TransitionConfigs/TransitionPresets';
 /**
  * Navigators
  */
-export { default as createStackNavigator } from './navigators/createStackNavigator';
+export {
+  default as createStackNavigator,
+  StackNavigator,
+  Props as StackNavigatorProps
+} from './navigators/createStackNavigator';
 
 export const Assets = [
   // eslint-disable-next-line import/no-commonjs
--- a/packages/stack/src/navigators/createStackNavigator.tsx
+++ b/packages/stack/src/navigators/createStackNavigator.tsx
@@ -9,6 +9,8 @@ import {
   StackRouterOptions,
   StackNavigationState,
   StackActions,
+  ParamListBase,
+  TypedNavigator,
 } from '@react-navigation/native';
 import StackView from '../views/Stack/StackView';
 import {
@@ -17,11 +19,11 @@ import {
   StackNavigationEventMap,
 } from '../types';
 
-type Props = DefaultNavigatorOptions<StackNavigationOptions> &
+export type Props = DefaultNavigatorOptions<StackNavigationOptions> &
   StackRouterOptions &
   StackNavigationConfig;
 
-function StackNavigator({
+export function StackNavigator({
   initialRouteName,
   children,
   screenOptions,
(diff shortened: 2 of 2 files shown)


packages/stack/src/navigators/createStackNavigator.tsx
```diff
@@ -17,11 +19,11 @@ import {
   StackNavigationEventMap,
 } from '../types';
 
-type Props = DefaultNavigatorOptions<StackNavigationOptions> &
+export type Props = DefaultNavigatorOptions<StackNavigationOptions> &
   StackRouterOptions &
   StackNavigationConfig;
 
-function StackNavigator({
+export function StackNavigator({
```
satya164, line 26: I think exposing this is confusing (there's no the difference between this and `Stack.Navigator` to the user, which one should they use). Why do you need it?

packages/stack/src/navigators/createStackNavigator.tsx
```diff
@@ -17,11 +19,11 @@ import {
   StackNavigationEventMap,
 } from '../types';
 
-type Props = DefaultNavigatorOptions<StackNavigationOptions> &
+export type Props = DefaultNavigatorOptions<StackNavigationOptions> &
```
satya164, line 22: I think this shouldn't be exported. You should use `React.ComponentProps<typeof Stack.Navigator>` or `React.ComponentProps<TypedStackNavigator<ParamList>['Navigator']>` in this case. Otherwise it lacks type information for params.

packages/stack/src/navigators/createStackNavigator.tsx
```diff
@@ -94,3 +96,13 @@ export default createNavigatorFactory<
   StackNavigationEventMap,
   typeof StackNavigator
 >(StackNavigator);
+
+export type TypedStackNavigator<
```
satya164, line 100: Can you rename it to `StackNavigatorType`?

packages/stack/src/index.tsx
```diff
@@ -6,7 +6,11 @@ import * as TransitionPresets from './TransitionConfigs/TransitionPresets';
 /**
  * Navigators
  */
-export { default as createStackNavigator } from './navigators/createStackNavigator';
+export {
+  default as createStackNavigator,
+  StackNavigator,
+  Props as StackNavigatorProps
+} from './navigators/createStackNavigator';
```
satya164, line 13: Can you move the type export to separate `export type` statement?

packages/stack/src/navigators/createStackNavigator.tsx
jtomaszewski: Should I also rename `TypedNavigator` to `NavigatorType`? (That's why I named it `TypedStackNavigator`, because `TypedNavigator` already existed.)

satya164: That's internal, so it doesn't matter what we name it. I think `TypedNavigator` is ok for internal use.

satya164:
Closing due to no activity

PR #8461, opened 2020-06-17
fix(TabRouter): When calculating next route name, use `initialRouteName` if the current route doesn't exist anymore

jtomaszewski:
Currently:

- if you mount/unmount routes, TabRouter will navigate to the first tab, instead of the tab specified in `initialRouteName` 
  - https://github.com/react-navigation/react-navigation/commit/e20cacad2d51434f86afc07f588a37c4910d498c fixes it
- if you change `initialRouteName` after `TabRouter` is mounted, and then mount/unmount routes, TabRouter will navigate to the old `initialRouteName`, instead to the one that is currently given to the `useNavigationBuilder` hook (through `<TabNav.Navigator initialRouteName={...} />`). 
  - https://github.com/react-navigation/react-navigation/commit/8a4a46ea4f2622f8dc280b99326e821c1431d786 fixes it

We need it because when user commits some router-unrelated action, we rearrange the tabs AND change the initial route, and we want the user land on a different tab that he landed initially before.

### DIFF
--- a/packages/core/src/__tests__/BaseNavigationContainer.test.tsx
+++ b/packages/core/src/__tests__/BaseNavigationContainer.test.tsx
@@ -1,10 +1,6 @@
 import * as React from 'react';
 import { act, render } from 'react-native-testing-library';
-import {
-  DefaultRouterOptions,
-  NavigationState,
-  Router,
-} from '@react-navigation/routers';
+import { NavigationState, Router } from '@react-navigation/routers';
 import BaseNavigationContainer from '../BaseNavigationContainer';
 import NavigationStateContext from '../NavigationStateContext';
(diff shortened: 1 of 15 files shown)


jtomaszewski:
Any tip on how would you see it to be implemented? Should I just do it in the same way, but name it as `defaultRouteName` (and later on we could slowly deprecate `initialRouteName`) ?

satya164:
Need to think about it, but probably there would be a way to override router methods (e.g. by passing a `router` prop) to implement custom behaviour.

satya164:
I'm closing this due to conflicts, but please open an RFC to discuss it more. In the future, please prefer opening an RFC when making such significant changes so we can discuss the API first before working on an implementation to avoid wasted effort.

PR #8465, opened 2020-06-17
Allow the `screenOptions` `header` to be a React component

### LINKED ISSUE #8463: Cannot use Hooks inside Header component
**Current Behavior**

- What code are you running and what is happening?

My Stack Navigator is defined as such (please note the `header` option):

```tsx
<StackNavigator.Navigator screenOptions={{ header: NavigationHeader }}>
  <DrawerNavigator.Screen name='Home' component={HomeScreen} />
  <DrawerNavigator.Screen name='Profile' component={ProfileScreen} />
</StackNavigator.Navigator>
```

I would think that `NavigationHeader` can be a React component, but apparently it has the same function signature as a React component while not being a React component. From the docs:

> Function that given `HeaderProps` returns a React Element, to display as a header.

- Include a screenshot or video if it makes sense.

The error that I get if I put a hook inside `NavigationHeader` is (in this case, I tried to `useMemo`):

```
Warrning: React has detected a change in the order of Hooks called by HeaderContainer. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://fb.me/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useContext                 useContext
2. useMemo                    useMemo
3. undefined                  useMemo
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    in HeaderContainer (at StackView.tsx:328)
    in CardStack (at StackView.tsx:435)
    in KeyboardManager (at StackView.tsx:433)
    in SafeAreaProviderCompat (at StackView.tsx:430)
    in RCTView (at StackView.tsx:429)
    in StackView (at createStackNavigator.tsx:82)
    in StackNavigator (at App.tsx:96)
    in NavigationStack (at SceneView.tsx:121)
    in StaticContainer
```
_(stack trace continues)_

**Expected Behavior**

- What do you expect should be happening?

I would expect the `header` option can be a React component. I'm specifically trying to `useMemo` on the `styleInterpolator` as it [appears that that result should be memoized](https://github.com/react-navigation/react-navigation/blob/1aadc79fb89177a2fff2dcd791d67a3c880009d0/packages/stack/src/views/Header/HeaderSegment.tsx#L116).

It also appears that the default `Header` component is mounted as a React component while the screen option is not? https://github.com/react-navigation/react-navigation/blob/b14fd9453d5b581966228f9cbb1f89b4fcecf52a/packages/stack/src/views/Header/HeaderContainer.tsx#L165

I would expect the two to be handled in the same way - both mounted as React components.

~~- Include a screenshot or video if it makes sense.~~

**How to reproduce**

https://snack.expo.io/Yvhp7KUKx

I've attempted to recreate the `screenOptions` with a custom Header that has a hook, however, the error is not manifesting in the Snack 🤷 

Looking at the code itself, it seems to verify my suspicion that I describe above - that it's not being rendered as a React component.

I'm unsure if there's some version mismatch between the Snack and my environment. I took the example from the website and added the `screenOptions` onto that.

Kindly let me know if this was helpful or if I can assist with reproducing this in other ways 🙏

**Your Environment**

| iOS or Android                 | Tested on iOS Simulator
| @react-navigation/native       | 5.5.1
| @react-navigation/stack        | 5.5.1
| react-native-gesture-handler   | 1.6.0
| react-native-safe-area-context | 0.7.3
| react-native-screens           | 2.2.0
| react-native                   | Expo 37.0.1
| react                   | 16.13.1
| expo                           | 37.0.0
| node                           | 12.14.0
| npm or yarn                    | yarn v1.22.4

tazsingh:
Fixes #8463

I've provided more information in the issue above.

In essence, the `header` in `screenOptions` is currently a "Function that given HeaderProps returns a React Element, to display as a header." according to the docs. A function that accepts props and returns a React Element is the signature of a React Function Component. Why not make the `header` officially a React Component in that case?

I stumbled upon this issue when trying to use a Hook inside a custom `header` component to see that I should instead write a function that returns a container component that then calls the hook inside of that... I think we can do better ✨

### DIFF
--- a/packages/stack/src/views/Header/HeaderContainer.tsx
+++ b/packages/stack/src/views/Header/HeaderContainer.tsx
@@ -162,7 +162,7 @@ export default function HeaderContainer({
                     : null
                 }
               >
-                {header !== undefined ? header(props) : <Header {...props} />}
+                {header !== undefined ? React.createElement(header, props) : <Header {...props} />}
               </View>
             </NavigationRouteContext.Provider>
           </NavigationContext.Provider>


satya164:
This is going to remount the header every time the reference changes, which will lose local state and cause perf issues. Just pass a function instead of a component.

tazsingh:
@satya164 A function would return a different identity on every invocation, which would lose local state and cause perf issues. https://github.com/react-navigation/react-navigation/issues/8463#issuecomment-645298848

`const NavigationHeader = () => ...` at the top level would be the same reference every time.

satya164:
It's impossible to ensure that people define it at top level. It's much easier to cause bugs by defining it inline.

satya164:
The thing is: both approaches have problems, but in one case, the bugs may go unnoticed and hard to catch, in other case, you catch the issue immediately. I think the later should be preferable.

satya164:
I also forgot to mention, by using a function, you can reference the component's props. If you make it a component which always needs to be defined at top level, it's impossible to access any props there. You can use context, but that way more work.

tazsingh:
I respectfully disagree - it's a render prop except it's not

Sure, you can write bad code lots of ways. React has great tooling to help folks identify what state is where and where perf issues are coming from. There are established patterns that put people into a good place and minimizes poor code.

In this particular case, sure, whatever, I'll return a React Element from a function. It may have been helpful to have something in the docs to the effect of the rationale you've given above as I would think it could be confusing to future travellers who see a function that takes `props` and returns an Element but can't be a Component when render props are a very common and understood thing 🤷 This way I personally wouldn't have spent 2 days debugging where a Hook mismatch error was coming from in my app, worked through stack traces to identify the line it was coming from in React Navigation, wouldn't have spent time writing a lengthy issue with reproduction steps in a Snack that has its own problems, wouldn't have spent time writing a PR, receive a diminutive 1 line response describing a very obvious function signature, and wouldn't be having this particular discussion (as lovely as it is to speak with you and hear your rationale!). Just providing feedback 🙏 Thank you for your kind work on this library!

sacru2red:
I think this is better way

PR #8473, opened 2020-06-18
fix: fix screen disappearing on Android

WoLewicki:
Added `collapsable={false}` to the dummy View in order for the Android to render screens properly. This issue is most probably similar to https://github.com/react-navigation/react-navigation/commit/9c06a92d092af150d653c3a2f7fdccd28090bb14 but fixes it on Android since the View seems to be removed from a native view hierarchy due to not drawing anything. To see the bug go to https://github.com/software-mansion/react-native-screens/issues/544.


kesha-antonov:
@satya164 will you release this fix for RNav 4?

PR #8486, opened 2020-06-21
chore: inform users about @react-navigation/native

developer239:
I used `react-navigation` a couple of times in the past. I don't use StackOverflow too much during my day and I know the API for 4.x. Two weeks ago I started a new project. I installed `react-navigation` as usual and because I didn't check the documentation. It took me two weeks to realize that the latest version is actually `v5`.

It would be nice to inform users that `react-navigation` was transformed into a mono repository and that they should install `@react-navigation/native` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "test": "jest",
     "prerelease": "yarn lerna run clean",
     "release": "yarn lerna publish",
-    "example": "yarn --cwd example"
+    "example": "yarn --cwd example",
+    "postinstall": "node -e \"try{require('./postinstall')}catch(e){}\""
   },
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.8.3",
--- a/postinstall.js
+++ b/postinstall.js
@@ -1,5 +1,6 @@
-const banner = 'https://github.com/react-navigation/react-navigation \u001b[33mwas transformed into monorepo. ' +
+const banner =
+  'https://github.com/react-navigation/react-navigation \u001b[33mwas transformed into monorepo. ' +
   'If you want to use the latest version install @react-navigation/native. \n\n You can find more information in our ' +
   'getting started guide: https://reactnavigation.org/docs/getting-started';
 
-console.log(banner)
+console.log(banner);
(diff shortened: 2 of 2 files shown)


satya164:
Thanks for the PR. I don't think whether the repo was transformed to monorepo or not is irrelevant to users.

I also don't want to add a log on installing the package as 4.x isn't deprecated and people who use it shouldn't be getting unnecessary warnings/logs. Also, yarn doesn't show such logs during the install, so it won't be visible to a lot of users regardless.

Besides, the `package.json` here is not the `package.json` that's installed, it's the monorepo's metadata.  So, this change isn't going to work. The packages reside under the `packages/` folder where all changes regarding packages should be made.

We already throw an error if you try to use v5 API when you have v4 installed. Feel free to send PRs improving the errors or adding additional checks there. The 4.x version exists under the 4.x branch.

You should really be reading the documentation when installing packages.

https://reactnavigation.org/blog/2020/02/06/react-navigation-5.0
https://reactnavigation.org/docs/upgrading-from-4.x

developer239:
@satya164 OK to make it simple: I run `yarn add react-navigation` and got `v4` and I did not know that there is v5 available. I know the API pretty well so it took me 2 weeks to realize that the application that I bootstrap is outdated.

Imagine how many tutorials there are on the internet. People who will follow them 6 months or even a couple of years from now will be misled because unless they will actually look at the README documentation they will not find out that they shouldn't use `react-navigation` anymore. It might not be deprecated but it will be in the future.

Maybe we can change the message and we can prevent it from being visible in the CI but I believe it is worth telling people that they should migrate to `@react-navigation/native`.

Try to remember how `babel` did it couple of years ago. 🙂

satya164:
> it is worth telling people that they should migrate to @react-navigation/native.

We don't recommend migrating if they don't need to. React Navigation 4 is not deprecated. It's perfectly fine to use it. For versions we no longer support, we already warn people via deprecation warnings.

I don't know what's the best way to announce new versions if you don't read the blog, but command line isn't a good place for announcements. People don't like unnecessary noise in the Terminal (just look at the core-js controversy) and it can hide important information. Regardless, it's literally impossible to show this message to people who are using Yarn.

PR #8488, opened 2020-06-21
chore: autofix eslint errors

developer239:
One of my PRs failed because of linting issues that could have been auto fixed. 🙂 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-commit": "yarn lint && yarn typescript && yarn test"
+      "pre-commit": "yarn lint --fix && yarn typescript && yarn test"
     }
   },
   "jest": {


satya164:
Thanks for the PR. Changing code on pre-commit hook isn't safe. It can mess up unrelated changes in files which the developer hasn't committed and makes impossible to partially commit files with `commit -p`. There are specific tools/workflows to handle such cases but they are just much more complicated.

Also, this change isn't going to work because you need to actually add the files to git before you commit, just fixing them isn't enough. Please make sure to test your changes before sending PRs.

> One of my PRs failed because of linting issues that could have been auto fixed.

I'm not sure how you even sent the PR since the pre-commit hook would've rejected the commit in the first place, unless you explicitly ignored the pre-commit hooj - in which case auto fix in pre-commit hook isn't going to help either.

developer239:
@satya164 You made some good points. Although I definitely did not force my commit I just did not bother to install dependencies.

Anyway, I believe that auto fixing on pre-commit hook is safe because it can mess up unrelated changes ONLY if the file was not fixed in the first place. This will never happen.

I would rather not waste any more time on this issue, although if you want I can help with `lint-staged` I can create another PR.

I. still believe that auto fixing errors is the least we can do for contributors. 🙂

PR #8491, opened 2020-06-22
feat: add an onReady callback to the container

satya164:


packages/native/src/NavigationContainer.tsx
```diff
@@ -46,13 +48,27 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
...
+  const isReady = rest.initialState != null || !isLinkingEnabled || isResolved;
+
+  const onReadyRef = React.useRef(onReady);
+
+  React.useEffect(() => {
+    onReadyRef.current = onReady;
+  });
+
+  React.useEffect(() => {
+    if (isReady) {
```
osdnk, line 66: shouldn't we prevent before calling it twice?

satya164: I think it will be fine because `isReady` will only change once and we have it in dep array.

packages/native/src/NavigationContainer.tsx
```diff
@@ -46,13 +48,27 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
...
   const linkingContext = React.useMemo(() => ({ options: linking }), [linking]);
 
-  if (rest.initialState == null && isLinkingEnabled && !isReady) {
+  const isReady = rest.initialState != null || !isLinkingEnabled || isResolved;
+
+  const onReadyRef = React.useRef(onReady);
+
+  React.useEffect(() => {
+    onReadyRef.current = onReady;
+  });
```
Bardiamist, line 63: Without `useEffect` should be more stable:

```
const onReadyRef = React.useRef(onReady);

onReadyRef.current = onReady;
```

satya164: That's not concurrent mode safe. Ref shouldn't be assigned during render.

packages/native/src/NavigationContainer.tsx
```diff
@@ -46,13 +48,27 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
...
 
-  const [isReady, initialState] = useThenable(getInitialState);
+  const [isResolved, initialState] = useThenable(getInitialState);
 
   React.useImperativeHandle(ref, () => refContainer.current);
 
   const linkingContext = React.useMemo(() => ({ options: linking }), [linking]);
 
-  if (rest.initialState == null && isLinkingEnabled && !isReady) {
+  const isReady = rest.initialState != null || !isLinkingEnabled || isResolved;
```
JDMathew, line 57: Wy did this change to `||` from `&&` ?

PR #8502, opened 2020-06-23
feat: rework linking configuration to be more strict

satya164:
The PR changes a few things about linking configuration:

- Moves the configuration for screens to a screens property so that it's possible to specify other options like `initialRouteName` for the navigator at root
- The nesting in the configuration needs to strictly match the shape of the navigation tree, it can't just rely on URL's shape anymore
- If a screen is not specified in the configuration, it won't be parsed to/from the URL (this is essential to handle unmatched screens)
- Treat `path: ''` and no specified path in the same way, unless `exact` is specified
- Disallow specifying unmatched screen with old format
- Add support for `initialRouteName` at top level
- Automatically adapt old configuration to new format


packages/core/src/getPathFromState.tsx
```diff
@@ -250,13 +282,28 @@ const createConfigItem = (
...
+
+  if (legacy) {
+    pattern =
+      config.exact !== true && parentPattern && config.path
+        ? joinPaths(parentPattern, config.path)
+        : config.path;
+  } else {
+    if (config.exact && config.path === undefined) {
+      throw new Error(
+        "A 'path' needs to be specified when specifying 'exact: true'. If you don't want this screen in the URL, specify it as empty string, e.g. 'path: ''."
```
WoLewicki, line 295:
```suggestion
        "A 'path' needs to be specified when specifying 'exact: true'. If you don't want this screen in the URL, specify it as empty string, e.g. `path: ''`."
```

packages/core/src/getPathFromState.tsx
```diff
@@ -59,15 +62,37 @@ const getActiveRoute = (state: State): { name: string; params?: object } => {
...
-  state?: State,
-  options: PathConfig = {}
+  state: State,
+  options?: Options
 ): string {
-  if (state === undefined) {
-    throw Error('NavigationState not passed');
+  if (state == null) {
+    throw Error(
+      "Got 'undefined' for the navigation state. You must pass a valid state object."
```
WoLewicki, line 70: Is it `undefined` or `null` then?

satya164: Yea, the types prevent it anyway so shouldn't be an issue

PR #8524, opened 2020-07-01
feat: emit gesture navigation events from stack view

martintreurnicht:
Allows you to subscribe to gesture navigation events, we have a custom keyboard that we want to hide and show when gesture is being used to navigate (same as native keyboard)

This PR has been reopened after https://github.com/react-navigation/react-navigation/pull/7800 was closed (`master` renamed to `main`)

### DIFF
--- a/packages/stack/src/types.tsx
+++ b/packages/stack/src/types.tsx
@@ -27,6 +27,18 @@ export type StackNavigationEventMap = {
    * Event which fires when a transition animation ends.
    */
   transitionEnd: { data: { closing: boolean } };
+  /**
+   * Event which fires when navigation gesture starts.
+   */
+  gestureStart: { data: undefined };
+  /**
+   * Event which fires when navigation gesture is completed.
+   */
+  gestureEnd: { data: undefined };
+  /**
+   * Event which fires when navigation gesture is canceled.
+   */
+  gestureCancel: { data: undefined };
 };
 
 export type StackNavigationHelpers = NavigationHelpers<
--- a/packages/stack/src/views/Stack/CardContainer.tsx
+++ b/packages/stack/src/views/Stack/CardContainer.tsx
@@ -46,6 +46,9 @@ type Props = TransitionPreset & {
   onPageChangeStart?: () => void;
   onPageChangeConfirm?: () => void;
   onPageChangeCancel?: () => void;
+  onGestureStart?: (props: { route: Route<string> }) => void;
+  onGestureEnd?: (props: { route: Route<string> }) => void;
+  onGestureCancel?: (props: { route: Route<string> }) => void;
   gestureEnabled?: boolean;
   gestureResponseDistance?: {
     vertical?: number;
@@ -95,6 +98,9 @@ function CardContainer({
   onPageChangeCancel,
   onPageChangeConfirm,
   onPageChangeStart,
+  onGestureCancel,
+  onGestureEnd,
+  onGestureStart,
   onTransitionEnd,
   onTransitionStart,
   renderHeader,
@@ -118,6 +124,20 @@ function CardContainer({
   const handleClose = () => {
     onTransitionEnd?.({ route: scene.route }, true);
     onCloseRoute({ route: scene.route });
+  };
+
+  const handleGestureBegin = () => {
+    onPageChangeStart?.();
+    onGestureStart?.({ route: scene.route });
+  };
+
+  const handleGestureCanceled = () => {
+    onPageChangeCancel?.();
+    onGestureCancel?.({ route: scene.route });
+  };
+
+  const handleGestureEnd = () => {
+    onGestureEnd?.({ route: scene.route });
   };
 
   const handleTransitionStart = ({ closing }: { closing: boolean }) => {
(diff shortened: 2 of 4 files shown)


martintreurnicht:
@satya164 I've re-opened the PR. I made the changes you requested based no my understanding of what you wanted me to do, decoupling gesture events from page change events

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -109,6 +115,9 @@ function CardContainer({
   React.useEffect(() => {
     onPageChangeConfirm?.();
   }, [active, onPageChangeConfirm]);
+  React.useEffect(() => {
+    onGestureEnd?.({ route: scene.route });
```
satya164, line 119: `onGestureEnd` needs to be called when gesture ends, but here seems like it's called if `scene`, `active` or `onGestureEnd` props change + on mount. I think you need to add a separate `onGestureEnd` callback in `Card` to handle it.

packages/stack/src/types.tsx
```diff
@@ -27,6 +27,18 @@ export type StackNavigationEventMap = {
    * Event which fires when a transition animation ends.
    */
   transitionEnd: { data: { closing: boolean } };
+  /**
+   * Event which fires when navigation gesture starts.
+   */
+  gestureStart: {};
```
satya164, line 33:
```suggestion
  gestureStart: { data: undefined };
```

packages/stack/src/types.tsx
```diff
@@ -27,6 +27,18 @@ export type StackNavigationEventMap = {
...
+  /**
+   * Event which fires when navigation gesture is completed.
+   */
+  gestureEnd: {};
```
satya164, line 37:
```suggestion
  gestureEnd: { data: undefined };
```

packages/stack/src/types.tsx
```diff
@@ -27,6 +27,18 @@ export type StackNavigationEventMap = {
...
+  /**
+   * Event which fires when navigation gesture is canceled.
+   */
+  gestureCancel: {};
```
satya164, line 41:
```suggestion
  gestureCancel: { data: undefined };
```

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -120,11 +129,22 @@ function CardContainer({
...
+
+  const handleGestureCanceled = () => {
+    onPageChangeCancel?.();
+    onGestureCancel?.({ route: scene.route });
+  };
+
   const handleTransitionStart = ({ closing }: { closing: boolean }) => {
     if (active && closing) {
       onPageChangeConfirm?.();
+      onGestureEnd?.({ route: scene.route });
```
satya164, line 145: `onGestureEnd` needs to explicitly be called on gesture end. transition could start due to various reasons, and doesn't mean that a gesture has ended.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -120,11 +129,22 @@ function CardContainer({
...
     } else {
-      onPageChangeCancel?.();
+      handleGestureCanceled();
```
satya164, line 147: starting a transition shouldn't assume that a gesture was cancelled

packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -405,6 +405,30 @@ export default class StackView extends React.Component<Props, State> {
       target: route.key,
     });
 
+  private handleGestureStart = ({ route }: { route: Route<string> }) => {
+    this.props.navigation.emit({
+      type: 'gestureStart',
+      data: {},
+      target: route.key,
+    });
```
satya164, line 413:
```suggestion
    this.props.navigation.emit({
      type: 'gestureStart',
      target: route.key,
    });
```

packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -405,6 +405,30 @@ export default class StackView extends React.Component<Props, State> {
...
+  };
+
+  private handleGestureEnd = ({ route }: { route: Route<string> }) => {
+    this.props.navigation.emit({
+      type: 'gestureEnd',
+      data: {},
+      target: route.key,
+    });
```
satya164, line 421:
```suggestion
    this.props.navigation.emit({
      type: 'gestureEnd',
      target: route.key,
    });
```

packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -405,6 +405,30 @@ export default class StackView extends React.Component<Props, State> {
...
+  };
+
+  private handleGestureCancel = ({ route }: { route: Route<string> }) => {
+    this.props.navigation.emit({
+      type: 'gestureCancel',
+      data: {},
+      target: route.key,
+    });
```
satya164, line 429:
```suggestion
    this.props.navigation.emit({
      type: 'gestureCancel',
      target: route.key,
    });
```

packages/stack/src/views/Stack/CardContainer.tsx
martintreurnicht: Ah, good catch. Yeah i did actually add the callback in `Card.tsx`, but for some reason got confused here. I've made the change

packages/stack/src/views/Stack/CardContainer.tsx
martintreurnicht: Ok, removing `onGestureEnd` call here, call back will be invoked from `Card.tsx`

packages/stack/src/views/Stack/CardContainer.tsx
martintreurnicht: Ok, makes sense. Reverting...

martintreurnicht:
> Thanks for the PR. I left some comments

Ok thanks for the feedback! I've made the changes you've suggested, please let me know if there is anything else

PR #8530, opened 2020-07-01
feat: add a `beforeRemove` event

satya164:
A lot of times, we want to prompt before leaving a screen if we have unsaved changes. Currently, we need to handle multiple cases to prevent this:

- Disable swipe gestures
- Override the back button in header
- Override the hardware back button on Android

This PR adds a new event which is emitted before a screen gets removed, and the developer has a chance to ask the user before closing the screen.

Example:

```js
React.useEffect(
  () =>
    navigation.addListener('beforeRemove', (e) => {
      if (!hasUnsavedChanges) {
        return;
      }

      e.preventDefault();

      Alert.alert(
        'Discard changes?',
        'You have unsaved changes. Are you sure to discard them and leave the screen?',
        [
          { text: "Don't leave", style: 'cancel', onPress: () => {} },
          {
            text: 'Discard',
            style: 'destructive',
            onPress: () => navigation.dispatch(e.data.action),
          },
        ]
      );
    }),
  [navigation, hasUnsavedChanges]
);
```

ezgif-2-09f96c4f63b5

**TODO**

- [x] Make it work when the screen gets removed by a parent navigator as well
- [x] Investigate unhandled action if discard was pressed on first screen of navigator

### DIFF
--- a/packages/core/src/__tests__/useOnAction.test.tsx
+++ b/packages/core/src/__tests__/useOnAction.test.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
-import { render } from 'react-native-testing-library';
-import type {
+import { act, render } from 'react-native-testing-library';
+import {
   Router,
   DefaultRouterOptions,
   NavigationState,
+  StackRouter,
 } from '@react-navigation/routers';
 import useNavigationBuilder from '../useNavigationBuilder';
 import BaseNavigationContainer from '../BaseNavigationContainer';
@@ -12,8 +13,19 @@ import MockRouter, {
   MockActions,
   MockRouterKey,
 } from './__fixtures__/MockRouter';
+import type { NavigationContainerRef } from '../types';
 
-beforeEach(() => (MockRouterKey.current = 0));
+jest.mock('nanoid/non-secure', () => {
+  const m = { nanoid: () => String(++m.__key), __key: 0 };
+
+  return m;
+});
+
+beforeEach(() => {
+  MockRouterKey.current = 0;
+
+  require('nanoid/non-secure').__key = 0;
+});
 
 it("lets parent handle the action if child didn't", () => {
   function CurrentRouter(options: DefaultRouterOptions) {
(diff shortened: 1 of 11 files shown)


packages/core/src/useOnAction.tsx
```diff
@@ -82,6 +92,45 @@ export default function useOnAction({
           onDispatchAction(action, state === result);
 
           if (state !== result) {
+            // @ts-expect-error: add this property to mark that we've already emitted this action
+            if (action[BEFORE_REMOVE_EMITTED] !== true) {
```
osdnk, line 96: let's try to use a dict here: 
Symbol('routesBeforeRemoveEventDispatched') :)

packages/core/src/__tests__/useOnAction.test.tsx
```diff
@@ -12,8 +13,19 @@ import MockRouter, {
   MockActions,
   MockRouterKey,
 } from './__fixtures__/MockRouter';
+import type { NavigationContainerRef } from '../types';
 
-beforeEach(() => (MockRouterKey.current = 0));
+jest.mock('nanoid/non-secure', () => {
+  const m = { nanoid: () => String(++m.__key), __key: 0 };
```
Trancever, line 19:
```suggestion
  return { nanoid: () => String(++m.__key), __key: 0 };
```

satya164: Return won't work. The variable is used inside the object.

packages/core/src/useOnPreventRemove.tsx
```diff
@@ -0,0 +1,93 @@
...
+const VISITED_ROUTE_KEYS = Symbol('VISITED_ROUTE_KEYS');
+
+export const shouldPreventRemove = (
+  emitter: NavigationEventEmitter<EventMapCore<any>>,
+  beforeRemoveListeners: Record<string, ChildBeforeRemoveListener | undefined>,
+  routes: Route<string>[],
+  action: NavigationAction
+) => {
+  // Call these in reverse order so last screens handle the event first
+  const reveredRoutes = [...routes].reverse();
```
Trancever, line 29:
```suggestion
  const reversedRoutes = [...routes].reverse();
```

packages/core/src/useOnPreventRemove.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const visitedRouteKeys: Set<string> =
+    // @ts-expect-error: add this property to mark that we've already emitted this action
+    action[VISITED_ROUTE_KEYS] ?? new Set<string>();
+
+  const beforeRemoveAction = {
+    ...action,
+    [VISITED_ROUTE_KEYS]: visitedRouteKeys,
+  };
+
+  for (const route of reveredRoutes) {
```
Trancever, line 40:
```suggestion
  for (const route of reversedRoutes) {
```

jaworek:
Has this feature been released? 
I'm using newest versions of react-navigation packages (bottom-tabs and native: 5.6.1, stack: 5.6.2) and I get error that this event does not exist. Even, when I used it nothing happens.

This event has been described in the documentation, but it is not mentioned in which 5.x version it has been added.
https://reactnavigation.org/docs/preventing-going-back

satya164:
It'll be released soon

jaworek:
Great, thanks! I was a little bit confused on why it was not working.

iMonk777:
Hi @satya164 

Can you please tell us when the fix is going to be released? I really need this event in my app :)

jaworek:
It is released in 5.7 version, which is already available.
Is it possible to mark this feature in documentation as 5.7+ instead of 5.x?

iMonk777:
Perfect, thank you @jaworek

tj-mc:
@jaworek I agree, I was quite confused looking at the 5.x docs, but not able to run this on 5.3. Lucky I stumbled across this thread.

micnguyen:
So glad I found this thread as well. I think @jaworek 's suggestion above is important. `5.x` is technically not correct for the current version of the docs on https://reactnavigation.org/docs/preventing-going-back/

dehypnosis:
Seems not wok in web platform

PR #8548, opened 2020-07-07
fix: add accessibilityState property

ssar504:
React Native 0.62 removes the deprecated `accessibilityStates` property and replaces it with an `accessibilityState` object instead. This PR adds the new property where needed, but leaves the old one in place for backwards compatibility. Without the change, the selected tab in BottomTabNavigator isn't announced properly in VoiceOver.

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabItem.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabItem.tsx
@@ -247,6 +247,7 @@ export default function BottomTabBarItem({
     testID,
     accessibilityLabel,
     accessibilityRole: 'button',
+    accessibilityState: { selected: focused },
     accessibilityStates: focused ? ['selected'] : [],
     style: [
       styles.tab,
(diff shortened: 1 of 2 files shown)


SleeplessByte:
(Adding this so it's searchable)

This should also fix the error:
```
React does not recognize the `accessibilityState` prop on a DOM element.
```

uxxman:
> (Adding this so it's searchable)
> 
> This should also fix the error:
> 
> ```
> React does not recognize the `accessibilityState` prop on a DOM element.
> ```
> 
> In order to remove that error from the console, perhaps a check can be added if the RN version is >= 0.62 (and thus **not** leave the old one in place)?

Still getting this on expo 38 with react-native 0.62

SleeplessByte:
@uxxman see my comment:

> In order to remove that error from the console, perhaps a check can be added if the RN version is >= 0.62 (and thus not leave the old one in place)?

They did leave the old one in place, so you'll still see the warning.

PR #8550, opened 2020-07-08
fix: make devtools global property configurable

kaladinlight:
The recent addition of `Object.defineProperty(global, 'REACT_NAVIGATION_REDUX_DEVTOOLS_EXTENSION_INTEGRATION_ENABLED')` resulted in the following type error: `TypeError: Cannot redefine property: REACT_NAVIGATION_REDUX_DEVTOOLS_EXTENSION_INTEGRATION_ENABLED` when attempting to mock `@react-navigation/native` in jest test suite. 

The secondary import of a react-navigation package containing the core package dependency throws a type error due to the property being non configurable. Update property to configurable to avoid the type error for use cases like module mocking in jest.

### DIFF
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -44,6 +44,7 @@ try {
           "Redux devtools extension integration can be enabled with the '@react-navigation/devtools' package. For more details, see https://reactnavigation.org/docs/devtools"
         );
       },
+      configurable: true,
     }
   );
 } catch (e) {


satya164:
Thanks for the PR, but I believe this should already suppress the error https://github.com/react-navigation/react-navigation/commit/1801a13323eff149fb6bc4e3c3f12422b401f178

kaladinlight:
> Thanks for the PR, but I believe this should already suppress the error [1801a13](https://github.com/react-navigation/react-navigation/commit/1801a13323eff149fb6bc4e3c3f12422b401f178)

Oh cool, didn't notice that. Pulling the new version to test and then we can close if that looks good. Will respond here. Thanks!

kaladinlight:
Awesome, that works out, sorry for the dupe!

PR #8557, opened 2020-07-09
fix: pass label position flag to label rendering in BottomTabBar

EyMaddis:
In the process of upgrading from v4, I noticed a regression. 
In the past, the function form of `tabBarLabel` did get an `orientation: 'landscape' | 'portrait'`, this is no longer the case.
However, when using a custom Text rendering, we need to apply a margin to the text in horizontal mode.

Since the orientation/horizontal state is decided based on internal heuristics, It is a huge pain with a high bug potential when reimplementing that detection myself.

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -63,7 +63,7 @@ export type BottomTabNavigationOptions = {
 
   /**
    * Title string of a tab displayed in the tab bar
-   * or a function that given { focused: boolean, color: string, orientation: 'horizontal' | 'vertical' } returns a React.Node to display in tab bar.
+   * or a function that given { focused: boolean, color: string, position: 'below-icon' | 'beside-icon' } returns a React.Node to display in tab bar.
    * When undefined, scene title is used. To hide, see tabBarOptions.showLabel in the previous section.
    */
   tabBarLabel?:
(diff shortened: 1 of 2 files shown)


packages/bottom-tabs/src/types.tsx
```diff
@@ -63,12 +63,16 @@ export type BottomTabNavigationOptions = {
...
+   * or a function that given { focused: boolean, color: string, orientation: 'horizontal' | 'vertical' } returns a React.Node to display in tab bar.
    * When undefined, scene title is used. To hide, see tabBarOptions.showLabel in the previous section.
    */
   tabBarLabel?:
     | string
-    | ((props: { focused: boolean; color: string }) => React.ReactNode);
+    | ((props: {
+        focused: boolean;
+        color: string;
+        orientation: 'horizontal' | 'vertical';
```
satya164, line 74:
```suggestion
        position: LabelPosition;
```

packages/bottom-tabs/src/views/BottomTabItem.tsx
```diff
@@ -30,7 +30,11 @@ type Props = {
    */
   label:
     | string
-    | ((props: { focused: boolean; color: string }) => React.ReactNode);
+    | ((props: {
+        focused: boolean;
+        color: string;
+        orientation: 'horizontal' | 'vertical';
```
satya164, line 36:
```suggestion
        position: LabelPosition;
```

packages/bottom-tabs/src/views/BottomTabItem.tsx
```diff
@@ -206,7 +210,11 @@ export default function BottomTabBarItem({
       );
     }
 
-    return label({ focused, color });
+    return label({
+      focused,
+      color,
+      orientation: horizontal ? 'horizontal' : 'vertical',
```
satya164, line 216:
```suggestion
      position: horizontal ? 'beside-icon' : 'below-icon',
```

packages/bottom-tabs/src/views/BottomTabItem.tsx
satya164: Also need to add the import for `LabelPosition` above

EyMaddis:
@satya164 thanks for the feedback, I added the changes and improved the documentation a bit.

PR #8561, opened 2020-07-09
feat: add a hook to update document title

satya164:


dehypnosis:
What a breaking change????
Can this be opt-outable?

Issue #1052, opened 2017-04-14
function for navigationOptions not working


bhoop:
Following the "Hello Mobile Navigation" example here: https://reactnavigation.org/docs/intro/, I reached the last part that uses a function for `navigationOptions` rather than a static object.

When I use this code, my title bar is blank:

    static navigationOptions = ({ navigation }) => ({ title: 'Chat With Someone' });

When I use this code, my title bar has the expected text in it:

    static navigationOptions = { title: 'Chat With Someone' };

My `package.json`, in case it's useful:

	"dependencies": {
		"react": "16.0.0-alpha.6",
		"react-native": "0.43.3",
		"react-navigation": "git+https://github.com/react-community/react-navigation.git#7edd9a7"
	},

iEricKoh:
I have same problem.

sibelius:
@bhoop I've tested against master do you still have the issue?

bhoop:
@sibelius When I change my `packages.json` dependency to:

    "react-navigation": "git+https://github.com/react-community/react-navigation.git"

    I get an error message `Unknown plugin "flow-react-proptypes"...`, which I assumed is related to #923 (the original reason I'm using the specific-commit dependency rather than the npm module). Should I file a separate issue for that?

sibelius:
@bhoop this is fixed on master

sibelius:
try with this commit: https://github.com/react-community/react-navigation/commit/4c55887990e1de2b92d400dafae19fb76c68da6a

bhoop:
@sibelius using that commit does indeed fix the issue, thanks.

HugoMatilla:
I had not the same error, but I wasn't able to show `headerRight` and other screen navigation options using the static object.

```js
static navigationOptions = {
    title: 'Home',
    headerRight: <Text>Info</Text>
  }
```

 As @sibelius points out and  @bhoop confirms, the commit `4c55887` fixes it.

seanmadi:
It looks like for some reason the function has changed from being a function that returns an object, to an object of functions. So this worked for me:
```
static navigationOptions = {
  title: ({ state }) => state.params.name,
};
```

jeronnex2:
I am having the same problem can you tell me how to use the commit in order to fix this?

PR #8570, opened 2020-07-10
feat: add Windows and macOS support

kaiguo:
react-native-gesture-handler doesn't have Windows/macOS implementation yet, use the web version as stubs for now to enable navigation features for Windows and mac.

Close https://github.com/microsoft/react-native-windows/issues/3884
Close https://github.com/microsoft/react-native-windows/issues/4044


elcssmouhsine:
could you also add react-native-macos support while at it :

    animationEnabled: Platform.OS !== 'web' && Platform.OS !== 'windows' && Platform.OS !== 'macos',

kaiguo:
> could you also add react-native-macos support while at it

Yep, added : )

elcssmouhsine:
react-native-gesture-handler does not support macos

[issue #1052](https://github.com/software-mansion/react-native-gesture-handler/issues/1052)

packages/drawer/src/views/Drawer.tsx
```diff
@@ -489,7 +495,9 @@ export default class DrawerView extends React.Component<Props> {
     // Check if the drawer width is available to avoid division by zero
     eq(this.drawerWidth, 0),
     0,
-    abs(divide(this.translateX, this.drawerWidth))
+    Platform.OS === 'windows' || Platform.OS === 'macos'
```
satya164, line 498: Maybe you could add a comment about why this is needed

packages/drawer/src/views/GestureHandlerCompat.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
```
satya164, line 1: Maybe this file could be named as `GestureHandlerStub` which would be more appropriate

packages/stack/src/views/GestureHandlerCompat.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
```
satya164, line 1: Maybe this file could be named as `GestureHandlerStub` which would be more appropriate

packages/stack/src/views/MaskedView.macos.tsx
```diff
@@ -0,0 +1,3 @@
+import StubbedMaskedView from './MaskedView';
```
satya164, line 1: Wouldn't this resolve to `MaskedView.native.tsx`? Maybe we need to rename `MaskedView.tsx` to `MaskedViewStub.tsx` and then re-export it.

```js
export * from './MaskedViewStub';
```

packages/drawer/src/views/Drawer.tsx
kaiguo: Moved the platform check to Overlay.tsx and added some comments.

kaiguo:
Thanks @satya164! Just pushed out another commit to address your feedbacks.

kaiguo:
Hey @satya164 do we know when the next release will be? Looks like the current release doesn't have this change and your refactor yet.

satya164:
I need to finish some other stuff before doing a release

Zorgatone:
Waiting for the release as well

lh9403:
Is MacOS supported now?

PR #8607, opened 2020-07-19
fix: replace url only when all the navigators are mounted

Federkun:
I suspect that the `getRootState` is called before potential other nested navigators are mounted.
What this PR try to solve is an issue with the initial url of the page that is replaced incorrectly. If I hit refresh on a page provided by a nested navigator (`/rootPage/nestedPage`), the url of the page will be replaced with the root's screen (`/rootPage`).

By deferring for a bit that operation, getRootState will provide the correct state, and `path` will have the nested path instead of the root screen path. I'm not sure if https://github.com/react-navigation/react-navigation/blob/f4180295bf22e32c65f6a7ab7089523cb2de58fb/packages/core/src/useNavigationBuilder.tsx#L376-L379 is related.

### DIFF
--- a/packages/native/src/__tests__/NavigationContainer.test.tsx
+++ b/packages/native/src/__tests__/NavigationContainer.test.tsx
@@ -99,6 +99,8 @@ it('integrates with the history API', () => {
     </NavigationContainer>
   );
 
+  jest.runAllTimers();
+
   expect(window.location.pathname).toBe('/feed');
 
   act(() => navigation.current?.navigate('Profile', { user: 'jane' }));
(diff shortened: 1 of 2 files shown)


Federkun:
Trying to replicate the problem in another repository, I realised what's the main cause. I don't mount my nested navigator together with the main one.

https://github.com/florello-tbox/react-navigation-example-of-wrong-url-replace/blob/master/packages/components/src/App.tsx#L27-L44

So, if I try to hit `nested/foo-bar` directly, i'm ( and I think correctly ) redirected to `nested`, because that path don't exists.. yet. Do you happen to have a suggestion for me to avoid this?

Federkun:
Can confirm that https://github.com/react-navigation/react-navigation/commit/af8b27414c8628570d946003f4fdff3341cb8954 works for me, thank you very much!

PR #8626, opened 2020-07-23
feat: allow style overrides for HeaderBackButton

andreibarabas:
Currently, if you use HeaderBackButton with the react-native-screens, there is an extra margin added that is not needed. This can be solved via the style overrides


felansu:
imagen
Maybe that error have relationship?

PR #8637, opened 2020-07-27
feat: Adds options for offset for useScrollToTop

mikaelbr:
Hi! Thanks for an awesome package. Just found the `useScrollToTop` hook which made my life much simpler.

There are some cases when scrolling to the top doesn't mean setting Y-axis to 0. There can be instances where you do some offset on the top due to animations, headers, or other content. I don't know how you feel about this feature, but thought I'd suggest it as a PR.

This introduces an optional options parameter to the custom hook for handling scroll to top. All it does is exposing the inner `ScrollOptions` and maps from differences for the different scrollable views.

I've set the options parameter as an optional without providing a default value to be able to check in the `scrollToTop` conditional wether or not options were provided. I'm not sure if this is necessary and if this causes some cases where you aren't able to scroll to top. Might be that we need to be more sophisticated with type signature?

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -14,6 +14,8 @@ type ScrollableWrapper =
   | { getNode(): ScrollableView }
   | ScrollableView;
 
+const defaultScrollOptions = { y: 0, animated: true };
+
 function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
   if (ref.current == null) {
     return null;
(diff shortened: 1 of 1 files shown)


satya164:
https://github.com/react-navigation/react-navigation/pull/6796

mikaelbr:
Bad searching on my part! Sorry for that. Thanks for suggestion for a solution. I thought about overriding the ref but figured it would be more hassle than this!

Perhaps this is a common enough pattern that it should be documented in https://reactnavigation.org/docs/use-scroll-to-top/ ?

PR #8665, opened 2020-08-01
Remove the now deprecated `useSafeArea` hook

codecog:
react-native-safe-area-context has deprecated the `useSafeArea` hook that sets the default insets in the bottom tabs component. `useSafeAreaInsets` is what is now in place and this PR updates the bottom-tabs package to accomodate that change.

This change will also update the react types packages as well as bumping the typescript version used with bottom tabs to 3.9.7

For reference: https://github.com/th3rdwave/react-native-safe-area-context/blob/master/src/SafeAreaContext.tsx#L134


satya164:
Thanks for the PR. However, dropping support for old versions is a breaking change and I can't merge this change until next major.

PR #8675, opened 2020-08-04
fix: Depricate scrollWithoutAnimationTo to support react-native 0.63.0 and later

reimertz:
Remove usage of depricated API and add deprication warning.


matthargett (commented):
can you add a test that captures and verifies the console output in this scenario?

PR #8700, opened 2020-08-10
fix: add aria-level to HeaderTitle

lakim:
We use `cypress-axe` to ensure accessibility compliance for the web version of our app (WCAG level A and AA) and it showed that the `HeaderTitle` is not compliant.

The rule that fails is `aria-required-attr` and it complains about the `h1` missing the `aria-level` attribute:
https://dequeuniversity.com/rules/axe/3.4/aria-required-attr?application=axeAPI

The solution is documented in `react-native-web`:
https://github.com/necolas/react-native-web/blob/master/packages/docs/src/guides/accessibility.stories.mdx#accessibilityrole-string

If you're interested, I could also add basic accessibility tests on `react-navigation` for the web using `cypress` and `cypress-axe`.

### DIFF
--- a/packages/stack/src/views/Header/HeaderTitle.tsx
+++ b/packages/stack/src/views/Header/HeaderTitle.tsx
@@ -21,6 +21,7 @@ export default function HeaderTitle({ tintColor, style, ...rest }: Props) {
   return (
     <Animated.Text
       accessibilityRole="header"
+      aria-level="1"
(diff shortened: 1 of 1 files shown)


lakim:
Other view using the [axe browser extension](https://www.deque.com/axe/browser-extensions/):

satya164 (approved):
Thanks!

PR #8703, opened 2020-08-10
test: move assertions into their own tests

oltrep:
This change makes the tests in NavigationStateUtils more specific by splitting tests with multiple assertions.

I also grouped the tests into describe blocks by the function they test. Let me know if that's something you want or not, I can revert it if needed.

This is my first contribution, but I plan to make more of these which would mostly consists on improving tests.


satya164 (approved):
Thanks! Seems there are some conflicts. Can you rebase it? :)

PR #8711, opened 2020-08-12
refactor: Use NavigateParams in CommonActions.navigate function signatures

andrekovac:
This PR improves the type of the `params` argument when using `CommonActions.navigate`.

Formerly it was just `object`, but the type `NavigateParams` type already existed, but just as a local type in  [core/src/getActionFromState.tsx](https://github.com/react-navigation/react-navigation/blob/1aadc79fb89177a2fff2dcd791d67a3c880009d0/packages/core/src/getActionFromState.tsx). It is now exported and used for the signature of the `CommonActions.navigate` action.

I opened a PR in the docs repo which describes this alternative function signature: https://github.com/react-navigation/react-navigation.github.io/pull/852

### DIFF
--- a/packages/core/src/getActionFromState.tsx
+++ b/packages/core/src/getActionFromState.tsx
@@ -1,10 +1,8 @@
-import type { PartialState, NavigationState } from '@react-navigation/routers';
-
-type NavigateParams = {
-  screen?: string;
-  params?: NavigateParams;
-  initial?: boolean;
-};
+import type {
+  PartialState,
+  NavigationState,
+  NavigateParams,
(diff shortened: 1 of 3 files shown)


satya164:
Thanks for the PR. The screen property can be specified when setting params with any action, not just navigate. So this change wouldn't be correct.

What we will need to do is to expose a helper type to specify child screens for nested navigators and document it.

satya164:
Closing as per the comment. Feel free to create an RFC about exporting a helper type if you have ideas regarding it https://github.com/react-navigation/rfcs

PR #8723, opened 2020-08-14
chore: enable faster bundling on preview builds

byCedric:
This enables a new interface with Metro, [outlined in this tweet](https://twitter.com/expo/status/1293613036482121728). It's still in a sort-of beta and that's why it's opt-in. We would love to see how this improves bundling on the React Navigation example app, and/or if you encounter anything special.

Feel free to close this as well if you don't want to do this right now!


satya164 (approved):
Thank you!

PR #8770, opened 2020-08-25
refactor: remove undefined assignment in switch router

oltrep:
`undefined` should be left to the javascript compiler.

Originally, I opened this PR because I thought there was an inconsistency between the stack and switch routers, in the way they handle params. Turns out it's all fine, but I found this case where `undefined` is assigned to a variable and I thought I could submit a small refactor for it.

### DIFF
--- a/packages/core/src/routers/SwitchRouter.js
+++ b/packages/core/src/routers/SwitchRouter.js
@@ -19,7 +19,7 @@ export default (routeConfigs, config = {}) => {
   const getCustomActionCreators =
     config.getCustomActionCreators || defaultActionCreators;
 
-  const initialRouteParams = config.initialRouteParams;
+  const { initialRouteParams } = config;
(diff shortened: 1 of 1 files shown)


packages/core/src/routers/SwitchRouter.js
```diff
@@ -71,13 +71,13 @@ export default (routeConfigs, config = {}) => {
...
         routeName,
-        params,
+        ...(params ? { params } : {}),
       };
     }
     return {
       key: routeName,
       routeName,
-      params,
+      ...(params ? { params } : {}),
```
satya164, line 80: This doesn't seem to change any behaviour and seems to be more code. Can you confirm?

oltrep: Since `params` can be `null` this will avoid having the property defined! It's how it's done in StackRouter IIRC :)

satya164 (approved):

PR #8778, opened 2020-08-27
test: add test case for StateUtils.pop

oltrep:
From the current implementation of `NavigationStateUtils.pop`, a new
test case was added to document the behavior when popping when the index
is at the first route and there are multiple routes.

Question: what is the reason why the top route does not get removed in that specific case? I'm not that familiar with react-navigation, but my basic intuition would probably have told me that popping in that situation would have removed the top-most route.

Thanks in advance for your feedback :)


satya164 (approved):

satya164:
Thanks!

PR #8780, opened 2020-08-27
test: document set params action for routers

oltrep:
These tests document the behavior of using set params action for the
different routers. Even if the action is named `SET_PARAMS`, it merges
the params with the previous one: this behavior matches the `setState`
method used in React's components.

### DIFF
--- a/packages/core/src/routers/__tests__/Routers.test.js
+++ b/packages/core/src/routers/__tests__/Routers.test.js
@@ -114,6 +114,51 @@ Object.keys(ROUTERS).forEach((routerName) => {
       );
       expect(state0.routes[state0.index].params.foo).toEqual(42);
     });
+
+    it('merges existing params when set params on existing state', () => {
+      const Screen = () => <div />;
+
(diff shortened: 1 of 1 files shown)


packages/core/src/routers/__tests__/StackRouter.test.js
```diff
@@ -2063,4 +2060,46 @@ describe('StackRouter', () => {
...
+
+    const newState = router.getStateForAction(
+      NavigationActions.setParams({ key: 'id-0', params: { b: 2 } }),
+      state
+    );
+
+    expect(newState.routes[newState.index].params).toEqual({ b: 2 });
+  });
+
+  it('should merge params when setting params during init', () => {
```
oltrep, line 2089: Is this the expected behavior? Let me know what should happen when getting the initial state from a set-params action. Right now, the current behavior is to merge so that's why I've written the test like this, but let me know if I need to change that too.

satya164 (approved):
Thanks!

Issue #8399, opened 2020-06-09
export type doesn't work


KJoobin:
**Current Behavior**

<img width="1198" alt="스크린샷 2020-06-09 오후 3 59 39" src="https://user-images.githubusercontent.com/45376611/84116204-55258900-aa6a-11ea-8d8d-43fa4eba88ef.png">

when i used stack navigation prop type, i has found error in 
```
import {StackNavigationProp} from '@React-navigation/stack'
```

and i see 

<img width="1680" alt="스크린샷 2020-06-09 오후 3 59 44" src="https://user-images.githubusercontent.com/45376611/84116325-8ef68f80-aa6a-11ea-8cbc-a0fc6ae591ec.png">
- What code are you running and what is happening?
- Include a screenshot or video if it makes sense.

**Expected Behavior**

when  remove type
```
export { 
    StackNavigationOptions,
    StackNavigationProp,
    StackScreenProps, 
    StackHeaderProps, 
    StackHeaderLeftButtonProps,
    StackHeaderTitleProps, 
    StackCardInterpolatedStyle, 
    StackCardInterpolationProps, 
    StackCardStyleInterpolator, 
    StackHeaderInterpolatedStyle, 
    StackHeaderInterpolationProps, 
    StackHeaderStyleInterpolator, 
} from './types';
```

it was work
what is the problem?
- What do you expect should be happening?
- Include a screenshot or video if it makes sense.

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | iOS
| @react-navigation/native       | 5.5.0
| @react-navigation/stack        | 5.4.1
| react-native-gesture-handler   | ~1.6.0
| react-native-safe-area-context | 0.7.3
| react-native-screens           | ~2.2.0
| react-native                   | in expo
| expo                           | ~37.0.3
| node                           |10.19.0
| npm or yarn                    | 1.16.0

raajnadar:
Upgrade typescript & @babel/core version clear the cache and run the app again the warning will go away.

Youn-ha:
Hi @raajnadar , I thank a lot for your informative reply. I tried running
```
yarn upgrade @babel/core --latest
yarn upgrade typescript --latest
yarn cache clean
```
though, the error remains.
I restarted my IDE and nothing changes.
Do you have any idea?
Thank you.

@react-navigation/stack/lib/typescript/src/index.d.ts
[...]

PR #8795, opened 2020-09-02
fix: export type error on typescript/babel projects

eleazar0425:
I found out that the #8399 is due to a bug with babel. You can read more about here: https://github.com/vercel/next.js/issues/7882. 
But actually we can fix this import error by just deleting the keyword `type` when exporting. Or using `export *` instead of declaring every specific type.
I must add that this issue is still happening even in a fresh new project, but this little change has fixed it.

### DIFF
--- a/packages/stack/src/index.tsx
+++ b/packages/stack/src/index.tsx
@@ -48,7 +48,7 @@ export { default as useGestureHandlerRef } from './utils/useGestureHandlerRef';
 /**
  * Types
  */
-export type {
+export {
   StackNavigationOptions,
   StackNavigationProp,
   StackScreenProps,


raajnadar:
This is not a fix.

satya164:
Thanks for the PR, but this is not a desirable change, so i'll close the PR.

There's no bug with Babel, you need to make sure that you're using latest version of metro babel preset as mentioned in docs https://reactnavigation.org/docs/troubleshooting#im-getting-syntaxerror-in-react-navigationxxxxxxtsx-or-syntaxerror-xxxreact-navigationxxxxxxtsx-unexpected-token

The linked next.js issue is exactly opposite of this, where the issue was fixed after adding `type`.

PR #8798, opened 2020-09-03
fix: useThenable doesn't always resolve

jiwansvmx:
When  app is run in debug mode and has deeplinks enabled, it's stuck at the fallback ui.  Noticed this when app runs in debug mode with chrome debugger, not an issue when safari inspector is connected.
Upon digging a little, found the following as the reason, I could be off:
`useThenable` doesn't resolve if the **input promise resolves after**
 ```js
 const [state, setState] = React.useState<[boolean, T | undefined]>([
    resolved,
    value,
  ]);
```
but **before the following  inner effect**  has run 
```js
React.useEffect(() => {
    let cancelled = false;
    const resolve = async () => {
      let result;
      try {
        result = await promise;
      } finally {
        if (!cancelled) {
          setState([true, result]);
        }
      }
    };
    if (!resolved) {
      resolve();
    }   
    
    return () => {
      cancelled = true;
    };
  }, [promise, resolved]);

```

### DIFF
--- a/packages/native/src/useThenable.tsx
+++ b/packages/native/src/useThenable.tsx
@@ -34,6 +34,10 @@ export default function useThenable<T>(create: () => PromiseLike<T>) {
 
     if (!resolved) {
       resolve();
+    } else {
+      if (!cancelled) {
+        setState([true, value]);
+      }
     }
 
     return () => {


satya164 (changes requested):
Thanks for the PR, but I don't see how this change fixes an issue. You're setting `resolved` state to `true` if it's already true.

As mentioned in the docs, there are many issues when using chrome debugger: https://reactnavigation.org/docs/troubleshooting#app-is-not-working-properly-when-connected-to-chrome-debugger, these are not bugs in React Navigation code.

jiwansvmx:
@satya164 

> Thanks for the PR, but I don't see how this change fixes an issue. You're setting resolved state to true if it's already true.

The scenario i am describing is where `resolved` isn't true initially,  thus `useThenable` returns false but `resolve` becomes `true` before the inner effect has ever run.

This fixes the issue because calling setState would rerun the hook .

b4stien:
I've been bitten by the same bug during testing, with `jest` and `react-test-renderer`, this is not an issue with chrome debugger but indeed a bug in React Navigation code.

What happens is what @jiwansvmx described, with:

1. Our `resolved` flag is initialized to `false` and `state.resolved` will be initialized as false as well because `promise` (https://github.com/react-navigation/react-navigation/blob/main/packages/native/src/useThenable.tsx#L10) has not been resolved yet
2. Our `promise` resolve, setting our `resolved` flag to `true`
3. The effect executes for the first time, bypassing the call to `setState` because our `resolved` flag is already set to `true`

Rewritting the effect to check `state` and not the `resolved` flag also works, it would be written as such: 

```
  React.useEffect(() => {
    let cancelled = false;

    const resolve = async () => {
      let result;

      try {
        result = await promise;
      } finally {
        if (!cancelled) {
          setState([true, result]);
        }
      }
    };

    if (!state[0]) {
      resolve();
    }

    return () => {
      cancelled = true;
    };
  }, [promise, state]);
```

If you're willing I can submit a PR with this code.

PR #8805, opened 2020-09-04
feat: add optional screens per navigator

WoLewicki:
Changes done here will work properly with https://github.com/software-mansion/react-native-screens/pull/624 merged and released. The documentation of `screensEnabled` and `activeLimit` props should also be added. It also enabled `Screens` in iOS stack-navigator by default.

New things:
- `screensEnabled` - prop for all navigators that can be switched by user. It controls if the `react-native-screens` are used by the navigator.
- `activeLimit` - prop that tells how many screens should stay active after transition. It can be set by user, defaults to  1 for normal mode and 2 for mode = “modal”.

TBD:
transparent screens could be used to decide if there should be more active screens (can be set by user too theoretically so not big of a problem)


jkadamczyk (approved):

packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -378,6 +387,9 @@ export default class CardStack extends React.Component<Props, State> {
       onGestureStart,
       onGestureEnd,
       onGestureCancel,
+      detachInactiveScreens = shouldUseActivityState
+        ? Platform.OS !== 'ios'
+        : Platform.OS !== 'ios' && mode !== 'modal',
```
satya164, line 392:
```suggestion
      detachInactiveScreens = Platform.OS === 'ios'
        ? false // Disable `react-native-screens` on iOS by default since it's buggy
        : shouldUseActivityState || mode !== 'modal', // Enable on new versions of screens or for non modals on older versions
```

satya164 (approved):

PR #8808, opened 2020-09-04
Fixes react-navigation/react-navigation#8807

### LINKED ISSUE #8807: item-specific style for DrawerItem
**Current Behavior**

- With React Navigation 5.x, there is no way to override the `DrawerItem` style for a specific item.

**Expected Behavior**

- `DrawerItem` should accept a `style` prop. One use case is to add a divider line below a specific drawer item.

I have a patch which I will turn into a PR very soon.

MarkCSmith:
feat: Support a style property in the drawer navigator's Screen options.


satya164:
Thanks the PR. I'm closing this since item style can already be specified in v5 (though not specific to screens), but this feature was added in v6 as `drawerItemStyle`

PR #8810, opened 2020-09-04
feat: Add support for an onItemPress property to DrawerItemList.

### LINKED ISSUE #8809: DrawerItemList should support an onItemPress callback function
**Current Behavior**

- With React Navigation 5.x an `onItemPress` hook is missing.

**Expected Behavior**

- `DrawerItemList` should accept an `onItemPress` prop: a function that can override the default navigation behavior. One use case is to make it easy to implement drawer items that do not navigate to another screen, e.g., to include an item that opens an URL in an external browser.

**How to reproduce**

I have a patch which I will turn into a PR very soon.

**Your Environment**

| Android or iOS                 | any
| @react-navigation/native       |5.7.3
| @react-navigation/drawer       |5.9.0
| react-native-reanimated        |1.13.0
| react-native-gesture-handler   |1.7.0
| react-native-safe-area-context |3.1.7
| react-native-screens           |2.10.1
| react-native                   |0.62.2
| expo                           |n/a
| node                           |14.5.0
| npm                    |6.14.5

MarkCSmith:
If the onItemPress function returns true, the default navigation
behavior is suppressed. Fixes #8809.


lanman2000:
Yes yes yes please get this PR in! It is a massive pain to always have to override the whole component for the item

ger88555:
I second this. This event is the only way to tell when a screen is accessed from the drawer menu, and not through the use of the ```navigate``` method, coming back from another screen, etc.

satya164:

The current version of darwer support a `drawerItemPress` event so I think this isn't needed anymore. Thanks for the PR.

PR #8816, opened 2020-09-07
feat: add detachInactiveScreens prop for bottom tabs and drawer

WoLewicki:
Added optional screens enabling for v4 bottom-tabs and drawer navigators.


matthargett (commented):
Can you add/update tests that pin the behavior of the prop getting set/retained correctly?

satya164 (changes requested):
Can you also update the prop names to match the changes we did in v5?

satya164 (approved):

satya164 (changes requested):
Can you make sure the CI passes?

PR #8819, opened 2020-09-07
chore: update docs website URLs to avoid the redirect

Simek:
After the docs website migration to the Docusaurus V2 the URL format have changed. Using the old URLs format results in redirect, which can be avoided by updating the docs website links. This PR updates all outdated URLs that I was able to find.

Additionally I have updated the Material Design website link which was changed too.

### DIFF
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -46,7 +46,7 @@ const getDefaultDrawerWidth = ({
   /*
    * Default drawer width is screen width - header height
    * with a max width of 280 on mobile and 320 on tablet
-   * https://material.io/guidelines/patterns/navigation-drawer/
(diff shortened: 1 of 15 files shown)


packages/drawer/src/views/DrawerView.tsx
```diff
@@ -46,7 +46,7 @@ const getDefaultDrawerWidth = ({
   /*
    * Default drawer width is screen width - header height
    * with a max width of 280 on mobile and 320 on tablet
```
Simek, line 48: It looks like the default values are no longer listed on the [linked page](https://material.io/components/navigation-drawer) nor on the [Android specific page](https://material.io/develop/android/components/navigation-view). 

Let me know if I should removed this link from the comment.

satya164 (approved):
Thank you for the PR

PR #8828, opened 2020-09-10
fix: prevent error if no setNativeProps

ethanneff:
Prevents this error: 

Screen Shot 2020-09-09 at 7 30 13 PM

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -14,7 +14,7 @@ import Color from 'color';
 
 import CardSheet from './CardSheet';
 import {
-  PanGestureHandler,n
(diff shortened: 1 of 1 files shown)


satya164:
Why/when does `setNativeProps` not exist?

ethanneff:
This error began to happen after I updated to react-navigation 5 from 4. My react native app is about 2 years old with all the latest dependencies. 

I'm not too sure what is causing this error. I just noticed it after updating to react-navigation 5

satya164:
Closing since there's no demo/explanation on how to repro the issue and I don't think ignoring it isn't the right solution as it can break the intended behavior. If you're still facing the issue, please open an issue with a demo so we can investigate how to fix it.

PR #8834, opened 2020-09-11
feat: add pressColor and pressOpacity props to drawerItem

petra-daneva:
- Include pressColor property that could override the default grey value presented in TouchableItem


satya164:
Thanks for the PR!

PR #8844, opened 2020-09-14
test: fix usage of routerTestHelper back action

oltrep:
Found this case were the key gets set to `{ key: null }` instead of just being `null`. Should I just call `back()` at this point? Because I don't think there is a difference between a `null` vs `undefined` key


matthargett:
Looks like CI failed due to a setup/network failure:
```Warning: Permanently added the RSA host key for IP address '140.82.113.3' to the list of known hosts.
Enter passphrase for key '/home/circleci/.ssh/id_rsa': 

Too long with no output (exceeded 10m0s)```

matthargett (approved):

satya164 (approved):

PR #8849, opened 2020-09-15
fix: add missing types to drawer export

chrisdrackett:
these are useful when making a custom navigator that you want to share the config and events of the current drawer.

### DIFF
--- a/packages/drawer/src/index.tsx
+++ b/packages/drawer/src/index.tsx
@@ -23,6 +23,8 @@ export { default as useIsDrawerOpen } from './utils/useIsDrawerOpen';
  * Types
  */
 export type {
+  DrawerNavigationConfig,
+   DrawerNavigationEventMap,
   DrawerNavigationOptions,
   DrawerNavigationProp,
   DrawerScreenProps,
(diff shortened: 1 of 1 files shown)


satya164:
If you're making a custom navigator, then I don't think you should be using the same types as the built-in drawer navigator. Also, importing drawer navigator would pull in all its dependencies which is unnecessary for your custom navigator.

I get that it maybe desirable for you to align the API, but you'll want to add new things to our navigator, and not support existing things, so the default type won't be desirable. If you add some prop and we later added a different one with same name, it's going to give you an error, which will force us to add new props under semver major.

chrisdrackett:
@satya164 my custom navigation is at the end of the day is just for our app (and works 99% the same as the drawer navigator) I just needed to update some things that I couldn't update without cloning the entirety of the drawer code.

satya164:
I'm just saying that exposing it would mean we'll have less flexibility in changing them since they can be considered breaking changes, which I'd like to avoid. Since you implemented a custom drawer, if suggest to copy the types as well.

You can also do something like this:

```ts
type DrawerProps = React.ComponentProps<ReturnType<typeof createDrawerNavigator<ParamListBase>>['Navigator']>
```

PR #8857, opened 2020-09-16
Fix: blink issue for floatingHeader (Android)

Goye:
Solves the blink issue when you are using a custom header (`headerMode="float"`), if could you please take a look 🙏  @Ashoat 

### Before: 
before

### After:
after

### DIFF
--- a/packages/stack/src/views/Stack/CardStack.tsx
+++ b/packages/stack/src/views/Stack/CardStack.tsx
@@ -420,7 +420,7 @@ export default class CardStack extends React.Component<Props, State> {
                     headerShown = isParentHeaderShown === false,
                   } = options;
 
-                  if (headerTransparent || headerShown === false) {
+                  if (!headerTransparent || headerShown === false) {
                     return true;
                   }
 


packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -420,7 +420,7 @@ export default class CardStack extends React.Component<Props, State> {
                     headerShown = isParentHeaderShown === false,
                   } = options;
 
-                  if (headerTransparent || headerShown === false) {
+                  if (!headerTransparent || headerShown === false) {
```
Ashoat, line 423: The logic here is deciding whether to position the header using `position: 'absolute'` or `position: 'relative'`. The reason that we check `headerTransparent` is because for a transparent header, we want to layout the header over whatever is in the background. `position: 'relative'` means that the header will take up its own vertical space, which means it won't work well for the transparent-header scenario.

Ashoat (changes requested):
Can you clarify a bit more about the issue you're experiencing? Do you know if the value of `isFloatHeaderAbsolute` is changing when you navigate back and forth?

From the way the app looks, it seems like you want to have an absolutely-positioned header at all times. Have you tried setting the `headerTransparent` option?

Ashoat:
In #8858, I noticed you observed:

> Seeing the diff between versions I saw that you are adding this validation `!headerTransparent` in the hasAbsoluteHeader but not in `isFloatHeaderAbsolute` and that is the problem

The reason for this is that `hasAbsoluteHeader`'s job is to determine where the `Card` needs to have a `marginTop`. We apply a `marginTop` when both of the following two conditions are met:

1. The header is not relatively positioned. A relatively positioned header does not need a `marginTop` since the header itself will push down the `Card`.
2. The header is not transparent. In the case of `headerTransparent` we want to allow the user to render something in the `Card` that will appear behind the header, so we avoid applying `marginTop`.

Note that `hasAbsoluteHeader` has to be determined individually for every single `Card`, but on the other hand we only determine `isFloatHeaderAbsolute` once for the whole `CardStack`. This is because we render a single float header for the whole stack. When rendering the float header, we decide `isFloatHeaderAbsolute` by checking the top two scenes in the scene stack. If neither of the top two scenes need an absolutely positioned header, then we render the header relatively.

But if there are more than two scenes in the stack, then it's possible that a scene earlier in the stack needs an absolutely positioned header, but the later scenes do not. In this situation, we apply a `marginTop` to the `Card` that needs an absolutely positioned header, even if the current header is not absolutely positioned, since when that `Card` will be visible it will need the `marginTop`.

My guess is the flicker you are seeing is occurring because the value of `isFloatHeaderAbsolute` is being flipped. Some ideas for you:

1. You could try and enforce that the header is always absolutely positioned. Have you tried setting the `headerTransparent` option for all of your screens?
2. If you need the header is be absolutely positioned sometimes, but relatively positioned at other times, you could see what happens if you only apply the `marginTop` when `isFloatHeaderAbsolute` flips. In other words, you could remove the `!headerTransparent` check from [this line](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Stack/CardStack.tsx#L589).

Goye:
Hey Ashoat,

I hope you are having a great week and going back to the question, yes I tried to set `headerTransparent` for all screens and this is happening and it is exactly for by this line:

```
hasAbsoluteHeader={
  isFloatHeaderAbsolute && !headerTransparent
}
```



These are the screenOptions that I am using it: 

```
const screenOptions: StackNavigationOptions = {
  ...screensTransition,
  gestureEnabled: false,
  headerShown: false,
  cardStyle: {
  },
````

This is my RootStack navigator:
```
<RootStack.Navigator
    initialRouteName="Home"
    headerMode="float"
    screenOptions={screenOptions}
>
    <RootStack.Screen
        name="Home"
        component={Home}
        options={getHeaderOptions('Home')}
    />
   <RootStack.Screen name="Auth" component={AuthStackScreen} />
</RootStack.Navigator>
```

And this is the AuthStackScreen: 
```
<AuthNavigatorContext.Provider value={props}>
   <AuthStack.Navigator screenOptions={screenOptions}>
     <AuthStack.Screen name="parentalGate" component={ParentalGate} />
   </AuthStack.Navigator>
</AuthNavigatorContext.Provider>
```

Regarding your assumption, I am sending the prop `headerShown: false` because we are using a custom header but as you said above when I am in the Home screen the value is `true` but when I am changing to another screen this validation `headerShown = isParentHeaderShown === false` is returning `false` so it is changing between relative and absolute and that is the reason for the flicker

When I am having a `headerMode === 'float'` it should not be always absolute?

Finally about your suggestion, right now the change that I made in the PR is forcing the position absolute so in my case is working fine.

Ashoat:
Yes, as mentioned in my comment above setting `headerTransparent` means the `marginTop` will not be applied. You will have to apply the `marginTop` yourself in each of your screens.

Goye:
so you think we can close the issue & PR? I opened the PR because in the version that I had (stack@5.4.1) that was not necessary @Ashoat

Ashoat:
Yeah, I think the PR should be closed. If the problem has been resolved feel free to close the issue as well.

PR #8862, opened 2020-09-17
tabBarBadge accept boolean type

bruno02100:
According to the docs the tabBarBadge accepts a boolean, it works, but the type does not accept it.

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -106,7 +106,7 @@ export type BottomTabNavigationOptions = {
   /**
    * Text to show in a badge on the tab icon.
    */
-  tabBarBadge?: number | string;
+  tabBarBadge?: number | string | boolean;
 
   /**
    * Accessibility label for the tab button. This is read by the screen reader when the user taps the tab.


packages/bottom-tabs/src/types.tsx
```diff
@@ -106,7 +106,7 @@ export type BottomTabNavigationOptions = {
   /**
    * Text to show in a badge on the tab icon.
    */
-  tabBarBadge?: number | string;
+  tabBarBadge?: number | string | boolean;
```
ccheney, line 109: this should actually be:

`tabBarBadge?: number | string | true;`

Passing in a `false` bool will actually still render the dot

ccheney (changes requested):

satya164:
Thanks for the PR. However, `tabBarBadge` doesn't accept a boolean.

Workshed:
@satya164 can i clarify this one - I was about to make the same PR. The docs here do indeed suggest that it accepts a boolean (and in testing it does indeed accept a boolean, or at least a `true` value): https://reactnavigation.org/docs/material-bottom-tab-navigator/#tabbarbadge

Workshed:
How else would you get a badge without a value inside it?

PR #8868, opened 2020-09-18
fix: typo in logic of getStateFromPath

Titozzz:
What just randomly reading, but I think there's a typo
```
['ab', 'aba'].sort((a, b) => {
      if (a.startsWith(b)) {
        return -1;
      }

      if (b.startsWith(a)) {
        return 1;
      }
});
(2) ["aba", "ab"]
['ab', 'a'].sort((a, b) => {
      if (a.startsWith(b)) {
        return -1;
      }

      if (b.startsWith(a)) {
        return 1;
      }
});
(2) ["ab", "a"]
```


satya164 (approved):
Thanks!

PR #8871, opened 2020-09-18
feat: allow editing styles on the main element of drawer

chrisdrackett:
I want to have a shadow on my drawer element on mobile but not on tablet. If I do this with a component wrapping the drawer I loose navigation state within the drawer content. But because overflow is `hidden` in this component and I can't change that I don't think it is currently possible to do without making a fully custom drawer.


satya164:
Hello, the PR got closed automatically due to some changes in the main branch. Would you be able to send a new PR with more info about what you mean by "If I do this with a component wrapping the drawer I loose navigation state within the drawer content"? Thank you.

PR #8875, opened 2020-09-20
feat: specify badge color

caseyg1204:
Optionally specify badge text color. We use a green color for badge that resulted in black text, but we want white text.


caseyg1204:
@satya164 can you please take a look at this when you have a moment?

cruzach:
@caseyg1204 I know this is really late to followup, but in case anyone else is also looking at this PR- you can set the text and background colors of the badges with [`tabBarBadgeStyle`](https://reactnavigation.org/docs/bottom-tab-navigator/#tabbarbadgestyle)

PR #8895, opened 2020-09-23
feat: add an unhandled action listener

ahmadj-levelbenefits:
Often developers miss these console messages: this allows missed routes to be emitted to whatever event logger users prefer.

### DIFF
--- a/packages/core/src/__tests__/BaseNavigationContainer.test.tsx
+++ b/packages/core/src/__tests__/BaseNavigationContainer.test.tsx
@@ -722,7 +722,7 @@ it("throws if the ref hasn't finished initializing", () => {
   render(element);
 });
 
-it("invokes the unhandled action listener with the unhandled action", () => {
+it('invokes the unhandled action listener with the unhandled action', () => {
   const ref = React.createRef<NavigationContainerRef>();
   const fn = jest.fn();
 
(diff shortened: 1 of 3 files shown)


packages/core/src/__tests__/BaseNavigationContainer.test.tsx
```diff
@@ -721,3 +721,39 @@ it("throws if the ref hasn't finished initializing", () => {
...
+
+  act(() => ref.current!.navigate('bar'));
+  act(() => ref.current!.navigate('baz'));
+
+  expect(fn).toHaveBeenCalledWith({
+    payload: {
+      name: 'baz',
+    },
+    type: 'NAVIGATE',
+  });
```
ahmadj-levelbenefits, line 758: Would anyone be able to tell me how to silence the warnings this test emits? The test output is otherwise pristine, and I'd like to preserve that.

satya164 (changes requested):
Thanks for the PR!

Just 2 small changes:
1. Can you rename `onUnhandledNavigationAction` to `onUnhandledAction`?
2. If a listener is specified, I think we should return instead of showing redbox, since they dev might want to handle it in a different way

ahmadj-levelbenefits:
Cool - to avoid a name collision prefixed the internal one with `default`.

satya164:
Thank you! Looks like the CI is failing due to prettier errors. Can you run `yarn lint --fix` and push the changes so the CI passes?

ahmadj-levelbenefits:
Ping @satya164

satya164 (approved):
Thanks!

ahmadj-levelbenefits:
Thank you!

PR #8920, opened 2020-09-29
fix: NavigationEvents subscribe events on new nav state

oltrep:
If the navigation prop changes, the NavigationEvents component was not
subscribing to the new navigation value. These changes fix this problem and add a
test to verify that behavior.

Let me know if that is something you would need in 5.x, I can submit a fix for that version too.


satya164 (approved):
Thanks!

satya164:
I think RN5 doesn't need a fix for this.

Issue #8821, opened 2020-09-08
Make react-native-vector-icons optional


pradel:
**Current Behavior**

I want to use `@react-navigation/material-bottom-tabs` without `react-native-vector-icons`, but the following error is popping up.

```
error: bundling failed: Error: Unable to resolve module `react-native-vector-icons/MaterialCommunityIcons` from `node_modules/@react-navigation/material-bottom-tabs/src/views/MaterialBottomTabView.tsx`: react-native-vector-icons/MaterialCommunityIcons could not be found within the project.
```

Currently you can use react-native-paper without `react-native-vector-icons` by using the `babel-plugin-optional-require` babel plugin. https://callstack.github.io/react-native-paper/getting-started.html

![Screenshot 2020-09-08 at 17 33 09](https://user-images.githubusercontent.com/5749437/92497295-7593e980-f1f9-11ea-9fbd-62fb6fd5a8cf.png)

**Expected Behavior**

I should be able to use this package without `react-native-vector-icons` as I do not render any icon.

This is the way react-native-paper is using to allow this behavior https://github.com/callstack/react-native-paper/blob/212aa73715f157e1a77f8738859a608a543ba04c/src/components/MaterialCommunityIcon.tsx#L14.

**How to reproduce**

Create a new react-native project (without expo) and just add and use `@react-navigation/material-bottom-tabs` without an icon.

**Your Environment**

| software                               | version |
| -------------------------------------- | ------- |
| iOS or Android                         | both
| @react-navigation/native               | 5.3.0
| @react-navigation/material-bottom-tabs | 5.2.2
| react-native-paper                     | 2.16.0
| react-native                           | 0.61.2
| expo                                   | none
| node                                   | 12.16.2
| npm or yarn                            | yarn

pradel:
The issue is still happening in the latest version of the package see https://github.com/react-navigation/react-navigation/blob/main/packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx#L4

rusakovic:
Confirm the issue.

software | version
-- | --
iOS or Android | iOS
@react-navigation/native | ^5.7.3
@react-navigation/material-top-tabs | ^5.2.16
react-native | https://github.com/expo/react-native/archive/sdk-38.0.2.tar.gz
expo | ^38.0.9
node | 12.16.2
npm or yarn | npm

```
[...]

PR #8936, opened 2020-10-06
feat: make react-native-vector-icons optional

kaushil111:
Make react-native-vector-icons optional #8821 
Referenced from https://github.com/callstack/react-native-paper/blob/4b26429c49053eaa4c3e0fae208639e01093fa87/src/components/MaterialCommunityIcon.tsx#L14

https://callstack.github.io/react-native-paper/getting-started.html
> To get smaller bundle size by excluding modules you don't use, you can use our optional babel plugin. The plugin automatically rewrites the import statements so that only the modules you use are imported instead of the whole library. Add react-native-paper/babel to the plugins section in your babel.config.js for production environment. It should look like this:
> ```
> module.exports = {
>   presets: ['module:metro-react-native-babel-preset'],
>   env: {
>     production: {
>       plugins: ['react-native-paper/babel'],
>     },
>   },
> };
> ```
> If you created your project using Expo, it'll look something like this:
> ```
> module.exports = function(api) {
>   api.cache(true);
>   return {
>     presets: ['babel-preset-expo'],
>     env: {
>       production: {
>         plugins: ['react-native-paper/babel'],
>       },
>     },
>   };
> };
> ```

### DIFF
--- a/packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
+++ b/packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { StyleSheet, Platform } from 'react-native';
 import { BottomNavigation, DefaultTheme, DarkTheme } from 'react-native-paper';
-import MaterialCommunityIcons from 'react-native-vector-icons/MaterialCommunityIcons';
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thank you!

Issue #8941, opened 2020-10-08
Linking prefixes for wild cards subdomain


hosseinmd:
I have a web url deep link  'https://*.example.com'
I couldn't define wild cards subdomain for prefixes
**Current Behavior**
- We should define all prefixes

**Expected Behavior**

- Ability to define wild cards subdomain

**How to reproduce**

- I think we could do it with regex

**Your Environment**

| @react-navigation/native       |5.7.5
| react-native                   |v63

PR #8942, opened 2020-10-08
feat: support wildcard string prefixes

hosseinmd:
https://github.com/react-navigation/react-navigation/issues/8941
Prefixes should be more flexible for situations like wild card subdomain. On android and IOS we can define wild cards by * but react-navigation does not work, In this PR I added support for RegExp Prefixes.

For Example
```js
{
  prefixes: [
    /^[^.s]+.example.com/g
 ],
}
```
I tested this work well.

### DIFF
--- a/packages/native/src/types.tsx
+++ b/packages/native/src/types.tsx
@@ -27,8 +27,16 @@ export type LinkingOptions = {
    * The prefixes are stripped from the URL before parsing them.
    * Usually they are the `scheme` + `host` (e.g. `myapp://chat?user=jane`)
    * Only applicable on Android and iOS.
+   *
+   * @example
+   *  {
+   *    prefixes: [
+   *      "https://example.com", // Exact
+   *      /^[^.s]+.example.com/g // Match with all subdomain and scheme,
+   *    ],
+   *  }
    */
-  prefixes: string[];
+  prefixes: (string | RegExp)[];
   /**
    * Config to fine-tune how to parse the path.
    *
--- a/packages/native/src/types.tsx
+++ b/packages/native/src/types.tsx
@@ -29,14 +29,12 @@ export type LinkingOptions = {
    * Only applicable on Android and iOS.
    *
    * @example
-   *  {
    *    prefixes: [
    *      "https://example.com", // Exact
-   *      /^[^.s]+.example.com/g // Match with all subdomain and scheme,
-   *    ],
-   *  }
+   *      "https://*.example.com" // Match with any subdomain
+   *    ]
    */
-  prefixes: (string | RegExp)[];
+  prefixes: string[];
   /**
    * Config to fine-tune how to parse the path.
    *
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -60,10 +60,7 @@ export default function useLinking(
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
       /**
-       * https is default schema, as mentioned in android documentation
-       * User can add prefixes like this `example.com`
-       */
-      const protocol = prefix.match(/^[^:]+:\/\//)?.[0] || 'https://';
+      const protocol = prefix.match(/^[^:]+:\/\//)?.[0] ?? '';
       const host = prefix.replace(protocol, '');
       const prefixRegex = new RegExp(
         `^${escapeStringRegexp(protocol)}${host
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -58,8 +58,9 @@ export default function useLinking(
 
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
-      if (checkIsPrefixMatch(url, prefix)) {
-        return url.replace(prefix, '');
+      const prefixRegex = new RegExp(`^${prefix.replace(/\*/g, '[^/]+')}`);
+      if (url.match(prefixRegex)) {
+        return url.replace(prefixRegex, '');
       }
     }
 
@@ -121,19 +122,4 @@ export default function useLinking(
   return {
     getInitialState,
   };
-}
-
-function checkIsPrefixMatch(
-  url: string,
-  prefix: string | RegExp
-): string | RegExp | false {
-  if (typeof prefix === 'string') {
-    if (url.startsWith(prefix)) {
-      return prefix;
-    }
-  } else if (url.match(prefix)) {
-    return prefix;
-  }
-
-  return false;
 }
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -131,7 +131,7 @@ function checkIsPrefixMatch(
     if (url.startsWith(prefix)) {
       return prefix;
     }
-  } else if (prefix.test(url)) {
+  } else if (url.match(prefix)) {
     return prefix;
   }
 
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -6,6 +6,7 @@ import {
   NavigationContainerRef,
 } from '@react-navigation/core';
 import type { LinkingOptions } from './types';
+import escapeStringRegexp from 'escape-string-regexp';
 
 let isUsingLinking = false;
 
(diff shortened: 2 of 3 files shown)


spacesuitdiver:
I wonder if we are better of with just a predicate function here instead then we can have whatever logic including your RegExp. Otherwise looks good.

hosseinmd:
> I wonder if we are better of with just a predicate function here instead then we can have whatever logic including your RegExp. Otherwise looks good.

Prefixes define just for remove unnecessary parts of the link, react-navigation just watching to routes and query, so host or schema does not matter to them, so as you mentioned we can put there a function which remove unnecessary parts and gives them what matters.
But I need to know what the members think.

spacesuitdiver:
You’re right, it’s be a filter, we’re on the same page though function seems a little more flexible as you can pull in 3rd party libraries too.

packages/native/src/types.tsx
```diff
@@ -27,8 +27,16 @@ export type LinkingOptions = {
    * The prefixes are stripped from the URL before parsing them.
    * Usually they are the `scheme` + `host` (e.g. `myapp://chat?user=jane`)
    * Only applicable on Android and iOS.
+   *
+   * @example
+   *  {
+   *    prefixes: [
+   *      "https://example.com", // Exact
+   *      /^[^.s]+.example.com/g // Match with all subdomain and scheme,
```
satya164, line 35: It's just an example, but still:

1. Why do we need `[^.s]+`? Wouldn't `.+` be enough?
1. The `.` needs to be escaped, otherwise it'll match any character
1. I'd rather mention the scheme in the example
1. The `/g` is not necessary since it'll only match once

```suggestion
   *      /^https?:\/\/.+\.example\.com/ // Match with any subdomain and scheme,
```

packages/native/src/useLinking.native.tsx
```diff
@@ -122,3 +122,18 @@ export default function useLinking(
...
+
+function checkIsPrefixMatch(
+  url: string,
+  prefix: string | RegExp
+): string | RegExp | false {
+  if (typeof prefix === 'string') {
+    if (url.startsWith(prefix)) {
+      return prefix;
+    }
+  } else if (prefix.test(url)) {
```
satya164, line 134: I'd avoid `regex.test` since `test` is stateful when used with `/g` and this regex is provided from outside the lib.
```suggestion
  } else if (url.match(prefix)) {
```

satya164 (changes requested):
Thanks for the PR. I'm not entirely sure if we should support a regex. It makes it more flexible for the user, however there are some issues:

1. By specifying a regex, it's no longer just a prefix since user can specify a regex to match anything. Maybe it should be a separate option instead like `matches` or something
2. Currently we don't do it, but for things like web integration, we would need to be able to understand the prefix, for example, say there's a subpath which we shouldn't overwrite, a regex would prevent us from doing it

Maybe it's better to support what's supported by iOS & Android: e.g. `https://*.example.com` could work, we would convert it into a regex internally like we do for the paths in the deep link config. It's not as flexible, but if we support everything the underlying platform supports, then it should be fine.

In future, if we have more use case requests, we can consider supporting a `matches` option for advanced cases.

packages/native/src/types.tsx
hosseinmd: In our case we want to support http too. However, your code is more readable but there is a bug if the url is something like https://other.com/?sub.example.com

hosseinmd: This regex could solve that`/^(https|http):\/\/[^\/]+.example.com/`

hosseinmd: So, for this `https://*.example.com` pattern, if we replace * with `[^\/]+` and add `^` to start of prefix we could make our regex internally, after the testing I will push it.

satya164: > In our case we want to support http too
`https?` matches both `https` and `http`
> there is a bug if the url is something like other.com/?sub.example.com
yeah, `.+` can be changed to `[^/]+` to match anything without `/`. changing to `[^.s]+` doesn't seem to match anything

hosseinmd:
> Thanks for the PR. I'm not entirely sure if we should support a regex. It makes it more flexible for the user, however there are some issues:
> 
> 1. By specifying a regex, it's no longer just a prefix since user can specify a regex to match anything. Maybe it should be a separate option instead like `matches` or something
> 2. Currently we don't do it, but for things like web integration, we would need to be able to understand the prefix, for example, say there's a subpath which we shouldn't overwrite, a regex would prevent us from doing it
> 
> Maybe it's better to support what's supported by iOS & Android: e.g. `https://*.example.com` could work, we would convert it into a regex internally like we do for the paths in the deep link config. It's not as flexible, but if we support everything the underlying platform supports, then it should be fine.
> 
> In future, if we have more use case requests, we can consider supporting a `matches` option for advanced cases.

I agree with you, this pattern (https://*.example.com) is more reliable, and we could document it nicely.

satya164:
Can you update the PR to support wildcard string instead of regexes?

hosseinmd:
> Can you update the PR to support wildcard string instead of regexes?

Done, please check it.

packages/native/src/useLinking.native.tsx
```diff
@@ -58,8 +58,9 @@ export default function useLinking(
 
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
-      if (checkIsPrefixMatch(url, prefix)) {
-        return url.replace(prefix, '');
+      const prefixRegex = new RegExp(`^${prefix.replace(/\*/g, '[^/]+')}`);
```
satya164, line 61: Constructing the regex this way would mean that user would be able to add random strings. Since we only support wildcard in place of segments of the URL, can you change it to something like this:

```suggestion
      const prefixRegex = new RegExp(
        `^(${prefix
          .split('.')
          .map((it) => (it === '*' ? '.+' : escape(it)))
          .join('\\.')})`
      );
```

We could do more complex stuff to support `*` anywhere while escaping the rest, but I think for now this is fine.

satya164 (changes requested):

packages/native/src/useLinking.native.tsx
```diff
@@ -58,8 +58,9 @@ export default function useLinking(
 
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
-      if (url.startsWith(prefix)) {
-        return url.replace(prefix, '');
+      const prefixRegex = new RegExp(`^${prefix.replace(/\*/g, '[^/]+')}`);
+      if (url.match(prefixRegex)) {
```
satya164, line 62: Since we're constructing the regex ourselves, `.test` is fine:
```suggestion
      if (prefixRegex.test(url)) {
```

packages/native/src/useLinking.native.tsx
hosseinmd: I tested that, doesn't work. But work when I removed escape from that.

hosseinmd: What about `.+` Don't you think it should be `[^/]+`

hosseinmd: I think if we want to use escape this should be equal.
`escape(str) === str`
But they are not equal

hosseinmd:
@satya164 I have committed your suggested code with little change, please check it.

packages/native/src/useLinking.native.tsx
```diff
@@ -58,8 +59,20 @@ export default function useLinking(
 
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
-      if (url.startsWith(prefix)) {
-        return url.replace(prefix, '');
+      /**
+       * https is default schema, as mentioned in android documentation
+       * User can add prefixes like this `example.com`
+       */
+      const protocol = prefix.match(/^[^:]+:\/\//)?.[0] || 'https://';
```
satya164, line 66: I would prefer no to add a schema by default unless user has specified it.
```suggestion
      /**
      const protocol = prefix.match(/^[^:]+:\/\//)?.[0] ?? '';
```

packages/native/src/useLinking.native.tsx
```diff
@@ -58,8 +59,17 @@ export default function useLinking(
 
   const extractPathFromURL = React.useCallback((url: string) => {
     for (const prefix of prefixesRef.current) {
-      if (url.startsWith(prefix)) {
-        return url.replace(prefix, '');
+      /**
```
satya164, line 62:
```suggestion
```

satya164 (approved):

PR #8968, opened 2020-10-21
fix(prefixes): add http or https for default protocol

hosseinmd:
Prefixes without a scheme are assumed as either http or https. For example, www.google.com matches both http://www.google.com and https://www.google.com.

https://developer.android.com/guide/navigation/navigation-deep-link

### DIFF
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -61,8 +61,12 @@ export default function useLinking(
     for (const prefix of prefixesRef.current) {
       const protocol = prefix.match(/^[^:]+:\/\//)?.[0] ?? '';
       const host = prefix.replace(protocol, '');
+      const schema = protocol
+        ? escapeStringRegexp(protocol)
+        : '(http|https)://';
+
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Is the behaviour same on both Android and iOS for a default scheme? Unless it is, I'd rather not add any scheme by default. It's easy enough to specify 2 prefixes.

hosseinmd:
I couldn't find anything about iOS, it seems the iOS schema is required, but our work does not depend on it, we are adding that to have simple prefixes.

hosseinmd:
With this we have a good doc and don't need to tell developers to add protocol to prefixes

satya164:
Thanks, but I believe it's better to be explicit than to implictely assume a scheme which may or may not be correct. It's very little work to add another prefix if you need it.

hosseinmd:
> Thanks, but I believe it's better to be explicit than to implictely assume a scheme which may or may not be correct. It's very little work to add another prefix if you need it.

That is not about me. Anyway, thank you.

PR #8974, opened 2020-10-22
fix: set needsOffscreenAlphaCompositing and update default android animation

### LINKED ISSUE #8696: Elevation artifacts when using ScaleFromCenterAndroid
**Current Behavior**

- When using TransitionPresets.ScaleFromCenterAndroid for android it has elevation artifacts

Video

**Expected Behavior**

- I've reopened a new issue, because https://github.com/react-navigation/react-navigation/issues/8676 was closed already
- Artifact should not be shown
- I've investigated this issue. Please see facebook/react-native#23090 (comment).
- ~I guess setting needsOffscreenAlphaCompositing on this view: https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Header/HeaderSegment.tsx#L313 at least during the animation would fix the problem.~ We need to set needsOffscreenAlphaCompositing on the whole screen, as more elements can have elevation set, so not only on the header view.

Sidenote: DefaultTransition does not check android version > Q yet but should use ScaleFromCenterAndroid for Q by default?!

**How to reproduce**

- See [expo snack](https://snack.expo.io/@domir/getting-started-%7C-react-navigation)

**Your Environment**

| iOS or Android                 | android
| @react-navigation/native       |  5.7.3
| @react-navigation/stack        | 5.9.0
| react-native-gesture-handler   |
| react-native-safe-area-context |
| react-native-screens           |
| react-native                   |
| expo                           |

satya164:
closes #8696

### DIFF
--- a/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
+++ b/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
@@ -289,8 +289,8 @@ export function forScaleFromCenterAndroid({
   );
 
   const opacity = progress.interpolate({
-    inputRange: [0, 0.8, 1, 1.2, 2],
-    outputRange: [0, 0.5, 1, 0.33, 0],
+    inputRange: [0, 0.75, 0.875, 1, 1.0825, 1.2075, 2],
+    outputRange: [0, 0, 1, 1, 1, 1, 0],
   });
 
   const scale = conditional(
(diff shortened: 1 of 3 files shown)


hanno-ferrannegre:
I will test this @satya164, if it works, you are a hero. If not, you are still one.

ferrannp:
@satya164 the hero 🙌 .

PR #8983, opened 2020-10-23
test: add navigation focus events tests

oltrep:
This PR adds a few unit tests on the NavigationFocusEvents. It asserts that events are emitted in the expected order in different cases.


satya164 (approved):
Thanks!

PR #8987, opened 2020-10-24
feat: add `getInitialURL` and `subscribe` options to linking config

satya164:
For apps with push notifications linking to screens inside the app, currently we need to handle them separately (e.g. [instructions for firebase](https://rnfirebase.io/messaging/notifications#handling-interaction), [instructions for expo notifications](https://docs.expo.io/push-notifications/receiving-notifications/)). But if we add a link in the notification to use for deep linking, we can instead reuse the same deep linking logic instead.

This commit adds the `getInitialURL` and `subscribe` options which internally used `Linking` API to allow more advanced implementations by combining it with other sources such as push notifications.

Example usage with Firebase notifications could look like this:

```js
const linking = {
  prefixes: ['myapp://', 'https://myapp.com'],
  async getInitialURL() {
    // Check if app was opened from a deep link
    const url = await Linking.getInitialURL();

    if (url != null) {
      return url;
    }

    // Check if there is an initial firebase notification
    const message = await messaging().getInitialNotification();

    // Get the `url` property from the notification which corresponds to a screen
    // This property needs to be set on the notification payload when sending it
    return message?.notification.url;
  },
  subscribe(listener) {
    const onReceiveURL = ({ url }: { url: string }) => listener(url);

    // Listen to incoming links from deep linking
    Linking.addEventListener('url', onReceiveURL);

    // Listen to firebase push notifications
    const unsubscribeNotification = messaging().onNotificationOpenedApp(
      (message) => {
        const url = message.notification.url;

        if (url) {
          // If the notification has a `url` property, use it for linking
          listener(url);
        }
      }
    );

    return () => {
      // Clean up the event listeners
      Linking.removeEventListener('url', onReceiveURL);
      unsubscribeNotification();
    };
  },
  config,
};
```

### DIFF
--- a/packages/native/src/types.tsx
+++ b/packages/native/src/types.tsx
@@ -50,6 +50,20 @@ export type LinkingOptions = {
    */
   config?: { initialRouteName?: string; screens: PathConfigMap };
   /**
+   * Custom function to get the initial URL used for linking.
+   * Uses `Linking.getInitialURL()` by default.
+   * Not supported on the web.
+   */
+  getInitialURL?: () => Promise<string | null | undefined>;
+  /**
+   * Custom function to get subscribe to URL updates.
+   * Uses `Linking.addEventListener('url', callback)` by default.
+   * Not supported on the web.
+   */
+  subscribe?: (
+    listener: (url: string) => void
+  ) => undefined | void | (() => void);
+  /**
    * Custom function to parse the URL to a valid navigation state (advanced).
    * Only applicable on Web.
    */
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -16,6 +16,22 @@ export default function useLinking(
     enabled = true,
     prefixes,
     config,
+    getInitialURL = () =>
+      Promise.race([
+        Linking.getInitialURL(),
+        new Promise<undefined>((resolve) =>
+          // Timeout in 150ms if `getInitialState` doesn't resolve
+          // Workaround for https://github.com/facebook/react-native/issues/25675
+          setTimeout(resolve, 150)
+        ),
+      ]),
+    subscribe = (listener) => {
+      const callback = ({ url }: { url: string }) => listener(url);
+
+      Linking.addEventListener('url', callback);
+
+      return () => Linking.removeEventListener('url', callback);
+    },
     getStateFromPath = getStateFromPathDefault,
   }: LinkingOptions
 ) {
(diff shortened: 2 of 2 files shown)


codecog:
One thing i find myself having to do with links is parse the incoming to and take out the parts that i need. An extension may be to add this as well.

Say you have a deep link to a chat id and the link is `chatapp://ChatScreen/01` and i want to be able to navigate to that screen. `getInitialUrl()` may return something like:
```
getInitialUrl = () => {
  // parse url

  return {
    screen: 'ChatScreen' // from parse
    params: { id: 01 }
  }
}
```
This isn't likely to cover every case, but more of an idea to simplify navigating to a specific screen for things that need to be handled for every app that wants deeplinking.

LauraBeatris:
This is awesome! It'll help a lot in order to execute some previously logic before navigating to the screen, such as saving some param on the Async Storage.

What I did at the time to solve this issue, was to create a custom router... But I think it's too much work and would simplify a lot if ``react-navigation`` handle this logic for us. 

Thanks!

satya164:
@codecog I'm not sure I understand.

If you have a link `chatapp://chat/01` and want to navigate to that screen, it's already possible with linking config without anything extra:

```js
const linking = {
  prefixes: ['chatapp://'],
  config: {
    screens: {
      ChatScreen: {
        path: 'chat/:id',
        parse: {
          id: Number
        }
      }
    }
  }
}
```

If want to have full control over how the link is parsed, you can provide a custom `getStateForPath` function and do the pasring there.

pzatorski:
This is really good thinking!

huttotw:
@satya164 thanks so much for this contribution, I think it is what I need when when integrating Branch. However, I am not able to get these overrides to work in my linking object, it is as if they are never called. Is there a prerequisite that I am missing?

satya164:
are you using the latest version?

huttotw:
@satya164 yes, here is my linking object:
```js
const linking = {
    config: {
      screens: {
        CheckIn: 'check-in',
        Tabs: {
          initialRouteName: 'Home',
          path: 'tabs',
          screens: {
            Chat: {
              initialRouteName: 'Chat',
              path: 'threads',
            },

            Home: 'home',
            LogIn: 'log-in',
            Search: {
              initialRouteName: 'Search',
              path: 'search',
              screens: {
                PropertyList: 'properties',
              },
            },
            Stay: {
              initialRouteName: 'Stay',
              path: 'stays',
              screens: {
                Guide: 'guides',
                GuideDetail: 'guides/:uid',
              },
            },
          },
        },
      },
    },
    getInitialURL: () => {
      console.log('getInitialUrl');
      return Linking.getInitialURL();
    },
    subscribe: (listener) => {
      console.log('subscribe');
      const onReceiveURL = ({ url }) => listener(url);

      // Listen to incoming links from deep linking
      Linking.addEventListener('url', onReceiveURL);

      return () => {
        // Clean up the event listeners
        Linking.removeEventListener('url', onReceiveURL);
      };
    },
    prefixes: [
      prefix,
    ],
  };
```

I do not get any of the console log statements.

huttotw:
I just have this code in the body of App.jsx before the render function, wherein I am passing `linking` to the `NavigationContainer` via the `linking` prop.

satya164:
then there's no reason it shouldn't work. try clearing metro bundler cache and check in node modules if you have correct updated files.

huttotw:
@satya164 ah yes, it was the metro bundler cache, thanks so much for pointing that out.

PR #8992, opened 2020-10-26
test: add tests to getEventManager

oltrep:
Decided to contribute back some tests I've written when porting over the implementation of getEventManager. Let me know if I can improve anything :)


satya164 (approved):
Thanks!

PR #9004, opened 2020-10-30
fix: let Bottom Tabs accessibility labels generated by the OS

MoOx:
The accessibility label "tab X of Y" is generated by the os, with the proper language. In order to do so, we need the OS to detect tab inside a tablist. This PR should handle this.
It can be tested on iOS via VoiceOver. Pressing a tab should say "{label}, tab X on Y".

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -260,13 +260,6 @@ export default function BottomTabBar({
               ? options.title
               : route.name;
 
-          const accessibilityLabel =
-            options.tabBarAccessibilityLabel !== undefined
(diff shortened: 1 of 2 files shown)


satya164 (approved):
Thanks!

MoOx:
Don't be too grateful, it doesn't work as expected. I am trying to track down the issue with react native, but my understanding is that react native doesn't properly inform iOS about accessibility traits (even if the doc say so).
I will let this PR open in case I found something to adjust this.

satya164:
@MoOx that's sad to know. lemme know if you figure it out and i'll be happy to merge.

satya164:
Closing due to cleanup PRs. Please send a new PR if you manage to find a solution.

Issue #8094, opened 2020-04-21
[android] focus dismiss after navigation back


MaxBolotnyi:
** PR that solves the issue **
https://github.com/react-navigation/react-navigation/pull/7931

**Current Behavior**
Here is the expo snack reproducing the issue.
https://snack.expo.io/@shonsy/keyboard-bug-v4

Here is also the video of what happens
[Screen Recording 2020-04-21 at 15.15.17.zip](https://github.com/react-navigation/react-navigation/files/4509902/Screen.Recording.2020-04-21.at.15.15.17.zip)

**Expected Behavior**

When you navigate back and immediately click on the text input - the focus should stay in the 'focused' state.

**How to reproduce**
1. Go to the snack link provided above.
2. Run the app on 'android';
3. Click on the 'Push screen' button.
4. Click on the 'back' button.
5. Immediately click on the text input with placeholder 'click here'

P.S.: Watch the video if you have any issues with repro.

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | android
| react-navigation      | 4.3.1
| @react-navigation/stack        | 2.3.10
| react-native-safe-area-context | 0.6.0
| react-native-screens           | 2.0.0
| react-native                   | 0.61.5
| expo                           | 37.0.0
| node                           | any
| npm or yarn                    | any

MaxBolotnyi:
> The versions mentioned in the issue for the following packages differ from the latest versions on npm:
> 
> * `@react-navigation/stack` (found: `2.3.10`, latest: `5.2.11`)
> 
> Can you verify that the issue still exists after upgrading to the latest versions of these packages?

yes, it reproduces because of the stack library, both with the react-navigation-stack and @react-navigation/stack

raajnadar:
Create a snack with v5

sa8ab:
Same issue react navigation 4

MaxBolotnyi:
@raajnadar 

[Here](https://snack.expo.io/@shonsy/keyboard-bug-v5) is a repro with the latest available libraries on the snack. However, the issue is present on both v5 and v4 and is not related directly to the version of the navigation, but to the KeyboardManager.ts in the **stack** package, I have opened a PR that is supposed to solve the issue.

Trancever:
@MaxBolotnyi I am pretty sure that the IOS transition preset that you use for Android is causing the issue. You can use the default one for android as a workaround for now.

[...]

PR #9008, opened 2020-10-30
fix: android textinput gets blurred after navigating back

marhaupe:
When navigating from ScreenA to ScreenB and then back to ScreenA,
react-navigation unconditionally calls `Keyboard.dismiss()`.
If the user is fast enough and taps on a `TextInput` after coming
back from ScreenB, the keyboard opens, just to be dismissed again.
This would also happen if some logic automatically focuses the
`TextInput` in ScreenA. This behaviour can be observed in
https://snack.expo.io/lTDZhVNhV.

### DIFF
--- a/packages/stack/src/views/KeyboardManager.tsx
+++ b/packages/stack/src/views/KeyboardManager.tsx
@@ -58,10 +58,12 @@ export default class KeyboardManager extends React.Component<Props> {
 
     const input = this.previouslyFocusedTextInput;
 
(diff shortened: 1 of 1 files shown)


marhaupe:
This fixes the demo in https://github.com/react-navigation/react-navigation/issues/8094 and the bug in our larger, proprietary application.

satya164 (approved):
Thanks!

PR #9013, opened 2020-11-01
feat: useRedo

linmic:
This hook is to address a common use case in React Native app where we tend to redo some actions when the view is _resurfaced from the background_ or _refocused_. The most obvious one that pops out of my head is to re-fetch the updated information from the source, the following is the example with [apollo-client](https://github.com/apollographql/apollo-client):

```javascript
import React from 'react';
import { View, Text } from 'react-native';
import { gql, useQuery } from '@apollo/client';
import { useRedo } from '@react-navigation/native';

const GET_DOGS = gql`
  query GetDogs {
    dogs {
      id
      breed
    }
  }
`;

const Dogs = () => {
  const { loading, error, data, refetch } = useQuery(GET_DOGS, {
    fetchPolicy: 'cache-and-network',
  });

  useRedo(refetch);

  if (loading) return 'Loading...';
  if (error) return `Error! ${error.message}`;

  return (
    <View>
      {data.dogs.map(({ id, breed }) => (
        <Text key={id}>{breed}</Text>
      ))}
    </View>
  );
};
```


satya164:
Thanks for the PR. However I'm not sure about handling app state in React Navigation. I wouldn't be merging the PR, but feel free to publish as a separate library, or propose this hook in https://github.com/react-navigation/hooks

However, there are some things incorrect in the code that you would want to fix first:

1. I think a name like `useAppStateAwareFocusEffect` or something like that makes more sense, redoing in your case is a subset of use cases.
2. You should follow [rules of hooks](https://reactjs.org/docs/hooks-rules.html) when writing code with hooks, for example, the dependency arrays in `useCallback` and `useEffect` should contain references which are used.
3. You can't memoize passed functions like that. The memoization needs to be done by the consumer because only they can specify a correct dependency array.
4. When you are providing an API for side-effects, you also need to handle cleanup, e.g. the effect should be cleaned up when the app becomes inactive. The `useFocusEffect` hook handles cleanup for blur and unmount.

linmic:
@satya164 Sure. Thanks for the feedback!

PR #9032, opened 2020-11-07
feat: add option to show a header in drawer navigator screens

satya164:
This commit adds new `header` and `headerShown` options in drawer navigator to be able to show a header, along with a bunch of header related options similar to stack navigator.

Historically, we have suggested to nest a stack navigator inside drawer navigator to render a header. While it works, it's not efficient to nest an entire navigator just for a header, considering it adds a lot of additional overhead from the code to handle animations, gestures etc. which won't ever be run in this case. It also increases the view hierarchy which has negative impacts on performance on Android, and could cause content not to render on older iOS devices.

Another issue with the approach is that since drawer navigator is at the root in this setup, it's possible to open drawer from every screen in the stack navigator, which usually isn't the expected behaviour. It's necessary to write additional code to disable the gesture to open drawer in all screens but first.

In addition, users also need to add a custom drawer icon to the header manually to be able to toggle the drawer

If drawer navigator could render its own header we'd avoid all these shortcomings as well as make the code simpler.

For now, I have implemented a new `Header` component in drawer since it's way simpler than stack navigator header. Though we may consider creating a shared UI package and add such components there which all our navigators could use.

The `Header` includes a button to toggle the drawer by default, and supports customization options such as showing custom left/right/title components. For this commit, I have kept `headerShown` to `false` by default coz I wasn't sure if it'd be a breaking change to start showing headers in drawers. Probably we can toggle it to `true` by default in next major.

The buttons to toggle drawer look like this by default:

**iOS**:
Simulator Screen Shot - iPhone 11 - 2020-11-07 at 05 03 54

**Android**:
Screenshot_20201107-052847


brentvatne (approved):
as discussed on messenger, it would be good to extract a header component to another library that we can share between stack and drawer (and others?) and that developers can use whenever they want to render a header detached directly

PR #9058, opened 2020-11-12
chore: replace community maskedview

Naturalclar:
replaced `@react-native-community/masked-view` with `@react-native-masked-view/masked-view` due to namespace change.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@expo/vector-icons": "^10.0.0",
     "@react-native-async-storage/async-storage": "^1.13.1",
-    "@react-native-community/masked-view": "0.1.10",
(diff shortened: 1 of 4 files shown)


satya164 (approved):
Thanks!

afestein:
I noticed this has been merged into the react-navigation:6.x branch for the upcoming release, which still looks a while off.

Without updating to the new namespace, I don't think we can update to Xcode 12 because [this fix](https://github.com/react-native-masked-view/masked-view/pull/80) isn't available in the old version.

Is there any reason we couldn't also merge this into the main branch?

satya164:
@afestein unfortunately changing peer dependencies is a breaking change. if you need this change, I'd suggest to use patch-package for now.

Issue #9098, opened 2020-11-24
openByDefault={true} does not open the drawer on routes other than "/" with expo-linking


sekonia:
**Current Behavior**
With routes defined (expo-linking) and passed to `NavigationContainer` as linking-prop, the drawer does not open initially on any route other than `/` on the web, even though `openByDefault` is set to true on `Drawer.Navigator`. This applies to all drawerTypes other than `permanent`.

```
import * as Linking from "expo-linking";
import { NavigationContainer } from "@react-navigation/native";
import React from "react";
import { createDrawerNavigator } from "@react-navigation/drawer";

const prefix = Linking.makeUrl("/");
const linking = {
    prefixes: [prefix],
    config: {
        screens: {
            Welcome: "welcome",
            PersonalInformation: "personal-information",
        },
    },
};

const App: React.FC = () => {
    const Drawer = createDrawerNavigator();
    return (
        <NavigationContainer linking={linking}>
            <Drawer.Navigator openByDefault={true}>
                <Drawer.Screen
                    name="Welcome"
                    component={Welcome}
                />
                <Drawer.Screen
                    name="PersonalInformation"
                    component={PersonalInformation}
                />
            </Drawer.Navigator>
        </NavigationContainer>
    );
};
registerRootComponent(App);
```

I have noticed, that functions from `DrawerRouter` like `openDrawer()` are only called when I load the page on `/`. On any other route, no functions from `DrawerRouter` are called and that's why `{type: "drawer"}` is not present in the history-object.

**Expected Behavior**
The drawer should open by default when I refresh the page on `/welcome` or `/personal-information`. Currently the drawer only opens on `/`.

**How to reproduce**
Define any linking object, set `openByDefault` to true and refresh the application on any route other than `/`.

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 |
| @react-navigation/native       | 5.8.10
| @react-navigation/drawer       | 5.11.4
| react-native-reanimated        | 1.13.0
| react-native-gesture-handler   | 1.7.0
| react-native-safe-area-context | 3.1.4
| react-native-screens           | 2.10.1
| react-native                   | 63.2
| expo                           | 39.0.2
| node                           | 12.18.4
| npm or yarn                    | yarn

sekonia:
Ok, that's interesting... I had no empty route defined in my linking object because I wanted every route to have a name. I didn't think that this was impossible.

This works
```
const linking = {
[...]

PR #9099, opened 2020-11-24
fix: consider openByDefault prop when rehydrating drawer state

sekonia:
openByDefault in connection with expo-links only works for `getInitialState()`. When a page of the application is refreshed, `getRehydratedState()` is called and the drawer stays closed. This should fix the issue.

### DIFF
--- a/packages/routers/src/DrawerRouter.tsx
+++ b/packages/routers/src/DrawerRouter.tsx
@@ -145,7 +145,7 @@ export default function DrawerRouter({
         routeParamList,
       });
 
-      if (isDrawerOpen(partialState) || openByDefault) {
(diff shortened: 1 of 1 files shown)


sekonia:
I have opened an issue earlier today that describes the problem: #9098

packages/routers/src/DrawerRouter.tsx
```diff
@@ -145,7 +145,7 @@ export default function DrawerRouter({
         routeParamList,
       });
 
-      if (isDrawerOpen(partialState)) {
+      if (isDrawerOpen(partialState) || openByDefault) {
```
satya164, line 148:
```suggestion
      if (partialState.history ? isDrawerOpen(partialState) : openByDefault) {
```

satya164:
Thanks!

PR #9104, opened 2020-11-27
feat: focus trap in active screen

RafikiTiki:
Right now when building app for web using react-navigation with react-native-web there is as problem with handling keyboard navigation making such app completely not compliant with WCAG spec.

The problem is that when user navigates from one screen to another the first one is still rendered (obviously, it's react-navigation way) and when he/her try to "move" through the second screen using keyboard (tabbing through focusable components), after navigating, focus is somehow reset to the previous screen and only after tabbing through it's hidden content the user is able to actually interact with the current screen.

This PR implements focus trap inside currently active screen, looping focus inside of it.


RafikiTiki:
Using screens wrapper component with `useFocusEffect` combined with similar focus trap implementation turned out to be more flexible in my use case, closing this PR.

RafikiTiki:
Reopening so that willing person can try to make this approach more generic :)

RafikiTiki:
Reopening so that willing person can try to make this approach more generic :)

satya164:
Cleaned up the PR a bit. So tested this a bit and for now what's missing is that it doesn't work well with nested navigators since the focus trap is always active on the innermost screen. Also, this would need to be addressed with other navigators as well.

PR #9169, opened 2020-12-23
docs: link directly to #react-navigation on Reactiflux

nickmccurdy:


nickmccurdy:
There seems to be an issue with the GitHub Actions

satya164:
Thanks for the PR. Unfortunately it looks like they removed the react-navigation channel from the discord for some reason :(

PR #9178, opened 2020-12-25
fix: warnings printing to console for un-exported members of react-native-screens

### LINKED ISSUE #8993: Attempted import error: 'shouldUseActivityState' is not exported from 'react-native-screens'.
After upgrading to 

```bash
    "@react-navigation/bottom-tabs": "^5.9.2",
    "@react-navigation/drawer": "^5.9.3",
    "@react-navigation/native": "^5.7.6",
    "@react-navigation/stack": "^5.9.3",
```

There are warnings after building, for an expo managed flow SDK 39, env setup are as below:

```bash
    "expo-cli": "^3.28.1",
    "react-native": "https://github.com/expo/react-native/archive/sdk-39.0.0.tar.gz",
    "react-native-screens": "~2.10.1",
```

image

It seems relates to this PR

https://github.com/react-navigation/react-navigation/pull/8805

My questions are:
- does this matter?
- Is this a missing type in the `react-native-screens`
- does this require to run with a specific version of `react-native-screens`?

@WoLewicki What am I missing here?

woodcockjosh:
- Fixes #8993
- Fixes failing lint

Will rebase from master if lint error is fixed before this is merged

### DIFF
--- a/packages/bottom-tabs/src/views/ResourceSavingScene.tsx
+++ b/packages/bottom-tabs/src/views/ResourceSavingScene.tsx
@@ -1,11 +1,9 @@
 import * as React from 'react';
 import { Platform, StyleSheet, View } from 'react-native';
-import {
-  Screen,
-  screensEnabled,
-  // @ts-ignore
-  shouldUseActivityState,
-} from 'react-native-screens';
+import { Screen, screensEnabled } from 'react-native-screens';
+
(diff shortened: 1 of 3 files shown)


satya164:
Thanks for the PR, but you can upgrade react-native-screens instead to get rid of the warning as the export is present there. We don't need custom handling for it.

PR #9201, opened 2021-01-06
chore: set displayName for ThemeContext

dulmandakh:
Set displayName for ThemeContext, so that it'll show ThemeContext.Provider instead of mere Context.Provider in React Devtools.


packages/native/src/theming/ThemeContext.tsx
```diff
@@ -3,5 +3,6 @@ import DefaultTheme from './DefaultTheme';
 import type { Theme } from '../types';
 
 const ThemeContext = React.createContext<Theme>(DefaultTheme);
+ThemeContext.displayName = 'ThemeContext';
```
satya164, line 6:
```suggestion

ThemeContext.displayName = 'ThemeContext';
```

satya164 (approved):
Thanks!

PR #9202, opened 2021-01-06
chore: set displayName for LinkingContext

dulmandakh:


packages/native/src/LinkingContext.tsx
```diff
@@ -4,5 +4,6 @@ import type { LinkingOptions } from './types';
 const LinkingContext = React.createContext<{
   options: LinkingOptions | undefined;
 }>({ options: undefined });
+LinkingContext.displayName = 'LinkingContext';
```
satya164, line 7:
```suggestion

LinkingContext.displayName = 'LinkingContext';
```

satya164 (approved):
Thanks!

Issue #9209, opened 2021-01-09
Impossible to use useLinkProps outside navigator


yairopro:
**Current Behavior**

- What code are you running and what is happening?
I try to build a link using `useLinkProps()` hook outside the main stack navigator. In my case, I build a link to home for the app's logo which is in the custom app bar.
Clicking on the link result an error:
````
Couldn't find a navigation object.
````
**Expected Behavior**

- What do you expect should be happening?
The link to work.

**How to reproduce**

**[Here's a full sandbox demo](https://codesandbox.io/s/react-navigation-uselinkprops-outside-navigator-sckfz)**.

Basically you just need to build your app like so:
````javascript
<NavigationContainer>
      <AppBar /> // <-- try to build a link from inside the AppBar component

      <Stack.Navigator screenOptions={{ headerShown: false }}>
        // screens ...
      </Stack.Navigator>
</NavigationContainer>
````
**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| @react-navigation/native       | 5.8.10
| react-native-web                   | 0.14.10

satya164:
fixed in https://github.com/react-navigation/react-navigation/commit/1d40279db18ab2aed12517ed3ca6af6d509477d2

yairopro:
Thank

PR #9210, opened 2021-01-10
Navigation container exports its ref interface as context.

yairopro:
This PR is to solve #9209 .

Navigation container exports its ref interface as context.
This context is used in useLinkTo & useLinkProps.


satya164:
Thanks for the PR. Closing since this is already implemented in 6.x

Issue #9213, opened 2021-01-11
NavigationContainer reset the ref each render


yairopro:
**Current Behavior**

The NavigationContainer.ref gives a different instance each time the parent render even though the ref function is the same each render (I prefer function rather than ref objects).

**Expected Behavior**

- What do you expect should be happening?
- Include a screenshot or video if it makes sense.

**How to reproduce**

Here's a snack about the problem:
https://snack.expo.io/@yairopro/react-navigation:-bug---navigationcontainer-ref-reset-every-time

In the snack i use a state setter. The setter is always the same function instance so the `ref` prop should not trigger it a second time.

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| @react-navigation/native       | 5.8.10
| react-native                   | *

satya164:
Closing since this is no longer the case in React Navigation 6.

yairopro:
🎉

PR #9214, opened 2021-01-11
Fix NavigationContainer ref prop

yairopro:
Fix for issue #9213.

The `ref` prop is added as dependency at the call of `React.useImperativeHandle()`.

### DIFF
--- a/packages/native/src/NavigationContainer.tsx
+++ b/packages/native/src/NavigationContainer.tsx
@@ -61,7 +61,7 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
 
   const [isResolved, initialState] = useThenable(getInitialState);
 
-  React.useImperativeHandle(ref, () => refContainer.current);
+  React.useImperativeHandle(ref, () => refContainer.current, [ref]);
 
   const linkingContext = React.useMemo(() => ({ options: linking }), [linking]);
 
--- a/packages/native/src/NavigationContainer.tsx
+++ b/packages/native/src/NavigationContainer.tsx
@@ -61,7 +61,7 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
 
   const [isResolved, initialState] = useThenable(getInitialState);
 
-  React.useImperativeHandle(ref, () => refContainer.current, [ref]);
+  React.useImperativeHandle(ref, () => refContainer.current, [refContainer.current]);
 
   const linkingContext = React.useMemo(() => ({ options: linking }), [linking]);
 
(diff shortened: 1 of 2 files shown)


satya164 (commented):
Thanks for the PR, but I don't understand the use case.

Refs contain mutable values. By contract, they can change anytime without a re-render. They aren't supposed to be stored in state. In function components, the only proper way to store a ref is to use the `useRef` hook.

Regarding the PR, the dependency array needs to contain the dependencies of the ref: i.e. the values which make up the ref (in this case `dispatch`, `resetRoot` etc.), not the ref to assign it to. You can verify this by looking at the suggestions from the ESLint plugin.

Anyway, it's impossible to provide a stable ref without additional complexity, and there's no reason to expect a mutable value to be stable. So I won't be merging the pull request.

yairopro:
I agree with you that the ref prop itself should not be in the dependency. Like every other hooks with dependencies as parameter, only mutable values needed by the factory function _(in this case, the factory to create the ref handle)_, should be set as dependency. **My bad**.
The thing is, the `useImperativeHandle()` having no dependency at all makes it generate the ref handle on every render, and reseting it in the ref prop. It's like the `useMemo()` without a dependencies array, it will recompute every render the value that should have been memoized from the previous render.

So clearly the solution would be to set at least an array of dependency.

PS: I need the store the handle in a state in order to be aware of its _state_, in case some fields have changed, and trigger side effects. Assigning it an object `{current}`, given from a `useRef()`, without being aware is not an option for me. But thanks for the proposition.

packages/native/src/NavigationContainer.tsx
```diff
@@ -61,7 +61,7 @@ const NavigationContainer = React.forwardRef(function NavigationContainer(
 
   const [isResolved, initialState] = useThenable(getInitialState);
 
-  React.useImperativeHandle(ref, () => refContainer.current);
+  React.useImperativeHandle(ref, () => refContainer.current, [refContainer.current]);
```
satya164, line 64: Refs cannot and shouldn't be used as a dependency since they contain mutable values. The dependency array should only contain immutable data which can trigger a re-render

satya164:
> The thing is, the useImperativeHandle() having no dependency at all makes it generate the ref handle on every render, and reseting it in the ref prop. ... it will recompute every render the value that should have been memoized from the previous render.
> So clearly the solution would be to set at least an array of dependency.

Recomputing doesn't matter. There's nothing expensive going on here that needs to be memoized.

Your changes don't use a proper dependency list, e.g. omits state which is going to result in incorrect value returned from `dangerouslyGetState`, as well as omits many other things used inside the ref. Specifying proper dependency array would mean it won't be stable without additional work.

The dependency array of specifying `refContainer.current` is also incorrect since it's a mutable value and shouldn't be used in a dependency array.

> I need the store the handle in a state in order to be aware of its state and trigger side effects. So, assigning it to an object (given from a useRef() ) without being aware is not an option for me. But thanks for the proposition.

Why exactly do you need to be aware of ref state? You shouldn't be doing side effects on ref re-assignments. Move them into `useEffect`. You should re-consider your approach regarding whatever you're doing. This is not the proper way of doing things in React.

PR #9227, opened 2021-01-17
Fix nested navigations

sharifhh:


packages/routers/src/StackRouter.tsx
```diff
@@ -396,6 +396,9 @@ export default function StackRouter(options: StackRouterOptions) {
             }
 
             if (index === -1 && action.payload.name !== undefined) {
+              if (!state.routeNames.includes(action.payload.name)) {
```
satya164, line 399: I think it would be good to move this to the top of this case, and change to something like:

```js
if (action.payload.name !== undefined && !state.routeNames.includes(action.payload.name)) {
  return null;
}
```

satya164 (changes requested):
Thanks for the PR. Can you also add a test for the case it fixes so that we can avoid such issues in future?

packages/routers/src/StackRouter.tsx
sharifhh: I am not sure if it won't any intended workflow other than that, what I did is I read the old code and added this if statement

sharifhh (commented):
Can someone help with the testing here? this is a critical bug guys, just create a nested navigators and navigate using this
navigation.navigate({key: 'TestKey', name: 'NestedNavigatorB', params: {test: 'test'}) and it will crash with the following: cannot access options for undefined object "screen"

satya164:
I added tests here and addressed the change request https://github.com/react-navigation/react-navigation/commit/fac5d56de5f592ce7a4472b56151fce25e829f21

PR #9237, opened 2021-01-21
Fix deep link not working on nested navigation

neoswallow:
Existing code works for simple navigation stack where all screens are under RootStack but fails with nested RootStack.


satya164 (changes requested):
Thanks for the pull request. But I don't understand what this fixes.

This check isn't for checking if all of the routes in the parsed state exist. It's also incorrect to check since there can be routes which haven't been navigated to yet and deep link should be able to navigate to them.

This check is merely for making sure that your linking config matches the shape of your navigator config.

neoswallow:
True, I cannot confirm that this fix will cover all possible navigation configuration as I do not have full understanding of react-navigation.

I do have an assumption that rootState contains all possible screen it can navigate to when it was called.
If navigation configuration changes, rootState should have updated state when this is called again.
So I currently does not see the case where this would fail.

Code in release does not check beyond top most stack.
This code should go through all stack levels known at the time of running.
Any change in navigation config will correctly be found when it's called again.

Let me know if I had wrong assumption about navigation config in rootState.

neoswallow:
As a side note, current latest release does break our app.
I had to revert back to 5.8.2 on @react-navigation/native.
I can confirm that the change works with our app (changes made on 5.9.0).
In our app, rootStack contains either onboarding stack containing a stack of screens for onboarding or drawer stack where it contains a tab with several tab screens along with other non-tab screens.
Deep link (on url and also on push notification) does work on both cases.
Existing latest release only checks top most level routeNames contain only one entry in both loggedIn and loggedOut state.
By going into one level deep, I was able to find routeNames with all navigable screens.
I went ahead and made the code to check recursively to consider cases with even deeper nested navigation config than our app.
So even though I can't say that this PR will handle all cases, this should handle more cases without breaking it.

satya164:
I'm closing since the assumptions are incorrect and there are a lot of merge conflicts already.

I don't see how adding more checks on top of an existing check can fix an issue - when the existing check preventing the navigation is still there, and it can cause more issues since the additional checks would fail in many valid cases. But it's hard to validate any of these since the PR doesn't include a test case or an example of what it fixes.

If you're still having problems please open an issue with a repro and I'll try to look at it when I can.

PR #9244, opened 2021-01-23
fix: correct typings for tabPress event

ethanneff:
## Context
https://reactnavigation.org/docs/material-bottom-tab-navigator/#events

## Before
Screen Shot 2021-01-23 at 3 19 33 PM

## After
Screen Shot 2021-01-23 at 3 21 42 PM

### DIFF
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -39,6 +39,7 @@ export type EventMapCore<State extends NavigationState> = {
   blur: { data: undefined };
   state: { data: { state: State } };
   beforeRemove: { data: { action: NavigationAction }; canPreventDefault: true };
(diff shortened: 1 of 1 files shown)


satya164:
https://reactnavigation.org/docs/typescript/#annotating-usenavigation

yerevin:
I'm sorry if that's not the correct way to bump that issue, but I also came into the same problem of tabPress event type not being defined and the changes from this pull request resolved the issue. 
I found also someone running into that problem and reporting that in Issue https://github.com/react-navigation/react-navigation/issues/9730 which has been already automatically closed.
If updating types in library itself is not the solution, can I ask for example of how navigation props should be annotated to make tabPress not throw type error as described in issue #9730?

Issue #9272, opened 2021-02-02
Duplicate tabBarIcon in react tree


myckhel:
@react-navigation/native:5.7.3

The implementation of `tabBarIcon` causes `tabBarIcon` to be buggy when used as functional component (component that compute some inner functions).

Given this example: 
```jsx
<Tab.Screen
    options={{
      tabBarIcon: ({focused}) => (
        <CountableInbox key='4663' focused={focused} />
      ),
    }}
    name="Inbox"
    component={InboxScreen}
  />


const CountableInbox = memo(({focused}) => {
  const navigation = useNavigation()
  
  const {
    state: {data: count},
    setState,
  } = useConversationCount();
  
  useEventListener({
    names: [RELOAD_UNREAD_CONVERSATION_COUNT],
    callback: () =>
      setState(({data, ...s}, merge) => merge(s, {data: (data || 0) + 1})),
  });

  useLayoutEffect(() => {
    navigation.setOptions({tabBarBadge: count});
  }, [count]);
  
  return <Icon focused={focused} Icon={Chat} />;
});
```

Since `tabBarIcon` is always mounted twice so the `CountableInbox` component will be created twice and will cause `useConversationCount` to make network request twice.

https://github.com/react-navigation/react-navigation/blob/860adbfd8b4c493dc3e28c99ad5d78699c2f8797/packages/bottom-tabs/src/views/TabBarIcon.tsx#L49

From then referenced file i am sure the `renderIcon` can be called once based on the active prop.

Please help refactor `TabBarIcon`

satya164:
https://github.com/react-navigation/react-navigation/pull/9273#issuecomment-774031982

myckhel:
Hello,
i request for this issue to be opened.

i think the issue has not been resolved and no helping suggestion was provided.

PR #9273, opened 2021-02-02
fix: refactored tabBarIcon

myckhel:
Fix for duplicate tabBarIcon issue https://github.com/react-navigation/react-navigation/issues/9272

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabItem.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabItem.tsx
@@ -232,19 +232,16 @@ export default function BottomTabBarItem({
       return null;
     }
 
-    const activeOpacity = focused ? 1 : 0;
-    const inactiveOpacity = focused ? 0 : 1;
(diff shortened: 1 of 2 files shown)


myckhel:
installation from forked repo:
`yarn add https://github.com/myckhel/react-navigation.git#react-navigation-bottom-tabs-v5.11.7-gitpkg`

satya164:
There are 2 icons overlaid to prevent any rendered images from flickering when changing. You shouldn't be using it to perform any side effects. Do those side effects inside your screen.

myckhel:
💔😔😒

I guess there is no other way to achieve the real time badge counter update i was trying to do.

Since the screen is not automatically focused so i won't be able to set the screen options for bagdeCount. 

Does this changes produces the image flickering issue?

Is there any other way we can try not to create 2 tabBarIcons?

myckhel:
Please can you respond to my questions?

PR #9283, opened 2021-02-05
fix: upgrade masked view dependency so users can use latest version

LouiseBC:
# Problem:
React Navigation has a dependency on an old release of masked-view, and all updates are now being made in a different namespace (@react-native-community -> @react-native-masked-view). In my project I need the latest version of masked-view because the older version causes a bug with Detox. Since the masked-view library has changed to a new namespace I can't simply use a newer version of the package than specified in the package.json, and they can't exist side by side because the native modules have the same name.

# Solution:
Update by switching to the dependency from the new namespace

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@expo/vector-icons": "^10.0.0",
     "@react-native-async-storage/async-storage": "^1.13.1",
-    "@react-native-community/masked-view": "0.1.10",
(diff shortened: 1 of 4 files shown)


satya164:
Thanks for the PR. But changing the package name is a breaking so we can't do it in a minor version. It'll be changed in v6.

For now, I'd suggest you to use [patch-package](https://www.npmjs.com/package/patch-package) to change the imports in the library.

PR #9379, opened 2021-03-02
[example] Upgrade expo-cli to v4.1.4

bntzio:
Upgrades to `expo-cli` v4.1.4 to fix the example app.

Prior to that version running the example app will throw the following error if users do not belong to the `react-navigation` organization.

Screen Shot 2021-03-01 at 11 58 49 PM

By upgrading to v4.1.4 (see https://github.com/expo/expo-cli/pull/3148) if users do not belong to the org the app fallbacks to offline mode, so they can successfully start the example app.

Screen Shot 2021-03-02 at 12 08 04 AM

Screen Shot 2021-03-02 at 12 12 15 AM


example/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
satya164, line 1: Why is this change necessary?

satya164 (commented):
Th

example/tsconfig.json
bntzio: This it’s automatically generated by expo when you start the app, I believe it finds for a tsconfig and if there’s none, it creates this for you.

satya164:
thanks for the PR. expo-cli is latest version on main branch.

PR #9405, opened 2021-03-10
fix: respect headerStatusBarHeight option in Stack

adamTrz:
Use `headerStatusBarHeight` that was passed in `options`

### DIFF
--- a/packages/stack/src/views/Header/Header.tsx
+++ b/packages/stack/src/views/Header/Header.tsx
@@ -46,7 +46,6 @@ export default React.memo(function Header({
     (state) => state.routes[0].key === route.key
   );
 
-
   const statusBarHeight =
     options.headerStatusBarHeight !== undefined
       ? options.headerStatusBarHeight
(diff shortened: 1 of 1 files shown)


packages/stack/src/views/Header/Header.tsx
```diff
@@ -46,8 +46,13 @@ export default React.memo(function Header({
...
   );
 
+
   const statusBarHeight =
-    (isModal && !isFirstRouteInParent) || isParentHeaderShown ? 0 : insets.top;
+    options.headerStatusBarHeight !== undefined
+    ? options.headerStatusBarHeight
+    : ((isModal && !isFirstRouteInParent) || isParentHeaderShown
+      ? 0
+      : insets.top);
```
satya164, line 55:
```suggestion
    options.headerStatusBarHeight !== undefined
      ? options.headerStatusBarHeight
      : (isModal && !isFirstRouteInParent) || isParentHeaderShown
      ? 0
      : insets.top;
```

packages/stack/src/views/Header/Header.tsx
```diff
@@ -46,8 +46,13 @@ export default React.memo(function Header({
     (state) => state.routes[0].key === route.key
   );
 
+
```
satya164, line 49:
```suggestion
```

Issue #9128, opened 2020-12-04
Web - Navigation is broken after going back and forth


yairopro:
**Current Behavior**

Here's the demo [from the blog ](https://reactnavigation.org/blog/2020/05/16/web-support)

![web-broken](https://user-images.githubusercontent.com/8110568/101173483-469be100-364b-11eb-8f2c-dcd69a7562ac.gif)

**Expected Behavior**

The navigation to not be broken.

**How to reproduce**

Doing the following actions will break the navigation stack on web:
1. backward
2. forward
3. backward again
4. Now the navigation is broken. You cannot go backward or forward anymore.

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| @react-navigation/native       | 5.8.10
| @react-navigation/stack        | 5.12.8

StellarOcean:
This issue is reproducible with the following code: 
```
import 'react-native-gesture-handler';
import React from 'react';
import { NavigationContainer, useLinkTo } from '@react-navigation/native';
import { View, Text, Button } from 'react-native';

import { createStackNavigator } from '@react-navigation/stack';

const linking = {
  prefixes: ['https://localhost'],
  config: {
    screens: {
      Main: {
        screens: {
          Home: '/Main/Home',
          Details: '/Main/Details'
        }
      },
      MyModal: '/MyModal'
    },
  }
}

function HomeScreen() {
  const linkTo = useLinkTo();

  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
      <Text style={{ fontSize: 30 }}>This is the home screen!</Text>
      <Button
        onPress={() => linkTo('/MyModal')}
        title="Open Modal"
      />
    </View>
  );
}

function DetailsScreen() {
  return (
    <View>
      <Text>Details</Text>
    </View>
  );
}

function ModalScreen() {
  const linkTo = useLinkTo();

  return (
    <View style={{ flex: 1, alignItems: 'center', justifyContent: 'center' }}>
      <Text style={{ fontSize: 30 }}>This is a modal!</Text>
      <Button onPress={() => linkTo('/Main/Home')} title="Dismiss" />
    </View>
  );
}


const MainStack = createStackNavigator();
const RootStack = createStackNavigator();

function MainStackScreen() {
  return (
    <MainStack.Navigator>
      <MainStack.Screen name="Home" component={HomeScreen} />
      <MainStack.Screen name="Details" component={DetailsScreen} />
[...]

PR #9408, opened 2021-03-11
fix: web navigation broken, when going back and forth in history

HeAtNet:
This fix resolves the #9128 issue
https://github.com/react-navigation/react-navigation/issues/9128

Until not merged, you can try this, by adding the following code to the package.json file's scripts property, and it will use the fixed code for @react-navigation/native.
`"postinstall": "git clone --branch web-navigation-fix https://github.com/HeAtNet/react-navigation.git temp_react_navigation && cd temp_react_navigation/packages/native && npm install @types/jest --save-dev && npm i && npm run prepare && rm node_modules -r -f && cd ../../.. && rm node_modules/@react-navigation/native -r -f && cp temp_react_navigation/packages/native node_modules/@react-navigation/ -r && rm temp_react_navigation -r -f",`


muqker:
This did indeed solve the issue for me. I hope the pull request gets accepted. For now, I am using it as a patch-package.

TikiTDO:
@satya164 Any chance this PR could get a review? This fixes a really nasty web navigation issue. The only task failing for this PR is the `welcome` github action, which appears to be erroring out immediately.

satya164:
I'll have some time to test this next week.

adamhari:
Hoping this makes it in soon, I am anxious to support browser navigation with react-native-web. This library works really well apart from that!

packages/native/src/useLinking.tsx
```diff
@@ -46,9 +46,16 @@ const createMemoryHistory = () => {
...
+        let currentIndex = items.findIndex((item) => item.id === id);
+        if (currentIndex < 0) {
+          currentIndex = 0;
+        }
 
-        return index > -1 ? index : 0;
+        // Fix createMemoryHistory.index variable's value, because it will go out of sync, when navigating
+        // with the browser. The 'popstate' event not sets the createMemoryHistory.index variable, but
+        // calls this function, so the corrrect value can be set.
+        index = currentIndex;
```
satya164, line 57: We shouldn't change the internal state inside a getter. Please move this logic to the `onPopState` function which is the appropriate place.

packages/native/src/useLinking.tsx
```diff
@@ -95,7 +102,12 @@ const createMemoryHistory = () => {
 
       const id = window.history.state?.id ?? nanoid();
 
-      if (items.length) {
+      if (items.findIndex((item) => item.id === id) < 0) {
```
satya164, line 105: This is also going to be `true` when `items.length` is`0`, causing the `else` clause in this condition to never get called. Should this check be after the `items.length` check?

satya164 (changes requested):
Thanks for the PR. Can you address the review? I'll merge after they are addressed.

Hirbod:
@HeAtNet could you add the requested change requests by satya?

packages/native/src/useLinking.tsx
HeAtNet: Yes, that is the appropriate place. I fixed.

packages/native/src/useLinking.tsx
HeAtNet: That's true. Furthermore the third statement is not needed, so I changed the conditions.

HeAtNet:
I made the requested changes.
The integration test failed and I don't know why, but I hope this is a good solution now.

TikiTDO:
@HeAtNet You may want to merge in main (and perhaps target main as the merge target).

takacskalman:
Any updates?

PofMagicfingers:
I think @HeAtNet needs to change the merge target to main, for this to work according to @TikiTDO

Hirbod:
@satya164 can you work with this PR or would you like me to create a new PR targeting main? This one is sitting here for a while now and web navigation with browser navigation buttons are broken ever since.

satya164:
Thank you for updating the PR. I rebased it and merged it in fb84805

Hirbod:
While this fix works fine while navigation back and forth now, it still throws a crash when you reach the initial screen. 
It works perfectly fine when you click back and forth on any other screen.
Bildschirmfoto 2021-09-28 um 11 20 42

PofMagicfingers:
I think we need to open a new issue, this one is not tracked anymore

PR #9412, opened 2021-03-11
fix: remove unnecessary conditions in the Badge component 

Stanislavja:
Remove unnecessary condition for boolean value.
Reorder if else condition for little improve.
Moved props with default value at the end.


satya164 (approved):
Thanks

PR #9416, opened 2021-03-13
fix: type checking for RouteConfig with TypedNavigators

mjpizz:
When I was setting up components for my Screens, I realized that there was no type checking that would catch if a particular Screen component was expecting the wrong `route` or `navigation` prop.

Looking more closely at `RouteConfig`, it looks like we can add stronger types for the `route` and `navigation` to allow more static typechecks when configuring routes. Totally open to feedback if I've missed something! It's possible that it was an intentional decision to omit types in this config.


satya164:
CI didn't run on this PR for some reason, are there no type errors after you change this? The main reason it's `any` is because otherwise it gives type errors with types such as `StackNavigationProp`, `CompositeNavigationProp` etc.

mjpizz:
Lint and tests all pass for me. However you're right, `tsc` is pretty unhappy with this change. I'll make another pass to see what would be required to tighten this up.

mjpizz:
Looks like the type errors are mostly in examples and tests. And at least a few of those tests simply under-specify their types, so (I think) those are genuine issues to fix in the tests. Will do a bit more investigation when I have more time.

JaapWeijland:
Sounds as a neat addition. Is this issue fixed by this?

https://github.com/react-navigation/react-navigation/issues/10153

satya164:

PR #9457, opened 2021-03-25
fix: remove calls to removed Keyboard.removeListener in useIsKeyboardShown

janicduplessis:
Keyboard.removeEventListener was restored in https://github.com/facebook/react-native/commit/035718ba97bb44c68f2a4ccdd95e537e3d28690c but not Keyboard.removeListener. In this case we can use the subscription object returned from addListener. The addListener method always returned a subscription object so no need to check if the return value is an object (like in other cases of addEventListener methods).

### DIFF
--- a/packages/bottom-tabs/src/utils/useIsKeyboardShown.tsx
+++ b/packages/bottom-tabs/src/utils/useIsKeyboardShown.tsx
@@ -9,6 +9,7 @@ export default function useIsKeyboardShown() {
     const handleKeyboardHide = () => setIsKeyboardShown(false);
 
     let subscriptions: EmitterSubscription[];
+
     if (Platform.OS === 'ios') {
       subscriptions = [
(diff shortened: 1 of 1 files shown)


packages/bottom-tabs/src/utils/useIsKeyboardShown.tsx
```diff
@@ -8,22 +8,21 @@ export default function useIsKeyboardShown() {
     const handleKeyboardShow = () => setIsKeyboardShown(true);
     const handleKeyboardHide = () => setIsKeyboardShown(false);
 
+    let subscriptions: EmitterSubscription[];
```
satya164, line 11:
```suggestion
    let subscriptions: EmitterSubscription[];

```

satya164:
Thanks

erickjth:
Hey guys, 
This is still an issue in `@react-navigation/bottom-tabs` version 5.11.15 using RN 0.65

PR #9458, opened 2021-03-25
feat: add props from native Screen component

WoLewicki:
PR adding options exposed to the `Screen` component of `react-native-screens` in https://github.com/software-mansion/react-native-screens/pull/863. It gives the user the ability to control orientation and status bar management per screen of the navigator. 

Changes:
- Added `NativeScreenTraitsProps` type to stack&drawer&tabs navigators
- Bump example and stack&drawer&tabs to use `react-native-screens` v3.3.0
- Hardcoded types from RNS library directly to native-stack to avoid cross-versioning issues
- Switched from using react-native's StatusBar to react-native-screens' in example app


satya164:

PR #9464, opened 2021-03-26
feat: improve useNavigationState typing

jemmyphan:


satya164 (changes requested):
Thank you for the PR. Can you use `ParamList` instead of `P`? We try to avoid single letter generics for better clarity. We have `T`, but it's for type inference rather than user provided like `ParamList`.

jemmyphan:
> Thank you for the PR. Can you use `ParamList` instead of `P`? We try to avoid single letter generics for better clarity. We have `T`, but it's for type inference rather than user provided like `ParamList`.

Thanks @satya164 for the review.
I have updated the PR, can you check it again?

satya164 (approved):

satya164:
Thanks

PR #9477, opened 2021-04-01
fix: check for screens enabled in ScreenContainer

WoLewicki:
Added a check for `screensEnabled` in `ScreenContainer` to make it behave the same as `Screen` in `ResourceSavingScreen.tsx`. Should resolve https://github.com/expo/expo/issues/12374.

### DIFF
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -6,7 +6,7 @@ import {
   ThemeContext,
   NavigationScreenProp,
 } from 'react-navigation';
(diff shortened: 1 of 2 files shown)


satya164 (approved):

alexandrius:
Is this patch published?

WoLewicki:
I think it was 2 days ago: https://github.com/react-navigation/react-navigation/commit/199baecff1d9a6212d5320e4ad440807f707711a

sanborN93:
How can i apply this patch if i am using react navigation 5?

Issue #9496, opened 2021-04-12
Stack Navigator back button disappears completely when keeping button pressed


michaelknoch:
**Current Behavior**
The header back button fades out completely when keeping the button pressed.

https://user-images.githubusercontent.com/5617793/114461646-45268a80-9be2-11eb-9a84-38b341d5fec2.mov

**Expected Behavior**
The button should fade out a bit but remain visible.

https://user-images.githubusercontent.com/5617793/114459987-7e122f80-9be1-11eb-9d72-47fbd265dcff.mov

**How to reproduce**
- clone https://github.com/michaelknoch/react-navigation-bug-repro
- `npm i`
- `cd ios && pod install`
- run ios project on device or simulator

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | ios
| @react-navigation/native       |5.9.4
| @react-navigation/stack        |5.14.4
| react-native-gesture-handler   |1.10.3
| react-native-safe-area-context |3.2.0
| react-native-screens           |3.1.0
| react-native                   |0.64.0
| expo                           |-
| node                           |v14.16.0
| npm or yarn                    |npm

michaelknoch:
fixed in https://github.com/react-navigation/react-navigation/pull/9498

PR #9498, opened 2021-04-13
fix default `pressOpacity` for `TouchableItem.ios.tsx`

michaelknoch:
fixes https://github.com/react-navigation/react-navigation/issues/9496

`this.props.pressOpacity` is used for the opacity animation. HeaderBackButton does not pass an explicit value for this prop so the default should be used. But the default does not exist because the name `activeOpacity` is invalid.


michaelknoch:
@satya164 can you tell me what it needs to get this in? I would really like to see this bug fixed in our apps.

satya164 (approved):
Thank you and sorry for the delay

PR #9502, opened 2021-04-15
Fix NavigationNavigator type error

magicismight:
```
export interface NavigationRouter<State = NavigationState, Options = {}> {
```


satya164 (approved):
Thanks for the PR

PR #9514, opened 2021-04-19
Improve params type

maulanasatyaadi:
Prevent type check error when use useRoute().params using typescript.
If I using ```const { params } = useRouter()``` there is alwas error if I put params data like ```params.id```, the error is like ```Property 'id' does  not exist on type 'object'```.


satya164:
Hey, thanks for the PR. It is intentional. You need to annotate with the correct type so that it doesn't have type errors.

https://reactnavigation.org/docs/typescript/#annotating-useroute

PR #9518, opened 2021-04-20
chore: update @types/react-native and remove ts-ignore

Naturalclar:
updated @types/react-native and removed ts-ignore where possible


satya164 (approved):
Thank you!

PR #9522, opened 2021-04-21
feat: specify type for `navigation` prop in `options`

Krisztiaan:
Typed `navigaton` for screen `options` instead of `any`


satya164:
Thanks for the PR. As you can see on the CI, this causes type errors. This is why it's currently `any`.

Krisztiaan:
May revisit this later to add `Parent` navigator typing to `createXYNavigator`s, enabling safe typing on these parameters and some more.

PR #9527, opened 2021-04-23
Change react-native-community to react-native-masked-view

FonDorn:
From React Native 64 can't use @react-navigation/stack correctly due to having dependency on the old masked-view version.
https://github.com/react-native-masked-view/masked-view/issues/111

This PR will fix that problem. Thanks.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "@expo/vector-icons": "^10.0.0",
     "@react-native-async-storage/async-storage": "^1.13.1",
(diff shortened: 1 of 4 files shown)


brentvatne:
we need to be careful if we roll this out to v5 because it's technically a breaking change and therefore more suitable for a major version bump

satya164:
Thanks for the PR. But as Brent pointed out, changing the package name is a breaking change so we can't do it in a minor version. It'll be changed in v6.

PR #9529, opened 2021-04-24
feat: add the 'drawerPress' event

ger88555:
Added the 'drawerPress' event, similar to the 'tabPress' event.

Navigation can be prevented with the  ```preventDefault``` method.


packages/drawer/src/views/DrawerItemList.tsx
```diff
@@ -27,6 +27,24 @@ export default function DrawerItemList({
...
     const focused = i === state.index;
+
+    const onPress = () => {
+      const event = navigation.emit({
+        type: 'drawerPress',
+        target: route.key,
+        canPreventDefault: true,
+      });
+
+      if (!focused && !event.defaultPrevented) {
```
satya164, line 38: Adding the `focused` check here is problematic. We still need to dispatch an action to close the drawer even when the screen is focused.

```suggestion
      if (!event.defaultPrevented) {
```

satya164 (changes requested):
Thanks for the PR. Please fill the pull request template when opening PRs.

satya164 (approved):
Thank you

PR #9540, opened 2021-04-29
Feat: Linking context

kopax-polyconseil:
Allow to export linking context so module for react navigation can access the linking configuration. It's a must.


satya164:
Thanks for the PR. Can you describe the motivation behind why you need it? What's your use case?

satya164:
Closing due to no reply

PR #9564, opened 2021-05-08
feat: add callback for tab index change

skipness:
I would like to know the new tab index when tab is changed so I created this PR

### DIFF
--- a/packages/material-top-tabs/src/types.tsx
+++ b/packages/material-top-tabs/src/types.tsx
@@ -209,6 +209,10 @@ export type MaterialTopTabNavigationConfig = Omit<
   | 'lazy'
 > & {
   /**
+   * Callback for tab index change
+   */
+  onIndexChange?: (index: number) => void;
+  /**
    * Function that returns a React element to display as the tab bar.
    */
   tabBar?: (props: MaterialTopTabBarProps) => React.ReactNode;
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. Can you describe the motivation behind this change? What's your use case?

skipness:
> Thanks for the PR. Can you describe the motivation behind this change? What's your use case?

Hello, thanks for replying!
In my case, I would like to perform list refresh when tab is focused. Combination of `useIsFocused` and `useScrollToTop` should be good but the list component is used all over the app, I don't want this two hooks stick with the list component. Instead, I created a list of refs in the container where the tab navigator is, get the ref with tab index then scroll the list. 
I think `onIndexChange` callback may come in handy in this situation(please let me know if my approach is correct) when some components are too generic, may be we should expose `onSwipeStart` and `onSwipeEnd` to user as well?

skipness:
any feedback will be great

khzouroussama (approved):
look good to me

khzouroussama:
this is very important as i got in the case where i built a wrapper component on top of this and i needed to keep track of tabIndex  and update of the navigation state is slow

skipness:
resolved conflicts

MonsterAnan:
Hi, any fixed?

satya164:
Hey, thanks a lot for the PR. But you can already check when the index changes using the `state` listener on the `screenListeners` prop https://reactnavigation.org/docs/navigation-events/#screenlisteners-prop-on-the-navigator

MonsterAnan:
@satya164 very thanks for your reply!
`state ` on the `screenListeners `is solved my problem!

satya164:
Thanks for the PR. However, you can get notified of state changes with the `state` event which already provides the same functionality: https://reactnavigation.org/docs/navigation-events/#state

PR #9592, opened 2021-05-16
Enable iPad Trackpad Two Finger back gesture in Stack Navigator

MrLoh:
This enables navigating back on iPadOS by swiping right with two fingers on a connected trackpad, which is the default behaviour of the native navigator. This could be made configurable, but I am not sure there is a reason to not have it enabled by default. Happy to discuss and adjust the PR.


satya164 (approved):
Thank you for the PR

Issue #9558, opened 2021-05-06
Tapping outside doesn't close the drawer on iOS


aschick82:
**Current Behavior**

- Running on React Native 0.63.3 and React Navigation Drawer 5.12.5
- When I tap on the overlay outside of drawer when it's open, the drawer doesn't go away. Also I am not able to open it by swiping from left. This line also exists in my main index.js file: 

**import 'react-native-gesture-handler';**

**Expected Behavior**

- Drawer to be closed by tapping outside on iOS

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | iOS
| @react-navigation/native       |   5.9.4
| @react-navigation/drawer       | 5.12.5
| react-native-reanimated        | 2.1.0
| react-native-gesture-handler   | 1.10.3
| react-native-safe-area-context | 3.2.0
| react-native-screens           | 3.1.1
| react-native                   | 0.63.3
| node                           | v14.15.3
| npm or yarn                    | yarn

aryalprakash:
I'm also facing this issue. It's working fine on Android though.

aryalprakash:
@Romick2005 Getting this error when installing version 1.13.2 of react-native-reanimated
`error: Error: Unable to resolve module ./reanimated2 from /../node_modules/react-native-reanimated/src/Animated.js:
`

Romick2005:
@aryalprakash Have you clean up after reanimated2? I mean remove "react-native-reanimated/plugin" in babel.config.js and of course "npx pod-install"? You should better go with 
`rm -rf node_modules && rm -rf package-lock.json && rm -rf ios/Pods && rm -rf ios/Podfile.lock && npm install --force && npx pod-install`

aryalprakash:
@Romick2005 Thanks. I had to use `npm start -- --reset-cache` even after doing those.
It's working now but there's some flickering when opening/closing drawer.

pinpointpanda:
I've also seen this in our project recently -- although we're not running latest `reanimated@2` yet. Will upgrade that tomorrow and if still present will provide a Snack and/or minimal repro.

tomwanzek:
I can confirm that the issue replicates for us on iOS after upgrading `react-native-reanimated` from 1.8.0 to 2.2.0.

[...]

PR #9594, opened 2021-05-17
#9558 Fix drawer close on typing outside and swipe action on iOS.

Romick2005:
This can be used to fix usage react-navigation v5 with react-native-reanimated v2, because Animated.event does not work for iOS and does not give any warnings or errors. In reanimated v2 we need to use hook useAnimatedGestureHandler that is not available as v5 uses component based drawer.
This fix also applicable for navigation v6, but for legacy drawer that do not use hooks.


satya164:
Thanks for the PR. This sounds like a bug in Reanimated 2? It is supposed to be backward compatible so the same code should work in v2 without any change

adderly:
This fixes the drawer swipe on IOS.

Romick2005:
Update as for today close outside works as expected. Drawer swipe does not work only in debug mode.

Winglonelion:
@Romick2005 Great works! 💯  But the main brain goes so far, could you resolve conflict?

Romick2005:
> @Romick2005 Great works! 💯 But the main branch goes so far, could you take a look to resolve conflict?

Done.

JohnBerd:
Thank you for this great job @Romick2005. Really need this to be merged, would it be a lot of work to get the code approval?

kg-currenxie:
Any news on this? 🤔

satya164:

PR #9667, opened 2021-06-11
test: upgrade to the new Playwright test-runner

mxschmitt:
Hey!

I made the following changes:

- Replaced Jest by the new Playwright test-runner
- Disabled the Jest linting rules for the Playwright e2e tests
- Rewrote the tests to the new test-runner
- Adjusted `Link.test.ts` which should be less flaky
- The tests run now across all three browsers: Chromium, Firefox, and WebKit

See here for reference about the new test-runner: https://playwright.dev/docs/test-intro

I extracted a fix for Netlify in #9668.

Let me know if you have any questions.

PR #9668, opened 2021-06-11
chore: add nvmrc for using LTS Node.js version

mxschmitt:
This should fix the Netlify builds too.
See here for reference: https://github.com/nvm-sh/nvm#nvmrc

The integration tests are failing but they are getting fixed in #9667.


satya164 (approved):
Thanks. I had no idea how to fix this,

PR #9672, opened 2021-06-14
fix: sort wildcard and :params 

zhigang1992:
We want to server user profile like https://github.com/zhigang1992 or https://twitter.com/zhigang1992 along side with screens in the app.

```typescript
const configs = {
  screens: {
    debugPanel: 'debugPanel',
    user: ':userSlug',
  },
}
getStateFromPath('/debugPanel', configs)
```

and

```typescript
const configs = {
  screens: {
    debugPanel: 'debugPanel',
    user: '*',
  },
}
getStateFromPath('/debugPanel', configs)
```

have different behaviors under Hermes engine, because we are only sorting '*'.

IMHO `:urlCode` is also a form of wildCard.

### DIFF
--- a/packages/core/src/__tests__/getStateFromPath.test.tsx
+++ b/packages/core/src/__tests__/getStateFromPath.test.tsx
@@ -2196,7 +2196,7 @@ it('tries to match wildcard patterns at the end', () => {
             path: '/bar/:id/',
             screens: {
               404: '*',
-              matches: ':params',
+              UserProfile: ':userSlug',
               Test: 'test',
             },
           },
--- a/packages/core/src/__tests__/getStateFromPath.test.tsx
+++ b/packages/core/src/__tests__/getStateFromPath.test.tsx
@@ -2196,6 +2196,7 @@ it('tries to match wildcard patterns at the end', () => {
             path: '/bar/:id/',
             screens: {
               404: '*',
+              matches: ':params',
               Test: 'test',
             },
           },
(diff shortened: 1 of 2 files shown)


packages/core/src/__tests__/getStateFromPath.test.tsx
```diff
@@ -2196,6 +2196,7 @@ it('tries to match wildcard patterns at the end', () => {
             path: '/bar/:id/',
             screens: {
               404: '*',
+              matches: ':params',
```
satya164, line 2199: Looks like a leftover?

satya164 (changes requested):
Thanks for the PR. Can you add some comments similar to those that were there previously?

packages/core/src/__tests__/getStateFromPath.test.tsx
zhigang1992: No, it's meant to be here. 

that `:params` should take lower priority compare to 'test'. Just like '*'?

zhigang1992: Since `:params` and `*` are both wildcards

satya164: Can you rename `matches` to some name like `Test1` etc.? `matches` looks like it's some option and not a screen name

zhigang1992:
@satya164 please take another look

satya164:
Thank you!

PR #9689, opened 2021-06-22
chore: upgrade to expo-github-actions v6

byCedric:
This version allows you to skip installing `expo-cli`. If you want to use EAS, this also got you covered! Having no CLI installed prevents the GH action from validating the `EXPO_TOKEN`, the [GH action handles this case](https://github.com/expo/expo-github-action/blob/main/src/tools.ts#L54) and I'm pretty confident this is set up properly 😄


satya164 (approved):
Thanks a lot

PR #9693, opened 2021-06-25
fix: import animated from react-native-reanimated

zhigang1992:


satya164:
Thanks for the PR, but stack doesn't use Reanimated.

zhigang1992:
Ah, I see. sry about that. Should have looked deeper?

In that case why do we need reanimated when setup react-navigation?

satya164:
@zhigang1992 you need it if you use drawer (and material top tabs in 4.x/5.x)

PR #9695, opened 2021-06-25
feat: use forwardRef with DrawerContent and DrawerContentScrollView

### LINKED ISSUE #9350: DrawerContentScrollView should use forwardRef
**Current Behavior**

DrawerContentScrollView does not forward refs so I can't implement any additional scroll behavior

**Expected Behavior**

DrawerContentScrollView should use forwardRef so ref gets forwarded to the internal ScrollView

   "@react-navigation/drawer": "^5.12.2",
    "@react-navigation/native": "^5.9.2",

alternacrow:
Closes #9350 

Fixed to use ref in DrawerContent and DrawerContentScrollView.
Please review.

### DIFF
--- a/packages/drawer/src/views/DrawerContent.tsx
+++ b/packages/drawer/src/views/DrawerContent.tsx
@@ -1,24 +1,25 @@
 import * as React from 'react';
+import type { ScrollView } from 'react-native';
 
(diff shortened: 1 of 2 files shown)


satya164 (approved):
Thanks for the PR

PR #9697, opened 2021-06-25
feat: allows hide specific tab in bottom tab bar

baptistemon:
allows hide specific tab in bottom tab bar

react-native-paper pull request : https://github.com/callstack/react-native-paper/pull/2791

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -104,6 +104,11 @@ export type BottomTabNavigationOptions = HeaderOptions & {
   headerShown?: boolean;
 
   /**
+   * Whether to show the tab.
+   */
+  isVisibleInTabBar?: boolean;
+
+  /**
    * Title string of a tab displayed in the tab bar
    * or a function that given { focused: boolean, color: string, position: 'below-icon' | 'beside-icon' } returns a React.Node to display in tab bar.
    *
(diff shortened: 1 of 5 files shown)


satya164:
Thanks for the PR, but I don't think it makes sense to have a tab that is not accessible to the user - and not to have any tabs selected in the tab bar.

I recommend nesting a stack navigator inside your tab navigator to have another screen that isn't part of the tab navigator.

baptistemon:
@satya164, the PR description is a bit light, I explain it a bit more :

The purpose of the option `isVisibleIntabBar` is the same as the option `tabBarButton: () => null`. I harmonize it between bottom TabBar and material bottom TabBar. (tabBarButton is not present in material bottom tab bar).

You may want to have a Home page in your `TabNavigator`, don't want to access it from the bottom tab bar but from a button in your page header. With the `isVisibleIntabBar` option, it is quite simple to do it.

satya164:
The purpose of a tab bar (or other navigation-related UI elements) is to let the user know where they are in the app. With a tab bar, people can be on a screen, switch to a different screen, and go back to where they were. This pattern makes it impossible to achieve such a common UX.

I don't think it makes sense to support and promote such patterns in the library. If you have some use case that cannot be achieved by nesting a stack navigator inside the tab navigator, then please open an RFC with some examples of existing apps which do this and rationale on why such a pattern is necessary https://github.com/react-navigation/rfcs

PR #9705, opened 2021-06-27
fix: blink while switching screens

Libesto:
Fix to prevent showing blank screen while screens switching process.
Currently we can see it using dark background color in screens.
This fix also makes clear work with eager loading of screens (`detachInactiveScreens={false}` `lazy={false}`).


satya164 (approved):
Thanks for the PR. Sounds like a smart fix.

PR #9738, opened 2021-07-14
fix: update isClosing in Stack Card

ouoru:
I noticed that accessing `closing` through `cardStyleInterpolator` would always return an Animated node with value 0. It looks like it isn't being updated anywhere, so I added it to the `animate` method.

I am using this functionality to allow screens to have different in and out transitions.

On a side note, I feel like this would be more useful as a boolean, instead of an Animated value.

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -184,6 +184,8 @@ export default class Card extends React.Component<Props> {
 
     this.lastToValue = toValue;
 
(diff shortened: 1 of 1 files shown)


todorone (approved):

satya164 (approved):
Thanks for the PR

satya164:
> On a side note, I feel like this would be more useful as a boolean, instead of an Animated value.

The reason behind an animated value is to avoid a re-render just before the animation starts, which could cause frame drops.

PR #9760, opened 2021-07-22
fix: Improve slow pop animations on stack

cristianoccazinsp:
Related https://github.com/software-mansion/react-native-screens/issues/1020

Can this be released in the 5.x branch and then ported to the 6.x ?  I see 6.x has changed significantly, so I would review if this is necessary in that branch.


cristianoccazinsp:
@satya164 why isn't this running the tests? I see a new major version is being released already, can you please review these tiny changes and see if they are fit for the new release? This greatly improves back navigation lag when the previous screen is heavy, and our app can't really have smooth transitions without it.

Elter71:
I see the issue was closed on 8 Sep. I think we can close this PR. Don't you think @cristianocca @satya164?

cristianoccazinsp:
Agreed. This is no longer needed as screens fixed it already.

PR #9773, opened 2021-07-28
fix: check current route in its children before checking all of them to handle navigation action

chuganzy:
Currently React Navigation checks if it can handle navigation action in following order as noted in `useOnAction`:

1. check if current router can handle
2. if not, check if parent can handle
3. if not, check if children can handle

In step 3, it checks the listeners from the newest order.
However this might be a problem especially in Tab navigator where it lazily initialize its children, so it always checks if the latest initialized tab index can handle the action first.

For example, if we made the navigator to be following (sorry if the drawing is not clear 😅 ), and wanted to go to `NoteDetail` from `AddNote`, it goes to the second tab regardless of the current selected tab once we open the second tab.

Group

To solve this, this PR makes handlers to be stored as objects, not arrays, and look up the handler using the current route key first. There were some places I was not too sure, so leaving comments by myself.

### DIFF
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -120,23 +120,23 @@ const BaseNavigationContainer = React.forwardRef(
           | NavigationAction
           | ((state: NavigationState) => NavigationAction)
       ) => {
-        if (listeners.focus[0] == null) {
+        const focus = Object.values(listeners.focus)[0];
+        if (focus == null) {
           console.error(NOT_INITIALIZED_ERROR);
         } else {
-          listeners.focus[0]((navigation) => navigation.dispatch(action));
+          focus((navigation) => navigation.dispatch(action));
         }
       },
       [listeners.focus]
     );
 
     const canGoBack = React.useCallback(() => {
-      if (listeners.focus[0] == null) {
+      const focus = Object.values(listeners.focus)[0];
+      if (focus == null) {
         return false;
       }
 
-      const { result, handled } = listeners.focus[0]((navigation) =>
-        navigation.canGoBack()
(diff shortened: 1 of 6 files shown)


packages/core/src/useFocusedListenersChildrenAdapter.tsx
```diff
@@ -24,7 +26,7 @@ export default function useFocusedListenersChildrenAdapter({
   const listener = React.useCallback(
     (callback: FocusedNavigationCallback<any>) => {
       if (navigation.isFocused()) {
-        for (const listener of focusedListeners) {
+        for (const listener of Object.values(focusedListeners)) {
```
chuganzy, line 29: Not sure if the order matters here.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -120,23 +120,23 @@ const BaseNavigationContainer = React.forwardRef(
           | NavigationAction
           | ((state: NavigationState) => NavigationAction)
       ) => {
-        if (listeners.focus[0] == null) {
+        const focus = Object.values(listeners.focus)[0];
```
chuganzy, line 123: Please see if these changes make sense.

chuganzy:
can anyone start to look at this?

yhkaplan:
@satya164 @kacperkapusciak Mentioning just in case

yhkaplan:
Hi @satya164 can you check this?

satya164:
Thanks for the PR. However, we don't want to make any changes to how this works since this behavior is being deprecated. More details in another similar PR: https://github.com/react-navigation/react-navigation/pull/10596#issuecomment-1171101159

PR #9774, opened 2021-07-29
feat: expose header height in native-stack

WoLewicki:
Added default header height in native-stack.

### DIFF
--- a/packages/elements/src/Header/getDefaultHeaderHeight.tsx
+++ b/packages/elements/src/Header/getDefaultHeaderHeight.tsx
@@ -4,7 +4,7 @@ import type { Layout } from '../types';
 
 export default function getDefaultHeaderHeight(
   layout: Layout,
-  modal: boolean,
+  modalPresentation: boolean,
   statusBarHeight: number
 ): number {
   let headerHeight;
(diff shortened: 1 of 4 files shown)


packages/elements/src/Header/getDefaultHeaderHeight.tsx
```diff
@@ -5,15 +5,16 @@ import type { Layout } from '../types';
 export default function getDefaultHeaderHeight(
   layout: Layout,
   modal: boolean,
-  statusBarHeight: number
+  statusBarHeight: number,
+  fullScreenModal = false,
```
satya164: I think this arg is unnecessary. Just pass `!fullScreenModal` for the `modal` argument. Can rename the argument to `modalPresentation` or something similar to make it more obvious.

satya164 (changes requested):

packages/native-stack/src/views/NativeStackView.tsx
```diff
@@ -86,6 +118,145 @@ const MaybeNestedStack = ({
...
+    presentation = 'card';
+  }
+
+  const isHeaderInPush = isAndroid
+    ? headerShown
+    : presentation === 'card' && headerShown !== false;
+
+  const isParentHeaderShown = React.useContext(HeaderShownContext);
+  const insets = useSafeAreaInsets();
+  const headerStatusBarHeight = isParentHeaderShown ? 0 : insets.top;
```
satya164, line 169: Is this correct in case of native stack? e.g. if it's nested inside of another stack (JS or native), will it hide the statusbar inset?

PR #9793, opened 2021-08-05
feat(elements): add style to SafeAreaProviderCompat props

aleppos:
When using SafeAreaProvider I can pass style directly and when I decide to SafeAreaProviderCompat I discover that style props is missing.

### DIFF
--- a/packages/elements/src/SafeAreaProviderCompat.tsx
+++ b/packages/elements/src/SafeAreaProviderCompat.tsx
@@ -1,5 +1,12 @@
 import * as React from 'react';
-import { Dimensions, Platform, StyleProp, View, ViewStyle } from 'react-native';
+import {
+  Dimensions,
+  Platform,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+} from 'react-native';
 import {
   initialWindowMetrics,
   SafeAreaInsetsContext,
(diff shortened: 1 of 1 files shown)


packages/elements/src/SafeAreaProviderCompat.tsx
```diff
@@ -23,19 +24,23 @@ const initialMetrics =
...
       {(insets) => {
         if (insets) {
           // If we already have insets, don't wrap the stack in another safe area provider
           // This avoids an issue with updates at the cost of potentially incorrect values
           // https://github.com/react-navigation/react-navigation/issues/174
-          return children;
+          return style == null ? (
+            children
+          ) : (
+            <View style={style}>{children}</View>
```
satya164, line 38: This `View` needs to have `flex: 1`, otherwise the behavior will be different based on whether there's a parent `SafeAreaProvider` or not.

packages/elements/src/SafeAreaProviderCompat.tsx
```diff
@@ -23,19 +24,23 @@ const initialMetrics =
...
+          );
```
satya164, line 39: Conditionally changing whether it's wrapped in a `View` based on a prop is problematic. Changing the `style` prop would cause any local state to be reset due to remount. This needs to be always wrapped in a `View`

satya164 (changes requested):
Thanks for the PR. Please fill the pull request template when opening PRs.

packages/elements/src/SafeAreaProviderCompat.tsx
aleppos: Done.

aleppos:
@satya164 can you take a look, thanks.

satya164 (approved):
Thanks

PR #9800, opened 2021-08-07
fix: fix useLinkTo crash when an object is passed as a parameter

LuckyLuke19:
When an object was passed to the `to` parameter in `useLinkTo`, an error message would be thrown (*Cannot find property 'root'*). Apparently, this was introduced in commit 1d40279, where the `root` variable was removed.


satya164 (approved):
Thank you for the PR

PR #9824, opened 2021-08-16
Fixing nullpointer exception

tdekoning:
We were encountering an error on this piece of code.

I don't have clear reproduction paths with an Expo snack, but if you look at the TypeScript types, you see that `nextDescriptor` can be undefined, causing the error on this line of code.


satya164:
Thanks

PR #9831, opened 2021-08-17
fix: allow each tab to control its inactive colors

### LINKED ISSUE #9830: [v6] Tabs cannot have different inactive colors
### Current behavior

All tabs have the same inactive colors (both tint and background) which is defined by the focused tab's options (`tabBarInactiveTintColor` and `tabBarInactiveBackgroundColor`). This was not true before 42c43ff. I'm not sure if this was ever intended or not, but we have been relying on this behavior since the original v6 prerelease. Our use case is that a particular tab may be disabled instead of just inactive and we want to visually distinguish between the two (not great for a11y, I know) regardless of which tab is currently focused. I think the behavior prior to 42c43ff seems more flexible since unfocused tabs can still base their inactive colors on the current navigation state (i.e. on which tab is focused).

### Expected behavior

All tabs can control their own inactive tint and background colors.

### Reproduction

https://snack.expo.dev/WJG6tQ_Kd

### Platform

- [X] Android
- [X] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack

### Environment

| @react-navigation/native               | 6.0.2
| @react-navigation/bottom-tabs          | 6.0.5
| @react-navigation/drawer               | 6.1.1
| @react-navigation/material-bottom-tabs | 
| @react-navigation/material-top-tabs    | 
| @react-navigation/stack                | 6.0.7
| @react-navigation/native-stack         | 
| react-native-tab-view                  | 3.0.1
| react-native-pager-view                | 
| react-native                           | 0.63.4
| expo                                   | 42.0.3
| node                                   | 14.17.0
| npm or yarn                            | yarn

sumnerwarren:
Resolves #9830. It may make sense to clarify these options a little more in the documentation?

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -146,9 +146,7 @@ export default function BottomTabBar({
     tabBarStyle,
     tabBarBackground,
     tabBarActiveTintColor,
-    tabBarInactiveTintColor,
     tabBarActiveBackgroundColor,
-    tabBarInactiveBackgroundColor,
   } = focusedOptions;
 
   const dimensions = useSafeAreaFrame();
@@ -347,9 +345,11 @@ export default function BottomTabBar({
                   testID={options.tabBarTestID}
                   allowFontScaling={options.tabBarAllowFontScaling}
                   activeTintColor={tabBarActiveTintColor}
-                  inactiveTintColor={tabBarInactiveTintColor}
+                  inactiveTintColor={options.tabBarInactiveTintColor}
                   activeBackgroundColor={tabBarActiveBackgroundColor}
-                  inactiveBackgroundColor={tabBarInactiveBackgroundColor}
+                  inactiveBackgroundColor={
+                    options.tabBarInactiveBackgroundColor
+                  }
                   button={options.tabBarButton}
                   icon={
                     options.tabBarIcon ??


sumnerwarren:
If maintainers are comfortable with this change, I can work on updating the bottom tab example (maybe specifically to show how to maintain the focused tab controls inactive tab colors behavior).

satya164:
Thanks for the PR, but having `activeTintColor` and `inactiveTintColor` to the whole tab bar as it allows you do things like changing the tab bar to dark for one specific screen etc. This was already the case in material top tabs, so it's better not to have different behavior here. For example See this example: https://github.com/react-navigation/react-navigation/blob/bc19ee761632f8bb5246e839bd2a438f59da6062/example/src/Screens/BottomTabs.tsx#L118

As per your use case, you can achieve it this way:
- the active and inactive color of icons for each tab can be customized in `tabBarIcon` by ignoring the `tintColor` pass by React Navigation and using your own
- the label's tint color can be customized for the screen itself by `labelStyle` option

Example:
```js
options={({ navigation, route }) => {
  const state = navigation.getState();
  const isFocused = state.routes[state.index].key === route.key;

  return {
    tabBarIcon: ({ focused, size }) => <MaterialCommunityIcons name="image-album" color={focused ? YourActiveColor : YourInactiveColor} size={size} />,
    labelStyle: { color: isFocused ? YourActiveColor : YourInactiveColor }
  };
 })
 ```

Lemme know if it works for you.

sumnerwarren:
Thanks for the quick reply. That will definitely work for our use case. I think the documentation is accurate for the current usage; we shouldn't have been relying on this undocumented behavior.

PR #9832, opened 2021-08-18
fix: content container in native web stack should fill parent

brentvatne:
`flex: 1` needs to be set on the `View` container sits between the `Screen` component and screen contents in `NativeStackView` otherwise content on tabs nested inside native stack will not be visible, along with other similar types of bugs.

Reproducible example: https://gist.github.com/brentvatne/273d34be0c937237eb353e4100c49949

Apply the fix from this diff to see it resolves the issue.

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.tsx
+++ b/packages/native-stack/src/views/NativeStackView.tsx
@@ -138,7 +138,9 @@ export default function NativeStackView({ state, descriptors }: Props) {
                 { display: isFocused ? 'flex' : 'none' },
               ]}
             >
-              <View style={contentStyle}>{render()}</View>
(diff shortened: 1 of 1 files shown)


packages/native-stack/src/views/NativeStackView.tsx
```diff
@@ -151,6 +153,9 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
   },
+  contentContainer: {
```
brentvatne, line 156: i created a separate style for this to express the intent more clearly but i don't see an issue with just reusing container, alternatively

satya164 (approved):
Thanks!

PR #9860, opened 2021-08-23
fix: do not transform views when the drawer type is permanent

netshade:
This PR removes `transform` style from container and drawer views when the drawer type is set to be `permanent`. It does so to prevent cases where the `transform` CSS rule being present in a `web` context may cause other elements to no longer work correctly - see [this Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=20574) regarding `position:fixed`. 

I tried to stick to what I think to be the *intent* here, in that if the drawer type is set to be permanent, then no animation of the scene container or the drawer is ever intended in the future, and therefore, not having any `transform` present is desirable.

### DIFF
--- a/packages/drawer/src/views/modern/Drawer.tsx
+++ b/packages/drawer/src/views/modern/Drawer.tsx
@@ -260,24 +260,27 @@ export default function Drawer({
   });
 
   const drawerAnimatedStyle = useAnimatedStyle(() => {
+    if (drawerType === 'permanent') {
+      return {};
+    }
     return {
       transform: [
         {
-          translateX:
-            drawerType === 'permanent' || drawerType === 'back'
-              ? 0
-              : translateX.value,
+          translateX: drawerType === 'back' ? 0 : translateX.value,
         },
       ],
     };
   });
 
   const contentAnimatedStyle = useAnimatedStyle(() => {
+    if (drawerType === 'permanent') {
+      return {};
+    }
     return {
       transform: [
         {
           translateX:
-            drawerType === 'permanent' || drawerType === 'front'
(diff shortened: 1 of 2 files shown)


netshade:
After some review, it also appears that an animated `opacity` value is also being applied to the drawer view in a case where no animation is in fact required. Changed here to in fact be the constant `1`, as no animation or value other than `1` should be possible when the drawer is set to permanent.

netshade:
Just checking on this - any chance I can get a review?

packages/drawer/src/views/modern/Drawer.tsx
```diff
@@ -260,24 +260,27 @@ export default function Drawer({
...
+          translateX: drawerType === 'back' ? 0 : translateX.value,
         },
       ],
     };
   });
 
   const contentAnimatedStyle = useAnimatedStyle(() => {
+    if (drawerType === 'permanent') {
+      return {};
+    }
```
satya164, line 278:
```suggestion
    }

```

packages/drawer/src/views/modern/Drawer.tsx
```diff
@@ -260,24 +260,27 @@ export default function Drawer({
   });
 
   const drawerAnimatedStyle = useAnimatedStyle(() => {
+    if (drawerType === 'permanent') {
+      return {};
+    }
```
satya164, line 265:
```suggestion
    }

```

satya164 (approved):
Thank you

satya164:
Merged in 2a88d0d

netshade:
Thanks!

Issue #9889, opened 2021-08-30
After resized window in other screen, when navigate to the old screen, it shows empty view


nikgraf:
### Current behavior

After resizing the window and then switch to another screen the view is empty.

Someone else posted the issue here: https://github.com/microsoft/react-native-macos/issues/817 and recorded the behaviour

![bug](https://user-images.githubusercontent.com/4313235/130565966-af6934b2-cd17-469f-bab0-25a4a3c24b9e.gif)

I could reproduce the exact same behaviour in my open source codebase: https://github.com/SerenityNotes/serenity-notes-clients/blob/main/app/src/components/Navigation.tsx#L249-L322

So far I verified it happens with the Drawer as well as BottomTabs in react-navigation 5 and 6.

### Expected behavior

The screen should not be empty.

### Reproduction

https://github.com/SerenityNotes/serenity-notes-clients/blob/main/app/src/components/Navigation.tsx#L249-L322

### Platform

- [ ] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [X] MacOS

### Packages

- [X] @react-navigation/bottom-tabs
- [X] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

```
    "@react-navigation/bottom-tabs": "^6.0.5",
    "@react-navigation/drawer": "^6.1.4",
    "@react-navigation/native": "^6.0.2",
    "@react-navigation/stack": "^6.0.7",
    "react-native": "0.63.4",
    "react-native-macos": "^0.63.37",
    "react-native-reanimated": "~2.2.0",
    "react-native-safe-area-context": "^3.2.0",
```

satya164:
I don't have any knowledge about react-native-macos, so I'd appreciate a PR to fix it if someone manages to debug it. This might be related to React Native Screens.

nikgraf:
@satya164 managed to find the bug and opened a PR with a solution

PR #9904, opened 2021-09-03
fix: resized view never re-appears on macos #9889

nikgraf:
I created the original issue, found the root cause and was able to fix it.

Original issue: https://github.com/react-navigation/react-navigation/issues/9889


satya164 (approved):
Thank you

Issue #8981, opened 2020-10-23
Roadmap for v6


satya164:
This is a list of potential changes we want to have for v6.

WIP branch: https://github.com/react-navigation/react-navigation/tree/6.x

**General**

- [x] Overwrite params instead of merging for all actions except `setParams`. This has been a source of confusion
- [x] API to provide default screen options (for custom navigators)
- [x] Warn on duplicate screen names across navigators
- [x] Don't expose `state` property on `route` prop (probably expose with a non-enumerable symbol property for internal APIs)
- [x] Drop legacy linking config, validate linking config against navigation state against state changes in development mode and warn about misconfiguration
- [ ] ~Perform validation of the linking config to identify errors such conflicting screens for patterns on initial render rather than on each invocation of `getStateFromPath`~
- [ ] ~Accept a function instead of a component in `createNavigatorFactory` (for custom navigators). This would let us automatically wrap them in `React.forwardRef` and allow exposing the `ref` on individual navigators as well~
- [x] Drop support for older versions of `react-native-safe-area-context` and `reacct-native-screens` to cleanup the code
- [x] Use `@react-native-masked-view/masked-view` instead of `@react-native-community/masked-view` (or perhaps drop it altogether? need to check if metro supports optional require now)
- [x] Remove the compat package from repo
- [x] Stop exporting `useLinking` hook
- [x] Drop support for older versions of TypeScript and add names inside tuples (e.g. for `navigate`)
- [ ] ~Warn when a navigation method was called inside render~
- [x] Drop usage of dimensions and use window metrics from safe area context
- [x] Migrate components to `Pressable`
- [ ] ~Use system colors in default theme~

**Stack**

- [x] Drop `headerMode="none"`. There's already `headerShown` option which is more granular, and it's better not to have 2 ways of doing the same thing
- [x] Add deprecation warning for `headerMode="none"` and convert it to `headerShown: false` under the hood
- [x] Use the modal presentation animation on iOS when `mode` is set to `modal`, and maybe a slide up animation (similar to twitter app) for Android
- [x] Streamline header's props, pass `route`, `descriptor` and `navigation` as direct props, make `scene` contain the animation. The `previous` prop can contain `scene`, and `descriptor` properties
- [ ] Investigate: support pre-loading screens, this means that if `routes` array contains more routes then what's in `index`, they wouldn't trimmed automatically, and those extra routes would be considered as preloaded by rendering them offscreen.

**Drawer**

[...]

PR #9910, opened 2021-09-07
feat: add useAnimatedTabbarVisible hook in element for  animated hide/show bottomtab

numandev1:
added `useAnimatedTabbarVisible` hook for supporting v5 `tabBarVisible` functionality in V6 #8981

animatedtabbarHook


losh11:
I'm not sure if I'm doing something wrong, but I have a modified/custom bottom tab bar, and using a button to show/hide as you did doesn't seem to do anything. Is this expected?

satya164 (commented):
Thanks for the PR, however, I'd suggest you publish this hook separately from your account.

I'd prefer to avoid adding this to the core to reduce the maintenance cost. There are also limitations with `setOptions` (can't use it in multiple screens etc.) which is ok for an app, but would be confusing to have such a hook in the core.

osliver:
this implementation works much better then recommendation from core team, but problem when you have some specific styles, like borderTopRightRadius, borderTopLeftRadius, backgroundColor, etc.

Issue #9870, opened 2021-08-25
transitionStart / transitionEnd events not firing


giantslogik:
### Current behavior

transitionEnd does not fire on pop/popToTop. 
transitionStart does not fire on replace/push.

### Expected behavior

transitionEnd / transitionStart  fire once for all navigation types pop / popToTop / replace / push.

### Reproduction

https://snack.expo.dev/vREs6SGpw

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [X] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.2
| @react-navigation/stack                | 6.0.7
| react-native-safe-area-context         | 3.3.0
| react-native-screens                   | 3.5.0
| react-native-gesture-handler           | 1.10.3
| react-native                           | 0.64.2
| npm or yarn                            | 7.20.0

giantslogik:
On the snack https://snack.expo.dev/vREs6SGpw. - Open the expo Log console to see the events firing.

This is duplicate of https://github.com/react-navigation/react-navigation/issues/8861 which was reported for navigation 5.x

giantslogik:
The root cause seems to be inconsistent use of target, sometimes its the previous screen, sometimes the new screen. It should be consistently the new screen. As per documentation it should be current screen. i.e. new screen.
   
"This event is fired when the transition animation ends for the current screen"

giantslogik:
PR with fix here: https://github.com/react-navigation/react-navigation/pull/9915

giantslogik:
Workaround that worked for my scenario. Don't add the listeners using `listeners` prop or `screenListeners`. They fire fine if you use `navigation.addListener` to listen to the event.

https://reactnavigation.org/docs/navigation-events/#navigationaddlistener

PR #9915, opened 2021-09-09
fix: transitionStart and transitionEnd events for stack navigator

giantslogik:
This fixes the onTransitionStart / onTransitionEnd events to fire correctly/consistently for all navigation actions and helper APIs. **Event are fired for the focussed screen**.

_Note: It isn't possible to fire the onTransitionEnd event for the blurred screen, because its callbacks are removed from useEventEmitter::emit before the transition completes. This PR does not emit onTransitionStart / onTransitionEnd events for the blurred screen.  The intended behaviour is unclear as the documentation uses the term "current screen", as well as says "e.data.closing - Boolean indicating whether the screen is being opened or closed."_ 

Issue: https://github.com/react-navigation/react-navigation/issues/9870


giantslogik:
See https://github.com/react-navigation/react-navigation/issues/9870

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -152,8 +165,19 @@ function CardContainer({
     } else {
       onPageChangeCancel?.();
     }
-
-    onTransitionStart?.({ route }, closing);
+    /*fix for https://github.com/react-navigation/react-navigation/issues/9870
+       We trigger transition events only on the incoming (current) screen
```
satya164, line 169: "current screen" doesn't mean incoming screen, it means the screen where you add the listener - that is animating/transitioning.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -152,8 +165,19 @@ function CardContainer({
...
+       */
+    if (previousRoute && closing) {
+      //route is closing send to the previous route for navigation.goBack() / navigation.pop / navigation.popToTop()
```
satya164, line 172: The `closing` property is not relevant to the previous route since the previous route isn't being closed, so having a listener fire in a screen that's not transitioning or closing would be confusing.

packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -152,8 +165,19 @@ function CardContainer({
...
+      onTransitionStart?.({ route: previousRoute }, !closing);
+    } else {
+      //send to route for navigation.navigate  / navigation.push  / navigation.replace
+      // Run the operation in the next frame so that the target (new route) is available in useEventEmitter::emit.
+      requestAnimationFrame(() => {
+        onTransitionStart?.({ route }, closing);
+      });
```
satya164, line 179: This is likely due to `componentDidMount` in `Card.tsx` firing before `useEffect` fires as it's async. So proper fix would be to refactor `Card.tsx` to function component so that we start animation in an `useEffect` to match it.

satya164 (changes requested):
Thanks for the PR. Seems there is some confusion regarding the wording in the docs. The events should fire for the screen it's relevant to, e.g. if screen A is transitioning, the event should fire for screen A.

This is how all events work in React Navigation, so it'd be inconsistent to change the behavior of these events. In addition, it'd also be a breaking change.

It would be great if you can send a PR to the docs to clarify it.

packages/stack/src/views/Stack/CardContainer.tsx
giantslogik: Thanks for clarification. I will update the docs.

giantslogik: Since both the current / previous screen (can) animate in a transition (cardStyleInterpolator supports this) . Was it intentended to send events to both screen assuming they had listeners?

packages/stack/src/views/Stack/CardContainer.tsx
giantslogik: There are multiple possible call stacks based on the kind of navigation action.

`Card.tsx`  `componentDidMount / componentDidUpdate / handleGestureStateChange`
`Card.tsx` `animate`
`CardContainer.tsx` `handleTransition`

`CardContainer.tsx` `handleTransition` is called directly from `Card.tsx`  `componentDidMount`  (via `Card.tsx` `animate`). Which `useEffect` are you referring to? the one in `CardContainer.tsx` isn't involved in the flow.

Card -> CardContainer callback registration uses a prop.

```
    <Card --snipped--
      onTransition={handleTransition}
```

packages/stack/src/views/Stack/CardContainer.tsx
giantslogik: These events are linked to screens animating and not to just the "top" screen. The typical use case is to do something linked to animation starting or ending. this could be on the next / previous screen depending on if the navigation is a pop/push.

 The design of these events is broken as callbacks are removed for useEventEmitter::emit before transitionEnd fires for a closed screen. The changes in this PR do NOT fix the events for closed screens. They do bring the events on par with the corresponding react navigation 3.x events.

giantslogik (commented):
Updated the docs and added a PR on the docs repo.

This PR assumes the events were to be delivered to animating screens both previous / next as they may be on stack pop/push. It brings the events on par with navigation 3.x events. (I am migrating 3.x code to 6.x )

There is core design issue with closing flag and transitionEnd, since a closing screen will never be still registered for callback at transitionEnd. The PR does not fix events with the closing screen.

Use case: The PR allows the top screen after transition to perform operation(s) after the transition animation ends.

satya164:
> There is core design issue with closing flag and transitionEnd, since a closing screen will never be still registered for callback at transitionEnd. The PR does not fix events with the closing screen.

That's only an issue if you add the listeners using `listeners` prop or `screenListeners`. They fire fine with you use `navigation.addListener` to listen to the event.

> It brings the events on par with navigation 3.x events. (I am migrating 3.x code to 6.x )

The event system was pretty broken in 3.x. The event system is completely redesigned and the behavior of events is way different from 3.x as a result. The animation system is also completely redesigned. It's not a goal to match the old behavior.

> This PR assumes the events were to be delivered to animating screens, both previous / next as they may be on stack pop/push.

The way it's originally implemented assumes that the transitioning screen here is the screen that starts the animation using the `Animated` API and transitions into focus state or transitions out of focus state. The previous screen animating or not entirely depends on the `cardStyleInterpolator` used, e.g. it won't be the case for `forSlideVerticalIOS`.

This is more of a low-level API to build higher-level things on top of them. I believe the current API is enough for those use cases. It's not ideal for consumers to directly use these APIs in their apps. Please prefer `InteractionManager.runAfterInteractions` if you need some code to be deferred until transition ends.

While I think both assumptions make sense depending on which way you look at it, more importantly, changing the behavior of this event a breaking change as libs like react-navigation-shared-element rely on these APIs and changing which screens they are called for breaks those libraries. Hence I cannot merge this pull request.

If there is a specific use case this is preventing, please open an issue about that use case so that it can be addressed separately without changing how the API works.

giantslogik:
@satya164 thanks for the detailed analysis.

For context : My usage doesn't fit your description of 'consumer' . I'm maintaining a library somewhat similar to [react-native-router-flux](https://github.com/aksonov/react-native-router-flux) . It's primary purpose is to have a centrally directed navigation API, where the transitions , and navigation (animation/state) is dependent on user action rather than a hierarchy of screens and navigators.  It's been in use for three years and is built over React navigation 3.x - and now migrated to React Navigation 6.x.

Use Case: I need to on transitionEnd (animation completing) 

a) trigger `CommonActions` to modify React Navigation state.
b) trigger code on the Top Screen of transition. (irrespective of, how it was navigated to , push/pop/replace)

I evaluated the two suggestions you made:
> They fire fine with you use `navigation.addListener` to listen to the event.
 
Adding global listeners via   navigationRef.addListener('transitionEnd'.....) did not work.  I believe addListener works only in context of a screen. This is where it deviates from event handling in 3.x and breaks for me.

>InteractionManager.runAfterInteractions
 Non global, and not specific to React Navigation screen transitions. I don't think this would trigger the top screen in all cases - push/pop/replace.

Can this use case be handled in the existing React Navigation 6.x somehow? Should i open an another issue?

satya164:
If you want this level of flexibility over the API, I'd suggest you use the exported components such as `StackView`, `BottomTabView` etc. directly where you can pass your own implementations of `navigation`, `descriptors` etc. which will allow you to implement custom behavior. These views are exported to make it possible to build such abstractions which may want to deviate from how React Navigation manages the navigation state and give you complete freedom over the API.

Another option is for you to wrap each screen a HOC so that you can listen to events on the screen and then you can re-emit those events as you see fit.

giantslogik:
I have a HOC-like approach working. I'm trying to avoid making breaking changes in the library requiring App rewrites.

Do you think it's possible to fix the `transitionEnd` event delivery when using `screenListener` or `listener` , ie changing the lifetime/lifecycle of those listeners to match `navigation.addListener`?

scarlac:
I don't know about 6.x and other platforms but 5.x on iOS is inconsistent. There isn't a great workaround beyond `setTimeout()` so this PR would be great to get in if it works as intended.

`transitionEnd` will trigger for normal `.navigate()` and native back button behavior presses by the user but never with `.goBack()` programmatically which is where it seems inconsistent and where the event is definitely needed.

roryabraham:
> [transitionStart and transitionEnd] is more of a low-level API ... It's not ideal for consumers to directly use these APIs in their apps. Please prefer InteractionManager.runAfterInteractions if you need some code to be deferred until transition ends.

It's worth noting here that on web `InteractionManager.runAfterInteractions` uses `requestAnimationFrame`, and thus does will interrupt a transition in progress

PR #9924, opened 2021-09-10
feat: add specific type for merging on navigation

weslleydsg:
When navigating to existing screen we should be able to send partial params. That's what merging params expects on navigate options. Therefore this pull request add this type when merge is true.

### DIFF
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -205,7 +205,13 @@ type NavigationHelpersCommon<
           name: RouteName;
           key?: string;
           params: ParamList[RouteName];
-          merge?: boolean;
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. However `merge: true` doesn't really guarantee that the screen actually exists so this won't be correct. It just means that params will be merged if the screen exists. If the screen doesn't exist, passing partial params will cause a crash. This is more likely to happen especially on the Web where the user can refresh the browser on any screen losing the previous screens.

In many cases, you know that the screen will exist, in which case I think explicitly ignoring the error is reasonable.

bryanltobing:
@weslleydsg how do you cast it?

PR #9928, opened 2021-09-11
chore: add `this` typing to `NavigationHelpersCommon`

Krisztiaan:
## Motivation

Add missing `this` context information to `NavigationHelpersCommon`

Helps avoid issues when `@typescript-eslint/unbound-method` is used.

```ts
const { navigate } = useNavigation(); // This is only valid with this change.
```

### DIFF
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -178,6 +178,7 @@ type NavigationHelpersCommon<
    * @param action Action object or update function.
    */
   dispatch(
+    this: void,
     action: NavigationAction | ((state: State) => NavigationAction)
   ): void;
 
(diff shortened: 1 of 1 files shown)


Krisztiaan:
@satya164 could you, or someone else possibly take a look? I'm around to make any changes requested on this, it'd be really nice to be able to drop our `patch-packages` on this

satya164 (commented):
Thanks for the PR.

However, I don't want to merge this change as this adds extra overhead when maintaining as I have to make sure to use this pattern everywhere to satisfy a lint rule that we don't use. This isn't a pattern I follow and I'll 99% forget to follow this in the future.

There are also more places that'll give you an error when using that needs to be updated (every router has its own methods), this file is just a subset of them.

Krisztiaan:
Thanks, understandable. I'll try and find a better way to do `this`.

Issue #9897, opened 2021-09-01
Typing issues with nested screens, parameters and the Linking configuration


xrash:
### Current behavior

There's no example of how to type a screen that has its own parameters _and_ a nested navigator together, considering the Linking configuration must also be typed. This is how I'm doing:

```typescript
export type ParamList = {
  ScreenA: undefined
  ScreenB: NavigatorScreenParams<ParamListInnerNavigator> & {
    someparam?: number,
  }
}
```

However this only works in TypeScript 4.0. When I upgrade to TypeScript 4.1 and onwards, type checking the Linking configuration breaks. I suspect this could be related to a change in recursive conditional types that was introduced in TypeScript 4.1.

If this is not the correct way of typing a screen with parameters  + nested navigator, please provide examples on how to do it correctly. If that is the correct way, please fix the typing issue.

#### Reproducing it

In order to reproduce it easily, please follow the steps below:

1. Clone this repo: https://github.com/xrash/NestedNavigationTypingExample
2. Run `yarn` to install dependencies
3. Run `yarn tsc` and see how it works nicely
4. Run `yarn add typescript@4.1.6` in order to upgrade TS
5. Run `yarn tsc` again; now it breaks

#### Further investigation

Looks like the type condition below goes from returning `true` to returning `false` when upgrading from TS 4.0 to TS 4.1:

```typescript
// @react-navigation/core/src/types.tsx:640
NonNullable<
    ParamList[RouteName]
  > extends NavigatorScreenParams<infer T, any>
 ```
 
 You can test it like this:
 
 ```typescript
type Check<T extends {}> = NonNullable<T> extends NavigatorScreenParams<infer F, any> ? true : false

type Result = Check<
  NavigatorScreenParams<{
    screenA: undefined
    screenB: undefined
  }> & { myparam1: string }
>
```

`Result` is `true` on `typescript@4.0.8`, and `false` on `typescript@4.1.6` and up.

### Expected behavior

It is expected to have examples on how to type nested screens with parameters - fully type checked, considering the Linking object. It is expected that changing TS version from 4.0 to 4.1 doesn't break things, or at least this should be mentioned somewhere I guess.

### Reproduction

https://github.com/xrash/NestedNavigationTypingExample

### Platform

- [X] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
[...]

PR #9948, opened 2021-09-17
fix: typing issues with nested screens, parameters and the Linking #9897

react-one:
**Fix for issue** #9897 

This PR  update NavigatorScreenParams type to fix " extends NavigatorScreenParams<infer T, any> " condition always returning false in PathConfigMap recursive type check
Please check #9897  for the issue description.

### DIFF
--- a/packages/native-stack/CHANGELOG.md
+++ b/packages/native-stack/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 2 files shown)


packages/core/src/types.tsx
```diff
@@ -599,7 +599,7 @@ export type TypedNavigator<
 };
 
 export type NavigatorScreenParams<
-  ParamList,
+  ParamList extends ParamListBase | NavigationState,
```
satya164, line 602: I'm confused why `ParamList` would be of type `NavigationState`

satya164 (changes requested):

packages/core/src/types.tsx
react-one: @satya164 yeh, it shouldn't be, it could extends ParamListBase only, i will fix that, thanks

PR #9951, opened 2021-09-19
feat: enhance native-stack header left props

yusufyildirim:
## Motivation 
I needed to implement a custom back button for our project that uses `native stack` of the react-navigation. The thing is, the only way seems to implement it with `headerLeft` prop but the native stack does not provide some basic props to make a custom back button implementation possible. What those `basic` props would be you might ask;

* **canGoBack**: To determine if we should show a back button
* **isInModal**: To determine the style of the back button. It's common to show a "X" button on modals instead of the regular left arrowed back button.
* **headerBackTitle**: To support iOS styled back button with a title

I checked the regular stack version and [there are plenty of props passing to the `headerLeft`](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Header/HeaderSegment.tsx#L150) so we can implement a custom back button easily.

I added the most crucial props to be able to implement a custom back button. We should probably enhance this more but I feel like this is a good start.

Waiting for your valuable feedback!

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -85,6 +85,27 @@ export type NativeStackHeaderProps = {
   navigation: NativeStackNavigationProp<ParamListBase>;
 };
 
+export type HeaderLeftProps = {
+  /**
+   * Tint color for the header.
+   */
+  tintColor?: string;
+  /**
+   * Label text for the button. Usually the title of the previous screen.
+   * By default, this is only shown on iOS.
+   */
+  headerBackTitle?: string;
+  /**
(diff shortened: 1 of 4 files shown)


hamzawaleed0102 (approved):
LGTM! 👍

satya164 (commented):
Thank you. I'll be removing `isInModal` for now since 1. the back button would already have this info since the option specified by the user, 2. we would need to standardize this between JS stack and native-stack, this should also work for nested stacks etc.

satya164:
Merged in ef41211

PR #9956, opened 2021-09-21
fix(export): Export BottomTabHeaderProps

Titozzz:


satya164:
Thanks! Merged in 8475481

Issue #9936, opened 2021-09-15
Allow disabling BackHandler subscription in `DrawerView `


parasharrajat:
### Current behavior

The current back handling behavior in the DrawerView when the screen is focused is to close or open the drawer. This behavior cannot be overridden by a screen in a `Drawer.Navigator`. According to the docs,

> Custom back button handling in other situations may not be supported at the moment (eg. A known case when this does not work is when you want to handle back button press in an open drawer. PRs for such use cases are welcome!).

### Expected behavior

`DrawerView` can be made more extensible by disabling the default BackHandler subscription entirely.

e.g. one should be able to delegate BackHandler responsibility to a Screen and determine whether to open the drawer or exit the application

### Reproduction

https://snack.expo.dev/@parasaharrajat/react-navigation---stack-navigator-example

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [X] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [X] @react-navigation/stack
- [X] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^6.0.0-next.13
| @react-navigation/drawer               | ^6.0.0-next.17 | 
| @react-navigation/stack                | ^6.0.0-next.25
| react-native-gesture-handler           | 1.9.0
| react-native                           | v0.63
| node                                   | v14
| npm or yarn                            | npm

parasharrajat:
@github-actions Yes. 

This issue basically a clone of the issues https://github.com/react-navigation/react-navigation/issues/9511.

parasharrajat:
Currently, I use the Nested Drawer navigator inside the Stack navigator screen. But when the screen containing the Drawer navigator is focused, it does not allow to go back via the hardware back button. 

if if we try to override the backhander inside that screen the Subscription inside Drawer takes precedence.

If we can hide the following subscription behind a flag then Stack navigator will properly handle the hardware back button.

```
   const subscription = enableBackHandler
+      ? BackHandler.addEventListener('hardwareBackPress', handleClose)
+      : undefined;

```

parasharrajat:
Does anyone think this is a good to have feature?

satya164:
How is your repro related to this issue in any way?

[...]

PR #9957, opened 2021-09-22
feat: Allow disabling BackHandler subscription in DrawerView

parasharrajat:
Fixes https://github.com/react-navigation/react-navigation/issues/9936

It allows disabling the BackHandler inside the Drawer navigator.


satya164 (changes requested):
Thanks for the PR, but please follow the pull request template and explain the exact motivation of the change to avoid back-n-forth.

parasharrajat:
I will update the description and other information as needed.

parasharrajat:
@satya164 I have updated the info and I am not sure why the test are failing. Could you please help? Thanks.

PR #9960, opened 2021-09-23
fix: make useScrollToTop work with react-native-largelist

doomsower:
This fix makes `useScrollToTop` work with `react-native-largelist`. Otherwise you get error `Argument 1 (CGFloat) of SpringScrollView.scrollTo must not be null`. Should not break anything, because it's still compatible with original `scrollTo` signature.


satya164 (approved):
Thank you

PR #9970, opened 2021-09-26
fix: fix navigation when going back and forth in history on web

### LINKED ISSUE #9128: Web - Navigation is broken after going back and forth

satya164:
closes #9408
fixes #9128

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -98,12 +98,16 @@ const createMemoryHistory = () => {
 
       const id = window.history.state?.id ?? nanoid();
 
-      if (items.length) {
-        items[index] = { path, state, id };
+      if (!items.length || items.findIndex((item) => item.id === id) < 0) {
+        // There are two scenarios for creating an array with only one history record:
+        // - When loaded id not found in the items array, this function by default will replace
+        //   the first item. We need to keep only the new updated object, otherwise it will break
+        //   the page when navigating forward in history.
+        // - This is the first time any state modifications are done
+        //   So we need to push the entry as there's nothing to replace
+        items = [{ path, state, id }];
       } else {
-        // This is the first time any state modifications are done
-        // So we need to push the entry as there's nothing to replace
-        items.push({ path, state, id });
+        items[index] = { path, state, id };
       }
 
       window.history.replaceState({ id }, '', path);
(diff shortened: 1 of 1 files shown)


AlastairTaft:
Copying over my comment from [here](https://github.com/react-navigation/react-navigation/issues/9128#issuecomment-941908166)

I now hit this error

error

version

packages/native/src/useLinking.tsx
```diff
@@ -98,12 +98,16 @@ const createMemoryHistory = () => {
...
+        //   the first item. We need to keep only the new updated object, otherwise it will break
+        //   the page when navigating forward in history.
+        // - This is the first time any state modifications are done
+        //   So we need to push the entry as there's nothing to replace
+        items = [{ path, state, id }];
       } else {
-        // This is the first time any state modifications are done
-        // So we need to push the entry as there's nothing to replace
-        items.push({ path, state, id });
+        items[index] = { path, state, id };
```
roryabraham, line 110: Because we're seeing [this error](https://github.com/react-navigation/react-navigation/pull/9970#issuecomment-941908688), that tells us that somehow the `items` array is actually getting in a state where it contains `undefined`.

Now, there's only four places the `items` array is written to: [line 84](https://github.com/react-navigation/react-navigation/blob/fb84805c889bbb7059e7e95592c004aea2a510d6/packages/native/src/useLinking.tsx#L84), [line 86](https://github.com/react-navigation/react-navigation/blob/fb84805c889bbb7059e7e95592c004aea2a510d6/packages/native/src/useLinking.tsx#L86), [line 108](https://github.com/react-navigation/react-navigation/blob/fb84805c889bbb7059e7e95592c004aea2a510d6/packages/native/src/useLinking.tsx#L108), and [line 110](https://github.com/react-navigation/react-navigation/blob/fb84805c889bbb7059e7e95592c004aea2a510d6/packages/native/src/useLinking.tsx#L110).

Reviewing these 1 by 1:

#### Line 84
`items = items.slice(0, index - 1)` -> I don't see any way this can add an `undefined` value to the array ❌ 

#### Line 86
`items.push({ path, state, id })` -> At worst, this could append an object like: `{path: undefined, state: undefined, id: undefined}`, but we wouldn't end up with `undefined` itself in the `items` array. So not here either ❌ 

#### Line 108
`items = [{ path, state, id }];` -> Similar to line 86, this can't cause `undefined` to be in the `items` array ❌ 

#### Line 110
`items[index] = { path, state, id };` -> This _can_ cause `undefined` to be in the `items` array if `index` is greater than the length of the array, like so:

```js
> let items = [];
undefined
> items[1] = 'Hello';
'Hello'
> items[0];
undefined
```

roryabraham: How the `index` is somehow becoming greater than the length of the `item` array? I do not know

Issue #8384, opened 2020-06-04
Create a CLI package to make it easier to install React Navigation packages


satya164:
Currently we need to install a lot of different packages, so installation ca get long. A CLI would allow us to streamline the experience.

Example usage:

```sh
# Install stack navigator
npx @react-navigation/cli add stack
```

This should install the navigator along with all of the required peer dependencies.

Aung-Myint-Thein:
Would also be nice if we can have a table of packages and their latest version numbers. I am trying to update from 5.0.7 but struggling to find out what is the latest version numbers. :D

saideepesh000:
@satya164, hey I would do a prototype kinda thing and send it to u.

saideepesh000:
@satya164, @brentvatne, @skevy  I kinda created what you mentioned, but just made to decide what they want to install, it isn't published to npm or uploaded to GitHub, if you wish I want to continue on this, I would do that. Thanks.

![Screen Recording (28-10-2020 18-19-52)](https://user-images.githubusercontent.com/43727167/97439785-a4eaeb00-194c-11eb-906f-f36eb983602a.gif)

satya164:
@saideepesh000 looks nice. I don't have much time right now, but can you open a PR so I can take a look when possible? You can create a new folder called `cli` under the `packages` folder.

For initial version I'd like to have what I mentioned so we can mention it as an one-liner in the documentation. The command should do the following:

- Based on the argument (e.g. `stack`), it would check for a package named `@react-navigation/stack` on npm, and fetch it's metadata, or display error if the package doesn't exist
- It can read the peer dependencies from the metadata and install all packages with one command
- It should use choose which command to use to install based on following:
	- Use `expo` if we find `expo` in the `dependencies` of `package.json`
	- Use `yarn` if the user has `yarn.lock` file
	- Use `npm` if the user has `package-json.lock`
    - If no lockfile is present, use `yarn` if it's installed, otheriwse `npm`
- For gesture handler, we need to add `import 'react-native-gesture-handler'`. Right now it's the only package which needs it, so we can hardcode it to check if it was installed, and print a message about adding it, or even add it the top of `index.js`/`index.ts` automatically. 

Selecting packages to install from a liist and updating what's installed can be secondary functionality. Let's discuss them after the basic functionality. Also, we'd likely want to not display packages such as `core` or `routers` since normally users don't need to install them and they can confuse people.

cc @brentvatne if you have some ideas, please comment

[...]

PR #9974, opened 2021-09-27
Create a CLI package to make it easier to install React Navigation packages (Fix issue: #8384)

react-one:
# Create a CLI package to make it easier to install React Navigation packages ( Fix Issue: #8384 )

Mainly for now, an `add` command that allows us to install a package, with auto-detection of what package manager is in use. And automatic installation of peer dependencies. And also auto adding of `import 'react-native-gesture-handler'`

- [x] Installing package by name by fetching metadata. And can precise version (ex: stack@6.0.0)
- [x] Display error if the package doesn’t exist

- [x] Pick up the right installer following the logic
- Pick user picked installer if --installer option is provided
- A check and pick expo (expo in package.json)
- If not check yarn.lock (pick yarn)
- If not check package.lock (pick npm)
- If not check if yarn is installed (pick yarn)
- If not pick npm

- [x] Installing peer dependencies read from the metadata.

- [x] Auto-add of import 'react-native-gesture-handler'
- - [x] Check import 'react-native-gesture-handler'. And checking for both index.js|ts and App.js|ts.
- - [x] Add it automatically if not already added with a confirmation prompt.
- - [x] Auto picks up import quote style from code. If none found default to a single quote.

<br>

Done too:

- [x] Using loader in cli (ora package) and confirmation prompt (enquirer package)

- [x] Implemented logging with levels (with --verbose and --debug cli options too)

- [x] Implemented tests for installer pick up. And for react-native-gesture-handle. Used memfs ( in-memory file system ) and fs mocking for tests.

- [x] initial doc ( doc file )


mocktarissa (approved):

packages/cli/src/utils/logger.ts
```diff
@@ -0,0 +1,79 @@
+export class Logger {
+  private _isVerbose = false;
+  private _isDebug = false;
+  private _isLogEnabled = true;
+
+  public setIsVerbose(enable: boolean): this {
```
satya164, line 6: Don't use `_` for private properties (it's used in JS as a convention as there's no private properties there), or `public` to match code conventions in the repo.

packages/cli/package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "devDependencies": {
+    "@types/jest": "^27.0.2",
+    "@types/node": "^16.9.4",
+    "@types/shelljs": "^0.8.9",
+    "jest": "^27.2.1",
+    "memfs": "^3.3.0",
+    "ts-jest": "^27.0.5",
+    "ts-node": "^9.0.0",
+    "typescript": "^4.4.3"
```
satya164, line 29: We shouldn't duplicate deps such as `typescript` or `jest` here as they are in the root project already. Also `ts-jest` doesn't seem necessary here.

satya164 (changes requested):
I left some comments, but it's a sizable PR so I haven't completed the review yet.

satya164:
thanks @react-one. can you post about the ux flow of the cli and features on our discord so others can provide feedback?

react-one:
- When executing commands using `shell.exec()`. The output wasn't colored. I fixed that by creating a custom exec function based on spawn. I used cross-spawn for cross-platform compatibility.
- And I profited and added colors for our logs.

PR #9982, opened 2021-09-29
fix: typings: when merge is true in navigate, allow partial params

christophehurpeau:
Allows to pass only part of parameters when navigate is used with merge: true

### DIFF
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -205,7 +205,13 @@ type NavigationHelpersCommon<
           name: RouteName;
           key?: string;
           params: ParamList[RouteName];
-          merge?: boolean;
+          merge?: false;
+        }
+      | {
+          name: RouteName;
+          key?: string;
+          params: Partial<ParamList[RouteName]>;
+          merge: true;
         }
   ): void;
 
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -210,7 +210,7 @@ type NavigationHelpersCommon<
       | {
           name: RouteName;
           key?: string;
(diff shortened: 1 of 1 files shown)


LRNZ09:
The whole `params` could also be omitted when `merge` is `true`, like:
```javascript
params?: Partial<ParamList[RouteName]>;
```

packages/core/src/types.tsx
```diff
@@ -205,7 +205,13 @@ type NavigationHelpersCommon<
           name: RouteName;
           key?: string;
           params: ParamList[RouteName];
-          merge?: boolean;
+          merge?: false;
+        }
+      | {
+          name: RouteName;
+          key?: string;
+          params: Partial<ParamList[RouteName]>;
```
LRNZ09, line 213: The whole `params` could also be omitted when `merge` is `true`, like:

```suggestion
          params?: Partial<ParamList[RouteName]>;
```

LRNZ09 (changes requested):

satya164:
Thank you for the PR. However, I can't merge this since this may lead to incorrect typings in certain situations. More details here: https://github.com/react-navigation/react-navigation/pull/9924#issuecomment-930018557

LRNZ09:
@satya164 I don't agree, if you say so the `merge` parameter is completely useless since you're going to replace the whole params of the screen anyway.

itsramiel:
Here comes patch-package again 😪

junkboy0315:
This is why we cant move to v6.

The remaining ways are patch packaging or ignoring it in over 100 places.😂😂

Issue #10000, opened 2021-09-30
Flipper Navigation plugin: Event table rows text not showing in dark theme


react-one:
### Current behavior

In Flipper react navigation plugin. We can clearly see that the events table rows text doesn't show in dark mode (color setting problem)

the problem was mentioned already in this issue : https://github.com/react-navigation/react-navigation/issues/9975

You can see and verify that in the illustrations below :

**Light Mode :** 

![soffice bin_utvjdFfmbW](https://user-images.githubusercontent.com/90264248/135495695-d54c7998-527f-43f5-8473-666d78472e8d.png)

**Dark Mode :** 

![soffice bin_oDVqOQelHp](https://user-images.githubusercontent.com/90264248/135495799-9f329ffa-11aa-4cf1-b15f-6aea2264cb40.png)

### Expected behavior

The interface of **the React navigation plugin** on flipper will be compatible with dark mode, and we can see the text in it.

### Reproduction

https://snack.expo.dev/lzLGkBvHq

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.4
| @react-navigation/native-stack         | 6.2.2
| @react-navigation/devtools  | 6.0.2
| react-native-safe-area-context         | 3.3.2
| react-native-screens                   | 3.7.2
| react                |  17.0.2
| react-native                           | 0.65.1
| node                                   |  14.17.6
| yarn                            |  1.22.11
| flipper   | 0.112.0 
|  react-navigation flipper plugin | 1.3.5

react-one:
Fixing Pull request: https://github.com/react-navigation/react-navigation/pull/10001

PR #10001, opened 2021-09-30
fix: flipper react navigation plugin ui rows text not showing in dark theme (Fix issue #10000)

react-one:
Fix for issue: https://github.com/react-navigation/react-navigation/issues/10000+

In [line 75](https://github.com/react-navigation/react-navigation/blob/main/packages/flipper-plugin-react-navigation/src/Logs.tsx#L75) 

We made the change to

```tsx
'color': props.selected && '#fff'
```

Because when the color property is not set (falsy value). flipper-plugin by default provide the right colors to the dark and light themes each. ( it get white in dark theme and black in light theme by default in such case ).<br>
So we set it to white only if selected (element with background highlight color). Otherwise leave it to the default.

In [line 128](https://github.com/react-navigation/react-navigation/blob/main/packages/flipper-plugin-react-navigation/src/Logs.tsx#L128) the color value is always black ('rgba(0, 0, 0, 0.85)'), so in Dark Mode the element "BlankslateText" would be unreadable, so leaving the default element color is the solution.

### DIFF
--- a/packages/flipper-plugin-react-navigation/src/Logs.tsx
+++ b/packages/flipper-plugin-react-navigation/src/Logs.tsx
@@ -72,14 +72,13 @@ const Row = styled.button<{ selected: boolean; faded: boolean }>((props) => ({
   'fontSize': theme.monospace.fontSize,
   'textAlign': 'left',
   'padding': `${theme.space.medium}px ${theme.space.large}px`,
-  'color': props.selected ? '#fff' : '#000',
+  'color': props.selected ? '#fff' : '',
   'backgroundColor': props.selected ? theme.primaryColor : 'transparent',
   'opacity': props.faded ? 0.5 : 1,
   'border': 0,
   'boxShadow': `inset 0 -1px 0 0 ${theme.dividerColor}`,
   'width': '100%',
   'cursor': 'pointer',
-
   '&:hover': {
     backgroundColor: props.selected
       ? theme.primaryColor
@@ -97,7 +96,6 @@ const JumpButton = styled.button({
   'cursor': 'pointer',
   'fontSize': theme.fontSize.small,
   'borderRadius': theme.borderRadius,
-
   '&:hover': {
     backgroundColor: 'rgba(0, 0, 0, 0.2)',
   },
@@ -125,7 +123,6 @@ const EmptyIcon = styled(CompassOutlined)({
 });
 
 const BlankslateText = styled.h5({
-  color: 'rgba(0, 0, 0, 0.85)',
   fontWeight: 600,
   fontSize: 16,
   lineHeight: 1.5,


packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -72,14 +72,13 @@ const Row = styled.button<{ selected: boolean; faded: boolean }>((props) => ({
   'fontSize': theme.monospace.fontSize,
   'textAlign': 'left',
   'padding': `${theme.space.medium}px ${theme.space.large}px`,
-  'color': props.selected ? '#fff' : '#000',
+  'color': props.selected ? '#fff' : '',
```
satya164, line 75: I think empty string is weird, it should specify proper color or a keyword (such as `inherit`) to make it clear what it's doing.

However, we need to migrate the plugin to the new architecture to have proper dark theme support if I understand correctly: https://fbflipper.com/docs/extending/sandy-migration/

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -72,14 +72,13 @@ const Row = styled.button<{ selected: boolean; faded: boolean }>((props) => ({
...
   'backgroundColor': props.selected ? theme.primaryColor : 'transparent',
   'opacity': props.faded ? 0.5 : 1,
   'border': 0,
   'boxShadow': `inset 0 -1px 0 0 ${theme.dividerColor}`,
   'width': '100%',
   'cursor': 'pointer',
-
```
satya164, line 82: Please revert this

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -97,7 +96,6 @@ const JumpButton = styled.button({
   'cursor': 'pointer',
   'fontSize': theme.fontSize.small,
   'borderRadius': theme.borderRadius,
-
```
satya164, line 100: Please revert this

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -125,7 +123,6 @@ const EmptyIcon = styled(CompassOutlined)({
 });
 
 const BlankslateText = styled.h5({
-  color: 'rgba(0, 0, 0, 0.85)',
```
satya164, line 128: This is to make the text appear lower opacity. If you're removing the color, it needs to add other styling like opacity to keep the same effect.

satya164 (changes requested):

react-one:
All notes were taken. 
Closing this PR because i included the fix on PR https://github.com/react-navigation/react-navigation/pull/10004. It's minimal and connected. Having a one PR make sense.

Issue #10003, opened 2021-10-01
Flipper Navigation plugin: Event table rows selection highlight not working


react-one:
### Current behavior

In the event table, selection highlight is not working. When selected, the selected item is not highlighted, only the first item in the list that remains selected.

You can see and verify that in the illustrations below :

![soffice bin_buDHKmmfKp](https://user-images.githubusercontent.com/90264248/135603677-64d53cec-92ef-4d3f-b39c-bc383dbe5126.png)

### Expected behavior

Selected item on the event table should be highlighted

### Reproduction

https://snack.expo.dev/lzLGkBvHq

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.4
| @react-navigation/native-stack         | 6.2.2
| @react-navigation/devtools             | 6.0.2
| react-native-safe-area-context         | 3.3.2
| react-native-screens                   | 3.7.2
| react                                  | 17.0.2
| react-native                           | 0.65.1
| node                                   | 14.17.6
| yarn                                   | 1.22.11
| flipper   					         | 0.112.0 
| react-navigation flipper plugin        | 1.3.5

react-one:
Fixing Pull request: https://github.com/react-navigation/react-navigation/pull/10004

react-one:
Fix issue : https://github.com/react-navigation/react-navigation/pull/10004

PR #10004, opened 2021-10-01
fix: flipper navigation plugin event table rows selection highlight not working and black theme text not showing (fix issue #10003, and issue #10000)

react-one:
I changed the PR to handle both of the two issues: https://github.com/react-navigation/react-navigation/issues/10000, https://github.com/react-navigation/react-navigation/issues/10003, and closed PR https://github.com/react-navigation/react-navigation/pull/10001

After review and redoing:

For migrating to Sandy, the architecture was already built on that.
I went and checked the whole package files and I had only to make a one not important change.

- Fixed the list selection problem, by following ant design guidelines. And where for the list, they suggested react-virtualized for virtual rendering. And following that, I picked up **react-window** which is a great lightweight alternative developed by the same author of react-virtualized. Lightweight in size (2ko vs 35ko+) and more optimized. The old List component that was used, is limited and is not fit for controlled rendering. Its usage doesn't support controlled selection and only a hacky way may be possible if even possible. Using ant design guidelines and using a virtual list through react-window make sense. And that’s following the guideline.

react-virtualized-auto-sizer package was used for auto-resize, as pointed by the doc of react-window. And it’s necessary.

- Add of a new log was done in the following way:
    - When the currently selected element is the last one. When a new log is added, automatically it would be selected and the scroll keeps moving to the bottom.
    - If not, then the log will be added last, but will not be selected and no scroll happen. The behavior is natural and desirable, (you can control what you want, like that and it's natural too).

- Assured the fading effect is going as was before, for the `reset this` button action.

- black theme text color fixed and by use of inherit.

Demo with black theme

https://youtu.be/FRvv81dKSaE

Demo with white theme

https://youtu.be/VWbW6Z7yRC8

AutoResize

https://youtu.be/bh7xl4_T_bs


packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -72,14 +73,12 @@ const Row = styled.button<{ selected: boolean; faded: boolean }>((props) => ({
   'fontSize': theme.monospace.fontSize,
   'textAlign': 'left',
   'padding': `${theme.space.medium}px ${theme.space.large}px`,
-  'color': props.selected ? '#fff' : '#000',
+  'color': props.selected ? '#fff' : '',
```
swrobel, line 76: Thanks for fixing this! It looks like this also includes the changes from #10001 which may cause approval issues for this PR. You probably want to make sure each PR is not dependent on the other.

swrobel (changes requested):

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -72,14 +73,12 @@ const Row = styled.button<{ selected: boolean; faded: boolean }>((props) => ({
...
   'backgroundColor': props.selected ? theme.primaryColor : 'transparent',
-  'opacity': props.faded ? 0.5 : 1,
```
satya164, line 77: Why is this removed?

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -21,10 +21,11 @@ export function Logs({ active, logs, index, resetTo }: Props) {
       <DataList
         style={{ height: '100%' }}
         items={logs}
-        onRenderItem={({ id, action }, _, i) => (
+        selection={selectedID || ''}
```
satya164, line 24: Empty string seems misleading. What's selected when passing an empty string? It should explicitly pass an ID (e.g. `selectedItem.id`)

satya164 (changes requested):

packages/flipper-plugin-react-navigation/src/Logs.tsx
react-one: I took time to get back and fix the issue well. i closed one PR, and had both changes go in this one. As the changes are minimal. I appreciate checking the point in the description. And i set some demo video. ( i was using an RDP, so the rendering is a one of an RDP )

react-one: I took time to get back and fix the issue well. i closed one PR, and had both changes go in this one. As the changes are minimal. I appreciate checking the point in the description. And i set some demo video. ( i was using an RDP, so the rendering is a one of an RDP )

AdamTyler:
This would be awesome to get this merged... Using the plugin in light mode burns my eyes! 😎

packages/flipper-plugin-react-navigation/src/Logs.tsx
react-one: I am not sure why its showing this line as removed. I pushed again a fix and that line was not changed

packages/flipper-plugin-react-navigation/src/Logs.tsx
```diff
@@ -1,46 +1,75 @@
...
 
 export function Logs({ active, logs, index, resetTo }: Props) {
   const [selectedID, setSelectedID] = React.useState<string | null>(null);
+  const listRef = React.useRef<List<Log[]>>();
 
   const selectedItem = selectedID
     ? logs.find((log) => log.id === selectedID)
     : logs[logs.length - 1];
 
+  const itemKey = (_index: number) => logs[_index].id;
```
satya164, line 23:
```suggestion
  const itemKey = (index: number) => logs[index].id;
```

satya164 (approved):
Thank you for the PR!

react-one:
@satya164 I rebased. And for `yarn.lock` didn't need any changes. Already the new dependencies were part of `main`.

henrymoulton:
@react-one / @satya164 can we get a release for this fix?

PR #10005, opened 2021-10-02
fix: fix type checking in linking config #9897

react-one:
This PR update `NavigatorScreenParams` type to fix type checking in linking config

```
export type PathConfigMap<ParamList extends {}> = {
  [RouteName in keyof ParamList]?: NonNullable<
    ParamList[RouteName]
  > extends NavigatorScreenParams<infer T, any>
    ? string | PathConfig<T>
    : string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>;
};
```

The check in `PathConfigMap` always return false, so whatever `ParamList[RouteName]` is, its type will be `string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>` 

Please check #9897 for the issue description.

### DIFF
--- a/packages/core/src/getActionFromState.tsx
+++ b/packages/core/src/getActionFromState.tsx
@@ -23,7 +23,7 @@ type NavigateAction<State extends NavigationState> = {
   type: 'NAVIGATE';
   payload: {
     name: string;
-    params?: NavigatorScreenParams<State>;
+    params?: NavigatorScreenParams<ParamListBase, State>;
     path?: string;
   };
 };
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
Thanks for the PR, but this doesn't seem like the correct fix. With the previous code, it would prevent unknown keys in the linking config. After this change, it doesn't seem to provide type-safety against that.

You can test this by adding some random screen name inside config in the provided repro:

```ts
        screens: {
          Test: 'test', // This should produce a type-error
          InnerNavigatorScreenA: 'inner-a',
          InnerNavigatorScreenB: 'inner-b',
        },
```

Issue #9975, opened 2021-09-27
Flipper DevTools error on PUSH, POP & GO_BACK actions


swrobel:
### Current behavior

The Flipper plugin now opens (thanks @zhigang1992!) and shows logs properly with NAVIGATE events, but as you can see in the screenshot & logs below, it displays an error in the righthand bar for PUSH, POP & GO_BACK actions. This error does not show in the React Native logs or Metro output. I verified this behavior on both the iOS & Android simulators.

<img width="1501" alt="Screen Shot 2021-09-26 at 10 15 16 PM" src="https://user-images.githubusercontent.com/157270/134848814-3e9c949b-a983-4418-ab48-0e1d795b58b4.png">

With that said, the action is still available below the error

<img width="791" alt="Screen Shot 2021-09-27 at 11 25 08 AM" src="https://user-images.githubusercontent.com/157270/134964772-4025eb5a-759d-4c9b-af18-674897d46dd7.png">

It's worth noting that in the screenshot immediately above this text, PUSH is the selected row, yet the top row is always highlighted in purple. It's non-obvious which row is highlighted. Furthermore, none of the rows below the top row are readable when the dark theme is enabled:

<img width="475" alt="Screen Shot 2021-09-27 at 11 25 46 AM" src="https://user-images.githubusercontent.com/157270/134964889-2341ab5b-00ea-4ca3-b456-5cd50aefa43c.png">

```
Error
at withStack (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:176976:43)
at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:176994:29)
at pushOrder (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:283999:22)
at _performTransitionSideEffects (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:65665:22)
at _receiveSignal (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:65607:45)
at onResponderRelease (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&:65503:34)
at apply (native)
[...]

PR #10006, opened 2021-10-03
fix: Flipper DevTools error on PUSH, POP & GO_BACK actions (fix: issue #9975)

react-one:
Fix: issue https://github.com/react-navigation/react-navigation/issues/9975

The problem as stated in the issue was hermes stack format being different and thus bad parsing.

To fix that we adapted the react native stack parsing files from:

https://github.com/facebook/react-native/blob/main/Libraries/Core/Devtools/parseErrorStack.js

https://github.com/facebook/react-native/blob/main/Libraries/Core/Devtools/parseHermesStack.js

Converted them to typescript and as pointed by this comment https://github.com/react-navigation/react-navigation/issues/9975#issuecomment-932215484

Getting the frames became:

```ts
const frames = parseErrorStack(stack)
   .slice(2)
   .filter((frame) => frame.file !== '[native code]');
```

Now all works well:

(In the illustration bellow hermes was in use)

image


react-one:
@satya164 Why the circleci integration tests keep failing randomly. with a server timeout error.

I experienced that before where pushing again had the check pass just ok.

image

satya164 (changes requested):
Thank you. The PR looks good. Can you change the file extensions to `.tsx`? As a convention, we use `.tsx` for all the TypeScript files in this repo.

satya164:
Not sure about what's going on with tests but I wouldn't worry about it for now since it's unrelated to the PR.

packages/devtools/src/parseErrorStack.ts
```diff
@@ -0,0 +1,52 @@
+import * as stacktraceParser from 'stacktrace-parser';
```
satya164, line 1: Since this code is vendored from React Native repository, please also include the copyright header.

packages/devtools/src/parseHermesStack.ts
```diff
@@ -0,0 +1,110 @@
+interface HermesStackLocationNative {
```
satya164, line 1: This also needs the copyright header from original file

satya164 (changes requested):

PR #10022, opened 2021-10-10
fix(drawer): hide overlay if drawer is not open

### LINKED ISSUE #9828: [v6] VoiceOver selects entire screen after short delay; not usable afterward
### Current behavior

Since upgrading to v6, we have noticed issues with VoiceOver on iOS. Upon enabled VoiceOver in our app, it will correctly select and read the header title for the current screen and then quickly select the entire screen (easier to see this in an app with dark mode) and then become unusable. The entire screen remains when you try to swipe navigate with VoiceOver. It is sometimes possible to navigate to another screen by double tapping, but this is inconsistent and does not seem to be related to where you double tap.

The same behavior can also be seen in the example app from this repository. In this case the left header button is selected when VoiceOver is enabled and then the whole screen is selected. Enabling the example dark theme first makes this easier to see.

### Expected behavior

VoiceOver will select and read the first element on the current screen and then allow normal navigation via swiping, tapping, double tapping, etc.

### Reproduction

https://github.com/react-navigation/react-navigation/tree/main/example

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/drawer
- [X] @react-navigation/stack

### Environment

These versions are taken from our app where we are experiencing this issue, but the versions installed from the example app in this repo (on main) also apply.

| @react-navigation/native               | 6.0.2
| @react-navigation/bottom-tabs          | 6.0.3
| @react-navigation/drawer               | 6.1.1
| @react-navigation/material-bottom-tabs | X
| @react-navigation/material-top-tabs    | X
| @react-navigation/stack                | 6.0.4
| @react-navigation/native-stack         | X
| react-native-tab-view                  | 3.0.1
| react-native-pager-view                | X
| react-native                           | 0.63.4
| expo                                   | 42.0.3
| node                                   | 14.17.0
| npm or yarn                            | yarn

grgr-dkrk:
fix #9828

From the fix at https://github.com/react-navigation/react-navigation/commit/3241190b19946c1cd0a744fb09a19d79ba683d74, the overlay containing `Pressable` is permanently displayed. This prevents VoiceOver from accessing the content under the overlay.
This PR prevents overlays from being generated while the drawer is not open.

https://user-images.githubusercontent.com/40130327/136688279-ae23c70e-946b-4651-801a-75eef7c73fec.mov


satya164:
are you sure this happens with the latest drawer? I can't reproduce the issue with accessibility inspector. the overlay is also positioned behind content using `zIndex` so it's weird that it covers the content

PR #10023, opened 2021-10-10
fix: replace deprecated Keyboard.removeListener() usage with subscription.remove() on v5

ahmaddehnavi:
replace deprecated Keyboard.removeListener() usage with subscription.remove()


jforaker:
LGTM!

jforaker:
A similar fix was [just merged](https://github.com/react-navigation/react-navigation/commit/51c49c50e706e93413bf52ea2694939aa9208972) into `4.x`. It would be great to get this in `5.x` as well! 

cc @satya164

osdnk:
Thanks a lot for this PR! 
This is already included in the newer version. We're not back-porting 5.x anymore. In this case, patch-package is the most we can offer,

PR #10031, opened 2021-10-11
fix: example CocoaPodsError invalid Podfile

leotm:
Causing pod install to abort.

Match [correct syntax](https://github.com/facebook/flipper/blob/main/react-native/ReactNativeFlipperExample/ios/Podfile).

Before

Screenshot 2021-10-11 at 14 17 49

After

Screenshot 2021-10-11 at 14 30 59


satya164 (approved):
Thank you!

PR #10054, opened 2021-10-15
feat(drawer): Accessibility; Background container acquires focus of VoiceOver on iOS

braandl:
When the User Taps into "non functional UI" on the Screen, the accessibility focus will be shifted to the `MaybeScreenContainer` view. This behavior is very uncomforable or the user, because the user will have to re-adjust the position of the VoiceOver focus.

I'd say this is a bug, because I can not see any reason why the user would want to focus on a container view that can not inherit any usable options?

Tested on iOS15 and Android 11.


braandl:
Hm, is there going to be any feedback on this? It is a quite minimal change with a huge effect on iOS apps.

satya164:
Thanks for the PR.

Have you investigated why `MaybeScreen` takes focus? It's just a regular `View` unless you're using `react-native-screens`. If the `ScreenContainer` from `react-native-screens` is taking focus for some reason, I think this needs to be fixed in that package since this can cause unintended behaviour in many other places, not just drawer.

braandl:
Ok, thank you for your feedback. I am currently not using react-native-screens but have been in the past. I will investigate.

Issue #10059, opened 2021-10-16
Drawer screens lacking onpress handler


qualitymanifest:
### Current behavior

Currently there is no way to pass an `onPress` handler to `Drawer.Screen`. There are many reasons why someone would want to do this, and I've found several people looking for this functionality on stackoverflow. For me personally, some of my `Drawer.Screen` components are `Stack.Navigator`s, and I want to be able to reset the stack when the drawer item is pressed.

I am aware of the ability to pass your drawer navigator a custom `drawerContent` where you can then pass it `DrawerItem`s with `onPress` handlers, and I am already doing that for some things like my logout button. Trying to use that for screens seems pretty cumbersome though, since it seems like then you have to implement all of the navigation and focused logic as well as drawer open/close state. Additionally, you can only place them before or after all of your regular `Drawer.Screen`s that are rendered via `DrawerItemList`

I solved this by adding an optional `onPress` that gets passed the normal `onPress` handler. That way the user can do what they want, and call the default `onPress` before or after (or not at all?)

My best guess as for why this functionality does not currently exist is that it can allow the user to do things that conflict with what react-navigation is doing internally when you tap on a drawer item, which could lead to bugs. If this is the case, maybe it could be called `dangerousOnPress`?

I suppose this is more of a feature request than a bug, but I have a PR incoming for this so it seemed to make more sense to post it here.

### Expected behavior

Drawer.Screens should have an `onPress` handler

### Reproduction

https://example.com

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [X] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |^6.0.4|
| @react-navigation/bottom-tabs          |
| @react-navigation/drawer               |^6.1.6|
| @react-navigation/material-bottom-tabs |
| @react-navigation/material-top-tabs    |
| @react-navigation/stack                |^6.0.9|
| @react-navigation/native-stack         |
| react-native-safe-area-context         |
| react-native-screens                   |
| react-native-gesture-handler           |
| react-native-reanimated                |
| react-native-tab-view                  |
| react-native-pager-view                |
[...]

PR #10060, opened 2021-10-16
feat(drawer): add drawer item onPress handler

qualitymanifest:
Closes https://github.com/react-navigation/react-navigation/issues/10059

Please see the issue for information on the problem and solution


satya164:

From a quick look at the diff, drawer currently supports a `drawerItemPress` event which addresses this use case, so this diff isn't needed anymore.

PR #10067, opened 2021-10-20
fix(core): removal of non-existing listener should not break updating navigation container ref

split:
Calling `removeListener` on event that doesn't have any listeners will add new key for that event to listeners with value `undefined`. After that attempts to set `ref.current` will fail with "Cannot read property 'forEach' of undefined" error. Fix is quite simple. Just needed to add check if there is listeners for the given event.

FYI @satya164. This is related to your recent change https://github.com/react-navigation/react-navigation/commit/acdde18d8938741fd27d8ff8c8249977e0cb03e7


satya164 (approved):
Thank you

PR #10110, opened 2021-11-07
feat(bottom-tabs): support tabBarShowIcon option in bottom-tabs navigator

haystome:


satya164:

From a quick look at the diff, this change will add additional maintenance overhead so currently, we don't want to merge this. If you strongly feel that this should be supported, please create a discussion with some examples for popular apps using this design.

Issue #10097, opened 2021-11-03
Set header height in headerStyle in native-stack?


gonzalo-rivas:
### Current behavior

I am migrating to react navigation 6 and I want to use the new standard native-stack library to upgrade the performance of these stacks. I have from react-navigation 5 this stack navigator (a snippet):

Screen1StackNavigator:
`  return (
    <Stack.Navigator
      screenOptions={{
        headerStyle: {
          height: 100,
          backgroundColor: 'black',
        },
      }}
    >
      <Stack.Screen
        name="Screen1"
        component={Screen1}
        options={{
          headerTitle: () => <Title title={'Screen1'} />,
        }}
      />
    </Stack.Navigator>
  )`

This snippet is in this tree of the main navigator:
MainTabsNavigator:

` <Tab.Navigator
      screenOptions={{
        lazy: true,
        headerShown: false,
      }}
      ...
    >
      <Tab.Screen
        name="Screen1"
        component={Screen1StackNavigator}
        options={{
          TabLabel: t('markets.title'),
        }}
      />...`

But with this native-stack, the header style cant set its height. I saw in the docs that the only props you can set are:

Style object for header. Supported properties:

backgroundColor

So I want to know if there is a hidden tricky way to set the height as I used to do in react navigation 5 with "default" stack.

### Expected behavior

Set the inner Stack Header Height to a specific height.

### Reproduction

-

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack

### Environment

"@react-navigation/bottom-tabs": "^6.0.9",
    "@react-navigation/drawer": "^6.1.8",
    "@react-navigation/elements": "^1.2.1",
    "@react-navigation/native": "^6.0.6",
    "@react-navigation/native-stack": "^6.2.5",
    "@react-navigation/stack": "^6.0.11",

tylercote:
+1 I am also seeing this issue going from v5 stack to v6 native-stack.  Setting height on the headerStyle is no longer applying.

matias-calvo-91:
I found a workaround for this issue.
If you use the `header` prop, you can set **height** specifying the whole component; like this:

```
<Stack.Screen
[...]

PR #10113, opened 2021-11-09
feat(native-stack): expose custom header height

Titozzz:


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -166,10 +166,8 @@ const SceneView = ({
 
   const isParentHeaderShown = React.useContext(HeaderShownContext);
   const parentHeaderHeight = React.useContext(HeaderHeightContext);
-  const headerHeight = getDefaultHeaderHeight(
-    useSafeAreaFrame(),
-    false,
-    insets.top
+  const [headerHeight, setHeaderHeight] = React.useState(
+    getDefaultHeaderHeight(useSafeAreaFrame(), false, insets.top)
```
satya164, line 170: I think we should do something like this:
```suggestion
  const defaultHeaderHeight = getDefaultHeaderHeight(
    useSafeAreaFrame(),
    false,
    insets.top
  );
    
  const [customHeaderHeight, setCustomHeaderHeight] = React.useState(defaultHeaderHeight)
  
  const headerHeight = header != null ? customHeaderHeight : defaultHeaderHeight;
```

This is so that we use correct height if custom header is removed later.

satya164 (changes requested):

satya164 (changes requested):
Thanks! Can you also add web implementation?

Titozzz:
> Thanks! Can you also add web implementation?

Not needed it's already done inside elements

tylercote:
@satya164 Is there anything else blocking this PR?  Several people are running into this issue as per this open issue https://github.com/react-navigation/react-navigation/issues/10097

satya164 (approved):
Thank you

PR #10138, opened 2021-11-22
fix(core): remove unsupported type annotation

Misaka-0x447f:
Remove unsupported type annotation in tsc version 3.9.10. In this version of tsc such annotation will produce a type error like this:
image


satya164:
> Remove unsupported type annotation in tsc version 3.9.10

Thanks for the PR. But as mentioned in getting started page, the minimum required TypeScript version is 4.1

https://reactnavigation.org/docs/getting-started/

PR #10165, opened 2021-11-29
feat(native-stack): add support for header background image

Johan-dutoit:
Add ability to use a background-image for native-stack.


satya164:
Thank you! Merged in https://github.com/react-navigation/react-navigation/commit/393773b688247456d09f397a794eff19424502f6

Issue #10064, opened 2021-10-19
TabBarStyle zIndex property doesn't have any effect


romancxx:
### Current behavior

I am trying to put a `LinearGradient` behind my bottom tab to make it fade over the screen. I recently upgraded to v6 and realized that this code is no longer working. To achieve that, I am using `zIndex` in the `tabBarStyle` of the `screenOptions` with a higher value than my linear. But whatever the value I pass, it doesn't change anything as it used to in v5. Now the linear is either on top of the bottom tab or under everything so not visible at all.

 
![Screenshot 2021-10-19 at 19 20 46](https://user-images.githubusercontent.com/21194324/137960421-577d6dcb-ef4c-4bdb-bef3-13fdafb69c94.png)

  

### Expected behavior

It used to look like that with v5 : 

![Screenshot 2021-10-19 at 19 17 42](https://user-images.githubusercontent.com/21194324/137960002-c07b2b31-bf68-4a19-9276-5d037e751ce6.png)

Here is a link reproducing the behavior I had in v5 : https://snack.expo.dev/@flowwwww/zindex-react-nav-v5

Maybe I'm trying to achieve this the wrong way but couldn't find any good example to put a view behind the bottom tab. 

### Reproduction

https://snack.expo.dev/@flowwwww/8ea5ba

### Platform

- [X] Android
- [X] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |  ^6.0.6
| @react-navigation/bottom-tabs          | ^6.0.9
| @react-navigation/stack                |^6.0.11
| @react-navigation/native-stack         | 6.2.5
| react-native-safe-area-context         |3.2.0
| react-native-screens                   |^3.8.0
| react-native                           | 0.64.1

Nazort:
Did you manage to find a fix @Tatsiana0607 ? 

The only solution I personally see would be to do a custom tab nav component instead of using the one provided by react-navigation, which is sad since it is just to fix a small detail ...

Tatsiana0607:
@Nazort No, unfortunately I didn't.
[...]

PR #10211, opened 2021-12-22
fix: bottom-tabs z-index problem (issue: #10064)

react-one:
## Issue

Fixing https://github.com/react-navigation/react-navigation/issues/10064 issue.

## Problem

The issue is due to a change in this PR [ #9793 ]  [line 42](https://github.com/aleppos/react-navigation/blob/12e4f2ded9e24c3cbff6dca116a54620cfebcd8a/packages/elements/src/SafeAreaProviderCompat.tsx#L42). Where the children were wrapped in a `View` Component. The `zIndex` behavior will not work since the `children` elements are not at the same level.

 Also Wrapping the children in a View is not necessary to achieve the wanted behavior of PR #9793,

Screenshot 2021-12-22 at 13 16 46

## Proposition

**I suggest** either

-  We revert the component to return children and not wrap them in a component as it is now in this PR.

- Or add a `flatChildren` prop and make the behavior conditional. Where there will be no View wrapping If `flatChildren` is `true`. That can give flexibility for the SafeAreaProviderCompat and make sure we do not break anything.

### DIFF
--- a/packages/elements/src/SafeAreaProviderCompat.tsx
+++ b/packages/elements/src/SafeAreaProviderCompat.tsx
@@ -39,7 +39,7 @@ export default function SafeAreaProviderCompat({ children, style }: Props) {
           // If we already have insets, don't wrap the stack in another safe area provider
           // This avoids an issue with updates at the cost of potentially incorrect values
           // https://github.com/react-navigation/react-navigation/issues/174
-          return <View style={[styles.container, style]}>{children}</View>;
+          return children;
         }
 
         return (


satya164 (changes requested):
Thanks for the PR, but this change isn't correct. This means `SafeAreaProviderCompact` will have wrapper `View` (from `SafeAreaProvider` in certain cases and no wrapper `View` in other cases. When it doesn't have the wrapper `View`, the `style` isn't going to be applied, which would produce different results for the styles in different scenarios

> Also Wrapping the children in a View is not necessary to achieve the wanted behavior of PR

The passed `style` prop isn't going to be applied to anything if just returning `children`, so the wrapper `View` is necessary.

PR #10218, opened 2021-12-25
feat(native-stack): Add Android header search bar

ChrisEdson:
In `react-native-screens` [3.10.0](https://github.com/software-mansion/react-native-screens/releases/tag/3.10.0), the header search bar was added to Android (PR [here](https://github.com/software-mansion/react-native-screens/pull/1166)).

The docs for `react-navigation` also need updating - but I'm unsure of the order of operations for this.

p.s. Merry Christmas 🎄

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -52,7 +52,7 @@
     "react": "~16.13.1",
     "react-native": "~0.63.4",
     "react-native-builder-bob": "^0.18.1",
-    "react-native-screens": "^3.3.0",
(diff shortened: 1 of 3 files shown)


rvasseur31 (approved):

satya164 (changes requested):
Thank you for the PR. Could you update the version number of react-native-screens in other packages as well so that we have a single version of the package in the repo?

satya164:
Merged in https://github.com/react-navigation/react-navigation/commit/e329c43334a01fa1282338cf519a3fc37dd453ad

PR #10224, opened 2021-12-28
Dismiss keyboard when pressing the drawer toggle button

florentroques:
Dismiss keyboard when pressing the drawer toggle button to avoid having to press 2 times  on drawer item to change page.


satya164:
Thanks for the PR, but this is a breaking change for people who want the keyboard to stay open. And This isn't the correct place to dismiss the code since it can't be overridden for those who want it.

You can call `Keyboard.dismiss` inside your code when drawer's open/close state changes: https://reactnavigation.org/docs/drawer-navigator/#checking-if-the-drawer-is-open

PR #10260, opened 2022-01-13
feat(native-stack): export NativeStackView to support custom routers on native-stack

AdityaPahilwani:
Currently, there is no way to create a custom navigator and router on top of **Native-Stack**

So by exporting **NativeStackView** along with its types from native-stack, we can use it along with **useNavigationBuilder** to create custom navigators​ with some additional functionality built on top of it.

We can also use **StackRouter** from **@react-navigation/native** to create customised routers and use it with the navigator we created on top of **NativeStackView**.

We can follow the same doc to extend NativeStackView with our own router or build additional functionality on top of it.
https://reactnavigation.org/docs/custom-navigators#extending-navigators.

### DIFF
--- a/packages/native-stack/src/index.tsx
+++ b/packages/native-stack/src/index.tsx
@@ -2,6 +2,7 @@
  * Navigators
  */
 export { default as createNativeStackNavigator } from './navigators/createNativeStackNavigator';
+export { default as NativeStackView } from './views/NativeStackView';
 
 /**
  * Types
@@ -11,4 +12,6 @@ export type {
   NativeStackNavigationOptions,
   NativeStackNavigationProp,
   NativeStackScreenProps,
+  NativeStackNavigationEventMap,
+  NativeStackNavigatorProps,
 } from './types';
--- a/packages/native-stack/src/index.tsx
+++ b/packages/native-stack/src/index.tsx
@@ -9,9 +9,9 @@ export { default as NativeStackView } from './views/NativeStackView';
  */
 export type {
   NativeStackHeaderProps,
+  NativeStackNavigationEventMap,
   NativeStackNavigationOptions,
   NativeStackNavigationProp,
-  NativeStackScreenProps,
-  NativeStackNavigationEventMap,
   NativeStackNavigatorProps,
+  NativeStackScreenProps,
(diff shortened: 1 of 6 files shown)


satya164 (changes requested):
Thanks for the PR. I think we don't need to export `NativeStackNavigatorProps` as you can already access it by `React.ComponentProps<typeof NativeStackView>`

Regarding exporting `NativeStackNavigationEventMap`, that's useful but currently, no other navigators export it (afaik). Could you add the export to other navigators as well? Then I'll be happy to merge the PR.

AdityaPahilwani:
Thanks for the review.
I have added exports for NavigationEventMap from other navigators.

packages/native-stack/src/index.tsx
```diff
@@ -2,13 +2,16 @@
...
 
 /**
  * Types
  */
 export type {
   NativeStackHeaderProps,
+  NativeStackNavigationEventMap,
   NativeStackNavigationOptions,
   NativeStackNavigationProp,
+  NativeStackNavigatorProps,
```
satya164, line 15:
```suggestion
```

packages/native-stack/src/index.tsx
```diff
@@ -2,12 +2,14 @@
  * Navigators
  */
 export { default as createNativeStackNavigator } from './navigators/createNativeStackNavigator';
+export { default as NativeStackView } from './views/NativeStackView';
```
satya164, line 5:
```suggestion

/**
 * Views
 */
 export { default as NativeStackView } from './views/NativeStackView';
```

satya164 (approved):
Thanks

packages/native-stack/src/index.tsx
```diff
@@ -3,11 +3,17 @@
  */
 export { default as createNativeStackNavigator } from './navigators/createNativeStackNavigator';
 
+/**
+ * Views
+ */
+ export { default as NativeStackView } from './views/NativeStackView';
+
```
satya164, line 10:
```suggestion
/**
 * Views
 */
export { default as NativeStackView } from './views/NativeStackView';

```

PR #10261, opened 2022-01-13
feat: bubble onTabPress and onLongTabPress

Gregoirevda:
Passing the onTabPress and onLongTabPress events as react-native-tab-view [does](https://github.com/satya164/react-native-tab-view/blob/main/src/TabBar.tsx#L427)


Gregoirevda:
Updated syntax from `fn?.()` to `if(fn) { fn () }`, but pipeline is not re-running

satya164:
Thanks for the PR, but this is intentional. Events are exposed using React Navigation's event system, similar to how other props are exposed using React Navigation's options. So we don't want to expose these props.

If you have a use case that absolutely requires this, then, please mention that when filling up the PR template when opening a PR

PR #10269, opened 2022-01-16
feat(material-top-tab): support inaccessible tabs

Gregoirevda:


example/src/Screens/MaterialTopTabs.tsx
```diff
@@ -41,6 +43,18 @@ export default function MaterialTopTabsScreen({
         component={Albums}
         options={{ title: 'Albums' }}
       />
+      <MaterialTopTabs.Screen
+        name="NoAccess"
+        component={Null}
```
Gregoirevda, line 48: component could be null, but required change in `useNavigationBuilder` in core

Gregoirevda:
Screenshot 2022-01-16 at 23 50 27

Adds support to add a non-accessible tab (eg: user has no access to)

satya164:
Please follow the PR template when sending PRs

PR #10288, opened 2022-01-23
fix: wrap native stack with NavigationContent

### LINKED ISSUE #10287: Navigates to screen on the wrong stack
### Current behavior

NativeStack has an issue that makes navigation get confused and navigate to the right screen but on the wrong stack.

You can check the reproduction URL to understand the case well but the important thing here is, we use NativeStacks inside TabNavigator and we have the *same screen* defined by the *same name* in the both `HomeStack` and `BrowseStack`.

As you can see on the recording, when we're on the HomePage if we click on the `Go To Collection Page` button, it navigates to the page without issue. Things get interesting when we switch to the `Browse` tab, go back to the `HomeTab` again, and click on the `Go To Collection Page` button once again. It doesn't navigate to the screen within the `HomeStack`, it does navigate within the `BrowseStack` instead.

https://user-images.githubusercontent.com/22980987/150687044-8fefb369-372c-4dcc-9e0b-3bb3275e3a5b.mov

#### Steps to Reproduce

- Open the app
- Navigate to `Browse` tab
- Navigate back to `Home` tab
- Click on `Go To Collection Page` button
- You won't see the `Collection Page`
- Navigate to `Browse` tab
- See `Collection Page` is opened there

### Expected behavior

Should navigate to the `CollectionPage` *within the stack* the navigation action happens.

### Reproduction

https://snack.expo.dev/@yusufyildirimmm/react-navigation-issue

### Platform

- [X] Android
- [X] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack

### Environment

| @react-navigation/native               | 6.0.0
| @react-navigation/bottom-tabs          | 6.0.0
| @react-navigation/native-stack         | 6.2.3
| @react-navigation/native-stack         | 3.3.2
| react-native-screens                   | 3.10.1
| react-native                           | 0.66.4
| expo                                   | 44

yusufyildirim:
This pull request fixes #10287.

## The Story
When we use `Link`, we actually use `useLinkProps` hook under the hood and it uses `navigation` object to`dispatch` events. That `navigation` object is a bit different than what we get when we use `useNavigation` hook.

`useNavigation` gets the navigation object [using `NavigationContext`](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/core/src/useNavigation.tsx#L16) while `useLinkProps` get its `navigation` object [using `NavigationHelperContext`](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/native/src/useLinkProps.tsx#L26). To my understanding, `NavigationHelperContext` wraps navigators (StackNavigator, TabNavigator etc.) while `NavigationContext` wraps pages. So both return valid navigation object but the target is different. Let's move on to the problem. 

## The Problem
Looks like every navigator wrapped with `NavigationContent` which is basically `NavigationHelpersContextProvider` as you can [see here](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/core/src/useNavigationBuilder.tsx#L606). Here's a couple of examples:
* [BottomTabNavigator](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx#L101)
* [DrawerNavigator](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/drawer/src/navigators/createDrawerNavigator.tsx#L99)
* [StackNavigator](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/stack/src/navigators/createStackNavigator.tsx#L110)

But `NativeStackNavigator` [is not wrapped with that](https://github.com/react-navigation/react-navigation/blob/b91c9b05ff96727f5fa6ef0bec51b5d7eac06600/packages/native-stack/src/navigators/createNativeStackNavigator.tsx#L67). This means, when we use `useLinkProps` we don't get our current native stack's navigation reference, we get the closest other navigator which is the `BottomTabNavigator` in our case. When an action is dispatched, it goes through `onAction` function provided by the `useOnAction` hook. It roughly tries to match the navigation action (i.e is that page defined in the navigator) if can't, starts trying the same thing with its parent navigator and goes like that.  Since it accesses to the `TabNavigator` instead of the current active `NativeStack` in our case, it starts that matching process from the wrong navigator and tries matching the route all the way up and ends up finding a match on the latest created stack which is BrowseStack.

Looks like every navigator wrapped with `NavigationContent` in [this PR](https://github.com/react-navigation/react-navigation/commit/1179d56c5008270753feef41acdc1dbd2191efcf) on May 10, 2021 but `NativeStackNavigator` landed into the repo on May 19, 2021 in [this PR](https://github.com/react-navigation/react-navigation/commit/a55d6a8d796fa2b9f46cfcaf71f05c6d87738072). This looks like a small mistake and it should be wrapped with `NavigationContent` too. This PR fixes that.

### DIFF
--- a/packages/native-stack/src/navigators/createNativeStackNavigator.tsx
+++ b/packages/native-stack/src/navigators/createNativeStackNavigator.tsx
@@ -25,18 +25,19 @@ function NativeStackNavigator({
   screenOptions,
   ...rest
 }: NativeStackNavigatorProps) {
-  const { state, descriptors, navigation } = useNavigationBuilder<
-    StackNavigationState<ParamListBase>,
-    StackRouterOptions,
-    StackActionHelpers<ParamListBase>,
-    NativeStackNavigationOptions,
-    NativeStackNavigationEventMap
-  >(StackRouter, {
-    initialRouteName,
-    children,
-    screenListeners,
-    screenOptions,
-  });
+  const { state, descriptors, navigation, NavigationContent } =
+    useNavigationBuilder<
+      StackNavigationState<ParamListBase>,
+      StackRouterOptions,
+      StackActionHelpers<ParamListBase>,
+      NativeStackNavigationOptions,
+      NativeStackNavigationEventMap
+    >(StackRouter, {
+      initialRouteName,
+      children,
+      screenListeners,
+      screenOptions,
+    });
 
   React.useEffect(
     () =>
@@ -64,12 +65,14 @@ function NativeStackNavigator({
   );
 
   return (
-    <NativeStackView
-      {...rest}
-      state={state}
-      navigation={navigation}
-      descriptors={descriptors}
-    />
+    <NavigationContent>
+      <NativeStackView
+        {...rest}
+        state={state}
+        navigation={navigation}
+        descriptors={descriptors}
+      />
+    </NavigationContent>
   );
 }
 


marktros (approved):
🚀

satya164 (approved):
Thank you for catching the issue and sending the PR

PR #10303, opened 2022-01-28
fix: back and forward does not work correctly, items might have empty indexes

murataka:
items array sometimes has empty indexes. This fixes the findixes functions to work correctly , when items array has an empty element.

Screen Shot 2022-01-28 at 22 48 18

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -50,7 +50,7 @@ const createMemoryHistory = () => {
       const id = window.history.state?.id;
 
       if (id) {
-        const index = items.findIndex((item) => item.id === id);
+        const index = items.findIndex((item) =>item && item.id === id);
 
         return index > -1 ? index : 0;
       }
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR< but the proper fix would be to fix the code to avoid empty indices rather than ignoring them - which would cause incorrect page to be shown

murataka:
> Thanks for the PR< but the proper fix would be to fix the code to avoid empty indices rather than ignoring them - which would cause incorrect page to be shown

Yes, definitely I agree with that , could you confirm this does not work by using the app ? 

When I tested it seems to work. 

It is not perfect , but if it seems to work. Better than getting crash logs for now .. Just let me know if it is problematic, and I shall have a look at it to find if we can handle the indexes in a better way.

murataka:
Sorry for annoying , but I really need to know exactly what I shall fix , If there's no problem , it will be just trying to fix a non existing problem , don't get me wrong , I already mentioned this seems to be a workaround , I would like to implement properly in an other post.  It would be a lot easier for you to find what's wrong , I've just used that library a few hours ago , for the first time..

satya164:
If the array has an empty item then it's a bug that needs to be fixed. Ignoring it just means the `index` tracking is going to return inconsistent values and it'd be even harder to find bugs.

I don't have the bandwidth to test it - since there isn't even a repro for the bug that this is adding workaround for. The first thing to do is to prepare a demo and document the exact steps to reproduce the issue.

Once there's a repro, I will check when I have free time.

murataka:
> 

Thanks for the reply, I am trying to fix it , I shall cancel this pull request and open an other pull request as soon as I finish that . I am new at expensify as an external contributor, but would like to see you there. See you soon !

Issue #10210, opened 2021-12-22
Drawer position is incorrect when changing drawerType from 'front' to 'permanent'


AlexMedyantsev:
### Current behavior

I am creating a tablet app with drawer. 

When tablet in Landscape mode i want drawerType to be 'permanent'. 
When tablet is in the Portrait mode i want drawerType to be 'front'.

to get Orientation i use custom hook that listens for orientation change and returns actual value
then in drawer settings i change drawerType according to the Orientation
`drawerType: Orientation === 'PORTRAIT' ? 'front' : 'permanent',`

If i open a drawer while tablet is in portrait mode and rotate tablet to landscape, then the drawer will be positioned incorrectly with huge left offset

https://user-images.githubusercontent.com/24460400/147084555-2bfb5792-df22-4e16-8c2a-fa3d95ef63c3.mp4

### Expected behavior

If i open a drawer while tablet is in portrait mode and rotate tablet to landscape, then the drawer will be positioned correctly without huge left offset

### Reproduction

https://github.com/AlexMedyantsev/DrawerOrientationBug

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [X] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |^6.0.6
| @react-navigation/drawer               |^6.1.8
| @react-navigation/stack                |^6.0.11
| react-native-safe-area-context         |^3.3.2
| react-native-screens                   |^3.10.1
| react-native-gesture-handler           |^2.1.0
| react-native-reanimated                |^2.3.1
| react-native                           |0.66.4
| node                                   |16.11.1
| npm or yarn                            |yarn 1.22.11

AlexMedyantsev:
At first i came up with this workaround

```
  const [landscapeDrawerType, setLandscapeDrawerType] = useState('permanent');

  // my custom hook that returns actual orientation value
  const Orientation = useOrientation()

  // hook for checking if it's first render
  const isFirstRender = useFirstRender();

  useEffect(() => {
    if (!isFirstRender ) {
      setLandscapeDrawerType('slide');
[...]

PR #10304, opened 2022-01-28
fix(drawer): fix drawer when changing drawertype (ex: front -> permanent)

shrief1234567899:
regarding to 
https://github.com/react-navigation/react-navigation/issues/10210

and after latest updates for 
https://github.com/react-navigation/react-navigation/commit/524657484e6b21ecc5754b63468dd00304a8e450

### DIFF
--- a/packages/drawer/src/views/legacy/Drawer.tsx
+++ b/packages/drawer/src/views/legacy/Drawer.tsx
@@ -558,7 +558,7 @@ export default class DrawerView extends React.Component<DrawerProps> {
                     drawerType === 'permanent'
                       ? // Reanimated needs the property to be present, but it results in Browser bug
                         // https://bugs.chromium.org/p/chromium/issues/detail?id=20574
-                        undefined
+                        [{ translateX: 0 }]
                       : [{ translateX: contentTranslateX }],
                 },
               ]}
@@ -619,7 +619,7 @@ export default class DrawerView extends React.Component<DrawerProps> {
                     drawerType === 'permanent'
                       ? // Reanimated needs the property to be present, but it results in Browser bug
                         // https://bugs.chromium.org/p/chromium/issues/detail?id=20574
-                        undefined
+                        [{ translateX: 0 }]
                       : [{ translateX: drawerTranslateX }],
                   opacity: this.drawerOpacity,
                 },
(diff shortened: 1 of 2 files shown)


satya164:
What exactly is this fixing?

sherif-cyber:
> What exactly is this fixing?

fixing white block in screen when
- open drawer as front drawerType
- switching from portrait to landscape mode
- change drawerType to be Permanent instead of front

satya164:
I already made a fix in https://github.com/react-navigation/react-navigation/commit/524657484e6b21ecc5754b63468dd00304a8e450, what does this do additionally?

sherif-cyber:
> I already made a fix in [5246574](https://github.com/react-navigation/react-navigation/commit/524657484e6b21ecc5754b63468dd00304a8e450), what does this do additionally?

i just changed **undefined transform** to be zero in translateX

because without this change, we still face an issue **(wrong position for drawer in middle of screen and white space beside it)** when changing drawerType within Runtime from **front** to be **permanent** as appeared in video.

satya164:
> because without this change, we still face an issue (wrong position for drawer in middle of screen and white space beside it) when changing drawerType within Runtime from front to be permanent as appeared in video.

Can you open a new bug report for the issue you encountered?

> i just changed undefined transform to be zero in translateX

I can see that in the commit, but the mentioned commit (and the code comments) already say why I used `undefined` and not a transform. So just changing it to `transform` is not a fix.

sherif-cyber:
> > because without this change, we still face an issue (wrong position for drawer in middle of screen and white space beside it) when changing drawerType within Runtime from front to be permanent as appeared in video.
> 
> Can you open a new bug report for the issue you encountered?
> 
> > i just changed undefined transform to be zero in translateX
> 
> I can see that in the commit, but the mentioned commit (and the code comments) already say why I used `undefined` and not a transform. So just changing it to `transform` is not a fix.

umm, i removed using translateX and now i am using empty array instead of it, and it's also working fine for me.

when we use undefined it cause a design bug **(same bug in mentioned video)**

sherif-cyber:
> > > because without this change, we still face an issue (wrong position for drawer in middle of screen and white space beside it) when changing drawerType within Runtime from front to be permanent as appeared in video.
> > 
> > 
> > Can you open a new bug report for the issue you encountered?
> > > i just changed undefined transform to be zero in translateX
> > 
> > 
> > I can see that in the commit, but the mentioned commit (and the code comments) already say why I used `undefined` and not a transform. So just changing it to `transform` is not a fix.
> 
> umm, i removed using translateX and now i am using empty array instead of it, and it's also working fine for me.
> 
> when we use undefined it cause a design bug **(same bug in mentioned video)**

can we merge it ? @satya164

adamhari:
I've been locked to 6.1.4 because the changes in 6.1.5 broke the drawer like this when switching the drawerType. This PR fixes it for me.

satya164 (approved):
Can confirm `[]` fixes the issue. thank you

PR #10306, opened 2022-01-29
fix: back and forward does not work correctly, items should not have empty indexes

murataka:
In findindexes function , there's no need to

`item?.`

as the indexes are now being managed properly , I left them just because further testing might be essential and if there's still a mistake in the commit , not to make it crash.

It is possibly safe not to 

`item?.`.

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -179,7 +179,7 @@ const createMemoryHistory = () => {
 
         const onPopState = () => {
           const id = window.history.state?.id;
-          const currentIndex = items.findIndex((item) => item && item.id === id);
+          const currentIndex = items.findIndex((item) => item?.id === id);
 
           // Fix createMemoryHistory.index variable's value
           // as it may go out of sync when navigating in the browser.
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -99,7 +99,7 @@ const createMemoryHistory = () => {
 
       const id = window.history.state?.id ?? nanoid();
 
-      if (!items.length || items.findIndex((item) => item && item.id === id) < 0) {
+      if (!items.length || items.findIndex((item) => item?.id === id) < 0) {
         // There are two scenarios for creating an array with only one history record:
         // - When loaded id not found in the items array, this function by default will replace
         //   the first item. We need to keep only the new updated object, otherwise it will break
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -50,7 +50,7 @@ const createMemoryHistory = () => {
       const id = window.history.state?.id;
 
       if (id) {
-        const index = items.findIndex((item) => item.id === id);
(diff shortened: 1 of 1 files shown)


murataka:
Hello, 
let me know if you think this might create any other issue , and any kind of reviews would help. 

I would like to fix any possible issue created by this commit , as much as I can.

Thanks !

murataka:
These comments were already explaining the below change .

```
       // There are two scenarios for creating an array with only one history record:
        // - When loaded id not found in the items array, this function by default will replace
        //   the first item. We need to keep only the new updated object, otherwise it will break
        //   the page when navigating forward in history.
        // - This is the first time any state modifications are done
        //   So we need to push the entry as there's nothing to replace
```
`items[index] = { path, state, id };` 
to
`items[0] = { path, state, id };`

packages/native/src/useLinking.tsx
```diff
@@ -50,7 +50,7 @@ const createMemoryHistory = () => {
       const id = window.history.state?.id;
 
       if (id) {
-        const index = items.findIndex((item) => item.id === id);
+        const index = items.findIndex((item) => item && item.id === id);
```
roryabraham, line 53:
```suggestion
        const index = items.findIndex((item) => item?.id === id);
```

packages/native/src/useLinking.tsx
```diff
@@ -99,7 +99,7 @@ const createMemoryHistory = () => {
 
       const id = window.history.state?.id ?? nanoid();
 
-      if (!items.length || items.findIndex((item) => item.id === id) < 0) {
+      if (!items.length || items.findIndex((item) => item && item.id === id) < 0) {
```
roryabraham, line 102:
```suggestion
      if (!items.length || items.findIndex((item) => item?.id === id) < 0) {
```

packages/native/src/useLinking.tsx
```diff
@@ -179,7 +179,7 @@ const createMemoryHistory = () => {
 
         const onPopState = () => {
           const id = window.history.state?.id;
-          const currentIndex = items.findIndex((item) => item.id === id);
+          const currentIndex = items.findIndex((item) => item && item.id === id);
```
roryabraham, line 182:
```suggestion
          const currentIndex = items.findIndex((item) => item?.id === id);
```

roryabraham (approved):

satya164 (changes requested):
This change doesn't make sense to me. The change seems to be that it's always going to replace the first item when calling replace regardless of what the index is which doesn't look correct and going to produce incorrect history.

murataka:
> This change doesn't make sense to me. The change seems to be that it's always going to replace the first item when calling replace regardless of what the index is which doesn't look correct and going to produce incorrect history.

i shall check if the other solution if still creates empty indexes. feels still there's something missing..

murataka:
https://github.com/react-navigation/react-navigation/pull/10312

solves the issue , and I did not get any empty indexes while testing it . 

Closing this pull request.

PR #10318, opened 2022-02-01
fix: set header to translucent when using searchbar

satya164:


packages/native-stack/src/views/HeaderConfig.tsx
```diff
@@ -185,7 +188,11 @@ export default function HeaderConfig({
         translucent={
           headerBackground != null ||
           // This defaults to `true`, so we can't pass `undefined`
-          headerTransparent === true
+          headerTransparent === true ||
```
WoLewicki, line 191: I'd add a check `headerLargeTitle === true ||` since large header also requires this prop to work correctly with `ScrollView`.

PR #10334, opened 2022-02-08
feat: pass route index and total length BottomTabBar tabBarBackground…

mgcrea:
**Motivation**

Would be great to pass a few props to the `tabBarBackground` render prop so that we can render custom animated backgrounds. Example:

Screenshot 2022-02-08 at 18 50 41

Did add `index` (aka. `focusedRouteIndex`) and `length` (aka. `totalRoutesLength`) to address this.

Open to rename props or refactor if you'd rather have another API.


satya164:
Can't you already use `useNavigationState` hook for that?

mgcrea:
Unfortunately it does not work:

```ts
  const [index, length] = useNavigationState((state) => [state.index, state.routes.length]);
```

inside my `<TabBarBackground />` components always returns incorrect values (a static `[1, 2]`) while I have 5 routes and playing with `state.index`.

satya164:
> @mgcrea Unfortunately it does not work

So let's fix that instead

Issue #10328, opened 2022-02-04
Preventing going back doesn't work with the native stack navigator


sdandois:
### Current behavior

I've followed the [preventing going back guide](https://reactnavigation.org/docs/preventing-going-back/) and it was working great. However, after that I've changed the stack navigator with the native stack navigator and the example no longer works. 

The behavior is that the event is still being thrown, but the `e.preventDefault()` action is not working.

### Expected behavior

I would expect the `e.preventDefault()` to prevent going back in the native stack navigator also.

### Reproduction

https://snack.expo.dev/Sj3BtIgCP

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.8
| @react-navigation/native-stack         | 6.0.5
| react-native-safe-area-context         | 3.3.2
| react-native-screens                   | ~3.8.0
| react-native-gesture-handler           | ~1.10.2
| react-native-reanimated                | ~2.2.0
| react-native-tab-view                  | ^3.0.0
| react-native-pager-view                | 5.4.6
| expo                                   | 43.0.0

sdandois:
Related https://github.com/software-mansion/react-native-screens/pull/801

WoLewicki:
Yeah, seems like the props from this PR are not added to v6 `native-stack`. cc @kacperkapusciak could you add it there too? Or maybe there is something that prevents us from doing so?

PR #10354, opened 2022-02-16
feat: add nativeBackButtonDismissalEnabled prop to native-stack

kacperkapusciak:
**Motivation**

This PR adds `nativeBackButtonDismissalEnabled` prop to `native-stack` v6. This prop enables the developers to prevent going back with the header back button on Android.

The implementation was discussed in _great_ detail in the `react-native-screens` repo in https://github.com/software-mansion/react-native-screens/pull/801

Fixes https://github.com/react-navigation/react-navigation/issues/10328

**Test plan**

To test the change you can use [this Snack](https://snack.expo.dev/Sj3BtIgCP)

### Before

https://user-images.githubusercontent.com/39658211/154259015-42f0ea11-96f2-47e4-a5c5-46cea14ab148.mov

### After

https://user-images.githubusercontent.com/39658211/154259086-0c2f5aa9-6f94-49a5-9e8d-a81a3e548c1c.mov

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -379,6 +379,16 @@ export type NativeStackNavigationOptions = {
    */
   gestureEnabled?: boolean;
   /**
+   * Boolean indicating whether, when the Android default back button is clicked, the `pop` action should be performed on the native
+   * side or on the JS side to be able to prevent it. Unfortunately the same behavior is not available on iOS since the behavior of
+   * native back button cannot be changed there. Defaults to `false`.
+   *
+   * Only supported on Android.
+   *
+   * @platform android
+   */
+  nativeBackButtonDismissalEnabled?: boolean;
+  /**
    * The type of animation to use when this screen replaces another screen. Defaults to `pop`.
    *
    * Supported values:
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -379,9 +379,8 @@ export type NativeStackNavigationOptions = {
    */
   gestureEnabled?: boolean;
   /**
-   * Boolean indicating whether, when the Android default back button is clicked, the `pop` action should be performed on the native
-   * side or on the JS side to be able to prevent it. Unfortunately the same behavior is not available on iOS since the behavior of
-   * native back button cannot be changed there. Defaults to `false`.
+   * Whether the dismissal of the screen on tap of the header back button on Android should be performed by the native or the
+   * navigation's side. Defaults to `false`.
    *
    * Only supported on Android.
    *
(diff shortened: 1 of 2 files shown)


WoLewicki (approved):

packages/native-stack/src/types.tsx
```diff
@@ -378,6 +378,15 @@ export type NativeStackNavigationOptions = {
...
   gestureEnabled?: boolean;
+  /**
+   * Whether the dismissal of the screen on tap of the header back button on Android should be performed by the native or the
+   * navigation's side. Defaults to `false`.
+   *
+   * Only supported on Android.
+   *
+   * @platform android
+   */
+  nativeBackButtonDismissalEnabled?: boolean;
```
satya164, line 389: Is there any use case for exposing this to the user in React Navigation? Why not have it always set to `false`?

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -200,6 +202,12 @@ const SceneView = ({
       onAppear={onAppear}
       onDisappear={onDisappear}
       onDismissed={onDismissed}
+      onHeaderBackButtonClicked={() => {
+        navigation.dispatch({
+          ...StackActions.pop(),
+          source: route.key,
```
satya164, line 208: This should also specify the `target` option. Let's move it to `NativeStackViewInner` component like the other callbacks

satya164 (changes requested):
There's a warning about `beforeRemove` in `NativeStackViewInner`. We also need to update the warning message there to mention that custom `headerLeft` is only needed on iOS.

packages/native-stack/src/types.tsx
kacperkapusciak: When you use the native back button (`nativeBackButtonDismissalEnabled: true`) native changes can go through to the JS side because a view on the JS side is detached after the transition. It is necessary for eg. animations and transition context.

https://user-images.githubusercontent.com/39658211/154917853-ba437a1c-9176-44b4-be76-4d67a9a15152.mov


When using navigation's go back (`nativeBackButtonDismissalEnabled: false`) these changes can't be accessed because the view is detached immediately on the JS side.

https://user-images.githubusercontent.com/39658211/154917877-99a4fefa-953b-4d0e-bbfd-b3c29efd2d86.mov

packages/native-stack/src/views/NativeStackView.native.tsx
kacperkapusciak: Fixed in [8421534](https://github.com/react-navigation/react-navigation/pull/10354/commits/84215349631d8ef71fc9a5159b385ff32b170a47)

nandorojo:
Are there plans to support `beforeRemove` on iOS without disabling gestures? It would be nice if `beforeRemove` would snap the swiped screen back into place when calling `preventDefault()`, similar to the way the native iOS modal slightly bounces down when `gestureEnabled: false`.

WoLewicki:
@nandorojo I think the `iOS` part was discussed in this PR: https://github.com/software-mansion/react-native-screens/pull/801#issuecomment-913602845 and it was kind of abandoned since then. It would be good to come back to this one day. cc @satya164

PR #10365, opened 2022-02-18
refactor: avoid concat to get configs

OctaBruzzesi:
**Motivation**

In order to have more redeable code, I've refactored how it gets the configs in `getStateFromPaths` to let TS obtain its type and avoid casting it


satya164:
Thanks for the PR. However `flatMap` is not supported in iOS 11 which React Native supports, so it's not something we can use.

PR #10392, opened 2022-02-28
feat: add keepInactiveScreensVisible prop for bottom-tabs

IjzerenHein:
Hey there. This PR adds a minimal addition to the `bottom-tabs` package in order to make screen transitions on tab-switching possible. The `keepInactiveScreensVisible` prop keeps all screens visible, allowing screens to implement their own show/hide logic. The example below shows a facebook like translate/fade animation when switching tabs.

**iOS**
https://user-images.githubusercontent.com/6184593/156014039-71b5b990-8c71-48bb-81ca-7b2f0570ab6b.mov

**Android**
https://user-images.githubusercontent.com/6184593/156015620-9b3a2b1b-4cb4-4ecb-ae6c-8f9a9c37f9c3.mp4

# Example code
https://gist.github.com/IjzerenHein/172826535bb7bf782b2a8d6381da42d8

# Test plan

- Ran particular code in a production app (see #example_code)
- `yarn typescript`
- `yarn lint`
- `yarn test`


packages/bottom-tabs/src/views/BottomTabView.tsx
```diff
@@ -123,7 +124,7 @@ export default function BottomTabView(props: Props) {
             <MaybeScreen
               key={route.key}
               style={[StyleSheet.absoluteFill, { zIndex: isFocused ? 0 : -1 }]}
-              visible={isFocused}
+              visible={isFocused || !!keepInactiveScreensVisible}
```
satya164, line 127: This needs to be a separate prop on `MaybeScreen` (e.g. `enabled`). The `visible` prop also sets other props such as disabling `pointerEvents` which wouldn't be appropriate to keep enabled for unfocused screens.

packages/bottom-tabs/src/types.tsx
```diff
@@ -272,6 +272,12 @@ export type BottomTabNavigationConfig = {
    * Defaults to `true` on Android.
```
satya164, line 272: This line seems to be a leftover from an old impl, let's remove it

satya164 (changes requested):
Thanks for the PR. I'd prefer to avoid a new prop and make `detachInactiveScreens={false}` work like this. That way it'll also match the behavior of stack navigator. If someone wants to disable `react-native-screens` specifically, they can already call `enableScrens(false)`

We'd also need to make this change in drawer navigator

hassan-bilal144:
oijnionoinoinonoinoinonononoinon

Issue #10398, opened 2022-03-03
Using initialRouteName in web make screen redirects to pages outside app


listiani13:
### Current behavior

I currently define initialRouteName in each routes so I don't have to handle the back button manually. Here's the implemented navigation structure:

```typescript
const linking: LinkingOptions<RootStackParamList> = {
  prefixes: [Linking.createURL('/')],
  config: {
    initialRouteName: 'Root',
    screens: {
      Root: {
        initialRouteName: 'TabOne',
        path: '',
        screens: {
          TabOne: {
            screens: {
              TabOneScreen: 'one',
            },
          },
          TabTwo: {
            screens: {
              TabTwoScreen: 'two',
            },
          },
        },
      },
      Stack: {
        path: 'Stack',
        initialRouteName: 'A',
        screens: {
          A: 'a',
          B: 'b',
        },
      },
      Modal: 'modal',
      NotFound: '*',
    },
  },
};
```

Steps to reproduce:
1. Open new tab
2. Enter deep link directly (e.g. http://localhost:19006/Stack/a)
3. Use back button
4. It then navigates to the initialRoute, then redirect to the previous page outside App (in my case, it redirects back to New Tab screen)

https://user-images.githubusercontent.com/24470609/156485734-d4de463c-6562-4317-bbad-bd9d217d7b02.mp4

### Expected behavior

I expect the back button to only pop the screen within the app (and not redirecting to pages outside app)

### Reproduction

https://github.com/listiani13/react-navigation-initial-route-name-bug

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [X] @react-navigation/stack
- [ ] @react-navigation/native-stack

### Environment

- [ x ] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.8 |
| @react-navigation/bottom-tabs          | 6.2.0 |
| @react-navigation/drawer               | 6.3.1 |
| @react-navigation/stack                | 6.1.1 |
| expo                                   | 44.0.6 |
| node                                   | 16.3.1 |
| npm or yarn                            | yarn |

listiani13:
I don't use these packages ^

[...]

PR #10414, opened 2022-03-09
fix: using initialRouteName in web, make screen redirects to pages outside app

listiani13:

**Motivation**
If we set initialRouteName and directly access the deep link, back button on the initial screen keeps redirecting us to page outside app.
Preview

https://user-images.githubusercontent.com/24470609/156485734-d4de463c-6562-4317-bbad-bd9d217d7b02.mp4

This PR added a guard to avoid going back outside the app when using App Back Button & supplying initialRouteName. Should address this issue https://github.com/react-navigation/react-navigation/issues/10398

**Test plan**

- Create patch for this fix
- Apply patch to the reproducible bug (I've applied it in here - https://github.com/listiani13/react-navigation-initial-route-name-bug/tree/poc-using-fix)

Preview:

https://user-images.githubusercontent.com/24470609/157353361-3cab3942-a861-48cd-8bde-a9114b466f87.mp4

It's not affecting native, because I only change `useLinking.ts` (not `.native.tsx`) 

> The code must pass tests.

✅

**Code formatting**

> Look around. Match the style of the rest of the codebase. Run `yarn lint --fix` before committing.

✅


packages/native/src/useLinking.tsx
```diff
@@ -592,7 +596,8 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+              // Add Math.max to guard going to < 0 index
+              await history.go(Math.max(historyDelta, 0));
```
listiani13, line 600: guarding the historyDelta here. Sometimes it returns -1, which makes it routes to outside the app when using initialRouteName (see issue #10398 for screenshot).

My thought process:
When using initialRouteName, we don't store the initialRouteName in the browser's history
This means when historyDelta = -1, it will try to go to the previous page in browser's history (instead of our specified route name)
To fix this, I added a guard so we make sure we never navigate outside App

yarn.lock
```diff
@@ -4409,6 +4409,20 @@
...
+  dependencies:
+    "@types/react" "*"
+
+"@types/react-window@^1.8.5":
+  version "1.8.5"
+  resolved "https://registry.yarnpkg.com/@types/react-window/-/react-window-1.8.5.tgz#285fcc5cea703eef78d90f499e1457e9b5c02fc1"
+  integrity sha512-V9q3CvhC9Jk9bWBOysPGaWy/Z0lxYcTXLtLipkt2cnRj1JOSFNF7wqGpkScSXMgBwC+fnVRg/7shwgddBG5ICw==
+  dependencies:
+    "@types/react" "*"
+
```
listiani13, line 4425: not sure why, but whenever i run yarn install, it generates this for yarn lock

packages/native/src/useLinking.tsx
satya164: This branch of code (`else if (historyDelta < 0) {`) only executes when the `historyDelta` is negative, it doesn't make sense to guard it here - it's ALWAYS going to call `history.go(0)` when it comes here. It's the same as not having this branch of the condition at all.

If this shouldn't happen, then this code should be removed and tested to see if this doesn't break anything.

satya164 (changes requested):

listiani13:
Hey @flodnv! Feel free to open similar PR! Sorry I currently didn't have capacity to address the requested changes. Thanks a lot!!

aneequeahmad:
@listiani13 i have opened another similar PR. Thanks for allowing 
https://github.com/react-navigation/react-navigation/pull/10558

PR #10415, opened 2022-03-09
feat: add ability to hide screen via getPathFromState in web (useful for modal)

listiani13:

**Motivation**

**Existing problem**
In web, we sometimes want to treat Modal just as modal (not screen with path). To do that we can leverage `getPathFromState` where we hide the Modal from path (to prevent user going directly into it). The code looks something like this. This can also be accessed in https://github.com/listiani13/react-navigation-initial-route-name-bug/pull/1/
image

But the current problem is, the link is still being pushed into our browser's history. Causing the native browser back button & forward button doesn't work as expected. It may be easier to understand when it's visualized. Here's the preview of this

https://user-images.githubusercontent.com/24470609/157356661-c0bf2f58-c220-47d5-afda-02484a585881.mp4

What's not expected?
1. we can close the modal using Browser Back Button
2. The browser's forward button is being enabled
3. If we pressed browser's forward button, we will stay still in that screen, causing confusing UX

To address this, I added a guard to not push to browser's history state if previous link & current link is the same. The path also contains the screen params, so I don't see why we can't do this - since the state must be the same if link is the same

**Test plan**

> Demonstrate the code is solid. Example: the exact commands you ran and their output, screenshots / videos if the pull request changes UI.

I create POC in here https://github.com/listiani13/react-navigation-initial-route-name-bug/pull/1. Here's the preview

https://user-images.githubusercontent.com/24470609/157357240-546633b3-63d6-4c1b-b36d-95d292cd548f.mp4

I only change `useLinking.ts` (not the native one). Have tested in web using the above scenario

> The code must pass tests.

✅

**Code formatting**

> Look around. Match the style of the rest of the codebase. Run `yarn lint --fix` before committing.

✅


satya164 (changes requested):
This is going to break existing apps since it's completely valid to have multiple screens of the same type and params (resulting in the same path) in a Stack navigator. This will also mess with the history integration since it's expected for the history to match the number of screens in the stack.

If you want your modal to not be in React Navigation's history, then I'd suggest not to use a React Navigation screen as modal and use another component instead.

PR #10469, opened 2022-04-01
feat: add an ID prop to navigators

satya164:
When we have nested navigators, we often may want control over which navigator should handle an action that we dispatch. This is often handled automatically, e.g. if you navigate to a screen named `Settings`, the navigator which contains the `Settings` screen will handle the action. But in some cases, it's not possible to figure out the navigator automatically.

Some example cases:

We may want to have nested drawers to have a drawer on the left, and one on the right - and still be able to open/close them. We can use `getParent` to achieve this:

```js
navigation.getParent().openDrawer();
```

However, this is prone to error if we nest another navigator inside (e.g. a stack). So we need a way to distinguish between which drawer we want to target.

An alternative way to distinguish different navigators is to specify the `key` of the navigator in the `target` field of the action:

```js
navigation.dispatch({
  ...DrawerActions.openDrawer(),
  target: 'keyOfTheDrawer',
});
```

However, this has a few problems:

- It's not straightforward to get the `key` of the navigator since the `key` is generated by React Navigation.
- It's not type-safe. Nothing ensures that the `key` is valid.

We may also want to call `setOptions` on a parent navigator. The options don't bubble up unlike the actions, so it's important make sure to call it for the correct navigator. We can use `getParent` in this case as well, but it's prone to the same set of issues as the previous case.

There are also no alternatives for this unlike specifying `key` for actions.

This commit adds a `id` prop to navigators. Example:

```js
<Drawer.Navigator id="LeftDrawer">{/* ... */}</Drawer.Navigator>
```

Then we can use it along with `getParent` as follows:

```js
navigation.getParent('LeftDrawer').openDrawer();
```

No matter how many nested navigators are added, we can always find the correct navigator this way.

To make this work with type-checking, we'd pass a third generic to our `NavigationProp` or `ScreenProps` types:

```ts
type LeftDrawerScreenProps<T extends keyof LeftDrawerParamList> =
  DrawerScreenProps<LeftDrawerParamList, T, 'LeftDrawer'>;
```

This isn't the best way to do this, for example, the `id` prop on the navigator itself isn't type-safe. But supporting it with a good API is a lot of work, so for now, I've decided to go with this alternative approach.

### DIFF
--- a/packages/core/src/useNavigationCache.tsx
+++ b/packages/core/src/useNavigationCache.tsx
@@ -32,7 +32,14 @@ type NavigationCache<
   EventMap extends Record<string, any>
 > = Record<
   string,
-  NavigationProp<ParamListBase, string, State, ScreenOptions, EventMap>
+  NavigationProp<
+    ParamListBase,
+    string,
+    string | undefined,
+    State,
+    ScreenOptions,
+    EventMap
+  >
 >;
 
 /**
@@ -133,6 +140,15 @@ export default function useNavigationCache<
         // FIXME: too much work to fix the types for now
         ...(emitter.create(route.key) as any),
         dispatch: (thunk: Thunk) => withStack(() => dispatch(thunk)),
+        getParent: (id?: string) => {
+          if (id !== undefined && id === rest.getId()) {
+            // If the passed id is the same as the current navigation id,
+            // we return the cached navigation object for the relevant route
+            return acc[route.key];
+          }
+
+          return rest.getParent(id);
+        },
         setOptions: (options: object) =>
           setOptions((o) => ({
             ...o,
--- a/packages/core/src/useNavigationHelpers.tsx
+++ b/packages/core/src/useNavigationHelpers.tsx
@@ -8,7 +8,7 @@ import {
 import * as React from 'react';
 
 import NavigationContext from './NavigationContext';
-import { NavigationHelpers, NavigationProp, PrivateValueStore } from './types';
+import { NavigationHelpers, PrivateValueStore } from './types';
 import UnhandledActionContext from './UnhandledActionContext';
 import type { NavigationEventEmitter } from './useEventEmitter';
 
@@ -17,6 +17,7 @@ import type { NavigationEventEmitter } from './useEventEmitter';
 PrivateValueStore;
 
 type Options<State extends NavigationState, Action extends NavigationAction> = {
+  id: string | undefined;
   onAction: (action: NavigationAction) => boolean;
   getState: () => State;
   emitter: NavigationEventEmitter<any>;
@@ -32,7 +33,13 @@ export default function useNavigationHelpers<
   ActionHelpers extends Record<string, () => void>,
   Action extends NavigationAction,
   EventMap extends Record<string, any>
->({ onAction, getState, emitter, router }: Options<State, Action>) {
+>({
+  id: navigatorId,
+  onAction,
+  getState,
+  emitter,
+  router,
+}: Options<State, Action>) {
   const onUnhandledAction = React.useContext(UnhandledActionContext);
   const parentNavigationHelpers = React.useContext(NavigationContext);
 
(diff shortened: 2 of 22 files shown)


packages/core/src/useNavigationCache.tsx
```diff
@@ -133,6 +140,13 @@ export default function useNavigationCache<
         // FIXME: too much work to fix the types for now
         ...(emitter.create(route.key) as any),
         dispatch: (thunk: Thunk) => withStack(() => dispatch(thunk)),
+        getParent: (id?: string) => {
+          if (id !== undefined && id === rest.getId()) {
+            return acc[route.key];
+          }
+
+          return rest.getParent(id);
+        },
```
brentvatne, line 149: just want to verify that this logic is correct - might be worth adding a comment to clarify intent, or maybe it's obvious to someone who is more familiar with the codebase these days ;)

brentvatne (approved):

brentvatne:
one thing that might trip people up is that there is no uniqueness constraint for IDs on navigators as proposed in this pr. we should be sure to document this

packages/core/src/useNavigationHelpers.tsx
```diff
@@ -82,12 +86,32 @@ export default function useNavigationHelpers<
...
+        if (id !== undefined) {
+          let current = navigationHelpers;
+
+          while (current && id !== current.getId()) {
+            current = current.getParent();
+          }
+
+          if (current == null) {
+            throw new Error(
+              `Couldn't find a parent navigator with the ID "${id}". Is this navigator nested under another navigator with this ID?`
```
thymikee, line 100: Worth adding a test case for this

thymikee (commented):
Updating documentation with a use case would be nice :)

satya164:
> @brentvatne one thing that might trip people up is that there is no uniqueness constraint for IDs on navigators as proposed in this pr. we should be sure to document this

Yeah, unfortunately since there can be multiple instances of the same navigator component, it's possible for the ID to get duplicated, so I think we can't enforce the uniqueness in runtime, just document it.

> @thymikee Updating documentation with a use case would be nice :)

I'm planning to add a guide with 2 drawers (left & right) when I get some time. It's one of the things people have been asking a lot so it'd be useful.

lauriharpf:
In the codebase I'm working on, this seems to introduce

```
error TS2345: Argument of type 'string' is not assignable to parameter of type '{ key: string; params?: { screen?: undefined; params?: undefined; initial?: undefined; path?: string | undefined; state: Readonly<{ key: string; index: number; routeNames: string[]; history?: unknown[] | undefined; routes: NavigationRoute<...>[]; type: string; stale: false; }> | PartialState<...> | undefined; } | .....'.
```

on

```typescript
const { navigate } = useNavigation<TabNavigationProp>()
// ...
{!!title && <Foo title={title} onPress={() => navigate('BAR')} />}
```

The Typescript compiler accepted this when using `@react-navigation/core` 6.1.1, but 6.2.0 introduces this behaviour. The route setup looks like this:

```typescript
<Root.Navigator initialRouteName="AppRoot">
    <Root.Screen name="AppRoot" component={Tabs} />
</Root.Navigator>

// ... In Tabs:

<Tab.Navigator>
  <Tab.Screen name='BAR' component={BarStackNavigator} />
</Tab.Navigator>
 
// ... In BarStackNavigator:
 
<BarStack.Navigator>
  <BarStack.Screen name='BAZ' component={BazScreen} />
</BarStack.Navigator>
```

and the typing for the `TabNavigationProp` looks like this:

```typescript
export type TabParams = {
    BAR: NavigatorScreenParams<BarStackParamList>
  }
  
export type TabNavigationProp = BottomTabNavigationProp<TabParams, keyof TabParams>
```

with the `BarStackParamList` typed as

```typescript
export type BarStackParamList = {
    BAZ: {
      something?: string
    }
}
```

In case it helps anyone, changing to

```typescript
{!!title && <Foo title={title} onPress={() => navigate('BAR', { screen: 'BAZ', params: {} })} />}
```

seems to eliminate the Typescript error. I'm quite unfamiliar with the [React Navigation Type checking with TypeScript](https://reactnavigation.org/docs/typescript/), so unsure why this happened after updating to 6.2.0.

PR #10516, opened 2022-04-15
useBottomTabBarHeight behaviour changed when it's not in the bottom tabs navigator

eminsr:
Using `useBottomTabBarHeight` hook out of the Bottom Tab Bar now warns the developer once instead of trowing an error.

**Motivation**

Using the `useBottomTabBarHeight` in a component (that is usable across the app) is difficult because that component might not be always inside a Bottom Tab Navigator. Current behaviour trows an error in this situation but in most use cases if there isn't a bottom tab bar you won't need its height either so 0 height would solve it.

**Test plan**

The warn-once warning instead of the error.

Simulator Screen Shot - iPhone 12 - 2022-04-15 at 10 31 07

**Code formatting**

Look around. Match the style of the rest of the codebase. Run `yarn lint --fix` before committing.

### DIFF
--- a/packages/bottom-tabs/src/utils/useBottomTabBarHeight.tsx
+++ b/packages/bottom-tabs/src/utils/useBottomTabBarHeight.tsx
@@ -3,7 +3,9 @@ import warnOnce from 'warn-once';
 
 import BottomTabBarHeightContext from './BottomTabBarHeightContext';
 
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR, but you can already use `React.useContext(BottomTabBarHeightContext)` directly in your code if you're building a component that can be used anywhere. No need for a new option. `useBottomTabBarHeight` is for common cases where it makes sense to throw.

PR #10519, opened 2022-04-16
Strongly type the `component` prop on `RouteConfigComponent`

thomasttvo:
**Motivation**

Having a Route Parameters system is good and `navigator.navigate` can type check just fine, but the screen itself is not guaranteed to match the correct route params signature. These changes will allow screens to declare a `route: RouteProp<RouteParams, 'route'>` prop and get type-checked properly.

**Test plan**

`SomeNavigator.Screen` should type-check properly:

```tsx
type RouteParams = {
   hasParam: { param: string },
   hasParam2: { param2: string },
   noParam: undefined,
}

// Incorrect route types are recognized
const Component: FC<{route: RouteProp<RouteParams, 'hasParam'>}> = () => </>
<Screen name="hasParam" component={Component} />  ✅ 
<Screen name="hasParam2" component={Component} />  ❌ 
<Screen name="noParam" component={Component} />  ❌

// Component doesn't need to accept the `route` prop. Case I
const Component: FC<{}>  = () => </>
<Screen name="hasParam" component={Component} /> ✅   
<Screen name="noParam" component={Component} />  ✅

// Component doesn't need to accept the `route` prop. Case II
const Component: <{ someProp?: string }> = () => </>
<Screen name="hasParam" component={Component} /> ✅   
<Screen name="noParam" component={Component} />  ✅

// A required prop that's neither `route` nor `navigation` will throw errors
const Component: <{ someProp: string }> = () => </>
<Screen name="hasParam" component={Component} /> ❌   
<Screen name="noParam" component={Component} />  ❌

// Can accept just the `navigation` prop
const Component: <{ navigation: MyNavigation }> = () => </>
<Screen name="hasParam" component={Component} /> ✅   
<Screen name="noParam" component={Component} />  ✅
```


satya164:
Thanks for the PR. However, this change produces type errors. Can you take a look at them?

thomasttvo:
> Thanks for the PR. However, this change produces type errors. Can you take a look at them?

I believe the master branch has these errors at the time, I'll check again with the latest version in master

thomasttvo:
Fixed types. Ran the example app and things seemed to work fine.

satya164:
Thanks for the updates.

> I'd also like to recommend that we should be testing types using // @ts-expect-error

Can you give an example of when it passes but shouldn't?

`@ts-expect-error` is too limiting, it doesn't allow checking for specific errors so even if there is a different type error then it'll be suppressed which isn't desirable.

packages/core/src/__tests__/types.test.tsx
```diff
@@ -0,0 +1,98 @@
+/* eslint-disable jest/expect-expect */
```
satya164, line 1: Can you move the type-related tests to the `types.check.tsx` file? They don't need to be run with Jest since there are no runtime checks here and Jest won't report type errors that we're testing.

example/src/Screens/LinkComponent.tsx
```diff
@@ -8,18 +8,14 @@ import {
   createStackNavigator,
   StackScreenProps,
 } from '@react-navigation/stack';
+import type { LinkComponentDemoParamList } from 'example/src/config';
```
satya164, line 11: Use relative import here

example/src/config.ts
```diff
@@ -0,0 +1,137 @@
+import type { NavigatorScreenParams } from '@react-navigation/core';
```
satya164, line 1: This should import from `@react-navigation/native`

satya164, line 1: I'd call this file `screens.tsx`. Also, we use `.tsx` extension for all of the files.

example/src/config.ts
```diff
@@ -0,0 +1,137 @@
...
+  },
+  NativeStackPreventRemove: {
+    title: 'Prevent removing screen in Native Stack',
+    component: NativeStackPreventRemove,
+    params: undefined,
+  },
+  LinkComponent: {
+    title: '<Link />',
+    component: LinkComponent,
+    params: paramsType<
```
satya164, line 116: I'd inline `paramsType` here. I'm not a huge fan of using dummy functions just to infer the types but it's fine for now.

example/src/Screens/SimpleStack.tsx
```diff
@@ -66,12 +66,13 @@ const NewsFeedScreen = ({
...
   navigation,
 }: StackScreenProps<SimpleStackParams, 'NewsFeed'>) => {
+  let x: 'Albums' | 'Article';
   return (
     <ScrollView>
       <View style={styles.buttons}>
         <Button
           mode="contained"
-          onPress={() => navigation.navigate('Albums')}
+          onPress={() => navigation.navigate(x)}
```
satya164, line 75: `x` is never assigned, so this will cause a runtime error

packages/core/src/types.tsx
```diff
@@ -442,21 +453,26 @@ export type ScreenListeners<
...
     }
   | {
       /**
        * Lazily get a React component to render for this screen.
        */
-      getComponent: () => React.ComponentType<any>;
+      getComponent: () => Component;
       component?: never;
       children?: never;
     }
```
satya164, line 478: I'd create a helper type for this to avoid adding these generics since `RouteConfigComponent` is part of the public API

```js
type ScreenComponentType<
  ParamList extends ParamListBase,
  RouteName extends keyof ParamList,
  Props = { route: RouteProp<ParamList, RouteName>; navigation: any }
> =
  | React.ComponentType<Props>
  | React.ComponentType<Partial<Props>>
  | React.ComponentType<{}>;

export type RouteConfigComponent<
  ParamList extends ParamListBase,
  RouteName extends keyof ParamList
> =
  | {
      /**
       * React component to render for this screen.
       */
      component: ScreenComponentType<ParamList, RouteName>;
      getComponent?: never;
      children?: never;
    }
  | {
      /**
       * Lazily get a React component to render for this screen.
       */
      getComponent: () => ScreenComponentType<ParamList, RouteName>;
      component?: never;
      children?: never;
    }
```

packages/core/src/types.tsx
```diff
@@ -171,7 +171,7 @@ export class PrivateValueStore<T extends [any, any, any]> {
   protected ''?: T;
 }
 
-type NavigationHelpersCommon<
+export type NavigationHelpersCommon<
```
satya164, line 174: Exporting this makes it part of the public API which I'd avoid. This should never be used directly, the `NavigationHelpers` type should be used instead.

satya164 (changes requested):

example/src/Screens/SimpleStack.tsx
thomasttvo: oops that was debug code

example/src/config.ts
thomasttvo: Hmm I was thinking the `as unknown as` part might be confusing. The function name helps conveying the meaning here. Lmk what you think. I'm not super opinionated about this

thomasttvo:
> Can you give an example of when it passes but shouldn't?

Here's an example:

```ts
type Params =
    | {
        a: string;
        b: string;
        c: string;
      }
    | {
        a: number;
        b: string;
      };

  const func = (p: Params) => undefined;
  expectTypeOf(func).toBeCallableWith({ a: 1, b: '2', c: '3' }); // should throw error but doesn't
```

satya164 (approved):
i don't really like to use `jest` syntax for these checks since they aren't unit tests. but i'm going to merge it and make those changes.

thanks a lot for your work on this

uloco:
@thomasttvo  @satya164 

This change breaks nesting stack navigators and using CompositeScreenProps to unify their types.

If I have a nested stack structure like this `LevelOneStack -> LevelTwoStack -> LevelThreeStack` and I am on a screen in the `LevelThreeStack`, I want to be be able to type the following:

* Navigate directly to screens in 'LevelThreeStack' without specifying the stack name props.navigation.navigate("ScreenName")
* Navigate to `LevelTwoStack` screens directly without specifying the `LevelOneStack`, `.navigate("LevelTwoStack", {screen: "ScreenName"})`
* Navigate to a screen in the top stack like `.navigate("LevelOneStack", {screen: "LevelTwoStack", params: {screen: "ScreenName"}})`

To accomplish this I am using the NavigatorScreenParams<T> type in the param list and nesting all the stack props with the parent ones with `CompositeScreenProps`. This used to work until you introduced this type.

I can provide a snack if this is not clear enough.

satya164:
@uloco please open a new issue with code examples (preferably on typescript playground or snack) to illustrate the issue.

satya164:
@uloco typescript playground allows you to add libraries for sure. you can add libraries in the plugins tab under "Custom npm modules".

uloco:
oh nice I did not know that! thanks

uloco:
@satya164 see: https://github.com/react-navigation/react-navigation/issues/11242

PR #10547, opened 2022-04-25
fix: replaced I18nManager.isRTL with 18nManager.getConstants().isRTL

mohammadgharouni:
react-native-web removed I18nManager.isRTL , we should migrate to use getConstants


satya164 (approved):
Thanks

Issue #10544, opened 2022-04-24
Align `tabBarLabel` option signature between Material Bottom Tabs and Material Top Tabs


fabriziocucci:
### Current behavior

It seems that the definition of the `tabBarLabel` option is not consistent between Material bottom and top tabs.

Specifically:
```
// packages/material-bottom-tabs/src/types.tsx
  tabBarLabel?: string;

// packages/material-top-tabs/src/types.tsx
  tabBarLabel?:
    | string
    | ((props: { focused: boolean; color: string }) => React.ReactNode);
```

Is there a specific reason for this inconsistency or would you accept a PR to give the Bottom Tabs option the same level of flexibility as the Top Tabs one? i.e. support passing a function that given `{ focused: boolean, color: string }` returns a `React.Node`.

### Expected behavior

Ideally, the `tabBarLabel` option of the two packages should be aligned:

```
// packages/material-bottom-tabs/src/types.tsx
  tabBarLabel?:
    | string
    | ((props: { focused: boolean; color: string }) => React.ReactNode);

// packages/material-top-tabs/src/types.tsx
  tabBarLabel?:
    | string
    | ((props: { focused: boolean; color: string }) => React.ReactNode);
```

### Reproduction

[TS Playground](https://www.typescriptlang.org/play?#code/PTAEBUAsFMCdtASwM6gHYHsAulFoOahYajwAOsGANKMtAjguAMpECeZ0qAJtAGZ5EWRBjSgMfIjCIB3EmQCGAYwDWC-FwBcAKBCgAVKACyCrHEQKANqABC2YgFsICgEbJNoSFixl3IfEKQAK4uAHRKGA7A8MpYALRoCgBuiPimImjR0LEJyanposAulhguwA4KeMCKqupc5abmVnEu9pFxWK7IwMiwSsBYHFyhWMgAHgDEADIAbDO6YIYmZrAW1uAYZM5uHl4+fsABOCHhkVk5iSlpwoUxSvGX+TeZxaUNVTVqGt0VK2sdmw6XR6fQGQ2QI3G0wA7ABObSgUBAA)

### Platform

- [ ] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [X] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
[...]

PR #10548, opened 2022-04-25
Align tabBarLabel option of Bottom Tabs to Top Tabs

fabriziocucci:
**Motivation**

See #10544.

**Test plan**

I hope the reviewer will suggest the best way to test this.

### DIFF
--- a/packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
+++ b/packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
@@ -25,8 +25,6 @@ type Props = MaterialBottomTabNavigationConfig & {
   descriptors: MaterialBottomTabDescriptorMap;
 };
 
-type Scene = { route: { key: string } };
-
 // Optionally require vector-icons referenced from react-native-paper:
 // https://github.com/callstack/react-native-paper/blob/4b26429c49053eaa4c3e0fae208639e01093fa87/src/components/MaterialCommunityIcon.tsx#L14
 let MaterialCommunityIcons: React.ComponentType<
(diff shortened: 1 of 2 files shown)


packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
```diff
@@ -177,14 +175,21 @@ function MaterialBottomTabViewInner({
...
-          : (route as Route<string>).name;
+        const label =
+          options.tabBarLabel !== undefined
+            ? options.tabBarLabel
+            : options.title !== undefined
+            ? options.title
+            : (route as Route<string>).name;
+
+        if (typeof label === 'string') {
+          return label;
```
satya164, line 189: This shouldn't return a string. It should use `renderLabel` only if the label is a function, otherwise use the `getLabelText` prop. Without this the styling for the label will be incorrect when a string is used.

satya164 (changes requested):

packages/material-bottom-tabs/src/views/MaterialBottomTabView.tsx
satya164: Using `getLabelText` will preserve the previous behavior (e.g. use React Native Paper's theming system)

fabriziocucci:
@satya164 any updates on this?

fabriziocucci:
@satya164 any updates on this?

satya164:
Thanks for updating the PR. I'm a bit concerned about the checks using `focusedOptions`, but internally using `options` for the route. Since they don't match it can cause unexpected behavior when the options of an unfocused tab don't match the options of a focused tab.

However, I can't think of a solution from the top of my head. Probably this is something that needs changes on Paper's side too.

PR #10558, opened 2022-05-01
fix: navigation outside app when history delta is -1

aneequeahmad:
Opening PR as the issue is with navigation which navigates outside the app when history navigation is -1. I had fixed it by adding another additional if check when history is -1 but adding Math.max() check makes more sense.

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -592,7 +592,8 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+              // Add Math.max to guard going to < 0 index
+              await history.go(Math.max(historyDelta, 0));
             }
 
             // Store the updated state as well as fix the path if incorrect


satya164 (commented):
Thanks for the PR but like I commented [here] (https://github.com/react-navigation/react-navigation/pull/10414#discussion_r836390780), this branch of code only runs when delta is less than 0 - so adding `Math.max` to 0 doesn't make any sense - it's always going to call `history.go(0)` which does nothing.

The proper fix is to make sure that the resulting length of the history doesn't go below 0.

aneequeahmad:
@satya164 actually the problem is `nextIndex = -1` if the url is pasted in browser and condition ` if (nextIndex !== -1 && nextIndex < currentIndex)` is not executing since the `history.backIndex({  '/r/92741863' });` is returning -1.

aneequeahmad:
@satya164 yeah changing `window.history.go(n)` ti `window.history.go(index)` is fixing the issue.

aneequeahmad:
The problem occurs when current index is like one greater than next index like `currentIndex=3 and nextIndex=2` and `window.history.go(n)` is navigating to n where `n=-1 // passed as parameter` it should be index which is calculated in `go()` function

frenkield:
Hi All. It seems like the root problem is that the browser history is super opaque and it's difficult to determine whether it's ok to call `window.history.go()` with a negative argument. I experimented a bit and it seems like one just needs to avoid truncating the browser history to fewer than 2 items. So I think the following condition on whether to `go()` or not solves the problem:
```
if (window.history.length + historyDelta >= 2) {
    await history.go(historyDelta);
}
```

When a page loads the browser history length is either 1 or 2 (or more) depending upon how the page is initiallly loaded. There are 3 cases:

1. If you open a new tab and then navigate to a URL the browser history length is 2. Calling `window.history.go()` with any negative value will cause the browser to leave the page.
2. If you (on macos) navigate to a URL from the terminal via `open [url]` the browser history length is 1. Again, calling `window.history.go()` with any negative value will cause the browser to leave the page.
3. If you arrive on a page after clicking a link or refresh a page, the browser history length can be anything (greater than 1 or 2). In this case it's ok to call `window.history.go()` with a negative value as long as that keeps the browser history length its minimum length (either 1 or 2).

So it seems like the solution is to just prevent truncating the browser history to fewer than 2 items. In other words, it's ok to call `window.history.go(historyDelta)` as long as `window.history.length + historyDelta >= 2`.

packages/native/src/useLinking.tsx
```diff
@@ -592,7 +592,8 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+              // Add Math.max to guard going to < 0 index
+              await history.go(Math.max(historyDelta, 0));
```
aldo-expensify, line 596: This is equivalent to `await history.go(0)` because `historyDelta < 0` in this part of the code

aneequeahmad: @aldo-expensify Ahhh yes actually the histroy length is calculated in `go()` like 
```
if(n === 0) {
return
}
```
and a `if n<0 ` case is handeled here https://github.com/react-navigation/react-navigation/pull/10558#:~:text=Also%20the%20index%20%3D%20n%20%3C%200%20%3F%20Math.max(index%20%2D%20n%2C%200)%20%3A%20Math.min(index%20%2B%20n%2C%20items.length%20%2D%201)%3B which make sure that the length of history doesn't go below 0.

satya164: > so index should be used in `window.history.go(index)`

the `go` method API accepts a delta to go forward or backward. it doesn't accept an index

jpaas:
This PR fixes some issues I was having with web navigation.

In my scenario, my top-level navigator is a drawer navigator. If I navigated directly to a screen in the drawer other than the initial screen, and then navigated to the initial screen, it would back me out of the app.

PR #10562, opened 2022-05-02
fix: prevent browser from leaving page when going back in history

frenkield:
**Motivation**

There are already 2 blocked pull requests that attempt to address this issue: #10558 and #10414.

Issue: In certain situations the browser will leave the page entirely when updating the route. So this fix prevents the browser from leaving the page but still allows executing `window.history.go()` if the browser history length is sufficiently long.

This solution is different from the other pull requests because the others entirely prevent going back in the history. This solution allows going back if the browser history length is sufficiently long.

This problem is also referenced in the Expensify application: https://github.com/Expensify/App/issues/8641

**Minimum history length is 2 (experimentally)**

When a page loads the browser history length is either 1 or 2 (or more) depending upon how the page is initiallly loaded. There are 3 cases:

1. If you open a new tab and then navigate to a URL the browser history length is 2. Calling `window.history.go()` with any negative value will cause the browser to leave the page.
2. If you (on macos) navigate to a URL from the terminal via `open [url]` the browser history length is 1. Again, calling `window.history.go()` with any negative value will cause the browser to leave the page.
3. If you arrive on a page after clicking a link or refresh a page, the browser history length can be anything (greater than 1 or 2). In this case it's ok to call `window.history.go()` with a negative value as long as that keeps the browser history length its minimum length (either 1 or 2).

So it seems like the solution is to just prevent truncating the browser history to fewer than 2 items. In other words, it's ok to call `window.history.go(historyDelta)`  when `historyDelta < 0` as long as `window.history.length + historyDelta >= 2`.


packages/native/src/useLinking.tsx
```diff
@@ -592,7 +592,13 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+
+              // If window.history.length is big enough then it's ok to go back. Otherwise,
+              // we can't do window.history.go(historyDelta) because that will cause the
+              // browser to leave the page/application.
+              if (window.history.length + historyDelta >= 2) {
```
satya164, line 599: `window.history.length` isn't a reliable way to check if we can go back in history. it contains the history of all the navigation in the tab - including pages from different URLs, if there are any pages in the forward stack, etc., and doesn't get reset after refreshing or going back.

the only way to check it is with the `items` array that we have created.

satya164 (changes requested):

packages/native/src/useLinking.tsx
frenkield: Ok yeah sorry, that's totally true.

So if I save the browser history length upon page load, then I can use that value in the condition:
```
if (window.history.length + historyDelta >= initialBrowserHistoryLength) {
  await history.go(historyDelta);
}
```

Might that be the solution?

satya164: > So if I store the browser history length upon page load

`history.length` also includes the forward stack like I mentioned, not just previous items - which means it won't return what you expect if the user presses the browser's back button at any point. `history.length` shouldn't be used at all.

A potential workaround to the relevant issue is to push new entries to the history if React Naigation's history has more than one item on page load. It can be done around like 527 where the URL is updated on initial load with `history.replace`.

frenkield: Ok makes sense. So can we just do nothing if we arrive at this point in the code? In other words:
```
...

if (nextIndex !== -1 && nextIndex < currentIndex) {
  // An existing entry for this path exists and it's less than current index, go back to that
  await history.go(nextIndex - currentIndex);
} else {
  // We couldn't find an existing entry to go back to...

  // So do nothing because, given that historyDelta is negative, we're in danger of leaving the page
}

...
```

Or are you looking for a solution that addresses the root cause of why it ends up in a situation where there's no existing entry to go back to?

satya164: I already suggested a workaround, please use that. It's not possible to address why it ends up in this branch - you already have a repro for when it happens and it can happen in many scenarios where history isn't linear.

frenkield: Ok thanks @satya164.

aneequeahmad:
@satya164 , if I understand correctly, wouldn't pushing new entries on page load be more confusing for the user, as the browser back button will now behave differently than it should.
    
The solution @frenkield  suggested on the other hand seems to work. Just that maybe it should be done inside the `history's "go"` function `(line 196)`, where `window.history.go` is called.

In case the user uses the browser back button, the user is taken to the previous page (EG : in case the user came to the app using a URL)

On the other hand, if in-app react-navigation back button is used (EG : app header button), it should not leave the app entirely, and the user should remain inside the app (and only close drawer for example).

So the following solution works in both cases IMO:
```    
    if (window.history.length + historyDelta >= initialBrowserHistoryLength) {
	    //This only happens when programatic react-navigation-back button is used
	    await window.history.go(historyDelta);
    }
```

satya164:
@aneequeahmad I already explained why using `window.history.length` wouldn't work - try using the forward button a couple of times in your browser and look at `window.history.length`.

Pushing to browser history is one workaround, and you can try to come up with other workarounds if you spend some time debugging and investigating. I haven't spent time debugging this to give a perfect resolution. Dealing with browser history isn't straightforward, in addition you also need to make yourself familiar with the existing code.

I don't have the bandwidth to debug this right now. If this issue is important to you, consider hiring my time via [my employer](https://www.callstack.com/) so that I can spend time debugging this and either fix it or support you by spending time on PR review.

aneequeahmad:
@satya164 yes thats maybe a workaround but IMO we should find perfect resolution to fix this issue. Yes, this issue is important i have filled the form to hire your time. We can figure out what is the best way to fix this issue. 

Also, the form didn't ask about who to hire. I  filled it and mentioned your name. Let me know the next steps in hiring your time. Looking forward to it. Thanks

jpaas:
This PR almost fixed some problems I was having. But not quite. #10558 worked better for me.

In my scenario, my top-level navigator is a drawer navigator. If I navigated directly to a screen in the drawer other than the initial screen, and then navigated to the initial screen, it would back me out of the app.

Here's where there this PR didn't quite work for me.
1. Open a new browser tab
2. Navigate to my /non-initial-route
3. Click a cancel button to return me to the initial route.
4. So far so good, it doesn't back me out.
5. Navigate a second time to /non-initial-route
6. Click cancel again
7. Uh oh it backs me out of the route.

This may seem like a strange little corner case that doesn't matter much, but if you are trying to write e2e tests using selenium or cypress, this will break them quite badly.

PR #10563, opened 2022-05-03
BugFix: Native stack custom header not respecting screen orientation …

Tregorian:
…property

**Motivation**

When using custom header component in native stack, navigating from that screen to another with no header but screenOption = { orientation: 'landscape' }, and from there navigation.goBack(), the screen stays in landscape on android.
With built-in header that doesn't happen.

**Test plan**

Need two screens, one with a custom header, second doesn't need a header. Second screen options should include orientation: 'landscape'.
Navigate from first screen to the second, after that navigation.goBack().
Previously on android the screen was left also on landscape after goBack, now it rotates back to the original position.

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -24,6 +24,7 @@ import type { ScreenProps } from 'react-native-screens';
 import {
   Screen,
   ScreenStack,
+  ScreenStackHeaderConfig,
   StackPresentationTypes,
 } from 'react-native-screens';
 import warnOnce from 'warn-once';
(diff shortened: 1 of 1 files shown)


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -216,25 +217,28 @@ const SceneView = ({
...
-                          ),
-                        }
-                      : undefined,
-                    options,
-                    route,
-                    navigation,
-                  })}
-                </View>
+                <>
+                  <ScreenStackHeaderConfig hidden />
```
satya164, line 221: Don't we need to pass the options too?

Tregorian: Do you mean other options for the <ScreenStackHeaderConfig /> or something else?

PR #10568, opened 2022-05-04
Fix potential prototype pollution attacks (alternate approach)

matias-la:
Prevent event listeners of type `__proto__` to pollute the global Object prototype. This is fixed by setting the objects' prototype to null, so they don't have a `__proto__` property.

**Motivation**

This fixes the same problem as  #10455, but requires less changes and makes it clearer that the code is secure. Both  #10455 and this PR should fix the problem, so any of those could be merged (but not both, as it would be redundant).

**Test plan**

I tested that the changes compile ok. Also checked that no property of the Object prototype (such as `toString()` of `hasOwnProperty()`) is being accessed on the affected objects, since this change will set them to undefined.


satya164:
Closing in favor of #10455

PR #10570, opened 2022-05-05
Prevent Object properties to be used as parsing functions

matias-la:

**Motivation**

`parseQueryParams()` could had unexpected behavior when parameter names where valid `Object.prototype` properties. For example:

```
> parseQueryParams('?toString=42', {})
[Object: null prototype] { toString: '[object Object]' }
```

The code above was returning `{toString: '[object Object]'}` instead of the expected `{toString: '42'}`.

**Test plan**

Added a unit test reproducing the bug.

### DIFF
--- a/packages/core/src/__tests__/getPathFromState.test.tsx
+++ b/packages/core/src/__tests__/getPathFromState.test.tsx
@@ -902,6 +902,54 @@ it('keeps query params if path is empty', () => {
   ).toEqual(path);
 });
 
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. Looks like CI is failing due to lint errors. Can you fix them?

matias-la:
Sorry about that. Fixed and rebased!

satya164 (approved):
Thanks for the PR

PR #10454, opened 2022-03-24
Add transitionDuration property to native stack

svbutko:
**Motivation**

`react-native-screens` support [transition duration](https://github.com/software-mansion/react-native-screens/pull/1259) for native stack navigator since [v3.11.0](https://github.com/software-mansion/react-native-screens/releases/tag/3.11.0) but it's simply not passed as an option to it in `react-navigation`. At this moment it works only on iOS.

**Test plan**

```
const screenOptions = {
  transitionDuration: 2000,
  animation: 'simple_push',
};

export const MainSubStackScreen = () => {
  return (
    <Stack.Navigator screenOptions={screenOptions}>
      <Stack.Screen name={'TestComponentScreen'} component={TestComponentScreen} />
      <Stack.Screen
        name={"SecondComponentScreen"}
        component={SecondComponentScreen}
      />
    </Stack.Navigator>
  );
};
```

Code doesn't pass TS tests due to availability of this property starting from `v3.11.0`, I'm not sure whether it's good to raise a minimum version or not just for a single property

PR #10578, opened 2022-05-10
feat: add missing props from react-native-screens to native-stack

hirbod:
**Motivation**
This PR adds some missing props from `react-native-screens` to `react-navigation/native-stack`.

Especially the relevant props for the StatusBar are important if you want to use the StatusBar and NavigationBar management of RNS.

I thought it would be easiest if I provided all the compatible props in a PR, which is why I included transitionDuration. This does supersede #10454

New props:
- homeIndicatorHidden (ios)
- navigationBarColor (android)
- navigationBarHidden (android)
- statusBarColor (android)
- statusBarTranslucent (android)
- gestureDirection (ios)
- animationDuration (ios)

I made sure that I only included props which work with Expo (RNS 3.11.1, Expo SDK 45), thats why I did not include 
- hideKeyboardOnSwipe
- gestureResponseDistance

**Test plan**

All tests passed.
- yarn typescript
- yarn lint
- yarn test
- `yarn example start` runs


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -190,13 +197,20 @@ const SceneView = ({
...
       replaceAnimation={animationTypeForReplace}
       stackPresentation={presentation === 'card' ? 'push' : presentation}
       stackAnimation={animation}
       screenOrientation={orientation}
       statusBarAnimation={statusBarAnimation}
       statusBarHidden={statusBarHidden}
       statusBarStyle={statusBarStyle}
+      statusBarColor={statusBarColor}
+      statusBarTranslucent={statusBarTranslucent}
+      swipeDirection={swipeDirection}
```
WoLewicki, line 212: What about the logic of `swipeDirection` that sets other props correctly if not applied? (see https://github.com/software-mansion/react-native-screens/pull/1260/files#diff-d97b3df9fc6a0f0846507a3c189b483ac83a59c21db8788fc8d08449fbeb7343R176) I think it should be added here too.

packages/native-stack/src/types.tsx
```diff
@@ -351,6 +375,21 @@ export type NativeStackNavigationOptions = {
...
+  statusBarTranslucent?: boolean;
+  /**
+   * Sets the direction in which you should swipe to dismiss the screen.
+   * When using `vertical` option, options `fullScreenSwipeEnabled: true`, `customAnimationOnSwipe: true` and `stackAnimation: 'slide_from_bottom'` are set by default.
+   * The following values are supported:
+   * - `vertical` – dismiss screen vertically
+   * - `horizontal` – dismiss screen horizontally (default)
+   * @platform ios
+   */
+  swipeDirection?: ScreenProps['swipeDirection'];
```
WoLewicki, line 392: Could we rename it here to `gestureDirection` to align with [JS stack](https://reactnavigation.org/docs/stack-navigator/#gesturedirection)?

packages/native-stack/package.json
```diff
@@ -52,7 +52,7 @@
     "react": "17.0.1",
     "react-native": "~0.64.3",
     "react-native-builder-bob": "^0.18.1",
-    "react-native-screens": "^3.3.0",
+    "react-native-screens": "^3.11.1",
```
WoLewicki, line 55: Could you bump the `react-native-screens` in all packages, not only here?

WoLewicki (commented):
I think it is OK to add all the missing props and it will just have no effect if someone uses Expo SDK.

packages/native-stack/src/types.tsx
hirbod: But this will need changes upstream as well, no? Or you want to alias this?

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -190,13 +197,20 @@ const SceneView = ({
...
+      transitionDuration={transitionDuration}
```
WoLewicki, line 213: I think it would be better to call it `animationDuration` here since the main prop is called `animation` so we have all props regarding animating between screens close.

hirbod:
> I think it is OK to add all the missing props and it will just have no effect if someone uses Expo SDK.

This will require that I install RNS 3.13.1, because some types are missing in 3.11.1, which might have negative side effects in Expo Go, or am I wrong?

packages/native-stack/src/types.tsx
WoLewicki: Nope, you can do it the same as with e.g. [`orientation`](https://github.com/react-navigation/react-navigation/blob/0b01cb0412d8c2cbc6538b0451dd1856474b7276/packages/native-stack/src/views/NativeStackView.native.tsx#L206)

packages/native-stack/src/views/NativeStackView.native.tsx
hirbod: So the change would be

`transitionDuration={animationDuration)`, right?

packages/native-stack/src/types.tsx
hirbod: Ok, thats what I thought.

packages/native-stack/src/views/NativeStackView.native.tsx
WoLewicki: Yes

hirbod:
Looks like `hideKeyboardOnGesture` is not released yet upstream. Same for `gestureResponseDistance`
@WoLewicki 

I bumped to 3.13.1

WoLewicki:
Hmm if those props are not present there, I think it is ok to stay on `3.11.0` since there is no difference in props between that version and `3.13.1` then, yeah?

hirbod:
Correct. hideKeyboardOnSwipe and gestureResponseDistance are only available in main branch right now, therefore I can't import `ScreenProps['gestureResponseDistance']`

WoLewicki:
We are planning on releasing new version soon, but we need to align some Fabric things first so I don't want to provide the exact date now. I think we can just submit another PR when it is done in order not to block this one.

hirbod:
Ok, gonna finish your reviews now, will be done in 3 minutes

packages/native-stack/src/views/NativeStackView.native.tsx
hirbod: Done

packages/native-stack/package.json
hirbod: Done

packages/native-stack/src/types.tsx
hirbod: Done

hirbod:
Alright, all requests done @WoLewicki

WoLewicki (approved):
Looks good to me 🚀  Thanks for the contribution! Let's now wait for @satya164 review and I think it is ready to be merged then :tada:

hirbod:
You're welcome, thanks for reviewing!

yarn.lock
```diff
@@ -17610,14 +17610,30 @@ react-native-safe-area-context@3.3.2:
...
-react-native-screens@^3.3.0, react-native-screens@^3.4.0, react-native-screens@~3.10.1:
+react-native-screens@^3.11.1:
+  version "3.13.1"
+  resolved "https://registry.yarnpkg.com/react-native-screens/-/react-native-screens-3.13.1.tgz#b3b1c5788dca25a71668909f66d87fb35c5c5241"
+  integrity sha512-xcrnuUs0qUrGpc2gOTDY4VgHHADQwp80mwR1prU/Q0JqbZN5W3koLhuOsT6FkSRKjR5t40l+4LcjhHdpqRB2HA==
+  dependencies:
+    react-freeze "^1.0.0"
+    warn-once "^0.1.0"
+
+react-native-screens@^3.4.0:
```
kirillzyusko, line 17621: Can't 3.4.0 and 3.11.1 be casted to a single 3.13.1?

Just curious why before we had three different libraries pointing to single library version, and now it seems we have 3 different versions of react-native-screens 🤔

hirbod: These are some typical monorepo issues. But we should not 3.13.1, since 3.11.1 is the version shipped with the latest Expo SDK 45. I am not sure where RNS 3.4.0 is coming from, usually this issues disappear when you delete the yarn.lock and reinstall it

hirbod: @WoLewicki 

```
➜  react-navigation git:(main) ✗ yarn why react-native-screens
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "react-native-screens"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "react-native-screens@3.11.1"
info Has been hoisted to "react-native-screens"
info Reasons this module exists
   - "workspace-aggregator-37e960f9-a2b1-4db0-b892-7f2b0e76bc0e" depends on it
   - Hoisted from "_project_#@react-navigation#bottom-tabs#react-native-screens"
   - Hoisted from "_project_#@react-navigation#drawer#react-native-screens"
   - Hoisted from "_project_#@react-navigation#stack#react-native-screens"
   - Hoisted from "_project_#@react-navigation#example#react-native-screens"
info Disk size without dependencies: "1.84MB"
info Disk size with unique dependencies: "1.92MB"
info Disk size with transitive dependencies: "1.92MB"
info Number of shared dependencies: 2
=> Found "@react-navigation/native-stack#react-native-screens@3.13.1"
info This module exists because "_project_#@react-navigation#native-stack" depends on it.
info Disk size without dependencies: "2.25MB"
info Disk size with unique dependencies: "2.33MB"
info Disk size with transitive dependencies: "2.33MB"
info Number of shared dependencies: 2
=> Found "react-native-reanimated#react-native-screens@3.10.2"
info This module exists because "_project_#@react-navigation#drawer#react-native-reanimated" depends on it.
info Disk size without dependencies: "1.78MB"
info Disk size with unique dependencies: "1.86MB"
info Disk size with transitive dependencies: "1.86MB"
info Number of shared dependencies: 2
✨  Done in 1.19s.
```

I'll try to regenerate the lock file

hirbod: But another issue is that react-native-reanimated adds a peer to 3.10.2. 
Should we add a yarn resolution to package.json ?

WoLewicki: @tomekzaw @piaskowyk is there a reason for this peer dep in `reanimated`?

hirbod: I fixed all the package.json @WoLewicki 
The last piece is reanimated.

```
➜  react-navigation git:(main) yarn why react-native-screens
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "react-native-screens"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "react-native-screens@3.11.1"
info Has been hoisted to "react-native-screens"
info Reasons this module exists
   - "workspace-aggregator-7ec20abb-c574-406e-97c1-14d823890d28" depends on it
   - Hoisted from "_project_#@react-navigation#bottom-tabs#react-native-screens"
   - Hoisted from "_project_#@react-navigation#drawer#react-native-screens"
   - Hoisted from "_project_#@react-navigation#native-stack#react-native-screens"
   - Hoisted from "_project_#@react-navigation#stack#react-native-screens"
   - Hoisted from "_project_#@react-navigation#example#react-native-screens"
info Disk size without dependencies: "1.84MB"
info Disk size with unique dependencies: "1.92MB"
info Disk size with transitive dependencies: "1.92MB"
info Number of shared dependencies: 2
=> Found "react-native-reanimated#react-native-screens@3.10.2"
info This module exists because "_project_#@react-navigation#drawer#react-native-reanimated" depends on it.
info Disk size without dependencies: "1.78MB"
info Disk size with unique dependencies: "1.86MB"
info Disk size with transitive dependencies: "1.86MB"
info Number of shared dependencies: 2
✨  Done in 1.24s.
```

kirillzyusko: @hirbod looks better 👍 But why ^3.4.0 is pointed to 3.10.2 now, whereas we have 3.11.1 in the project? I thought ^3.4.0 means 3.x library (according to npm sematic versioning), so it should re-use 3.11.1 version? Or I'm missing something?

hirbod: Not sure. 3.4.0 is not being hoisted regarding to yarn why. It's 3.11.1 and 3.10.2 because of rea

kirillzyusko (approved):

yarn.lock
hirbod: @WoLewicki looks like the peer has been removed in 2.8.0
https://github.com/software-mansion/react-native-reanimated/commit/5601bc62cd5e1e2b4a1c41ce3e9d7059207a37d9

Shall I bump REA to 2.8.0? It is also the shipped version with SDK 45

WoLewicki: Yeah, I think it is a good idea to bump it too if nothing breaks then

hirbod: @WoLewicki upgrading REA to 2.8.0 fixes the issues but the example app will stuck on SplashScreen. Not sure if thats really the scope of this PR for me to upgrade to Expo SDK 45 now? I could do this in another PR

WoLewicki: That's weird. Does the app crash then?

hirbod: It does look like that. It just stuck on the SplashScreen as soon as I bump REA to 2.8.0. Guess some work has to be done (upgrading to Expo 44). I think there had been changes with auto installing which don't work with SDK 44

brentvatne:
can you open a pr to the website repo (https://github.com/react-navigation/react-navigation.github.io) to update the docs for this as well?

hirbod:
Sure, but might need a couple of days since I am caught up with work currently

hirbod:
@brentvatne done.
https://github.com/react-navigation/react-navigation.github.io/pull/1151

Please feel to correct the text directly if you're not happy with wording.

Chasty:
are you going to merge this PR? @hirbod maybe we need to update react native screens to the latest version.

`"react-native-screens": "^3.13.1"`

hirbod:
@brentvatne any chance to get some traction here? :)

kacperkapusciak:
Hi @hirbod 👋 
Would you mind resolving conflicts on this PR as I don't have permissions to push to forks?

Thanks

hirbod:
Hi @kacperkapusciak,

I resolved the conflicts. Once the checks are done we should be good to go :)

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -190,13 +216,20 @@ const SceneView = ({
             false
           : gestureEnabled
       }
+      homeIndicatorHidden={homeIndicatorHidden}
```
kacperkapusciak, line 219:
```suggestion
      homeIndicatorHidden={autoHideHomeIndicator}
```
After discussing with @satya164 we've decided to name this `autoHideHomeIndicator` as it's impossible to hide it permanently. It just prefers to stay hidden.

packages/native-stack/src/types.tsx
```diff
@@ -351,6 +375,21 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   statusBarStyle?: ScreenProps['statusBarStyle'];
+  /**
+   * Sets the translucency of the status bar. Defaults to `false`.
+   *
+   * @platform android
+   */
+  statusBarTranslucent?: boolean;
```
kacperkapusciak, line 383: This prop doesn't seem to work at all in my emulator neither in expo managed nor bare workflow. Maybe something with my device but could you double-check this? 🙏 

I'll test try to test this more

kacperkapusciak (changes requested):
Is it possible for you to allow maintainers to make changes to this branch? That would help me greatly 🙏 

Thanks!

hirbod:
@kacperkapusciak can you tell me where I can activate this?

packages/native-stack/src/views/NativeStackView.native.tsx
hirbod: Ok, keep in mind that we need to update this in my documentation PR as well.

packages/native-stack/src/types.tsx
hirbod: It was working for when I tested it, but only when you're not using StatusBar from react-native or expo IIRC

yarn.lock
```diff
@@ -18202,6 +18202,14 @@ react-native-screens@~3.11.1:
...
     warn-once "^0.1.0"
 
+react-native-screens@~3.11.1:
+  version "3.11.1"
+  resolved "https://registry.yarnpkg.com/react-native-screens/-/react-native-screens-3.11.1.tgz#9bca9968986ca9195cb1e7e6fca37543bde64ecb"
+  integrity sha512-ziQqVm97tNtovacyHwNmDwJPb8n9CqwsfttXx2p5Hk7wUWemDcPAX0ZJ/nNnGMSq2p2QPhPjjUpr3qKXuES0sQ==
+  dependencies:
+    react-freeze "^1.0.0"
+    warn-once "^0.1.0"
+
```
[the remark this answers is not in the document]
hirbod: Not 2, even 3 :D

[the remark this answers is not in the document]
hirbod: Should be resolved now

cenzovit:
Any update on getting the remaining missing props into native-stack? Looks like react-native-screens has since released versions with the final missing props added.

hirbod:
Yeah, we could make a new PR with

```
hideKeyboardOnSwipe
gestureResponseDistance
```

BlueBazze:
Might not be an issue. And not sure if i have messed anything up.
Using the option `navigationBarColor`.
If its set to an RGB value, the app crashes with the message `java.lang.RuntimeException: java.lang.reflect.InvocationTargetException`
If its set to an RGBA value, its works as 

Navigating to a screen with that option, it is as if the navigationBar is being set to position: absolute
Meaning that any content that is positioned at the bottom of your screen is now located behind the navigationBar.

Also using the option `navigationBarHidden`
Navigating to a screen with this option, it looks fine at first. But after the first tap anywhere on the screen causes the navigationBar to appear again.

hirbod:
@BlueBazze you should file this as issue in the react-native-screens repo with a reproduction.

PR #10585, opened 2022-05-12
Fix useReanimatedTransitionProgress throwing errors after upgrade to v6

aleqsio:
**Motivation**

Currently, trying to use [useReanimatedTransitionProgress](https://github.com/software-mansion/react-native-screens/blob/main/guides/GUIDE_FOR_LIBRARY_AUTHORS.md#usereanimatedtransitionprogress) from react-native-screens results in the following error:
`Error: Couldn't find values for reanimated transition progress. Are you inside a screen in Native Stack?`

**Reproduction**:
https://github.com/aleqsio/use-reanimated-transition-repro/blob/master/App.tsx

As @WoLewicki  pointed out, this is due to react navigation v6 not using a `Screen` from `ScreenContext`, instead using the default import. The fix is to use the screen from context if available. 

After this change the reproduction app works as expected.

---
Screenshots:

simulator_screenshot_1124636B-7D6A-44B2-9703-8A0F60DD59C9

simulator_screenshot_8BCDF351-86C5-4BF6-919E-9B19FEA152BD

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -22,8 +22,8 @@ import {
 } from 'react-native-safe-area-context';
 import type { ScreenProps } from 'react-native-screens';
 import {
-  ScreenContext,
   Screen as DefaultScreen,
+  ScreenContext,
   ScreenStack,
   StackPresentationTypes,
 } from 'react-native-screens';
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -22,7 +22,8 @@ import {
 } from 'react-native-safe-area-context';
 import type { ScreenProps } from 'react-native-screens';
 import {
-  Screen,
+  ScreenContext,
+  Screen as DefaultScreen,
   ScreenStack,
   StackPresentationTypes,
 } from 'react-native-screens';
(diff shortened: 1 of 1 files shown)


hirbod:
`useReanimatedTransitionProgress` has not been added to react-navigation yet. If you want to use it right now, you need to import `createNativeStackNavigator` from `react-native-screens/native-stack` and add the Provider for it.

I've talked about this topic with @satya164 while I was working on #10578 and he said:

> Transition progress and native back button dismiss require additional consideration to properly integrate/align with React Navigation's APIs, so it's not straightforward. And I don't have enough time to think about them right now.

hirbod (approved):

WoLewicki:
@hirbod since you can already use `useTransitionProgress` hook in v6, I don't see any problems with adding `useReanimatedTransitionProgress`, since they are both imported from `react-native-screens` anyway. We can then think about the proper API in `react-navigation` if needed. cc @satya164 wdyt about it?

hirbod:
Yes, reasonable. I don't see any issues with your PR.

hirbod:
> @hirbod since you can already use `useTransitionProgress` hook in v6, I don't see any problems with adding `useReanimatedTransitionProgress`, since they are both imported from `react-native-screens` anyway. We can then think about the proper API in `react-navigation` if needed. cc @satya164 wdyt about it?

I don't see any issues as well, I just gave some context that I was about to add support for it but @satya164 had concerns.

satya164:
I think it'd be better to fix this in `react-native-screens` instead so that it's not necessary to use `ScreenContext` for consumers. I had a discussion with @WoLewicki about it.

kacperkapusciak:
Sent a PR fixing this on rn-screens side https://github.com/software-mansion/react-native-screens/pull/1539 👍

kacperkapusciak:
Closing this as https://github.com/software-mansion/react-native-screens/pull/1539 got merged in rn-screens.

Issue #10527, opened 2022-04-18
Pushing a screen from outside of bottom tab navigation opens the screen in previous active bottom tab.


swizes:
### Current behavior

My root stack

```
<Stack.Navigator>
      <Stack.Screen
        name="BottomTabNavigation"
        component={BottomTabNavigation}
      />
      <Stack.Screen
        name={'ScreenWithoutTab'}
        key={'ScreenWithoutTab'}
        component={ScreenWithoutTab}
      />
</Stack.Navigator>
```

My BottomTab

```
<Tab.Navigator
      initialRouteName={'Home'}>
      <Tab.Screen
        name="Home"
        key={'Home'}
        component={HomeStack}
      />
      <Tab.Screen
        name="Settings"
        key={'Settings'}
        component={SettingsStack}
      />
</Tab.Navigator>
```

HomeStack

```
<Stack.Navigator
      initialRouteName={'HomeScreen'}>
      <Stack.Screen name="HomeScreen" key={'HomeScreen'} component={Home} />
      <Stack.Screen name="TestScreen" key={'TestScreen'} component={Test} />
</Stack.Navigator>
```

SettingsStack
```
<Stack.Navigator
      initialRouteName={'HomeScreen'}>
      <Stack.Screen name="HomeScreen" key={'HomeScreen'} component={Home} />
      <Stack.Screen name="TestScreen" key={'TestScreen'} component={Test} />
</Stack.Navigator>
```

Problematic Scenario

1. I open the app and the bottom tab navigation appears with initial tab **HOME**
2. I switch to **Settings Tab**
3. I switch to **Home Tab**
4. I press on a button on Home Screen to open the screen **ScreenWithoutTab**
5. In **ScreenWithoutTab** screen, i press a button which calls following functions

```
navigation.push('TestScreen')
navigation.goBack()
```

6. The screen opens at **Settings** Tab. But my latest active tab was **Home** as you can see at **Step 3.**

If i dont do the **Step 2.**, everything works as expected the **TestScreen** opens at **Home** tab

### Expected behavior

If I open a screen from the stack which is above bottom tab navigation and when a screen (which is nested in bottom tab) is pushed from that screen, the current tab should try to open the screen not previous selected tab.

### Reproduction

Steps to reproduce described above.

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/bottom-tabs
[...]

PR #10596, opened 2022-05-17
draft fix of issue 10527 - Prioritise active tabs when navigating

dorisaward:
**Motivation**

Potential fix for https://github.com/react-navigation/react-navigation/issues/10527

- The focused tab should be prioritised over the most recently opened tab.

**Test plan**

- Use with the minimal repro in this snack: https://snack.expo.dev/@liam/react-navigation-issue-10527-minimal-repro

The change must pass lint, typescript and tests. - Currently does not pass these. Raising as a draft PR to see if this would be a viable option before updating tests etc.


packages/core/src/useKeyedChildListeners.tsx
```diff
@@ -1,31 +1,39 @@
...
-      type: T,
-      key: string,
-      listener: KeyedListenerMap[T]
+    <ListenerType extends keyof KeyedChildListeners>(
+      type: ListenerType,
+      key: string | undefined,
+      listener: MapValueType<KeyedListenerMap[ListenerType]> | undefined
     ) => {
-      keyedListeners[type][key] = listener;
+      const keyToSet = key !== undefined ? key : 'undefined';
```
liamjones, line 32: This is one area we're not sure about the best way to handle. 

https://github.com/react-navigation/react-navigation/pull/10596/files#diff-8186269a9d18bc03947df26c6d0acaabe753c63a7bf0bec13280b7f06aecd266R621 states that sometimes the route.key can be missing when calling useOnAction which means this could now be called without a key (whereas before it was going to useChildListeners which didn't need key).

For now, we've been assigning the hardcoded key of `undefined` which is probably not ideal. For someone who knows the codebase better; in what situations is route.key missing at the above-referenced line? Would there only ever be one listener without a key or would we need to cope with there potentially being multiple listeners with no key?

Yupeng-li:
Hi, I updated the pull request. All existing tests passed but I didn't add a new test to cover the fix. Is the test required before you could accept the pull request? If so, would you please advise the preferred way of constructing the new test? Any code examples are appreciated. Thank you.

satya164:
Thank you for the PR. However, we can't merge this:
1. This is a breaking change as it changes behavior of existing code
2. We don't recommend relying on this, we recommend explicitly navigating to the nested screen you want: https://reactnavigation.org/docs/nesting-navigators/#navigating-to-a-screen-in-a-nested-navigator

In React Navigation 7, this will be disabled by default and require using the API to navigate to a screen in a nested navigator. It'll be eventually removed in another version.

liamjones:
@satya164 

> This is a breaking change as it changes behavior of existing code

Agreed but arguably that's correct if this is a "bug" fix? I can't see a user ever expecting "Navigate on the last lazy initialised tab" as a behaviour. I _can_ see them expecting "navigate on the tab I was just on before this modal" as one. This seems to be a common behaviour for shared screens across tabs in multiple apps I've used.

> We don't recommend relying on this, we recommend explicitly navigating to the nested screen you want: https://reactnavigation.org/docs/nesting-navigators/#navigating-to-a-screen-in-a-nested-navigator

This is disappointing to hear - is there a specific reason why this is being dropped? Having to specify parent navigators kills some of the flexibility/ease-of-use of the project. For example, currently, a view just knows "I want to navigate to X" and ask the system to do it. If the navigators are rearranged later on it doesn't matter to the view, it needs no knowledge of this.

If this feature is dropped then there's going to be a hard dependency between the views wanting to navigate and the structure of the navigators.

Maybe I'm approaching this from the wrong direction?;

The only reason we've run into this issue is that the docs recommend that you move those screens to a parent navigator when you wish to hide a tab bar in certain screens: https://reactnavigation.org/docs/hiding-tabbar-in-screens/. As such, that's what we've been trying to do. We could go down the route of displaying the modals inside the tab bar navigator but hiding it via style changes, etc. Can I ask - what is the reason the documented method is the recommended one? It says it's recommended but doesn't explain why.

satya164:
> For example, currently, a view just knows "I want to navigate to X" and ask the system to do it. If the navigators are rearranged later on it doesn't matter to the view, it needs no knowledge of this.

Rearranging navigators will break such code right now since the navigator itself needs to be rendered first to be able to navigate to a screen inside of it. In some scenarios, it may already be rendered (like a tab navigator with `lazy` set to `false`), but it couples the code to logic specified elsewhere which has a bigger chance of breaking stuff by changing seemingly unrelated code.

In addition, this doesn't work with the type system, so the TypScript can't help with identifying mistakes.

Specifying the parent navigator makes it work even if the navigator wasn't rendered previously. It also makes it work with TypeScript which will help with identifying the places where you need to update the code after rearranging the navigators, which is much more reliable than what's there right now.

liamjones:
> Rearranging navigators will break such code right now since the navigator itself needs to be rendered first to be able to navigate to a screen inside of it.

Ah, okay. I can see I've been making some assumptions based on the way our app works specifically, thanks for the clarification. 👍

satya164:
> It says it's recommended but doesn't explain why

Changing the screen size by hiding the tab bar during the animation can causes issues with the animation and also doesn't work with `native-stack`.

PR #10600, opened 2022-05-20
Push history items to history when react-navigation history length > 1

aneequeahmad:
### Prevent the app to navigate outside app when going back in history 

This PR is created after some discussion on other PR's like https://github.com/react-navigation/react-navigation/pull/10562.

**Motivation**

We came to the conclusion that we should push new entries to the history if React Navigation's history has more than one item on page load. It will keep the history updated and prevents the app to exit if `app-navigation` or `browser` back or forward buttons are used. https://github.com/react-navigation/react-navigation/pull/10562#discussion_r863224387.

@satya164 Let's fix this issue. It is an urgent need to fix the root cause as due to this `bug` there are many navigation issues on different platforms(mobileWeb, Desktop, Web). Your help in fixing the issue is highly appreciated. Thanks

**Test plan**

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -25,9 +25,11 @@ type HistoryRecord = {
   path: string;
 };
 
+let isNewEntriesPushed: boolean = false
+let items: HistoryRecord[] = [];
+
 const createMemoryHistory = () => {
   let index = 0;
-  let items: HistoryRecord[] = [];
 
   // Pending callbacks for `history.go(n)`
   // We might modify the callback stored if it was interrupted, so we have a ref to identify it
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Does this PR fix the bug where, on web, if you land on /tab2 (initial load) and then press /tab1 (the default tab, according to the `"linking"` parameter, mapping paths to screens), it goes to a blank page (as in, goes back to before you loaded the initial tab.

If this message is too cryptic, no worries, I'll open a new issue with example, but if you know what I am talking about, I then don't need to do that ;)

aneequeahmad:
@SleeplessByte You're right, it also solves other navigation issues like navigating to subLinks, First interaction after page loads etc. Have some code to push in this PR. Will update this `PR` today.

aneequeahmad:
@SleeplessByte no need to merge this PR. The other [PR](https://github.com/react-navigation/react-navigation/pull/10601) has been merged which has commits that fixes browser leaving the page when pressing back when there was an initial route. 

Closing this PR.

satya164:
Hey thanks for the PR, we discussed these changes but decided to go with a different approach for fixing the issue. Thank you for spending time on debugging and opening a PR.

Issue #10481, opened 2022-04-04
Web: Selecting first tab unexpectedly navigates back when both linking and initialRouteName are used


jrc:
### Current behavior

1. Create a new app and paste the code from https://reactnavigation.org/docs/tab-based-navigation/#a-native-stack-navigator-for-each-tab
2. Add a `linking` config to the navigation container:
    ```
    const linking = {
        prefixes: [
            /* your linking prefixes */
        ],
        config: {
            screens: {
                HomeStack: {
                    screens: {
                        Home: 'home',
                        Details: 'home/details',
                    },
                },
                SettingsStack: {
                    screens: {
                        Settings: 'settings',
                        Details: 'settings/details',
                    },
                },
            },
        },
    };
    ```
    
4.  Set `initialRouteName` on the tab navigator to be the second tab:
    ```
    <NavigationContainer linking={linking}>
    <Tab.Navigator initialRouteName="SettingsStack">
        <Tab.Screen name="HomeStack" component={HomeStackScreen} />
        <Tab.Screen name="SettingsStack" component={SettingsStackScreen} />
    </Tab.Navigator>
    </NavigationContainer>
    ```
    (optional) Also fix up the tab screen names to not be ambiguous (buggy sample code).
5. Run the app. The app defaults to the second tab (SettingsStack) as expected.
6. Click on the first tab (HomeStack). Instead of switching to the first tab, the browser navigates back instead, possibly even going outside the app.

Attaching screen recording demonstrating the behavior.

Notes:
The problem only seems to occur if the `initialRouteName` is used, causing the window location to change.
* Does not reproduce if `linking` is not set.
* Does not reproduce if `initialRouteName` is not set or is set to the first tab.
* Can't reproduce in Snack as you can't explicitly reload the app in the browser (as far as I can tell).

https://user-images.githubusercontent.com/334386/161596478-dd03ed4b-0466-4470-bca1-52f3bb33787d.mov

Possible related to
* https://github.com/react-navigation/react-navigation/issues/10398 (but here the back button is never used)
* https://github.com/react-navigation/react-navigation/issues/10152

### Expected behavior

The app should switch to each tab normally.

### Reproduction

https://snack.expo.dev/@john_gronska/setting-initialroutename-on-navigation-incorrectly-navigates-back-when-selecting-first-tab

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

[...]

PR #10601, opened 2022-05-22
prevent `history.go()` navigation outside bounds

marcaaron:
**Motivation**

As the length of a browser's history cannot be used to accurately determine whether a user will be taken out of a react-navigation app as a workaround this PR will guard against unexpected behavior by instead moving only within the upper or lower bounds of the memory history. If we are trying to navigate to an index that is beyond what we have in memory history this is probably a sign that our `state.history` or `state.routes` are out of sync with the memory history and we will calculate a delta that will not bring us where we expect.

I am interested in looking into the scenarios where `state.history` or `state.routes` are not in sync - but it seems unclear what the expected behavior should be.

**Issue:**

https://github.com/react-navigation/react-navigation/issues/10481

**Test plan**

**Automated**

- As part of this PR `createMemoryHistory()` was exported so it a unit test could be created and simulate state changes that trigger changes to the memory history.

**Manual**

- Copy the minimal reproduction similar to the linked issue into a new managed expo app with navigation template.
- Run the example website.
- Navigate directly to the alternate screen (not the `initialRouteName`) via browser's url in a new tab to ensure fresh browser history.
- Navigate back to the `Home` route.
- Verify that you are not navigated out of the app.

<details>
 <summary>Minimal Reproduction Code</summary>
 
```js
const linking = {
  prefixes: [Linking.createURL('/')],
  config: {
    initialRouteName: 'Home',
    screens: {
      Home: {
        path: 'home',
      },
      Test: {
        path: 'test',
      },
    },
  },
};

export default function Navigation() {
  return (
    <NavigationContainer
      linking={linking}
    >
      <Navigator />
    </NavigationContainer>
  );
}

const Tab = createBottomTabNavigator();

const TestScreen = ({route, to}) => {
  const linkTo = useLinkTo();
  return (
    <View style={{padding: 20}}>
      <Text>1. In a new tab navigate directly to /test via browser url bar</Text>
      <Text>2. Navigate to home via button</Text>
      <Text style={{fontSize: 24, padding: 20}}>Current Route: {route.name}</Text>
      <Button title={to} onPress={() => linkTo(to)}/>
    </View>
  );
}

function Navigator() {
  return (
    <Tab.Navigator initialRouteName="Home">
      <Tab.Screen name="Home" component={(props) => <TestScreen {...props} to="/test"/>} />
      <Tab.Screen name="Test" component={(props) => <TestScreen {...props} to="/home" />} />
    </Tab.Navigator>
  );
}
```
</details>

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -25,7 +25,7 @@ type HistoryRecord = {
   path: string;
 };
 
-const createMemoryHistory = () => {
+export const createMemoryHistory = () => {
   let index = 0;
   let items: HistoryRecord[] = [];
 
@@ -44,6 +44,9 @@ const createMemoryHistory = () => {
   };
 
   const history = {
+    get length(): number {
+      return items.length;
+    },
     get index(): number {
       // We store an id in the state instead of an index
       // Index could get out of sync with in-memory values if page reloads
@@ -123,14 +126,25 @@ const createMemoryHistory = () => {
     go(n: number) {
       interrupt();
 
+      // To guard against unexpected navigation out of the app we will assume that browser history is only as deep as the length of our memory
+      // history. If we don't have an item to navigate to then update our index and navigate as far as we can without taking the user out of the app.
+      const nextIndex = index + n;
+      const lastItemIndex = items.length - 1;
+      if (n < 0 && !items[nextIndex]) {
+        // Attempted to navigate beyond the first index. Negating the current index will align the browser history with the first item.
+        n = -index;
+        index = 0;
+      } else if (n > 0 && nextIndex > lastItemIndex) {
+        // Attempted to navigate past the last index. Calculate how many indices away from the last index and go there.
+        n = lastItemIndex - index;
+        index = lastItemIndex;
+      } else {
+        index = nextIndex;
+      }
+
       if (n === 0) {
         return;
       }
-
-      // We shouldn't go back more than the 0 index (otherwise we'll exit the page)
-      // Or forward more than the available index (or the app will crash)
-      index =
-        n < 0 ? Math.max(index - n, 0) : Math.min(index + n, items.length - 1);
 
       // When we call `history.go`, `popstate` will fire when there's history to go back to
       // So we need to somehow handle following cases:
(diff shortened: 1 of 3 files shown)


packages/native/src/useLinking.tsx
```diff
@@ -123,15 +126,26 @@ const createMemoryHistory = () => {
...
+      }
+
       if (n === 0) {
         return;
       }
 
-      // We shouldn't go back more than the 0 index (otherwise we'll exit the page)
-      // Or forward more than the available index (or the app will crash)
-      index =
-        n < 0 ? Math.max(index - n, 0) : Math.min(index + n, items.length - 1);
```
marcaaron, line 133: @satya164 I _think_ this logic is incorrect. If we set have a negative `n` delta (i.e. navigating back) then we will set the index to either `0` or `index - delta`, but since `delta` is negative we should be adding `index + n` not `index - n` ?

e.g. say we have 2 items and our index is `1` and our delta is `n = -1` then our new index would be `2` or `0` when the correct index should be `0`.

Screen Shot 2022-05-22 at 6 57 34 AM

Comment also seems out of date as setting the index does not prevent us from navigating outside the bounds since we are always calling `window.history.go(n)` with an unmodified delta.

https://github.com/react-navigation/react-navigation/blob/1e1bc84b64ca0f740e2d1b7ff531d64efdb56887/packages/native/src/useLinking.tsx#L210

aneequeahmad: @marcaaron yes, you're right this logic is incorrect as `delta` is negative and we should be adding `index + n` i figured it out as well during [debugging](https://github.com/Expensify/App/issues/7618#:~:text=there%20are%20a%20few%20other%20issues%20that%20i%20have%20found%20in%20react%2Dnavigation). Was about to raise a PR to fix this after the other [fix](https://github.com/react-navigation/react-navigation/pull/10600). 

Thank you for raising this PR 🙇

kacperkapusciak:
Hi @marcaaron 👋 
Amazing work with the PR! 🏅 

I had to hide the `history.length` API that you've exposed, which is a complete list of items in the back & forward stack. This is something that the end-users would virtually never need and we were afraid that that could only create confusion.

I've also extracted the `createMemoryHistory` to a separate file but it's only for cosmetics 😃  

Thank you for your contribution!

kacperkapusciak (approved):

PR #10602, opened 2022-05-23
add `homeIndicatorHidden` to native stack navigator options

michaelknoch:
**Motivation**
Allows hiding the home indicator for native navigators, groups or views. This is based on https://github.com/software-mansion/react-native-screens/pull/1267 from @WoLewicki which was released with [react-native-screens@3.11.0](https://github.com/software-mansion/react-native-screens/releases/tag/3.11.0)

**Test plan**
I tested it with the demo app and with our own project. Works like charm and makes [react-native-home-indicator](https://github.com/flowkey/react-native-home-indicator) obsolete

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -52,7 +52,7 @@
     "react": "17.0.1",
     "react-native": "~0.64.3",
     "react-native-builder-bob": "^0.18.1",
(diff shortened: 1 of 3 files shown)


michaelknoch:
can be closed when https://github.com/react-navigation/react-navigation/pull/10578 goes in

PR #10607, opened 2022-05-26
Explicitly set a boolean value for the `detachInactiveScreens` prop

mstaicu:
The value of the `detachInactiveScreens`, with which one renders the navigator returned from the call to `createStackNavigator` is not kept between renders, i.e. it is rendered with the correct value but then it falls back to `undefined`. This has some unintended consequences down the hierarchy of components and is currently crashing our React Native app when we seek to open the app from any deep links

- createStackComponent -> 
  - StackView -> 
  - CardStack -> ( if `detachInactiveScreens` is undefined, then its default value is derived from platform checks )
  - MaybeScreenContainer -> ( renders the `Screens.ScreenContainer` and `Screens.Screen` from `software-mansion/react-native-screens` with the `enabled` set incorrectly
  - ScreenContainer (`software-mansion/react-native-screens`) -> unpacks the props and completely [ignores](https://github.com/software-mansion/react-native-screens/blob/main/src/index.native.tsx#L340) the `ENABLE_SCREENS`'s default value as we're rendering this component with a value. Calling `enableScreens` to set the  `ENABLE_SCREENS` [has no effect](https://github.com/software-mansion/react-native-screens/blob/main/src/index.native.tsx#L49), thus we cannot toggle to not opt-in to the new architecture and we get crashes

### DIFF
--- a/packages/stack/src/navigators/createStackNavigator.tsx
+++ b/packages/stack/src/navigators/createStackNavigator.tsx
@@ -123,6 +123,7 @@ function StackNavigator({
         state={state}
         descriptors={descriptors}
         navigation={navigation}
+        detachInactiveScreens={rest.detachInactiveScreens ?? false}
       />
     </NavigationContent>
   );


satya164:
Thanks for the PR.

> Values set on a navigator for the detachInactiveScreens prop are not kept between renders, i.e. the value of the prop toggles from the initial value with which the navigator was rendered, then it falls back to undefined

Can you elaborate on what this means? I'm not sure what you mean by toggles. The default value is set based on the platform which will never change for a given platform.

Looks like you're defaulting `detachInactiveScreens` to `false` in your PR which isn't what we want. The default is supposed to be `true`. You can already explicitly specify `detachInactiveScreens={false}` in your navigator to control this behavior.

satya164:
Then you should specify `detachInactiveScreens` prop on the navigator. Changing the default value inside the library isn't an appropriate solution.

mstaicu:
Totally agree with you, but theres something going on that I don't understand. If I specify a value for `detachInactiveScreens`, that value gets lost between renders, and this is the part where I can't understand why it is doing that, hence why I proposed this change. The way I uncovered this behaviour is with the obvious logging in the `createStackComponent`.

So this issue that I'm experiencing would not be related to this framework?

kacperkapusciak:
Hi @mstaicu :wave:

> This would allow us to be able to use the `enableScreens` function to toggle `react-freeze` support

`react-freeze` support is toggled with `enableFreeze` function imported from react-native-screens.
```
import { enableFreeze } from "react-native-screens";
```
 You don't need to call `enableScreens` for that.

> We do not specify detachInactiveScreens on a stack navigator, then it will always go to undefined and that will make the platform checks take care of the default value, something that we do not want

This is the exact behaviour *we* want in react-navigation. You can customize this behaviour by setting `detachInactiveScreens={false}` on the navigator

> The way I uncovered this behaviour is with the obvious logging in the createStackComponent.

As you didn't provide any repro (😢) I created the simplest one myself

<details>
<summary>`detachInactiveScreens={false}` on Stack navigator</summary>
<br>

```jsx
import { createStackNavigator } from '@react-navigation/stack';

function HomeScreen({ navigation }) {
  return (
      <Text>Home Screen</Text>
      <Button
        title="Go to Details"
        onPress={() => navigation.navigate('Details')}
      />
    </View>
  );
}

function DetailsScreen({ navigation }) {
  return (
      <Text>Details Screen</Text>
      <Button
        title="Go to Details... again"
        onPress={() => navigation.push('Details')}
      />
    </View>
  );
}

function App() {
  return (
    <NavigationContainer>
      <Stack.Navigator detachInactiveScreens={false} initialRouteName="Home">
        <Stack.Screen name="Details" component={DetailsScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

export default App;

```
</details>

I `console.log'd` the `detachInactiveScreens` value in the `CardStack.tsx` and it doesn't seem to get lost or changed between renders.

image

It also seems impossible to lost/change this value between renders as it is set IN the render method.

I think the current behaviour in `react-navigation` is the expected behaviour. You can always use [patch-package](https://www.npmjs.com/package/patch-package) to patch any dependency in your app.

Cheers

mstaicu:
Totally agree with you @kacperkapusciak, and yes, I forgot to include a copy of the offending code, though your example seems to closely replicate my environment. In that case, I suggest we close this pull request as it answers my questions. I would really like to thank you for taking the time in looking into this, both you and @satya164

PR #10633, opened 2022-06-13
Add `Screen` option as `displayName`

nandorojo:
Small internal change to allow a custom `Screen` component with the `displayName` property.

Sorry for the `as any` – it's the only acceptable type that I know of here.

[This sandbox](https://codesandbox.io/s/reverent-night-r14syo?file=/src/App.js) demonstrates the `displayName` feature. The `ReadChildren` component that it loops through its children and lists their display names.

## **Motivation**

While this isn't necessary as a documented API, it will allow `expo-auto-navigation` to inject its own `Screen` component without needing a hacky wrapper function.

## **Test plan**

This doesn't affect any UI, so as long as it passes the TS types, it should be fine.

### DIFF
--- a/packages/core/src/useNavigationBuilder.tsx
+++ b/packages/core/src/useNavigationBuilder.tsx
@@ -77,7 +77,7 @@ const getRouteConfigsFromChildren = <
     ScreenConfigWithParent<State, ScreenOptions, EventMap>[]
   >((acc, child) => {
     if (React.isValidElement(child)) {
-      if (child.type === Screen) {
+      if (child.type === Screen || (child.type as any)?.displayName === 'Screen') {
         // We can only extract the config from `Screen` elements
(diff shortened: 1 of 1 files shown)


nandorojo:
Looks like this may not work, since the expo's navigation may not pass props directly to the outer-most child, but rather to a nested component.

In that case, this would likely require a feature with the option to create a custom `Screen`. I'm not sure how this would work with React's top-down architecture though.

satya164:
When you do `<Screen name="something" component={Something} />`, you're not actually rendering anything. It's purely a syntactic sugar for configuration objects. Wrapping it isn't supported since wrapping an object in another component wouldn't make sense.

For exposing APIs for your libraries and frameworks, I'd recommend using a configuration object instead of using `Screen` component to make it more obvious.

Issue #10609, opened 2022-05-26
useHeaderHeight is wrong on iPad in landscape mode


outaTiME:
### Current behavior

The `useHeaderHeight` doesn't take the size of the status bar on iPad (when in landscape mode), but it does on iPhone. 

Using the `getDefaultHeaderHeight` as follows works correctly on all platforms:

```js
const headerHeight = getDefaultHeaderHeight(
  useSafeAreaFrame(),
  false,
  insets.top
);
```

### Expected behavior

`useHeaderHeight` should take into account the size of the status bar on iPad as it does on other platforms

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [x] @react-navigation/​native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^6.0.10
| @react-navigation/bottom-tabs          | ^6.3.1
| @react-navigation/stack                | ^6.2.1
| @react-navigation/native-stack         | ^6.6.2
| react-native-safe-area-context         | 4.2.4
| react-native-screens                   | ~3.11.1
| react-native-gesture-handler           | ~2.2.1
| react-native-reanimated                | ~2.8.0
| expo                                   | ^45.0.4
| node                                   | 16.14.2
| yarn                            | 3.2.0

outaTiME:
Here I created a PR that solves this problem, I hope to have news about it soon 🙏

https://github.com/react-navigation/react-navigation/pull/10652

PR #10652, opened 2022-06-23
fix: use top inset for header height on ipad

outaTiME:
This fix uses the top inset for iPad header height calculation in the native stack.
There is a misplaced condition that uses `&&` when it should be `||`.
I hope to have this fix in some version of `react-navigation` soon.

Fixes: #10609


satya164 (approved):
Thank you

PR #10674, opened 2022-07-01
fix: explicitly set gestureDirection default value in native-stack

kacperkapusciak:
** Motivation **

This PR cleans up native-stack's new props docs and sets `gestureDirection` default value depending on `presentation` prop.

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -111,22 +111,22 @@ type SceneViewProps = {
   index: number;
   descriptor: NativeStackDescriptor;
   previousDescriptor?: NativeStackDescriptor;
+  nextDescriptor?: NativeStackDescriptor;
   onWillDisappear: () => void;
   onAppear: () => void;
   onDisappear: () => void;
   onDismissed: ScreenProps['onDismissed'];
-  gestureDirectionOverride?: ScreenProps['swipeDirection'];
 };
 
 const SceneView = ({
   descriptor,
   previousDescriptor,
+  nextDescriptor,
   index,
   onWillDisappear,
   onAppear,
   onDisappear,
   onDismissed,
-  gestureDirectionOverride,
 }: SceneViewProps) => {
   const { route, navigation, options, render } = descriptor;
   const {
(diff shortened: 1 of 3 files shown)


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -142,20 +142,24 @@ const SceneView = ({
...
     animation,
     customAnimationOnGesture,
     fullScreenGestureEnabled,
+    gestureDirection,
     presentation = 'card',
   } = options;
 
+  if (gestureDirection === undefined) {
+    gestureDirection = presentation === 'card' ? 'horizontal' : 'vertical';
+  }
```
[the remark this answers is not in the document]
kacperkapusciak: Thanks! Fixed in [358d28c](https://github.com/react-navigation/react-navigation/pull/10674/commits/358d28ca152054ec1b9896157f11db953ef08d1b)

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -332,6 +328,13 @@ function NativeStackViewInner({ state, navigation, descriptors }: Props) {
           ? descriptors[previousKey]
           : undefined;
 
+        // workaround for rn-screens where gestureDirection has to be set on both
+        // current and previous screen - software-mansion/react-native-screens/pull/1509
+        const { gestureDirection } = descriptor.options;
+        if (gestureDirection && previousDescriptor) {
+          previousDescriptor.options.gestureDirection = gestureDirection;
```
satya164, line 335: We can't mutate this since it's expected that a prop updating would trigger a re-render, however, a mutation wouldn't trigger a re-render.

Similar to `previousDescriptor`, you can check `nextDescriptor` and do:

```js
const nextGestureDirection = nextDescriptor?.options.gestureDirection;
const gestureDirection = nextGestureDirection != null ? nextGestureDirection : descriptor.options.gestureDirection;
```

kacperkapusciak: Yep, you're 100% right. Fixed in [07a1a54](https://github.com/react-navigation/react-navigation/pull/10674/commits/07a1a54625cfb8cf456f3da33c0af14ef7683956)

satya164 (approved):

PR #10688, opened 2022-07-11
fix: typo in useFocusEffect with async function error

lipidity:
Unmatched parentheses in the sample code included in the error message when an async function is passed to `useFocusEffect`.

**Motivation**

People who copy/paste from the error message and adapt it to their requirements would have a syntax error.

**Test plan**

Pass an async function to useFocusEffect, check that the parens are balanced in the sample code shown in the error message.


kacperkapusciak (approved):
Thanks!

PR #10693, opened 2022-07-12
fix: add helpful error when using legacy implementation with reanimated v3

kacperkapusciak:
**Motivation**

As Reanimated 3 will not support Reanimated 1 syntax anymore we need to show a helpful message describing the problem. 

The blog post [Announcing Reanimated 3](https://blog.swmansion.com/announcing-reanimated-3-16167428c5f7)

**Test plan**

Change should not have an effect on applications that use Reanimated 1 & 2. 
When using Reanimated 3 with `useLegacyImplementation` enabled there should be an error explaining the problem.

### DIFF
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -84,6 +84,16 @@ function DrawerViewBase({
   // @ts-expect-error: the type definitions are incomplete
   useLegacyImplementation = !Reanimated.isConfigured?.(),
 }: Props) {
+  // Reanimated v3 dropped legacy v1 syntax
+  const legacyImplemenationNotAvailable =
+    Reanimated.interpolateNode === undefined;
+
+  if (useLegacyImplementation && legacyImplemenationNotAvailable) {
+    throw new Error(
+      'Legacy drawer implementation is not available with Reanimated 3 as it no longer includes support for Reanimated 1 legacy API.'
+    );
+  }
+
   const Drawer: React.ComponentType<DrawerProps> = useLegacyImplementation
     ? require('./legacy/Drawer').default
     : require('./modern/Drawer').default;


packages/drawer/src/views/DrawerView.tsx
```diff
@@ -84,6 +84,16 @@ function DrawerViewBase({
   // @ts-expect-error: the type definitions are incomplete
   useLegacyImplementation = !Reanimated.isConfigured?.(),
 }: Props) {
+  // Reanimated v3 dropped legacy v1 syntax
+  const legacyImplemenationNotAvailable =
+    require('react-native-reanimated').interpolateNode === undefined;
+
+  if (useLegacyImplementation && legacyImplemenationNotAvailable) {
+    throw new Error(
+      'Legacy drawer implementation is not available with Reanimated 3 as it no longer includes support for Reanimated 1 legacy API.'
```
satya164, line 93: Let's adjust it a bit
```suggestion
      'The `useLegacyImplementation` prop is not available with Reanimated 3 as it no longer includes support for Reanimated 1 legacy API. Remove the `useLegacyImplementation` prop from `Drawer.Navigator` to be able to use it.'
```

packages/drawer/src/views/DrawerView.tsx
```diff
@@ -84,6 +84,16 @@ function DrawerViewBase({
   // @ts-expect-error: the type definitions are incomplete
   useLegacyImplementation = !Reanimated.isConfigured?.(),
 }: Props) {
+  // Reanimated v3 dropped legacy v1 syntax
+  const legacyImplemenationNotAvailable =
+    require('react-native-reanimated').interpolateNode === undefined;
```
satya164, line 89: I think there's another problem with the check, `interpolateNode` won't exist for people using Reanimated 1 package so we need to somehow check for that, maybe something like `require('react-native-reanimated').useSharedValue === undefined && require('react-native-reanimated').interpolateNode === undefined`

satya164 (approved):

PR #10696, opened 2022-07-14
fix: getComponent dynamic loading component optimization

wkeylin:
**Motivation**
Loading components dynamically on the web with [`loadable-component`](https://loadable-components.com/), the `getComponent` method creates a problem: the rendering of a new page causes the previous page to reload the component

### DIFF
--- a/packages/core/src/SceneView.tsx
+++ b/packages/core/src/SceneView.tsx
@@ -115,9 +115,10 @@ export default function SceneView<
     ]
   );
 
-  const ScreenComponent = screen.getComponent
-    ? screen.getComponent()
-    : screen.component;
+  const ScreenComponent = React.useMemo(
+    () => screen.getComponent?.() ?? screen.component,
+    [screen]
+  );
 
   return (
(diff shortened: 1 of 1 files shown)


kacperkapusciak:
Hi @wkeylin,
could you prepare a simple example demonstrating the effect of your change? 

Steps to reproduce the problem that you have would be incredibly helpful.

Thanks

wkeylin:
> Hi @wkeylin, could you prepare a simple example demonstrating the effect of your change?
> 
> Steps to reproduce the problem that you have would be incredibly helpful.
> 
> Thanks

I create a example: https://codesandbox.io/s/react-native-web-forked-f9v1uz?file=/src/App.js
### Steps:
1. In PageA will console: `pageA useEffect `
2. Go To PageB will console: 
```
pageA useEffect return 
pageA useEffect 
pageA useEffect return 
pageA useEffect 
pageB useEffect 
```
It looks like dynamically loaded components will cause the previously loaded components to be re-mounted

PR #10708, opened 2022-07-20
Fix linking path

ercote:
**Motivation**

At the moment, a linking url without a path (with or without an ending slash) cannot match the linking configuration, as it is ignored.

Consider the following linking configuration:
```
{
  prefixes: ['https://myapp.com/'],
  config: {
    screens: {
      MyScreen: '',
    },
  },
};
```
If the received linking url is equal to `https://myapp.com/` it won't be matching `MyScreen` as the empty path is ignore. To be able to match an empty path, we need to change the path validation (see change in `useLinking.native.tsx`).

The `getStateFromPath` function is already able to match an empty string. Tests have been added to clearly expose the expected behaviour. 

**Test plan**

Use a deeplink url without a path in your app and match that deeplink in your linking configuration using an empty path.

### DIFF
--- a/packages/core/src/__tests__/getStateFromPath.test.tsx
+++ b/packages/core/src/__tests__/getStateFromPath.test.tsx
@@ -893,6 +893,48 @@ it('returns matching screen if path is empty', () => {
   ).toEqual(changePath(state, '/'));
 });
 
+it('returns matching screen if path is only slash', () => {
+  const path = '/';
+  const config = {
+    screens: {
+      Foo: {
+        screens: {
+          Foe: 'foe',
+          Bar: {
+            screens: {
+              Qux: '',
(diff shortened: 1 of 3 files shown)


satya164 (approved):
Thanks for the PR

intergalacticspacehighway:
@satya164 @ercote this change resets the navigation state if the app is redirected from another app. 

IMO root URL ("" or /) deeplink should be handled when app is opened first time, if it's already opened then it should not reset the navigation state. I can do a PR. Let me know if this is expected.

You can check the below video to see the usecase. Navigation resets on redirection from another app.

https://user-images.githubusercontent.com/23293248/192725315-3d563283-6dd8-499c-be6d-ea729145f064.MP4

PR #10721, opened 2022-07-29
BugFix: safely get the route for onIndexChange

jafar-jabr:
**Motivation**

sometimes `onIndexChange` throws exception "can not read property `name` of undefined"

**Test plan**

Not a big changes, it will not break nothing.


kacperkapusciak:
Hey @jafar-jabr :wave:
Thanks for the PR! It looks like the CI is failing with linting errors. Could you fix them?

You can use `yarn lint` to run the linter locally.

Cheers

satya164 (changes requested):
Thanks for the PR. However, we can't merge this as this is a workaround that masks the actual issue: i.e. the `index` is out of range.

Please open an issue with a repro that replicates this issue instead.

PR #10723, opened 2022-07-30
fix: fixed `focused` value in `tabBarIcon`

kyawthura-gg:
**Motivation**

When accessing `focused` in `tabBarIcon`, it returns both `true` and `false`. It should return either true or false depending on focus. This pull request fixes this problem. 

**Test plan**

Add `focused` prop [here](https://github.com/kyawthura-gg/react-navigation/blob/12be3abb3f067c3f3e7741516eda57c542c9dedc/example/src/Screens/BottomTabs.tsx#L25) and try `console.log(focused)`. It will return either `true` or `false`.


satya164:
This is intentionally done so add to not unmount components when focus changes to avoid any potential flicker.

PR #10728, opened 2022-08-02
feat: expose tab press events

zarubond:
Exposing the press event on tab buttons

**Motivation**

Press event can be useful in number of cases, in my case I need the timestamp of the native event occurrence 

**Test plan**
Listen for the event on tabPress

`
<Tab.Navigator
                screenListeners={params => {
                    return {
                        tabPress: event => { /*event will contain data property with original press event*/
}
}...
`

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -25,11 +25,14 @@ export type BottomTabNavigationEventMap = {
   /**
    * Event which fires on tapping on the tab in the tab bar.
    */
-  tabPress: { data: undefined; canPreventDefault: true };
+  tabPress: {
+    data: GestureResponderEvent | React.MouseEvent<HTMLElement, MouseEvent>;
+    canPreventDefault: true;
+  };
   /**
    * Event which fires on long press on the tab in the tab bar.
    */
-  tabLongPress: { data: undefined };
+  tabLongPress: { data: GestureResponderEvent };
 };
 
 export type LabelPosition = 'beside-icon' | 'below-icon';
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -11,6 +11,7 @@ import {
 import React from 'react';
 import {
   Animated,
+  GestureResponderEvent,
   LayoutChangeEvent,
   Platform,
   StyleProp,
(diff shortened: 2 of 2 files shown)


satya164:
Why is `Date.now()` in your handler not enough?

zarubond:
I need a precise time of the user click event on the screen (we use it for performance measuring). Date.now() gives me only time after the event was handled by JS and bubbled through react-navigation which can be hundreds of ms after the user clicks on the screen. To get the precise time I need access to the native event inside MouseEvent with the exact time of user click.

satya164:
Can you clarify why you need the native timestamp?

zarubond:
We need a native timestamp to measure exactly how long it takes from user click to actual mounting of the tab`s content. This helps us to measure consistency and possible improvements in application responsiveness. Any other way like Date.now() will add/remove hundreds of milliseconds to the measuring results which and the end results unusable.

satya164:
I don't get why it will add hundreds of milliseconds. The event is emitted before the navigation action on the press - it should take hundreds of milliseconds for the press callback in JS to be called after the actual press. The component is mounted after the navigation action. So getting timestamp in the event should allow you to measure how long it takes to mount the component.

If something is preventing you from getting the event before the component is mounted, that's something we should fix.

zarubond:
The nativeEvent timestamp is recorded when user clicks on native OS element, then that information has to go to React Native, then over RN bridge, then to RN button callback, at which point it is handled by react-nativigation BottomTabBar, which then emits an event with press information and only after that it gets to the tabPress callback. This process is asynchronous and there is no guarantee when I receive the tab press event callback. The application can be also busy while handling the user event or the OS can be busy with some other task, which prolongs the time from users click on OS element to actual handling of the event in tabProcess.

To get precise measuring of app performance, I rely need the time when user clicked on native OS element, not when it got handled in JS.

satya164:
> then that information has to go to React Native, then over RN bridge, then to RN button callback, at which point it is handled by react-nativigation BottomTabBar

This won't take hundreds of milliseconds

> The application can be also busy while handling the user event or the OS can be busy with some other task, which prolongs the time from users click on OS element to actual handling of the event in tabProcess

This can happen, but if the goal is to measure the time it takes to render the component then including time then it would be misleading to measure.

Anyway, I don't want to include the whole event object in the `data` parameter. It couples us into exposing which pressable element (one example is how it exposes different types for web and native) we use and prevents us from exposing any other data in the `data` parameter in the future. I'd prefer to only expose a timestamp property.

If it's really necessary to get the native press timing, the timestamp has to be exposed for all of the other events that expose a press event (such as drawer press, etc.), alternatively, I'd prefer performance measurement-related stuff to be exposed to the devtools package rather than via the public API.

zarubond:
I think the native event should be exposed in every user event callback, like its done on react components by default. But I understand that you do not want to expose it though data parameter. So I am going to withdraw this PR.  

I have found alternative solution for my particular use-case. Where I use tabBarButton callback where I return Pressable component and hook directly on the onPress there to get the underlying native event.

Thank you very much for the feedback!

PR #10753, opened 2022-08-12
fix: choose correct route config when there are multiple configs for for a screen

### LINKED ISSUE #10748: Linking: Params are not parsed correctly when a screen appears in multiple configs
### Current behavior

I have this config:
```js
const config = {
  screens: {
    Tabs: {
      screens: {
        Sortiment: {
          path: "sortiment",
          screens: {
            Category: {
              path: "k/:category?",
              exact: true
            }
          }
        },
        Home: {
          path: "home",
          screens: {
            Category: "category"
          }
        }
      }
    }
  }
};
```
And I'm trying to parse this path:
```js
const path = "k/categoryParam?x=x";
```
The state object that is returned when I call `getStateFromPath(path, state)` looks like this:
```js
{
    "routes": [
        {
            "name": "Tabs",
            "state": {
                "routes": [
                    {
                        "name": "Sortiment",
                        "state": {
                            "routes": [
                                {
                                    "name": "Category",
                                    "path": "k/categoryParam?x=x",
                                    "params": {
                                        "x": "x"
                                    }
                                }
                            ]
                        }
                    }
                ]
            }
        }
    ]
}
```
Notice how `categoryParam` is not included in the routes params. If you remove the second path config inside the "Home" config, the params are parsed correctly.

I tracked the problem down to this line:
https://github.com/react-navigation/react-navigation/blob/838842e85fffd56a3a64a1f4e969318998217c7a/packages/core/src/getStateFromPath.tsx#L287

In this case there are two configs that can be found here, and the wrong one is used to construct the params.

### Expected behavior

`params` should include the category param and look like this:
```
{
     "category": "categoryParam",
     "x": "x"
}
```

### Reproduction

https://codesandbox.io/s/npm-playground-forked-pnv716?file=/src/index.js

### Platform

- [X] Android
- [X] iOS
- [X] Web
- [X] Windows
- [X] MacOS

### Packages

### Environment

| node                                   | v16.13.2
| yarn                            | 1.22.15

bviebahn:
This change fixes #10748. I am not very familiar with the code and used some trial and error to find a fix that passes all the existing tests as well as the test that I have added. Let me know if this is an appropriate fix or not or if I missed something.


satya164:

Issue #10768, opened 2022-08-18
Drawer Overlay component on iOS affects automated UI testing


vincentjames501:
### Current behavior

When using Drawer Navigator on iOS (with `drawerType !== 'permanent'`), the `Overlay` component is present at all times which makes certain UI testing drivers unhappy and unable to "click" the correct component. You can see the view by using XCode and going to `Debug > View Debugging > Capture View Hierarchy`

<img width="848" alt="image" src="https://user-images.githubusercontent.com/396866/185433515-2990cb6f-2eb0-45c8-b20d-59c949dad3d0.png">

### Expected behavior

This overlay component should only be present when the menu is opening or opened.

### Reproduction

https://expo.dev/@react-navigation/react-navigation-example

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [X] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [ ] @react-navigation/​native-stack

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.11
| @react-navigation/bottom-tabs          | 6.3.2
| @react-navigation/drawer               | 6.4.3
| @react-navigation/stack                | 6.2.2
| react-native-safe-area-context         | 3.3.2
| react-native-screens                   | 3.10.2
| react-native-gesture-handler           | 2.5.0
| react-native-reanimated                | 2.9.1
| react-native                           | 0.67.4
| node                                   | 16.15.0
| yarn                            | 1.22.19

vincentjames501:
There is no reproduction steps necessary, it's reproducible in any iOS drawer component usage with gestures enabled.

PR #10769, opened 2022-08-18
Only show Overlay when gesture is in progress or open for #10768

vincentjames501:

**Motivation**

See #10768

**Test plan**

After this change, the Overlay component is only shown when the gesture is started or drawer is opened. This makes iOS automated testing utilities happy.


satya164:

I took a quick look at the diff and it's not correct to remove the overlay when gesture is not active. The overlay would still need to show up when opening without gesture, or after finishing the gesture until the animation completes. Feel free to send a new PR if this issues are addressed - and please provide a test plan.

PR #10792, opened 2022-08-27
fix: add constraints needed for TypeScript 4.8

SimenB:
**Motivation**

The new TypeScript 4.8 gives the following errors:

```
../../node_modules/@react-navigation/core/lib/typescript/src/getPathFromState.d.ts:5:28 - error TS2344: Type 'ParamList' does not satisfy the constraint '{}'.

5     screens: PathConfigMap<ParamList>;
                             ~~~~~~~~~

  ../../node_modules/@react-navigation/core/lib/typescript/src/getPathFromState.d.ts:3:22
    3 declare type Options<ParamList> = {
                           ~~~~~~~~~
    This type parameter might need an `extends {}` constraint.

../../node_modules/@react-navigation/core/lib/typescript/src/types.d.ts:519:138 - error TS2344: Type 'T' does not satisfy the constraint '{}'.

519     [RouteName in keyof ParamList]?: NonNullable<ParamList[RouteName]> extends NavigatorScreenParams<infer T, any> ? string | PathConfig<T> : string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>;
                                                                                                                                             ~

  ../../node_modules/@react-navigation/core/lib/typescript/src/types.d.ts:519:108
    519     [RouteName in keyof ParamList]?: NonNullable<ParamList[RouteName]> extends NavigatorScreenParams<infer T, any> ? string | PathConfig<T> : string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>;
                                                                                                                   ~
    This type parameter might need an `extends {}` constraint.
```

Adding the suggested constraints fixes the issue.

https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#unconstrained-generics-no-longer-assignable-to

**Test plan**

Build using TypeScript 4.8


SimenB:
#10793 is needed for this to work

satya164 (changes requested):
Thank you for the PR. Looks like there is a syntax error with the change. Can you take a look?

satya164:
@SimenB I don't think prettier update will be enough. There seems to be an error in parsing in tests etc. Then I'm afraid it may be a breaking change.

SimenB:
It's only breaking for your own tooling, not consumers of the TS files (I think).

`eslint-config-satya164` is holding back `@typescript-eslint/parser` from a working versions, but upgrading it introduces a bazillion `babel/no-invalid-this` warning. Perhaps you can update the dependency on `main` yourself?

satya164:
Hello, the PR got closed automatically due to some changes in the main branch. Would you be able to send a new PR?

PR #10804, opened 2022-08-30
feat: allow adjusting numberOfLines prop on DrawerItem

tarikpnr:
**Motivation**

The `Text` component in `DrawerItem` has non configurable prop `numberOfLines` is set to 1. It is not optimal when `DrawerItem` has considerably long text. This pull request adds `numberOfLines` property to the `DrawerItem` props so that it can be configurable. Defaults to 1.


satya164:
Hello, the PR got closed automatically due to some changes in the main branch. Would you be able to send a new PR with link to material design guidelines showing this? Thank you.

PR #10814, opened 2022-09-04
fix: `PanGestureHandlerGestureEvent` is a type and should be exported as such

renchap:
**Motivation**

`PanGestureHandlerGestureEvent` is exported as a value, but is a type. This makes it included in the transpiled published code, which is incorrect and makes `esbuild` error it tries to consume the code (for example with https://microsoft.github.io/rnx-kit/docs/tools/metro-serializer-esbuild).

**Test plan**

There should not be any special test plans here, and the issue only happens when a non-standard RN/Metro setup is used. I tried the fix on my app using the ESBuild serializer mentioned above and it fixed the issue.


satya164 (approved):
Thank you

PR #10817, opened 2022-09-06
refactor!: drop support for key property in navigate

satya164:
BREAKING CHANGE: Previously, you could specify a route `key` to navigate to, e.g.
`navigation.navigate({ key: 'someuniquekey' })`. It's problematic since `key` is an internal
implementation detail and created by the library internally - which makes it weird to use. None
of the other actions support such usage either.

In addition, we have already added a better API (`getId`) which can be used for similar use
cases - and gives users full control since they provide the ID and it's not autogenerated by the
library.

So this change removes the `key` option from the `navigate` action.

### DIFF
--- a/packages/routers/src/CommonActions.tsx
+++ b/packages/routers/src/CommonActions.tsx
@@ -15,21 +15,12 @@ export type Action =
     }
   | {
       type: 'NAVIGATE';
-      payload:
-        | {
-            key: string;
-            name?: undefined;
-            params?: object;
-            path?: string;
-            merge?: boolean;
-          }
-        | {
-            name: string;
-            key?: string;
-            params?: object;
-            path?: string;
-            merge?: boolean;
-          };
+      payload: {
+        name: string;
+        params?: object;
+        path?: string;
+        merge?: boolean;
+      };
       source?: string;
       target?: string;
     }
@@ -50,17 +41,12 @@ export function goBack(): Action {
   return { type: 'GO_BACK' };
 }
 
-export function navigate(
-  options:
-    | { key: string; params?: object; path?: string; merge?: boolean }
-    | {
-        name: string;
-        key?: string;
-        params?: object;
-        path?: string;
-        merge?: boolean;
-      }
-): Action;
+export function navigate(options: {
+  name: string;
+  params?: object;
+  path?: string;
+  merge?: boolean;
+}): Action;
 // eslint-disable-next-line no-redeclare
 export function navigate(name: string, params?: object): Action;
 // eslint-disable-next-line no-redeclare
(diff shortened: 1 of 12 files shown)


packages/routers/src/CommonActions.tsx
```diff
@@ -70,9 +56,9 @@ export function navigate(...args: any): Action {
   } else {
     const payload = args[0] || {};
 
-    if (!payload.hasOwnProperty('key') && !payload.hasOwnProperty('name')) {
+    if (!payload.hasOwnProperty('name')) {
```
kacperkapusciak, line 59: We could maybe provide a helpful error for people upgrading from v6 to v7 in here:

```jsx
    if (!payload.hasOwnProperty('name')) {
      let message = 'You need to specify a name ...';

      if (payload.hasOwnProperty('key')) {
        message = 'Helpful message that key was removed in v7. ' + message;
      }

      throw new Error(message);
    }
```

packages/routers/src/__tests__/CommonActions.test.tsx
```diff
@@ -3,6 +3,6 @@ import * as CommonActions from '../CommonActions';
 it('throws if NAVIGATE is called without key or name', () => {
```
kacperkapusciak, line 3:
```suggestion
it('throws if NAVIGATE is called without name', () => {
```

To avoid accumulating legacy debt we should probably change it here as well

packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -328,7 +329,18 @@ export default class StackView extends React.Component<Props, State> {
...
+          ...state.routes.filter((r) => r.key !== route.key),
+          route,
+        ];
+
+        return CommonActions.reset({
+          ...state,
+          routes,
+          index: routes.length - 1,
+        });
+      });
```
kacperkapusciak, line 343: Previous code `navigation.navigate(route);` worked because the route has a `key` prop and because we're removing navigate by key in this PR you had to kinda reimplement this in place. Smart! But I would be much happier if there would be a unit test covering that if possible 😅

satya164: I don't think it's currently possible to have a unit test for this, most of the logic in stack doesn't have tests right now and adding tests would require a lot of work.

packages/routers/src/CommonActions.tsx
satya164: I didn't bother to add an error message since we don't really throw for other breaking changes so having errors for only a few isn't of much help. Hopefully, the migration guide + TypeScript will help catch these breaking changes.

PR #10820, opened 2022-09-06
fix!: align onReady callback and navigationRef.isReady

satya164:
BREAKING CHANGE: Previously, the `onReady` prop and `navigationRef.isReady()` work slightly
differently. The `onReady` callback fired when `NavigationContainer` finishes mounting and deep links is resolved. The `navigationRef.isReady()` method additionally checks if there are any navigators rendered - which may not be `true` if the user is rendering their navigators conditionally inside a
`NavigationContainer`.

This changes `onReady` to work similar to `navigationRef.isReady()`. The `onReady` callback will now fire only when there are navigators rendered - reflecting the value of `navigationRef.isReady()`.

### DIFF
--- a/example/src/index.tsx
+++ b/example/src/index.tsx
@@ -22,6 +22,7 @@ import {
   StackScreenProps,
 } from '@react-navigation/stack';
 import { createURL } from 'expo-linking';
+import * as SplashScreen from 'expo-splash-screen';
 import * as React from 'react';
 import {
   Dimensions,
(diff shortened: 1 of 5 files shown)


kacperkapusciak (approved):

yairopro:
The  [v6 doc says](https://reactnavigation.org/docs/navigation-container/#onready)
> You can use it for: Making sure that the ref is usable
🤥
Maybe it should warn the opposite..

PR #10824, opened 2022-09-07
chore: update yarn

SimenB:
**Motivation**

Why not 😀  I just ran `yarn set version stable`

**Test plan**

CI


netlify.toml
```diff
@@ -3,4 +3,3 @@
   base    = "/"
   publish = "example/web-build"
   command = "yarn example expo build:web"
-  environment = { YARN_VERSION = "3.2.2" }
```
SimenB, line 6: a global version of yarn exists in the env, so there's no need to define this

satya164:
Hello, the PR got closed automatically due to some changes in the main branch. Would you be able to send a new PR if this is still relevant? Thank you.

PR #10825, opened 2022-09-07
Create LICENSE

AmmarArnt:
**Motivation**

When working on an app and trying to list the 3rd party libraries that we use, give tribute to them and list their licenses, I have identified that the license is missing on the github (only to be found on the npmjs site). Therefore, to make it stand out and easier for other developers, I am proposing this change, so that it matches the license that is stated on the npmjs repository.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 React Navigation
+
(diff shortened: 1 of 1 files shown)


satya164:
We have `LICENSE` files in every package so I don't think this change is necessary

PR #10846, opened 2022-09-16
feat(drawer): add drawer status events

mdneyazahmad:
**Motivation**

Issue: https://github.com/Expensify/App/issues/8292

We have a text input on a screen of a drawer. When the screen gets focus. It focusses the input and this creates an issue in the safari browser. It seems like when animation is ongoing we should not focus.

This PR adds two events `drawerOpened` and `drawerClosed`. It will be fired, when the animation completes. This will be the right time to focus the text input.

**Test plan**

Test with the code below and observe it notifies when the drawer is either opened or closed for both implementation (legacy and modern). Remove `useLegacyImplementation` from the code to test modern implementation. 

I see there are some issues. It fires when the app loads. I will create another PR to fix that issues.

<details>
<summary>Code example for testing</summary>

```javascript
import { Text, View } from 'react-native';
import { createDrawerNavigator } from '@react-navigation/drawer';

function HomeScreen({ navigation }) {
  React.useEffect(() => {
    const listners = [
      navigation.addListener('drawerOpened', () => alert('Drawer opened!')),
      navigation.addListener('drawerClosed', () => alert('Drawer closed!')),
    ];

    return () => {
      listners.forEach((unsubs) => unsubs());
    };
  });
  return (
      <Text>Drawer status events</Text>
    </View>
  );
}

const Drawer = createDrawerNavigator();

  return (
    <NavigationContainer>
      <Drawer.Navigator initialRouteName="Home" useLegacyImplementation>
        <Drawer.Screen name="Home" component={HomeScreen} />
      </Drawer.Navigator>
    </NavigationContainer>
  );
}

```
</details>

https://user-images.githubusercontent.com/77761491/190592387-efeef429-e1f6-40c6-8dec-30ac19ac7c17.mov

### DIFF
--- a/example/types.check.tsx
+++ b/example/types.check.tsx
@@ -121,7 +121,13 @@ export const FeedScreen = ({
   expectTypeOf(navigation.addListener)
     .parameter(0)
     .toEqualTypeOf<
-      'focus' | 'blur' | 'state' | 'beforeRemove' | 'drawerItemPress'
+      | 'focus'
+      | 'blur'
+      | 'state'
+      | 'beforeRemove'
+      | 'drawerItemPress'
+      | 'drawerOpened'
+      | 'drawerClosed'
     >();
 
   expectTypeOf(navigation.getState().type).toEqualTypeOf<'drawer'>();
--- a/packages/drawer/src/types.tsx
+++ b/packages/drawer/src/types.tsx
@@ -243,6 +243,14 @@ export type DrawerNavigationEventMap = {
    * Event which fires on tapping on the item in the drawer menu.
    */
   drawerItemPress: { data: undefined; canPreventDefault: true };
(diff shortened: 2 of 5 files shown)


satya164:
Thanks for the PR. I'd suggest you to use `InteractionManager.runAfterInteractions` to schedule work to happen after animations instead.

satya164:
The two code snippets aren't equivalent. In the snipper using `InteractionManager`, you run it only once the first time screen is mounted. In the second snippet you focus every time the animation finishes - which isn't the same.

If you want to focus the input when the screen receives focus, you need to use `useFocusEffect` instead of `useEffect`. You can also get the drawer's status in the `state` event or `useDrawerStatus` hook.

PR #10878, opened 2022-09-23
fix(core): give slugs a higher priority than wildcard routes

EvanBacon:

**Motivation**

When a linking configuration has both a wildcard route and a slug route, the wildcard is getting a higher priority than the slug making the slug inaccessible. This change sorts the slug with a higher priority than the wildcard routes.

**Test plan**

Given the following linking configuration:

```json
{
    "prefixes": [
        "http://localhost:19000/"
    ],
    "config": {
        "screens": {
            "404": {
                "path": "*"
            },
            "slug": {
                "path": ":slug"
            },
            "index": {
                "path": ""
            }
        }
    }
}
```

- linking to `/foobar` will now show the `slug` screen instead of the `404` screen.
- linking to `/foo/bar` will now show the `404` screen.
- linking to `/` will now show the `index` screen.


satya164 (changes requested):
Let's add some tests for this case

satya164:

PR #10879, opened 2022-09-23
fix(core): use more predictable path segments from state

EvanBacon:
**Motivation**

Provide more expected path segments when routing to screens that all match the same path.

**Test plan**

Given the following linking config (assuming a tab navigator):

```json
{
    "prefixes": [
        "http://localhost:19000/"
    ],
    "config": {
        "screens": {
            "(tab)": {
                "path": "",
                "screens": {
                    "deep": {
                        "path": "*"
                    },
                    "dynamic": {
                        "path": ":dynamic"
                    },
                    "index": {
                        "path": ""
                    }
                }
            }
        }
    }
}
```

- Navigating to `dynamic` will now use the path `/` instead of `/undefined`
- Navigating to `deep` will now use the path `/` instead of `/*`
- Manually routing to `/foo/bar` will take you to `deep`
- Manually routing to `/foobar` will take you to `dynamic`
- Reloading on any of the pages will take you back to `index` as it has the highest priority for `/`.


packages/core/src/getPathFromState.tsx
```diff
@@ -180,19 +180,19 @@ export default function getPathFromState<ParamList extends {}>(
...
           const name = getParamName(p);
 
-          // We don't know what to show for wildcard patterns
-          // Showing the route name seems ok, though whatever we show here will be incorrect
-          // Since the page doesn't actually exist
+          // This can occur when a wildcard matches all routes after a given segment
+          // and the provided path was `/`.
           if (p === '*') {
-            return route.name;
+            return '';
```
satya164, line 186: I think this may be more appropriate:

```suggestion
            return route.path ?? '';
```

packages/core/src/getPathFromState.tsx
```diff
@@ -180,19 +180,19 @@ export default function getPathFromState<ParamList extends {}>(
...
           const name = getParamName(p);
 
-          // We don't know what to show for wildcard patterns
-          // Showing the route name seems ok, though whatever we show here will be incorrect
-          // Since the page doesn't actually exist
+          // This can occur when a wildcard matches all routes after a given segment
+          // and the provided path was `/`.
           if (p === '*') {
-            return route.name;
+            return route.path ?? '';
```
satya164, line 186: Let's leave `route.name` for now
```suggestion
            return route.path ?? route.name;
```

packages/core/src/getPathFromState.tsx
```diff
@@ -180,19 +180,19 @@ export default function getPathFromState<ParamList extends {}>(
...
           }
 
           // If the path has a pattern for a param, put the param in the path
           if (p.startsWith(':')) {
             const value = allParams[name];
 
-            if (value === undefined && p.endsWith('?')) {
-              // Optional params without value assigned in route.params should be ignored
+            if (value == null) {
```
satya164, line 193: We should add some tests for this change

satya164 (changes requested):

satya164:

Issue #10909, opened 2022-10-04
Conflicting peer dependency react-native-reanimated@"3.0.0-rc.3"


hannojg:
### Current behavior

We have a client using `@react-navigation/drawer@6.5.0` and wants to use the latest react-native-reanimated version 3.

In `@react-navigation/drawer@6.5.0` `package.json` it says:

```json
"peerDependencies": {
    "react-native-reanimated": ">= 1.0.0",
```

However, a semver check with `react-native-reanimated@3.0.0-rc.3` fails against `>= 1.0.0"`.

The full npm install error:

```text
➜ npm i
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR! 
npm ERR! While resolving: @react-navigation/drawer@6.5.0
npm ERR! Found: react-native-reanimated@3.0.0-rc.3
npm ERR! node_modules/react-native-reanimated
npm ERR!   react-native-reanimated@"3.0.0-rc.3" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react-native-reanimated@">= 1.0.0" from @react-navigation/drawer@6.5.0
npm ERR! node_modules/@react-navigation/drawer
npm ERR!   @react-navigation/drawer@"6.5.0" from the root project
npm ERR! 
npm ERR! Conflicting peer dependency: react-native-reanimated@2.10.0
npm ERR! node_modules/react-native-reanimated
npm ERR!   peer react-native-reanimated@">= 1.0.0" from @react-navigation/drawer@6.5.0
npm ERR!   node_modules/@react-navigation/drawer
npm ERR!     @react-navigation/drawer@"6.5.0" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
```

Not an issue with using yarn though.

### Expected behavior

Versions are compatible. One fix would be to add in the meantime this fix:

```json
"peerDependencies": {
    "react-native-reanimated": ">= 1.0.0 || >= 3.0.0-rc",
```

### Reproduction

https://github.com/hannojg/react-navigation-rea-v3-rc-install-repro

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [X] @react-navigation/​drawer
[...]

PR #10910, opened 2022-10-04
fix: #10909 reanimated v3 not installable

hannojg:
**Motivation**

When trying to install the `rc` version of reanimated v3 there is an error as described in #10909

**Test plan**

Installation testing works


kacperkapusciak:
Closing as explained in https://github.com/react-navigation/react-navigation/issues/10909#issuecomment-1266955497

PR #10902, opened 2022-10-03
fix(native-stack): headerTransparent option for custom header

hukpo:
NativeStack doesn't support the custom floating header because headerTransparent wasn't used.

This PR fixes #10822

PR #10919, opened 2022-10-09
fix: fix headerTransparent with custom header in native-stack

### LINKED ISSUE #10822: Transparent not working in custom header
### Current behavior

``` js
const HeaderMain = ({navigation, options, route, back}) => {
	
	return (
		<View style={[styles.container, {
			backgroundColor: "red",//DefaultTheme.colors.background,
			position: 'absolute',
			top: 0,
			left: 0,
			right: 0,
			zIndex: 10000,
			width: "100%",
			height: 60
		}]}

		>
			{back ? 
				<TouchableOpacity
					onPress={() => navigation.goBack()}
				>
					<Icon name='arrow-forward-outline' size={25} color="#000" />
				</TouchableOpacity>
			: null}
			<Text numberOfLines={1} style={styles.headerText}>{options.title ? options.title : /*route.name ? route.name :*/ ''}</Text>
		</View>
	);
}
```
When i add position: 'absolute' to styles of container, Header disappears!

### Expected behavior

I want to header is transparent but this not and just disappears.

### Reproduction

0

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/​native-stack

### Environment

| @react-navigation/native               |6.0.12
| @react-navigation/native-stack         |6.8.0
| react-native-safe-area-context         |4.3.3
| react-native-screens                   |3.17.0
| react-native                           |0.69.5

satya164:
Test plan:

Tested by rendering a dummy `View` for the header:

```js
options={{
  header: () => (
    <View
      style={{
        backgroundColor: 'rgba(0, 0, 0, 0.5)',
        height: 56,
      }}
    />
  ),
  headerTransparent: true,
}}
```

Fixes #10822
Closes #10902

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -147,6 +147,7 @@ const SceneView = ({
     header,
     headerBackButtonMenuEnabled,
     headerShown,
+    headerTransparent,
     autoHideHomeIndicator,
     navigationBarColor,
     navigationBarHidden,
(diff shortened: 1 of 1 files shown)


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -433,4 +441,14 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
   },
+  scene: {
+    flex: 1,
+    flexDirection: 'column-reverse',
```
kacperkapusciak, line 446: Why don't you just reorder components in JSX instead? Then you could reuse the `container` style too

satya164: @kacperkapusciak reordering that way works differently - i.e. it reflects the order of elements in the view hierarchy. in the view hierarchy, the element that comes later is positioned on top of the previous one. so here we place the custom header after the screen in JSX so that it's on top of the content (especially when absolutely positioned) - so that the touches on the header aren't blocked by the content.

a similar effect can be achieved by using `zIndex` style which does this reordering in the view hierarchy afaik, but often it doesn't seem to work properly (especially on Android) so we use this method. the same approach is also used in all other navigators which render a header.

kacperkapusciak (approved):

PR #10920, opened 2022-10-10
fix: issue 10915

### LINKED ISSUE #10915: Clicking web browser back -> forward breaks navigation
### Current behavior

This bug was initially reported in https://github.com/react-navigation/react-navigation/issues/9128, but was closed with a partial/incomplete fix.

I think I have found the root cause with @dan-doyon-endear which I will outline below.

@mrfr0g already did great work putting together repro steps so I'll paste his directions & repo here.

Steps:

- Click button which navigates to a route with a parameter, { id: 123 }
- Click the browser's back button
- Click the browser's forward button
- Click the browser's back button

A console error is raised and the navigation is broken.

178563479-8238cdc0-1c64-4f8b-a332-d23b51e70ed3

### Expected behavior

The web browser forward & back buttons should behave the same as navigating forward/back from inside of the application.

### Cause

There are two different bugs at play:

#### Bug 1

The first issue, which is what we observe in the console error, is that this code branch is invalid and should never be reached

https://github.com/react-navigation/react-navigation/blob/ff8d8a31d83f76129be5a8ee0a95ef2919eb1240/packages/native/src/useLinking.tsx#L271-L273

Even though the type hints suggest that `state` can be `unknown`, this case is not actually handled in the base router, which assumes that the payload is a valid state object.

https://github.com/react-navigation/react-navigation/blob/ff8d8a31d83f76129be5a8ee0a95ef2919eb1240/packages/routers/src/BaseRouter.tsx#L38-L39

This should really be fixed, but it's only a symptom of the root cause and not the main bug at play here.

#### Bug 2

The main bug is located in the [createMemoryHistory.tsx](https://github.com/endearhealth/react-navigation/blob/main/packages/native/src/createMemoryHistory.tsx) file.

If you read through that file, it uses an internal `index` variable to keep the current position in the history stack. What *should* happen is that going back a page decrements the index, and going forward increments the index, etc. This works fine for internal app navigation, because internal nav uses the function ``history.go(n: number) {...}`` which updates the index value as expected. But, if you click the browser forward/back buttons, a different code path is triggered from an event listener and it never updates the index value:

https://github.com/endearhealth/react-navigation/blob/53f0ae78e3948f1ee4a43ac1cba324c6af1493f8/packages/native/src/useLinking.tsx#L218-L226

This was kind of half-acknowledged when the `history.index` property was added. It fixed cases where the code was checking `history.index` externally to get the index, but importantly it didn't fix any of the other `history` methods which are still relying on the internal `index` variable:

https://github.com/endearhealth/react-navigation/blob/53f0ae78e3948f1ee4a43ac1cba324c6af1493f8/packages/native/src/createMemoryHistory.tsx#L32-L44

When the forward/back browser nav buttons are clicked, a callback is triggered that eventually ends up running `history.replace({ path, state })` with the new screen. But because the index is incorrect when this function is called, you end up with a corrupted history stack that will break in unexpected ways (see bug 1)

The fix is straightforward: recalculate the history index inside of the `history.replace()` method, the same way that the `history.index` property is doing elsewhere.

```javascript
replace({ path, state }: { path: string; state: NavigationState }) {
      interrupt();

      // Re-calculate the *true* index here instead of relying on the existing index to be correct
      let index = items.findIndex(item => item.id === id)

      if (!items.length || index < 0) {
        items = [{ path, state, id }];
        index = 0;
      } else {
        items[index] = { path, state, id };
      }

      window.history.replaceState({ id }, '', path);
    },
```

There are probably other similar bugs here. The internal `history.index` variable really should be removed completely. But patching `history.replace()` is the enough to fix the issue at hand for us.

### Reproduction

https://github.com/mrfr0g/stack-navigation-web-back-forward-bug

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

### Environment

| @react-navigation/native               | 6.0.11
| @react-navigation/bottom-tabs          | 6.3.3
| @react-navigation/material-top-tabs    | 6.2.3
| @react-navigation/native-stack         | 6.8.0
| react-native-safe-area-context         | 4.3.1
| react-native-screens                   | 3.15.0
| react-native-gesture-handler           | 2.5.0
| react-native-reanimated                | 2.9.1
| react-native-tab-view                  | 3.1.1
| react-native-pager-view                | 5.4.24
| react-native                           | 0.69.5
| expo                                   | 46.0.9

ilteoood:

**Motivation**

Closes #10915

**Test plan**

Reproduction explained inside the issue

cc: @simoneb


simoneb:
anybody around to allow the workflow run and review the PR?

crutch12:
Bump

satya164:
Hello, the PR got closed automatically due to some changes in the main branch. Would you be able to send a new PR with proper description and test plan? Thank you.

robbporto:
Not sure why this was closed. This is an important bugfix.

Can we open it again?

PR #10928, opened 2022-10-14
[Drawer] refactor: use GestrueDetector to fix onStart bug 

### LINKED ISSUE #10927: [Drawer] High `swipeEdgeWidth` value causes keyboard to gets constantly closed
### Current behavior

When using `Drawer` with e.g. `swipeEdgeWidth: 500` pressing inside a `TextInput` is closing the Keyboard (instead of keeping the focus inside the `TextInput`

https://user-images.githubusercontent.com/16821682/195829963-b492aea4-d0a1-4564-ae02-9fa964e0448a.mov

### Expected behavior

- I expect the Keyboard not to close when pressing again inside the `TextInput`
- The problem is with `react-native-gesture-handler` calling `onStart` before the activation criteria is met https://github.com/react-navigation/react-navigation/blob/a2e5f45c0ec603bb276c2e27595b4137fa3a85eb/packages/drawer/src/views/modern/Drawer.tsx#L183-L188 
- The issue on RNGH is (kind of) documented here: https://github.com/software-mansion/react-native-gesture-handler/issues/1225
- The best solution imho for this is to migrate to the new `GestureDetector`, which has a `onBegin` and a `onStart`, where `onStart` is really only called when the activation criteria is met.
  - PR for this change: https://github.com/react-navigation/react-navigation/pull/10928 
- Another solution might be to move the `onGestureStart` to `onActive`
  - PR, if you prefer this proposal over the other: https://github.com/react-navigation/react-navigation/pull/10929

### Reproduction

https://github.com/hannojg/react-navigation-swipeEdgeWidth-reproduction

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/​drawer

### Environment

| @react-navigation/native               | 6.0.13
| @react-navigation/drawer               | 6.5.0
| @react-navigation/stack                | 6.3.2
| @react-navigation/native-stack         | 6.9.1
| react-native-safe-area-context         | 4.4.1
| react-native-screens                   | 3.18.2
| react-native-gesture-handler           | 2.7.1
| react-native-reanimated                | 2.11.0
| react-native                           | 0.70.3
| node                                   | 16.15.1
| npm or yarn                            | 1.22.19

hannojg:
**Motivation**

In #10927 I showed how `onStart` gets called although the activation criteria isn't met. This causes issues.
This PR fixes the issue by migrating to the new `GestureDetector` API which has a separate `onBegin` and `onStart` method, where `onStart` will only be called if the activation criteria is really met (see here https://github.com/software-mansion/react-native-gesture-handler/issues/1225) 

Fixes #10927 

**Test plan**

- Set a `console.log` in `onStart` in `Drawer.tsx`
- In `example/src/index.tsx` add to the Drawer `screenOptions.swipeEdgeWidth: 500,`
- Now press anywhere in the example app 
 - Without these changes you will read `onStart`, although the user just taped and the activation criteria wasn't met
 - With these changes this won't happen, only when the user gesture met the criteria


satya164 (changes requested):
Thanks for the PR! However, this can't be merge to current stable version since it's a breaking change - as `GestureDetector` is only available from gesture handler 2.x.

Can you target the 7.x branch with PR instead? In addition, we need to bump the minimum peer dependency version on gesture handler across all packages in the repo. For drawer, we also need to do this change for the legacy implementation as well.

hannojg:
Hey @satya164, i just:

- Changed the base to `7.x` and rebased the branch onto it
- Updated the peerDependency of RNGH to be `>= 2.0.0`
- I also wanted to update the legacy implementation, but as mentioned [here](https://github.com/react-navigation/react-navigation/pull/10929#issuecomment-1280373158), the legacy implementation is already working correctly

packages/drawer/src/views/modern/Drawer.tsx
```diff
@@ -317,14 +321,7 @@ export default function Drawer({
 
   return (
     <DrawerProgressContext.Provider value={progress}>
-      <PanGestureHandler
-        activeOffsetX={[-SWIPE_DISTANCE_MINIMUM, SWIPE_DISTANCE_MINIMUM]}
-        failOffsetY={[-SWIPE_DISTANCE_MINIMUM, SWIPE_DISTANCE_MINIMUM]}
```
kacperkapusciak, line 322: I think you missed this prop during the migration

packages/drawer/src/views/modern/Drawer.tsx
```diff
@@ -207,11 +207,15 @@ export default function Drawer({
...
+    })
+    .onFinalize(() => {
       runOnJS(onGestureFinish)();
-    },
-  });
+    });
+  // @ts-expect-error Ref types incompatible yet, needs to get fixed upstream
+  panGesture.config = {
+    ...panGesture.config,
+    ...gestureHandlerProps,
```
kacperkapusciak, line 217: Hmm, this raises some more problems.

Currently, [`gestureHandlerProps`](https://reactnavigation.org/docs/drawer-navigator/#gesturehandlerprops) allows users to customise `PanGestureHandler` simply by passing props to it (with `React.ComponentProps<typeof PanGestureHandler>` signature).

After some consultation with @j-piasecki we've come to conclusion there's no guarantee whatsoever that props coming from v1 are going to work with RNGH v2. Or maybe the other way around - there are for sure going to be more props in RNGH 2.

We would have to adjust the API on how users can customize gestures in React Navigation v7.

The solution @j-piasecki and I propose is to drop the `gestureHandlerProps` prop in favor of the `panGesture` prop. Users would be able to pass `Gesture.Pan()` with options the user wants to add or override. This then would have to be merged right in this selected line.

kacperkapusciak (changes requested):

packages/drawer/src/views/modern/Drawer.tsx
hannojg: oh yeah def, that sounds like a clean API. Shall I make that change in the props, or will you do at some point in the `7.x` branch?

osdnk:
@hannojg Thanks. I think we would like to proceed with that PR. Is there a chance you can update the branch and make it up-to-date with the main (which serves as 7.x)?

satya164:
Closing in favor of https://github.com/react-navigation/react-navigation/pull/11776

PR #10929, opened 2022-10-14
fix: only call onGestureStart when criteria met in drawer

### LINKED ISSUE #10927: [Drawer] High `swipeEdgeWidth` value causes keyboard to gets constantly closed

hannojg:

**Motivation**

Fixes #10927.

**NOTE:** I already put up a PR for the issue https://github.com/react-navigation/react-navigation/pull/10928. This PR is just in case you don't like the proposal I made there.

**Test plan**

See other PR for test plan


satya164 (changes requested):
Thanks for the PR. Can you make the same change in legacy implementation as well?

hannojg:
@satya164 thx for reviewing this. I noticed that in the legacy implementation it already works correctly.
 `Keyboard.dismiss` and `toggleStatusBar` are only called when the activation criteria is met:


https://user-images.githubusercontent.com/16821682/196108081-80d154e2-68fb-48b3-aeec-f6c6b963c134.mov

kacperkapusciak (approved):

PR #10936, opened 2022-10-16
refactor!: improve the API for Link component

satya164:
Initially the `Link` component was designed to work with path strings via the `to` prop. But it has few issues:

- The path strings are not type-safe, making it easy to cause typos and bugs after refactor
- The API made navigating via screen name more incovenient, even if that's the preferred approach

This revamps the API of the `Link` component to make it easier to use. Instead of `to` prop, it now accepts `screen` and `params` props, as well as an optional `href` prop to use instead of the generated path.

e.g.:

```js
<Link screen="Details" params={{ foo: 42 }}>Go to Details</Link>
```

This also drops the `useLinkTo` hook and consolidates into the `useLinkBuilder` hook - which lets us build a `href` for a screen or action for a path.


packages/native/src/useLinkBuilder.tsx
```diff
@@ -78,5 +81,35 @@ export default function useLinkBuilder() {
...
+        : getStateFromPath(href, options?.config);
+
+      if (state) {
+        const action = getActionFromState(state, options?.config);
+
+        if (action !== undefined) {
+          return action;
+        } else {
+          return CommonActions.reset(state);
+        }
```
kacperkapusciak, line 103:
```suggestion
        return action !== undefined ? action : CommonActions.reset(state);
```
This one could be packed into one-liner

packages/native/src/Link.tsx
```diff
@@ -1,33 +1,34 @@
...
+      e: React.MouseEvent<HTMLAnchorElement, MouseEvent> | GestureResponderEvent
+    ) => void;
+  } & (TextProps & { children: React.ReactNode });
 
 /**
  * Component to render link to another screen using a path.
  * Uses an anchor tag on the web.
  *
- * @param props.to Absolute path to screen (e.g. `/feeds/hot`).
+ * @param props.to Screen to navigate to (e.g. { screen: 'Feeds', params: { sort: 'hot' } }).
```
kacperkapusciak, line 18: JSDoc here and in `useLinkProps` needs to be updated

packages/native/src/useLinkProps.tsx
```diff
@@ -93,33 +109,33 @@ export default function useLinkProps<
           );
         }
       } else {
-        linkTo(to);
+        // @ts-expect-error: This is already type-checked by the prop types
```
kacperkapusciak, line 112: I don't like this expect error. Could we make it work somehow?

packages/bottom-tabs/src/views/BottomTabItem.tsx
```diff
@@ -120,28 +120,29 @@ type Props = {
 };
 
 export default function BottomTabBarItem({
```
kacperkapusciak, line 122: As we are in here: please rename `BottomTabBarItem` -> `BottomTabItem` to make it consistent with the rest of the codebase. In code, it's used as `BottomTabItem` same as the name of the file.

packages/bottom-tabs/src/views/BottomTabItem.tsx
```diff
@@ -120,28 +120,29 @@ type Props = {
...
       // We need to use `onClick` to be able to prevent default browser handling of links.
       return (
         <Link
           {...rest}
-          to={to}
+          href={href}
+          action={CommonActions.navigate(route.name, route.params)}
           style={[styles.button, style]}
           onPress={(e: any) => {
             if (
```
kacperkapusciak, line 148: Not to this PR but a general question - Why this and `LinkPressable` component in the DrawerItem isn't reused and extracted to the `elements` package but copied over instead?

packages/drawer/src/views/DrawerItem.tsx
```diff
@@ -84,29 +88,32 @@ type Props = {
...
-  if (Platform.OS === 'web' && to) {
+  if (Platform.OS === 'web') {
     // React Native Web doesn't forward `onClick` if we use `TouchableWithoutFeedback`.
     // We need to use `onClick` to be able to prevent default browser handling of links.
     return (
       <Link
         {...rest}
-        to={to}
+        href={href}
+        action={CommonActions.navigate(route.name, route.params)}
```
kacperkapusciak, line 116: If require `route` here we could build `buildHref(route.name, route.params)` right here as well in place rather than prop drilling it a few levels

PR #10967, opened 2022-11-01
fix: appending text from the drawer when coping the last text in the screen using a triple click in Chrome

osama256:
When selecting the last text in the screen using a triple click in Chrome browser, the text inside the drawer got selected as well.

**Motivation**

Selecting the last text in the screen using triple click should work fine in all browsers.

this issue wasn't exist in previous versions but later we removed something belong to Chrome debugger which made this issue.

**Test plan**

https://user-images.githubusercontent.com/51829206/199290813-60ad2ded-47f7-4b54-9231-c710ab457283.mp4

### DIFF
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -80,9 +80,12 @@ function DrawerViewBase({
   detachInactiveScreens = Platform.OS === 'web' ||
     Platform.OS === 'android' ||
     Platform.OS === 'ios',
-  // Reanimated 2 is not configured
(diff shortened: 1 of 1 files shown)


satya164:
`global.nativeCallSyncHook` is for chrome debugger which is entirely unrelated to running the app in the browser.

this change is incorrect, when running the app in the browser, `global.nativeCallSyncHook` will always be `undefined` resulting in the logic resulting in `true`, thereby always forcing legacy implementation.

If you need to use the legacy implementation, then pass `useLegacyImplementation={true}` to your drawer explicitly.

PR #10984, opened 2022-11-08
feat: expose tabBarGap option in material top tabs

mlecoq:
**Motivation**

the new `tabBarGap` option helps to configure `gap` option of `react-native-tab-view` : see https://github.com/satya164/react-native-tab-view#gap

**Test plan**

Just set a tabBarGap option and check that there is a gap between tabs


satya164:

Issue #10845, opened 2022-09-15
[iOS/native-stack] Header shadow not shown if header transparent is true


alamothe:
### Current behavior

On iOS by default, header shadow is shown, both in iOS apps and in `react-navigation`. 

However, in `react-navigation` when we set that header is transparent:
```tsx
    <Stack.Navigator screenOptions={{
		  headerTransparent: true,
		  headerBlurEffect: "systemChromeMaterialLight",
    }}>
```
Suddenly header shadow is not shown anymore. This is inconsistent with native iOS apps, e.g. iMessage.

### Expected behavior

Whether header shadow is shown or not, should not depend on `headerTransparent` property.

It should be controlled by `headerShadowVisible` property, like this library intended.

### Reproduction

https://snack.expo.dev/@alamothe/no-header-shadow

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [X] @react-navigation/​native-stack

### Environment

- [X] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.12
| @react-navigation/native-stack         | 6.8.0
| react-native-safe-area-context         | 4.2.4
| react-native-screens                   | 3.11.1
| react-native-gesture-handler           | 2.2.1
| react-native-reanimated                | 2.8.0
| react-native                           | 0.68.2
| expo                                   | 45.0.8
| node                                   | v16.17.0
| npm or yarn                            | 8.15.0

maging-studio:
If you set `headerTransparent` to `true` and set the background color close to transparent, for example `rgba(255,255,255,0.1`), then the bottom shadow will be displayed correctly. 

This config fixes the issue
```
screenOptions={{
    headerShadowVisible: true,
    headerBlurEffect: 'regular',
    headerStyle: {
        backgroundColor:
            Platform.OS === 'ios' //
                ? 'rgba(255,255,255,0.1)'
                : 'white',
    },

    headerLargeTitle: true,
    headerLargeTitleShadowVisible: false,
    headerLargeStyle: {
        backgroundColor: colors.grayBgColor,
    },
}}
```

[...]

PR #10993, opened 2022-11-10
HeaderConfig: Updates hideShadow to allow native iOS behavior to be applied

objectivecosta:
**Motivation**

As reported in [#10845](https://github.com/react-navigation/react-navigation/issues/10845), the behavior that native software in the iOS ecosystem applies is that even transparent headers _can_ have a border in the NavigationBar/TopBar. In React Navigation, it was assumed that `headerTransparent=true` would also imply `hideShadow=true`, which would prevent this behavior (and therefore, make the behavior of this library inconsistent with native apps)

This fix allows the behavior to continue on Android, while allowing the native iOS behavior to be present (also providing the option to hide it forcefully).

Issue here: https://github.com/react-navigation/react-navigation/issues/10845

**Test plan**

- Configure a NativeStack to have a transparent header on iOS
- Observe that this does not affect the `hideShadow` property on iOS
- Observe that behavior on Android is unchanged

| Bug  | Expected |
:-------------------------:|:-------------------------:
  |  

The change must pass lint, typescript and tests: ✅

### DIFF
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -180,7 +180,8 @@ export default function HeaderConfig({
         hideShadow={
           headerShadowVisible === false ||
           headerBackground != null ||
-          headerTransparent
+          // On iOS, even transparent headers have a border (called shadow in this property)
+          (Platform.OS !== 'ios' && headerTransparent) 
         }
         largeTitle={headerLargeTitle}
         largeTitleBackgroundColor={headerLargeStyleFlattened.backgroundColor}
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -181,7 +181,7 @@ export default function HeaderConfig({
           headerShadowVisible === false ||
           headerBackground != null ||
           // On iOS, even transparent headers have a border (called shadow in this property)
-          (Platform.OS !== 'ios' && headerTransparent) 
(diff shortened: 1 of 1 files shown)


objectivecosta:
I don't know how to request a review here, but this one should be straightforward...

pinging @satya164 (the one I see approving most PRs here)

objectivecosta:
(Sorry if this is not appropriate, newbie to OSS contributing)

objectivecosta:
@satya164 Fair enough! Question: given that the behavior should be consistent between platforms, shouldn't the variable apply **a shadow**, and not a border? Having the shadow property not apply a shadow seems confusing callers.

Would you be up for a property that allows border and shadow to be differentiated (maybe in a different PR, after this one?)

objectivecosta:
@satya164 Did the changes on this one, will try and find time for another one separating shadow and border properly (so it doesn't look like iOS applying border is a workaround) if you agree to it ;)

satya164:
I think it's fine, a shadow can look like a border depending on the height. The underlying platform calls it a shadow as it was a bigger shadow at some point - so it really depends on iOS version.

satya164 (approved):

objectivecosta:
@satya164 True that! Pre-iOS 6 it was indeed a proper bigger shadow. Good point!

objectivecosta:
Thanks for the quick review and the good observations :D

objectivecosta:
@satya164 I'm currently working off of a specific commit for the project that I have that uses this, so I was wondering whether this will be inserted into the next release slot, and when that is...

satya164:
It'll be soon

PR #10996, opened 2022-11-14
feat: extend and modified navigation bottom tabbar position on user screen

josemak25:
feat: extend and modified navigation bottom tabbar position on user screen

**Motivation**

This PR extends the bottom navigation with some dynamic styles which helps position the bottom navigation in 3 ways which is left | right | bottom largely controlled by `tabBarPosition` prop passed the bottom navigation.

**Test plan**

Tested by rendering a bottom tabs `BottomTabs.Navigator`.

```js
const BottomTabs = createBottomTabNavigator<BottomTabParams>();

<BottomTabs.Navigator tabBarPosition="left">
```
Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-14 at 17 16 18
Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-14 at 17 16 07
Screenshot 2022-11-14 at 5 17 01 PM
Screenshot 2022-11-14 at 4 21 25 PM
Screenshot 2022-11-14 at 5 16 52 PM
Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-14 at 16 21 27


packages/bottom-tabs/src/types.tsx
```diff
@@ -288,6 +290,11 @@ export type BottomTabNavigationConfig = {
    * Style object for the component wrapping the screen content.
    */
   sceneContainerStyle?: StyleProp<ViewStyle>;
+
+  /**
+   * Bottom tab-bar position on the window.
+   */
+  tabBarPosition?: TabBarPosition;
```
satya164, line 297: This needs to be an `option` and not a prop. Let's it to `BottomTabNavigationOptions`

packages/bottom-tabs/src/types.tsx
```diff
@@ -326,3 +334,13 @@ export type BottomTabBarButtonProps = Omit<
...
+
+export type TabBarPositionMapConfig = Record<
+  TabBarPosition,
+  Record<'tabBar' | 'content' | 'tabBarItem', StyleProp<ViewStyle>>
+>;
+
+export type TabBarContainerPositionMapConfig = Record<
+  TabBarPosition,
+  StyleProp<ViewStyle>
+>;
```
satya164, line 346: These types are very specific to the logic inside the tab bar, so they should be located where they are used

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
   const tabBarBackgroundElement = tabBarBackground?.();
 
+  const tabBarPositionMap: TabBarPositionMapConfig = {
+    right: {
+      tabBarItem: {
+        height:
+          Platform.OS === ('ios' || 'android')
+            ? tabBarHeight / 1.5
+            : DEFAULT_TABBAR_HEIGHT,
+        marginTop: COMPACT_TABBAR_HEIGHT / 2,
```
satya164, line 275: The height of tab items is unrelated to the tab bar height. They shouldn't be calculated based on what the tab bar height is. Ideally, the height of each item would be dynamic based on the size of the icon and label and not hard coded.

The margins should be hard-coded to a constant which is not dependent on tab bar height. I'd prefer them to be `marginVertical` instead of `marginTop` so they are spread between each item.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+    tabBarPosition,
   });
 
   const tabBarBackgroundElement = tabBarBackground?.();
 
+  const tabBarPositionMap: TabBarPositionMapConfig = {
+    right: {
+      tabBarItem: {
+        height:
+          Platform.OS === ('ios' || 'android')
```
satya164, line 272: This equality check is incorrect. `'ios' || 'android'` will always evaluate to `'ios'`, so the check is `Platform.OS === 'ios'`. Why is this condition based on iOS or not?

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+            ? tabBarHeight / 1.5
+            : DEFAULT_TABBAR_HEIGHT,
+        marginTop: COMPACT_TABBAR_HEIGHT / 2,
+        flex: Platform.OS === ('ios' || 'android') ? 0 : 0.07,
```
satya164, line 276: Where's `0.07` coming from?

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+      },
+      tabBar: {
+        paddingHorizontal: 5,
+        borderLeftColor: colors.border,
+        paddingTop: insets.top + tabBarHeight,
```
satya164, line 281: Why does the `tabBar` need a top padding equal to its height?

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+      },
+      content: {
+        flex: 1,
+        width: tabBarHeight,
+        maxWidth:
+          Platform.OS === ('ios' || 'android')
+            ? DEFAULT_TABBAR_HEIGHT
+            : DEFAULT_TABBAR_HEIGHT * 1.3,
+      },
+    },
```
satya164, line 317: This code is almost the same as the `right` style apart from the `borderRightColor` and `borderRightWidth`. We should define those 2 conditionally instead of duplicating the whole style.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+        paddingTop: insets.top + tabBarHeight,
+        borderRightWidth: StyleSheet.hairlineWidth,
+      },
+      content: {
+        flex: 1,
+        width: tabBarHeight,
+        maxWidth:
+          Platform.OS === ('ios' || 'android')
+            ? DEFAULT_TABBAR_HEIGHT
+            : DEFAULT_TABBAR_HEIGHT * 1.3,
```
satya164, line 315: Why are both `width` and `maxWidth` needed? Also, where does `1.3` come from?

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -249,18 +260,87 @@ export default function BottomTabBar({
...
+
   return (
     <Animated.View
       style={[
-        styles.tabBar,
+        tabBarPositionMap[tabBarPosition].tabBar,
         {
+          left: 0,
+          right: 0,
+          bottom: 0,
```
satya164, line 341: The code conventions in this repo are to keep static styles in `StyleSheet` and only define dynamic styles inside the component. Let's keep that convention.

satya164 (changes requested):

packages/bottom-tabs/src/types.tsx
josemak25: Alright

packages/bottom-tabs/src/types.tsx
josemak25: Got you

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: Removing the height and using flex makes the tabBarItem looks like this, this is because the tabBarItem has no height and its element height is undefined

```js
   tabBarItem: {
        flex: 0,
        borderWidth: 1,
        marginVertical: COMPACT_TABITEM_GAP,
      },
```

Screenshot 2022-11-16 at 10 19 11 AM

```js
   tabBarItem: {
        flex: 1,
        borderWidth: 1,
        marginVertical: COMPACT_TABITEM_GAP,
      },
```

Screenshot 2022-11-16 at 10 18 58 AM

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: This has been removed from the previous comment od using dynamic height

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: This was a mistake I added it to help calculate the issue we have with getting the screen header height 

This should be 
```js        
paddingTop: insets.top + headerHeight, // but since we can't access the headerHeight from bottom tabs we just use only the inset
```

packages/bottom-tabs/src/views/BottomTabBar.tsx
satya164: > Removing the height and using flex makes the tabBarItem looks like this

It shouldn't be using `flex`. The main point is that the height should be based on what's inside the item without hardcoding, or if it's hardcoded to a specific height then it shouldn't be based on the height of tab bar at bottom. It needs to be its own thing based on what looks good in horizontal tab bar.

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: the tabBar uses the calculated tabBarHeight as its width.
the reason for adding a minor restriction to its width at the max `DEFAULT_TABBAR_HEIGHT` which is 49 is to prevent the tabBar from extending further than that.
this is because if the width grows more than that, it takes a very large part of the screen which makes the content screen small and not so presentable.
Also, the calculation for the 1.3 for web etc was to increase the maxWidth for web since there screens are way bigger but I can make this a constant for better understanding rather than multiplying a random number

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: got you

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: Removing the flex property still returns the same result as flex: 0

I totally understand why we shouldn't user-defined heights cause when I increase the size of the icon with a defined height on the tabBarItem it seems to cut the tabBarItem text out.

Just not sure how to approach this problem.
But I am thinking of using layouts to calculate the children (icon and text) passed to the tabBarItem to get it's height and use that as the tabBarItemHeight this way we have dynamic height.
What do you think of this approach @satya164 ?

packages/bottom-tabs/src/views/BottomTabBar.tsx
josemak25: Got you, this has been fixed with a different approach

packages/bottom-tabs/src/views/BottomTabBar.tsx
satya164: > But I am thinking of using layouts to calculate the children (icon and text) passed to the tabBarItem to get it's height and use that as the tabBarItemHeight this way we have dynamic height.

Calculating layouts is asynchronous so it won't produce good results. As I wrote, if it can't be dynamic and is hardcoded then it needs to be its own value, not based on the bottom tab's height.

packages/bottom-tabs/src/views/BottomTabBar.tsx
satya164: > Also, the calculation for the 1.3 for web etc was to increase the maxWidth for web since there screens are way bigger

We shouldn't have different values for Web & native. The screen size can be big on native apps too (e.g. on a tablet). The use case for this feature is that user will use tabs on the side mainly on bigger screens and use bottom tabs on smaller screens.

josemak25: got you

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -377,15 +446,9 @@ export default function BottomTabBar({
 }
 
 const styles = StyleSheet.create({
-  tabBar: {
+  layout: {
     left: 0,
     right: 0,
     bottom: 0,
-    borderTopWidth: StyleSheet.hairlineWidth,
-    elevation: 8,
```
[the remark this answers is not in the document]
josemak25: thanks for pointing this out, taking a look at it again **elevation** was missing.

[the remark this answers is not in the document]
josemak25: > This option, which sets shadow on Android, seems to be missing

Thanks for pointing this out, taking a look at it again **elevation** was missing.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -29,7 +33,14 @@ type Props = BottomTabBarProps & {
   style?: Animated.WithAnimatedValue<StyleProp<ViewStyle>>;
 };
 
+type TabBarPositionMapConfig = Record<
+  TabBarPosition,
+  Record<'tabBar' | 'content' | 'tabBarItem', StyleProp<ViewStyle>>
```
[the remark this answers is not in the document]
josemak25: > 

Yea true thanks

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -58,6 +69,8 @@ const shouldUseHorizontalLabels = ({
       case 'below-icon':
         return false;
     }
+  } else if (tabBarPosition !== 'bottom') {
+    return false;
   }
```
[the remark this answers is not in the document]
josemak25: got you

satya164:
Will continue in https://github.com/react-navigation/react-navigation/pull/11578

PR #11002, opened 2022-11-16
chore: migrate to lerna-lite for proper yarn berry support

satya164:
We had a couple of issues with our Lerna setup:

- After publishing, the `yarn.lock` gets out of date. There is a new option to sync lock file, so it should fix the issue.
- We previously had build failures because lerna ignores `devDependencies` when building dependency graph, the `graphType: 'all'` option should address that, and we don't need to manually build before running publish
- `expo run:ios` now installs pods as well, so we can remove the step installing pods


kacperkapusciak (approved):

satya164:
@ghiscoding Thank you for all the work. I'll check those options out. Adding contributors to changelogs sounds really great!

PR #11007, opened 2022-11-16
[WIP] Support embedding into existing website

davispuh:
**Motivation**

We want to embed React Native application into our existing website which has header and footer.

Unfortunately currently `@react-navigation/stack` doesn't support this because `absolute` positioning is used which causes that those elements are not taken into account for `height` calculation. Basically end result is that whole tree has `height` of 0 and visually you see just blank page.

The fix seems pretty straightforward which I did in first commit of this PR - just replace  `absolute` positioning with `relative`. But unfortunately that breaks some things so in next commits I am fixing those issues.

Also note that we want single scrollbar and not double scrollbars (ie. main website and embedded app)

This PR is not complete and some components still need fixing.

Let us know if this is the right way to approach this issue and we're open to any suggestions/recommendations.

**Test plan**

This is still work in progress. So far I have tested this PR only in browser using Firefox 102

I tested with Example app from this repo.

To test how it would look like when embedded into website

`$ expo.js customize:web`
```diff
diff --git a/example/web/index.html b/example/web/index.html
index 67f728ce..d109f4c2 100644
--- a/example/web/index.html
+++ b/example/web/index.html
@@ -111,6 +111,12 @@
         </div>
       </form>
     </noscript>
-    <div id="root"></div>
+    <div>
+      <h1>Website's header</h1>
+      <div>
+        <div id="root"></div>
+      </div>
+      <h2>Some footer</h2>
+    </div>
   </body>
 </html>

```

When embedded into existing website:
* [x] **Main drawer**
* [x] **Native Stack**
* [x] **Simple Stack**
* [x] **Modal Stack**
* [x] **Regular + Modal Stack**
* [ ] **Float + Screen Header Stack** - *works almost fine, just have header issue when you "Push Feed" and then "Navigate to Album"*
* [ ] **Transparent Stack** - *works fine, except "Show Dialog" doesn't really work, it's added at bottom of page*
* [x] **Header Customization in Stack**
* [x] **Header Customization in Native Stack**
* [ ] **Bottom Tabs** - *completely broken*
* [ ] **Material Top Tabs** - *very broken*
* [ ] **Material Bottom Tabs** - *completely broken*
* [ ] **Dynamic Tabs** - *completely broken*
* [x] **Master Detail**
* [x] **Auth Flow**
* [x] **Prevent removing screen in Stack**
* [x] **Prevent removing screen in Native Stack**
* [x] **`<Link />`**

Note for some components there's a little element flicker (you can see what was behind for split-second) when clicking back arrow button.

When used as only app (current behavior):
* [x] **Native Stack**
* [ ] **Simple Stack** - *works fine, only header is sticky which wasn't before, but maybe that's fine?*
* [x] **Modal Stack**
* [ ] **Regular + Modal Stack** - *works fine, same issue with header being sticky when it wasn't before*
* [x] **Float + Screen Header Stack**
* [ ] **Transparent Stack** - *works fine, but same thing with having sticky header*
* [x] **Header Customization in Stack**
* [x] **Header Customization in Native Stack**
* [ ] **Bottom Tabs** - *pretty broken*
* [ ] **Material Top Tabs** - *works almost fine, only height is taken from longest tab even if it's not visible*
* [ ] **Material Bottom Tabs** - *pretty broken*
* [ ] **Dynamic Tabs** - *pretty broken*
* [x] **Master Detail**
* [ ] **Auth Flow** - works fine, but looks bit different than before, maybe that's fine?
* [ ] **Prevent removing screen in Stack** - *works fine, but same issue with header being sticky when it wasn't before*
* [x] **Prevent removing screen in Native Stack**
* [ ] **`<Link />`** - *works fine, but same issue with header being sticky when it wasn't before*

Also note that overscroll behavior at top and bottom changes a bit comparing how it was before and also this new sticky header is bit wobbly than solid position like it was before but that seems related to how browsers implement it. 

And here's example how it would look like if React Website embedded React Navigation's Example app 😄 

attels


davispuh:
While implementing this change, I did a lot of experimenting with trying various things. In the end seems those other changes are not needed and are not included in this PR but I'll leave [that diff here anyway for a reference](https://gist.github.com/davispuh/e141fb1ebfa220dccb6a8c1e593db80c)

satya164:

davispuh:
It's never good idea to rewrite history/force-push :stuck_out_tongue: 
Anyway I'll just submit rebased PR here at some point in future, you can simply reopen this PR.

Also we're still looking for any thoughts if such change would be acceptable and if this is right way to approach it.
In short term we'll be patching it locally before deployment and currently finishing this PR is in backlog.

davispuh:
Ehhh, I rebased this in same branch but this PR still didn't reflect that change so I submited new PR #11499

PR #11020, opened 2022-11-21
chore: remove ios and android folders

kacperkapusciak:
## Motivation

This PR removes prebuild `ios` and `android` folders. By default, react-navigation uses Expo Go for development and testing. Prebuild in the react-navigation repo is only used for testing the flipper plugin and this causes confusion (the project is prebuilt but we use Expo Go either way) and additional maintenance.

We motivate this decision by having less code in the repo, less maintenance, and less confusion about project setup.


example/package.json
```diff
@@ -8,12 +8,13 @@
     "client": "NODE_OPTIONS=--openssl-legacy-provider expo start --dev-client",
     "web": "NODE_OPTIONS=--openssl-legacy-provider expo start --web",
     "web:export": "NODE_OPTIONS=--openssl-legacy-provider expo export:web",
-    "android": "expo run:android",
-    "ios": "run-s pod-install run:ios",
+    "android": "run-s check:prebuild run:android",
+    "run:android": "expo run:android",
+    "ios": "run-s check:prebuild run:ios",
     "run:ios": "expo run:ios",
-    "pod-install": "pod-install",
```
satya164, line 14: The dependency can be removed as well

scripts/check-prebuild.js
```diff
@@ -0,0 +1,21 @@
...
+
+const prebuildFolders = ['ios', 'android'];
+
+const notExist = prebuildFolders.some((folder) => {
+  const folderPath = path.join(__dirname, '../example', folder);
+  return !fs.existsSync(folderPath);
+});
+
+if (notExist) {
+  const message = `Native source code folders ${prebuildFolders.join(
```
satya164, line 16: Maybe instead of throwing we can just run `expo prebuild`?

CONTRIBUTING.md
```diff
@@ -63,6 +63,20 @@ Run the e2e tests by:
...
+cd example/ && npx expo prebuild
+```
+
+Before compiling iOS in bare workflow remember to uncomment these lines in `example/ios/Podfile`:
+
+```ruby
+if !ENV['CI']
+  use_flipper!()
+end
+```
```
satya164, line 78: We could automatically do this in the script

scripts/check-prebuild.js
kacperkapusciak: Fixed in [145d4a3](https://github.com/react-navigation/react-navigation/pull/11020/commits/145d4a39101e0d04320edde9b105279a333fae6f)

CONTRIBUTING.md
kacperkapusciak: Fixed in [145d4a3](https://github.com/react-navigation/react-navigation/pull/11020/commits/145d4a39101e0d04320edde9b105279a333fae6f)

I want to let you know this flow is a bit flaky - it has to fail at least once.

It runs `npx expo prebuild` and it fails after a while because it can't see flipper. The same script enables `use_flipper` right after that.

On the second run of the command, it works as expected. It kinda has to be like that because we can't do something in the middle of the `npx expo prebuild` command 😢

kacperkapusciak: Fixed in [c06a4c6](https://github.com/react-navigation/react-navigation/pull/11020/commits/c06a4c621b369ada479499f3684887a248cfb9c4)

example/package.json
kacperkapusciak: Fixed in [fe7926a](https://github.com/react-navigation/react-navigation/pull/11020/commits/fe7926a8981836ae2fe6ed6d08f2531b8f3b5dc5)

scripts/check-prebuild.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+/* eslint-disable import/no-commonjs */
```
satya164, line 3: lets add a overrides in the eslint config to disable commonjs

kacperkapusciak: Fixed in [a54cc25](https://github.com/react-navigation/react-navigation/pull/11020/commits/a54cc253ee77c91f19d54fbb302ef666ddaf0886)

PR #11022, opened 2022-11-21
fix: add accessibility props to NativeStack screens

mdjastrzebski:
**Motivation**

Current Stack navigator (and other navigations) have accessibility props like `importantForAccessbility` & `accessibilityElementsHidden` for screens that are not currently visible on the screen. However, NativeStack navigator does not have it, which breaks new React Native Testing Library [`includeHiddenElements`](https://callstack.github.io/react-native-testing-library/docs/api-queries#includehiddenelements-option) query options which allows for ignoring elements hidden from accessibility.

As far as I understand this works correctly with iOS/Android screen readers because of using native primitives which are understood by screen readers/assistive tech. This PR includes the two accessibility props which should cause no harm to React Navigation user but would support RNTL option.

**Test plan**

Change does not affect the UI or API, it only applies to adding `importantForAccessbility` & `accessibilityElementsHidden` a11y props to NativeStack screens.

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -46,7 +46,7 @@
   },
   "devDependencies": {
     "@react-navigation/native": "^6.0.13",
-    "@testing-library/react-native": "^7.2.0",
+    "@testing-library/react-native": "^11.5.0",
     "@types/react": "~18.0.0",
     "@types/react-native": "~0.69.1",
     "del-cli": "^3.0.1",
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -237,6 +237,7 @@ const SceneView = ({
     : parentHeaderBack;
 
   const isRemovePrevented = preventedRoutes[route.key]?.preventRemove;
+  const isFocused = navigation.isFocused();
 
   return (
     <Screen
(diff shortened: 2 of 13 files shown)


packages/native-stack/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "devDependencies": {
     "@react-navigation/native": "^6.0.13",
-    "@testing-library/react-native": "^7.2.0",
+    "@testing-library/react-native": "^11.5.0",
```
satya164, line 49: This should be updated for all packages across the monorepo

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -237,6 +237,7 @@ const SceneView = ({
     : parentHeaderBack;
 
   const isRemovePrevented = preventedRoutes[route.key]?.preventRemove;
+  const isFocused = navigation.isFocused();
```
satya164, line 240: `navigation.isFocused()` shouldn't be used during render. The appropriate one is `useIsFocused` but it can't be used here since this isn't a screen. To handle it properly, we need to pass `focused` prop to `SceneView` from the parent.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -278,6 +279,8 @@ const SceneView = ({
       onNativeDismissCancelled={onNativeDismissCancelled}
       // this prop is available since rn-screens 3.16
       freezeOnBlur={freezeOnBlur}
+      accessibilityElementsHidden={!isFocused}
+      importantForAccessibility={isFocused ? 'auto' : 'no-hide-descendants'}
```
satya164, line 283: Does React Native Screens support these props? I don't see it in their typedefs. I think proper place for these would be the `<View style={styles.scene}>`

satya164 (changes requested):

packages/native-stack/src/views/NativeStackView.native.tsx
mdjastrzebski: Passed `focused` from the parent using following check: `const isFocused = state.index === index;`

satya164 (approved):

Issue #9499, opened 2021-04-14
Web - URL updates inconsistently when navigating to same screen in drawer with new params


marcaaron:
**Current Behavior**

When navigating to a drawer screen using `linkTo()` the URL in the browser inconsistently updates

https://user-images.githubusercontent.com/32969087/114649636-95eac180-9c7c-11eb-8f03-da179f312ecc.mp4

**Expected Behavior**

When we call `linkTo()` with a route and particular params that those params appear in the url.

**How to reproduce**

https://snack.expo.io/@marcaaron/react-nav---inconsistent-drawer-url

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | N/A
| @react-navigation/native       | 5.9.4
| react-native                   | 0.63.2
| expo                           |
| node                           |
| npm or yarn                    |

marcaaron:
Interestingly, this issue does not occur if we avoid `linkTo()` and instead call `props.navigation.navigate('Test', {id: 2})` from within the `drawerContent` component.

marcaaron:
Looked into this a bit further. It seems like when we are calculating the `historyDelta` here we are also including a drawer route in the history stack 

https://github.com/react-navigation/react-navigation/blob/47e371609d8c7e0344e2e7030d001e2d60c258b6/packages/native/src/useLinking.tsx#L497-L503

![Screen Shot 2021-04-14 at 11 44 53 AM](https://user-images.githubusercontent.com/32969087/114784700-4c9b8000-9d17-11eb-8e75-5673398d87f9.png)

This ends up giving us a `historyDelta` of `-1` which leads us to call `history.go(-1)` here

https://github.com/react-navigation/react-navigation/blob/47e371609d8c7e0344e2e7030d001e2d60c258b6/packages/native/src/useLinking.tsx#L520-L523

mrsimply:
I experience this very same behavior while using a StackNavigator when initializing the Web App at a "nested" route.

E.g. opening the app directly with `/customers/123`, from there navigatin to `/customers` with navigate(...) and then to `/customers/456`, it will still show me `/customers/123` in the browsers address bar while it correctly calls and shows `/customers/456` internally. 

Anyone experiencing this?

Tyler-Fairchain:
[...]

Issue #10992, opened 2022-11-10
Multiple Navigation dispatches that include a replace break browser back


Jackman3005:
### Current behavior

When dispatching multiple navigation actions at a time (especially when one is a replace) the browser back functionality is broken:

```javascript
navigation.dispatch(
  StackActions.replace("Screen2")
);
navigation.push("Screen3");
```
Going from Screen 1 to Screen 3 by replacing screen 1 and adding screen 2 behind screen 3 results in the correct state at the end where there are two routes ['screen 2', 'screen 3']. Using `navigation.pop()` works as expected and it would pop up to screen 2. When hitting the browser back from screen 3 I am taken back to screen 1.

### Expected behavior

Browser back should perform the same as `navigation.pop()`.

I think the issue lies in these [lines of native/src/useLinking.tsx](https://github.com/react-navigation/react-navigation/blob/1846de6bd8247992286d39ee76e65f27debb1754/packages/native/src/useLinking.tsx#L366-L409). The delta logic seems to only be accurate when one change is being processed at a time. 

Maybe this logic needs to be executed for each state change regardless of render cycle?

Not sure tbh how it all works together.

### Reproduction

https://snack.expo.dev/@jcoy/react-navigation-multiple-dispatch-browser-history-bug

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [X] @react-navigation/​stack
- [ ] @react-navigation/​native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.4 |
| @react-navigation/stack                | 6.2.1 |
| react-native-safe-area-context         | 4.2.4 |
| react-native-screens                   | 3.11.1 |
| react-native-gesture-handler           | 2.2.1 |
| react-native-reanimated                | 2.8.0 |
| react-native                           | 0.68.2 |
| expo                                   | 45.0.0 |
| node                                   | 16 |
| npm or yarn                            | yarn |

Jackman3005:
For anyone who finds this because they are experiencing the same issue. I was able to resolve it by delaying firing my second action.

```javascript
navigation.dispatch(StackActions.replace("Screen2"));
setTimeout(() => {
[...]

PR #11035, opened 2022-11-27
fix: #9499, url updates inconsistently when navigation to the same screen

racsmith277:

addresses issue: https://github.com/react-navigation/react-navigation/issues/9499

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -389,7 +389,7 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+              await history.push({ path, state });
             }
 
             // Store the updated state as well as fix the path if incorrect


packages/native/src/useLinking.tsx
```diff
@@ -389,7 +389,7 @@ export default function useLinking(
               // We couldn't find an existing entry to go back to, so we'll go back by the delta
               // This won't be correct if multiple routes were pushed in one go before
               // Usually this shouldn't happen and this is a fallback for that
-              await history.go(historyDelta);
+              await history.push({ path, state });
```
satya164, line 392: This change is not correct. We don't want to always push new screens, it should go forward to back depending on whether the user navigated forward to backward.

satya164 (changes requested):

packages/native/src/useLinking.tsx
racsmith277: this is a fallback for the case where the user is navigating to the same screen (e.g. with different route params). it needs to be pushed in order for params to update appropriately.

satya164: Even for fallback it's not correct. URL doesn't need to be pushed for params to update. It could also replace the params after the url updates.

Jackman3005: What's the alternative in this situation? I agree that `push` may not be perfect, but `go` is also not perfect and causing other issues right now. The issue @satya164 describes for `push` is that in the fallback scenario if it was actually a replace, we would end up with more routes in browser history than intended (correct?). But at least the correct URL will be shown for the screen they are on, as opposed to what is happening now with `history.go(...)`.

Jackman3005: This may be a slight tangent, but I feel like we are operating with insufficient knowledge to make the correct changes to the browser history. We only observe the state as it is and try to compare to the previous state without knowing what operations were performed. Although sometimes that's all the knowledge we'd actually have (like when a reset occurs), but most of the time we could know what action was intended (navigate, replace, push, pop, etc.). Would it help if we were listening for actions in the happy case and updating history with the improved knowledge we get from knowing which action was dispatched to effect the state change, then cover edge cases like reset by doing a previous state to current state comparison?

I believe the line mentioned in this change also is the cause for [this issue I submitted](https://github.com/react-navigation/react-navigation/issues/10992) "Multiple Navigation dispatches that include a replace break browser back".

osdnk:
I think we didn't come to any reasonable conclusion so I will close it for now.

PR #11038, opened 2022-11-28
feat: expose tabBarGap option in material top tabs

mlecoq:
This PR replaces #10984

**Motivation**

the new `tabBarGap` option helps to configure `gap` option of `react-native-tab-view` : see https://github.com/react-navigation/react-navigation/tree/main/packages/react-native-tab-view#gap

**Test plan**

Just set a tabBarGap option and check that there is a gap between tabs


satya164 (approved):
Thank you for the PR

PR #11048, opened 2022-11-29
Fix: pass text component to label renderer

SeyoungJoo:

**Motivation**
Pass in the original text component rendered with a string label in to the react component override to ensure complete customisability and flexibility of accessibility concern. This ensures we do not need to restyle and redesign the react component override to match what the text would be if we had passed in a string instead.

**Test plan**
We would aim to use this in the consumer like so:
```
const tabBarLabel = (props: {
    focused: boolean
    color: string
    position: LabelPosition
    textComponent: React.ReactElement
}) => {
    const dynamicBalance = "123" // <-- From Redux
    return React.cloneElement(props.textComponent, {
        accessibilityLabel: `Account ${dynamicBalance}`,
    })
}
```

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabItem.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabItem.tsx
@@ -33,6 +33,7 @@ type Props = {
         focused: boolean;
         color: string;
         position: LabelPosition;
+        textComponent: React.ReactElement;
       }) => React.ReactNode);
   /**
    * Icon to display for the tab.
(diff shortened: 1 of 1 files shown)


satya164:
Hey @SeyoungJoo, thanks for the PR. However this approach isn't ideal, `React.cloneElement` is problematic as it doesn't provide any type-safety and it also makes it easy to break the element.

Can you provide more details about the a11y concerns? Tab navigator already lets you specify a `accessibilityLabel`: https://reactnavigation.org/docs/bottom-tab-navigator/#tabbaraccessibilitylabel

SeyoungJoo:
Hey @satya164. yeah, it allows accessiblityLabel but it's not responsive. I need to use dynamic value for accessibilityLabel and it doesn't allow me to 😞 I'll discuss about `React.cloneElement` with @KallumParr. Thanks :)

PR #11049, opened 2022-11-29
Refactor/modify code to md3 standards

josemak25:
refactor: modify existing code to use md3 new header and drawer standards for android

**Motivation**

This PR extends and modifies the side drawer and screen header elements of the react-navigation by following the new standards set by MD3

https://m3.material.io/components/top-app-bar/overview
https://m3.material.io/components/navigation-drawer/overview

**Test plan**

Start the example application and use an android phone to test, changes can only be seen on android as this PR only address android latest MD3 specifications


osdnk:
Thanks for those changes. It looks like it's still unfinished with no progress indication. We're happy to follow po on that.

Issue #11045, opened 2022-11-28
useScrollToTop only works with the last tab navigator when nesting top and bottom tab navigators


anthonyguay:
### Current behavior

I'm opening a new issue since [the original one](https://github.com/react-navigation/react-navigation/issues/8586) is 2 years old and doesn't have much discussion. The [proposed fix](https://github.com/react-navigation/react-navigation/pull/9434) has not been merged and the branch has just been closed. The issue is still occurring.

### Expected behavior

When tapping on a bottom tab, the screen scrolls back to the top. However, when adding a second tab navigator at the top of the screen, tapping the bottom tap does not automatically scroll to top anymore. Only tapping the top navigation tab will scroll the screen up. It is expected that the bottom tab tap will still scroll the screen to the top, just like tapping the status bar.

Twitter and TikTok are two famous apps having top and bottom navigation tabs, and tapping on the bottom tap always scroll the screen to top. Other apps have what appears to be top tabs, but is in fact "filters" for the same screen. For the user, it appears similar, so the fact that it doesn't scroll here appears to be a bug.

### Reproduction

https://snack.expo.dev/ll8bTFpxv

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [X] @react-navigation/​native-stack
- [ ] react-native-tab-view
- [ ] flipper-plugin-react-navigation

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |
| @react-navigation/bottom-tabs          |
| @react-navigation/drawer               |
| @react-navigation/material-bottom-tabs |
| @react-navigation/material-top-tabs    |
| @react-navigation/stack                |
| @react-navigation/native-stack         | `6.0.13`
| react-native-safe-area-context         |
| react-native-screens                   |
| react-native-gesture-handler           |
| react-native-reanimated                |
| react-native-tab-view                  |
| react-native-pager-view                |
| react-native                           |
| expo                                   |
| node                                   |
| npm or yarn                            |

PR #9434, opened 2021-03-18
fix(useScrollToTop): handling multiple tab bars

Gregoirevda:
Fixes #8586
If a screen is inside a bottom tab bar and a top tab bar, only a click on the first parent (top tab bar) will scroll to top. Clicking on the bottom tab bar won't.
With this fix, all parents being a tab bar add an event listener to scroll to top. It will only scroll to top when the current view is inside one of the tab bars.

Issue #8586, opened 2020-07-14
useScrollToTop only works with the last 'tab' navigator when nesting top and bottom tab navigators


iirovi:
**Current Behavior**

When a top tab navigator is nested in a bottom tab navigator, the `useScrollToTop` hook only scrolls to top when pressing the current top tab, not the bottom tab. 
https://snack.expo.io/ll8bTFpxv

This is because these lines in the code for `useScrollToTop`
<img width="628" alt="Screen Shot 2020-07-14 at 12 21 37" src="https://user-images.githubusercontent.com/1415868/87453679-de336100-c5d0-11ea-95b4-70d6bdc62fee.png">

**Expected Behavior**

I would expect that `useScrollToTop` would work for both tab bars, so that I could tap on either to scroll my current scrollview to top. 

**How to reproduce**

https://snack.expo.io/ll8bTFpxv

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| @react-navigation/native       | ^5.7.0
| @react-navigation/material-top-tabs       | ^5.2.13
| @react-navigation/bottom-tabs       | ^5.7.1
| react-native                   | 0.62
| expo                           | 36.0.0
| node                           | 12.16.1
| npm or yarn                    | 1.22.4

iirovi:
> The versions mentioned in the issue for the following packages differ from the latest versions on npm:
> 
> * `@react-navigation/native` (found: `5.2.4`, latest: `5.7.0`)
> * `@react-navigation/material-top-tabs` (found: `5.1.15`, latest: `5.2.13`)
> * `@react-navigation/bottom-tabs` (found: `5.3.4`, latest: `5.7.1`)
> 
> Can you verify that the issue still exists after upgrading to the latest versions of these packages?

Yes, issue still persist after upgrading (I edited the description and the snack to use the latest versions)

anhkieet:
I faced the same issue as you. The useScrollToTop only work for top tab

iirovi:
In case anyone stumbles into this issue later, I ended up implementing a custom solution, hooking up to the tab press listener of my bottom tab, and firing off an event from there if the tab is focused and in the first element of my stack navigator, and having my view containing the scrollable handle that event. 

```
<BottomTab.Screen
  name="Home"
  component={HomeComponent}
  listeners={({navigation, route}) => ({
    tabPress: () => {
      const isAtFirstScreenOfStack = !route.state || route.state?.index === 0;
[...]

PR #11063, opened 2022-12-01
Fix useScrollToTop not working when nesting multiple tab navigators

anthonyguay:
**Motivation**

This PR answers this open issue: https://github.com/react-navigation/react-navigation/issues/11045

When tapping on a bottom tab, the screen scrolls back to the top. However, when adding a second tab navigator at the top of the screen, tapping the bottom tap does not automatically scroll to top anymore. Only tapping the top navigation tab will scroll the screen up. It is expected that the bottom tab tap will still scroll the screen to the top, just like tapping the status bar.

**Context**

This issue has previously been reported [here](https://github.com/react-navigation/react-navigation/issues/8586) by [@iirovi](https://github.com/iirovi), and a pull request to fix the issue has been proposed by [@Gregoirevda](https://github.com/Gregoirevda) [here](https://github.com/react-navigation/react-navigation/pull/9434). Recent changes on main automatically closed the pull request, so I'm posting his solution here again at @satya164's [request](https://github.com/react-navigation/react-navigation/pull/9434#issuecomment-1328345015).

We tested this solution on our app and it works perfectly well. Previously, tapping the bottom tab would not scroll back to the top when a top tab bar was present, but this fixes the issue. You can find the package versions in [the open issue](https://github.com/react-navigation/react-navigation/issues/11045).

**Behavior example**

Twitter has both top and bottom navigation bars. When tapping on the bottom tap, the screen scrolls back to the top. Tapping the top tab also scroll to the top.

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -109,7 +109,7 @@ export default function useScrollToTop(
     });
 
     return () => {
-      unsubscribers.forEach(unsubscribe => unsubscribe());
+      unsubscribers.forEach((unsubscribe) => unsubscribe());
     };
   }, [navigation, ref, route.key]);
 }
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -57,7 +57,7 @@ export default function useScrollToTop(
     // We need to find the closest tab navigators and add the listener there
     while (currentNavigator) {
       if (currentNavigator.getState().type === 'tab') {
-        tabNavigators = tabNavigators.concat(currentNavigator);
+        tabNavigators.push(currentNavigator);
       }
       currentNavigator = currentNavigator.getParent();
     }
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -50,25 +50,25 @@ export default function useScrollToTop(
   const route = useRoute();
 
   React.useEffect(() => {
-    let tabNavigators: NavigationProp<ReactNavigation.RootParamList>[] = [];
+    let tabNavigations: NavigationProp<ReactNavigation.RootParamList>[] = [];
     let currentNavigation = navigation;
 
-    // The screen might be inside another navigator such as stack nested in tabs
-    // The screen might be in multiple tab bars (as a top tab bar and a bottom tab bar and we want both events to work)
-    // We need to find the closest tab navigators and add the listener there
+    // If the screen is nested inside multiple tab navigators, we should scroll to top for any of them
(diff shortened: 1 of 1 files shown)


packages/native/src/useScrollToTop.tsx
```diff
@@ -49,53 +49,62 @@ export default function useScrollToTop(
...
-      current = current.getParent();
+    let tabNavigators = [];
+    let currentNavigator = navigation;
+    
+     // The screen might be inside another navigator such as stack nested in tabs
+     // The screen might be in multiple tab bars (as a top tab bar and a bottom tab bar and we want both events to work)
+     // We need to find the closest tab navigators and add the listener there
+     while (currentNavigator) {
+      if(currentNavigator.getState().type === 'tab') {
+        tabNavigators = tabNavigators.concat(currentNavigator);
```
satya164, line 60: mutation is fine for local variables
```suggestion
        tabNavigators.push(currentNavigator);
```

packages/native/src/useScrollToTop.tsx
```diff
@@ -101,10 +106,10 @@ export default function useScrollToTop(
           });
         }
       );
-    })
+    });
 
     return () => {
-        unsubscribers.map(unsubscribe => unsubscribe())
+      unsubscribers.map((unsubscribe) => unsubscribe());
```
satya164, line 112: `forEach` is more appropriate since we aren't mapping the array to another array
```suggestion
      unsubscribers.forEach(unsubscribe => unsubscribe());
```

satya164 (changes requested):

packages/native/src/useScrollToTop.tsx
```diff
@@ -49,53 +49,67 @@ export default function useScrollToTop(
   const route = useRoute();
 
   React.useEffect(() => {
-    let current = navigation;
+    let tabNavigators = [];
+    let currentNavigator = navigation;
```
satya164, line 53: these are `navigation` objects and not `navigator` so I'd change the naming

satya164 (changes requested):
Thanks for the PR. Looks like TS is failing. Can you fix the errors?

packages/native/src/useScrollToTop.tsx
anthonyguay: Thanks for pushing commits with your comments! I changed the variable name and I fixes TS by declaring type explicitly. Warnings are silenced locally, can we run automated tests?

anthonyguay:
> @anthonyguay thanks for getting this is main. We'll finally be able to remove our patch 👍

All thanks to you @Gregoirevda ! We're also going to be able to remove the patch 👍

PR #11069, opened 2022-12-02
fix: bump query-string version

krudos:
fix vulnerability in decode-uri-component, dependency of query-string, fixed in version 7.1.3
more info
https://security.snyk.io/vuln/SNYK-JS-DECODEURICOMPONENT-3149970


satya164 (approved):
Thank you

PR #11073, opened 2022-12-03
refactor: use consistent naming for useBottomTabBarHeight

goleary:
discrepancy makes it harder for consumers of this package to understand the internals

I've removed the template because of how simple this change is. Let me know if y'all would like me to expound on my reasoning for opening this.


satya164 (approved):
Thank you

Issue #11026, opened 2022-11-21
Hash/Fragment is stripped away on Web


nikgraf:
### Current behavior

When opening the example with the URL `https://react-navigation-example.netlify.app/native-stack/article/gandalf#key=something` the `#key=something` is stripped away.

### Expected behavior

When opening the example with the URL https://react-navigation-example.netlify.app/native-stack/article/gandalf#key=something the URL stays the same.

### Reproduction

https://github.com/react-navigation/react-navigation/tree/main/example

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [X] @react-navigation/​native-stack

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |   6.0.14

Gregoirevda:
@nikgraf You're reading the url from `getStateFromPath` ?

nikgraf:
Nope, should I used it in this case? Would this prevent the replace and stripping off the hash/fragment part?

nikgraf:
great, will try that today

@satya164 do you have a hint where I would need to look to preserve the hash? I would try to create a PR

satya164:
@nikgraf I think `useLinking.tsx` would be the best place to do it

nikgraf:
was solved with the PR

PR #11078, opened 2022-12-06
fix: keep hash/fragment section for initial render

nikgraf:
**Motivation**

Fixes this issue that I discovered https://github.com/react-navigation/react-navigation/issues/11026

**Test plan**

1. Run the example
2. Visit http://localhost:19006/simple-stack/article/gandalf#test
3. The with #test part still exists
4. Note that if you change the hash it will be gone (this is expected)

### DIFF
--- a/packages/native/src/createMemoryHistory.tsx
+++ b/packages/native/src/createMemoryHistory.tsx
@@ -84,6 +84,10 @@ export default function createMemoryHistory() {
 
       const id = window.history.state?.id ?? nanoid();
 
+      // Need to keep the hash part of the path if there was no previous history entry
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR. I think we should also preserve the hash on URL updates (if the screen didn't change). For now, we could only check if the `path` didn't change and preserve the hash.

Can you move this logic to `history.replace` function where it preserves the hash if path is the same?

nikgraf:
@satya164 force pushed a fix that works no as you described and in History.replace

satya164 (approved):
Thank you

PR #11097, opened 2022-12-13
Add the ability to "retain" a dismissed screen, to support iOS Picture In Picture overlays.

jaredly:
([Feature request fwiw](https://react-navigation.canny.io/feature-requests/p/support-picture-in-picture-for-video-with-native-stack-navigator)) ([corresponding react-native-screens PR](https://github.com/software-mansion/react-native-screens/pull/1665))

Thanks for making this wonderful library! I've really enjoyed using it. I'm transitioning a large app from custom native navigation code (written before we adopted react-native) to react-navigation, and the only thing I'm having trouble with is iOS Picture-In-Picture.

If you're unfamiliar, Picture-in-Picture (PIP) refers to popping a video out into an "overlay" that persists as you navigate around the app, until you either close it or "restore" it, which results in the originating screen being pushed back onto the stack, and the video sliding back into its original location.

Here's a little video of that in action, for context:

https://user-images.githubusercontent.com/112170/207445170-f9fc77cd-cebe-44c9-b95e-d8338ad587c5.mp4

There are two things needed to support PIP well:
1) the ability to dismiss a screen without letting the resources be garbage collected. (when the originating ViewController is garbage collected, the PIP window automatically closes). Additionally, the screen must not be unmounted on the react side of things, so that events will still be handled properly (critically, the `onRestoreUserInterfaceForPictureInPictureStop` callback for react-native-video).
2) the ability to bring back the screen that the PIP overlay originated from, in response to the user tapping on the relevant icon in the overlay.

I've made a pull-request to `react-native-screens` adding a `hidden` prop to the `Screen` component, which allows a screen to stay in the react tree (so the resources are retained & events still propagate), but be ignored by the NavigationStack.

On the react-navigation side, we need an additional flag on a given `route`, to allow a screen to be (a) marked for retention (e.g. if a user dismisses this screen, hang on to it but mark it "hidden"), (b) restored from the hidden state, and (c) dropped from retention (which, if the screen is hidden, means it is removed, and if it is still in the visible stack, it just has the flag cleared).

## Test code and steps to reproduce

Here's a minimal repo illustrating the problem behavior: https://github.com/jaredly/rnsdemo
And a video:
https://user-images.githubusercontent.com/112170/207732805-7750c1e8-7276-4f51-9244-f86ec312092b.mp4

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -61,7 +61,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-safe-area-context": ">= 3.0.0",
-    "react-native-screens": ">= 3.0.0"
+    "react-native-screens": ">= 3.19.0"
   },
   "react-native-builder-bob": {
     "source": "src",
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -61,7 +61,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-safe-area-context": ">= 3.0.0",
-    "react-native-screens": ">= 3.19.0"
+    "react-native-screens": ">= 3.0.0"
   },
   "react-native-builder-bob": {
     "source": "src",
--- a/packages/native-stack/src/index.tsx
+++ b/packages/native-stack/src/index.tsx
@@ -9,6 +9,11 @@ export { default as createNativeStackNavigator } from './navigators/createNative
 export { default as NativeStackView } from './views/NativeStackView';
 
 /**
+ * Contexts
+ */
+export { RetainContext } from './views/RetainContext';
+
+/**
  * Types
  */
 export type {
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -116,6 +116,7 @@ const MaybeNestedStack = ({
 
 type SceneViewProps = {
   index: number;
+  hidden: boolean;
   focused: boolean;
   descriptor: NativeStackDescriptor;
   previousDescriptor?: NativeStackDescriptor;
@@ -134,6 +135,7 @@ const SceneView = ({
   descriptor,
   previousDescriptor,
   nextDescriptor,
+  hidden,
   onWillDisappear,
   onAppear,
   onDisappear,
@@ -253,6 +255,8 @@ const SceneView = ({
     <Screen
       key={route.key}
       enabled
+      // @ts-ignore
+      hidden={hidden}
       style={StyleSheet.absoluteFill}
       customAnimationOnSwipe={customAnimationOnGesture}
       fullScreenSwipeEnabled={fullScreenGestureEnabled}
@@ -392,79 +396,125 @@ type Props = {
   descriptors: NativeStackDescriptorMap;
 };
 
+// const PinnedRoutesContext = React.createContext<{
+//   pin: () => void;
+// }>({
+//   pin() {},
+// });
+
 function NativeStackViewInner({ state, navigation, descriptors }: Props) {
   const { setNextDismissedKey } = useDismissedRouteError(state);
+  // const [pinnedRoutes, setPinnedRoute] = React.useState<
+  //   Props['state']['routes'][0] | null
+  // >(null);
 
   useInvalidPreventRemoveError(descriptors);
 
+  const currentKeys: { [key: string]: boolean } = {};
+  state.routes.forEach((route) => (currentKeys[route.key] = true));
+
+  const currentState = React.useRef(state);
+  currentState.current = state;
+
+  // const pin = React.useMemo(
+  //   () => ({
+  //     pin() {
+  //       setPinnedRoute(() => {
+  //         const routes = currentState.current.routes;
+  //         return routes[routes.length - 1];
+  //       });
+  //     },
+  //     restore() {
+  //       setPinnedRoute((prev) => {
+  //         // umm so if the route is still in the stack, do something different
+  //         if (prev) {
+  //           navigation.dispatch({
+  //             ...StackActions.push(prev.name, prev.params),
+  //             source: prev.key,
+  //             target: prev.key,
+  //           });
+  //         }
+  //         return null;
+  //       });
+  //     },
+  //   }),
+  //   []
+  // );
+
   return (
+    // <PinnedRoutesContext.Provider value={pin}>
     <ScreenStack style={styles.container}>
-      {state.routes.map((route, index) => {
-        const descriptor = descriptors[route.key];
-        const isFocused = state.index === index;
-        const previousKey = state.routes[index - 1]?.key;
-        const nextKey = state.routes[index + 1]?.key;
-        const previousDescriptor = previousKey
-          ? descriptors[previousKey]
-          : undefined;
-        const nextDescriptor = nextKey ? descriptors[nextKey] : undefined;
+      {(state.retained || [])
+        .map((route) => ({ route, hidden: true }))
+        .concat(state.routes.map((route) => ({ route, hidden: false })))
+        .map(({ route, hidden }, index) => {
+          const descriptor = descriptors[route.key];
+          const isFocused = state.index === index;
+          const previousKey = state.routes[index - 1]?.key;
+          const nextKey = state.routes[index + 1]?.key;
+          const previousDescriptor = previousKey
+            ? descriptors[previousKey]
+            : undefined;
+          const nextDescriptor = nextKey ? descriptors[nextKey] : undefined;
 
-        return (
-          <SceneView
-            key={route.key}
-            index={index}
-            focused={isFocused}
-            descriptor={descriptor}
-            previousDescriptor={previousDescriptor}
-            nextDescriptor={nextDescriptor}
-            onWillDisappear={() => {
-              navigation.emit({
-                type: 'transitionStart',
-                data: { closing: true },
-                target: route.key,
-              });
-            }}
-            onAppear={() => {
-              navigation.emit({
-                type: 'transitionEnd',
-                data: { closing: false },
-                target: route.key,
-              });
-            }}
-            onDisappear={() => {
-              navigation.emit({
-                type: 'transitionEnd',
-                data: { closing: true },
-                target: route.key,
-              });
-            }}
-            onDismissed={(event) => {
-              navigation.dispatch({
-                ...StackActions.pop(event.nativeEvent.dismissCount),
-                source: route.key,
-                target: state.key,
-              });
+          return (
+            <SceneView
+              key={route.key}
+              index={index}
+              hidden={hidden}
+              focused={isFocused}
+              descriptor={descriptor}
+              previousDescriptor={previousDescriptor}
+              nextDescriptor={nextDescriptor}
+              onWillDisappear={() => {
+                navigation.emit({
+                  type: 'transitionStart',
+                  data: { closing: true },
+                  target: route.key,
+                });
+              }}
+              onAppear={() => {
+                navigation.emit({
+                  type: 'transitionEnd',
+                  data: { closing: false },
+                  target: route.key,
+                });
+              }}
+              onDisappear={() => {
+                navigation.emit({
+                  type: 'transitionEnd',
+                  data: { closing: true },
+                  target: route.key,
+                });
+              }}
+              onDismissed={(event) => {
+                navigation.dispatch({
+                  ...StackActions.pop(event.nativeEvent.dismissCount),
+                  source: route.key,
+                  target: state.key,
+                });
 
-              setNextDismissedKey(route.key);
-            }}
-            onHeaderBackButtonClicked={() => {
-              navigation.dispatch({
-                ...StackActions.pop(),
-                source: route.key,
-                target: state.key,
-              });
-            }}
-            onNativeDismissCancelled={(event) => {
-              navigation.dispatch({
-                ...StackActions.pop(event.nativeEvent.dismissCount),
-                source: route.key,
-                target: state.key,
-              });
-            }}
-          />
-        );
-      })}
+                setNextDismissedKey(route.key);
+              }}
+              onHeaderBackButtonClicked={() => {
+                navigation.dispatch({
+                  ...StackActions.pop(),
+                  source: route.key,
+                  target: state.key,
+                });
+              }}
+              onNativeDismissCancelled={(event) => {
+                navigation.dispatch({
+                  ...StackActions.pop(event.nativeEvent.dismissCount),
+                  source: route.key,
+                  target: state.key,
+                });
+              }}
+            />
+          );
+        })}
     </ScreenStack>
+    // </PinnedRoutesContext.Provider>
   );
 }
 
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -40,6 +40,7 @@ import useDismissedRouteError from '../utils/useDismissedRouteError';
 import useInvalidPreventRemoveError from '../utils/useInvalidPreventRemoveError';
 import DebugContainer from './DebugContainer';
 import HeaderConfig from './HeaderConfig';
+import { RetainContext, useRetainContext } from './RetainContext';
 
 const isAndroid = Platform.OS === 'android';
 
@@ -116,11 +117,11 @@ const MaybeNestedStack = ({
 
 type SceneViewProps = {
   index: number;
-  hidden: boolean;
   focused: boolean;
   descriptor: NativeStackDescriptor;
   previousDescriptor?: NativeStackDescriptor;
   nextDescriptor?: NativeStackDescriptor;
+  hidden: boolean;
   onWillDisappear: () => void;
   onAppear: () => void;
   onDisappear: () => void;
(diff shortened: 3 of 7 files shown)


packages/native-stack/package.json
```diff
@@ -61,7 +61,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-safe-area-context": ">= 3.0.0",
-    "react-native-screens": ">= 3.0.0"
+    "react-native-screens": ">= 3.19.0"
```
jaredly, line 64: I set this to a fictitious future version, because it relies on a change to react-native-screens

satya164: @jaredly the peer dependency shouldn't be bumped, the package is still compatible with react native screens 3, only the new feature needs a newer version so that can be documented separately

satya164 (changes requested):
Hey! Thanks for the PR.

I'm not sure about the approach, this is highly specific to the native stack and native features so I don't think it belongs in the router.

Native Stack should expose this with a different API such as a custom hook and store the additional data regarding screens that need to be retained in its internal state.

It'd also be useful to see some code snippets of how this may be used to understand the APIs needed.

jaredly:
@satya164 cool, thanks for the feedback! I'll try the custom hook direction.

jaredly:
@satya164 So in order to implement screen restoration, I will need to be able to specify the `key` of the screen that I'm adding "back on" to the stack, such that it doesn't get a `nanoid()` key. Would you rather that be a separate action, or an argument added to the `PUSH` action? (or to the `.payload` of the `PUSH` action)

satya164:
> Would you rather that be a separate action, or an argument added to the PUSH action? (or to the .payload of the PUSH action)

Since this is very specific to native-stack, there shouldn't be any changes to routers or the core package. Ideally, native stack would expose a method from its hook that you can call which would internally use reset action to add the route back to the state - so your key will be preserved in this case.

jaredly:
oh cool, yeah that will work great

packages/native-stack/src/index.tsx
```diff
@@ -8,6 +8,11 @@ export { default as createNativeStackNavigator } from './navigators/createNative
  */
 export { default as NativeStackView } from './views/NativeStackView';
 
+/**
+ * Contexts
+ */
+export { RetainContext } from './views/RetainContext';
```
jaredly, line 14: let me know if you want this to go into a `/contexts/` directory, or somewhere else

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -369,74 +373,86 @@ function NativeStackViewInner({ state, navigation, descriptors }: Props) {
...
+              onDismissed={(event) => {
+                navigation.dispatch({
+                  ...StackActions.pop(event.nativeEvent.dismissCount),
+                  source: route.key,
+                  target: state.key,
+                });
+
+                if (!retainedKeys.includes(route.key)) {
+                  setNextDismissedKey(route.key);
+                }
```
jaredly, line 435: this warning fires spuriously when restoring a previously dismissed screen

jaredly:
@satya164 ok I've got it working 😅  this repo https://github.com/jaredly/rnsdemo with the `native-stack` dependency pointed at a local checkout of this branch, has Picture in Picture working!

https://user-images.githubusercontent.com/112170/213299800-593727e1-b07b-4a7b-ba6c-d40e8ffcfb45.mp4

packages/native-stack/src/views/NativeStackView.native.tsx
satya164: @jaredly yea this one has some issue, need to address it later

satya164:
Thanks @jaredly for updating the MR! I'll try to review it when I get some free time

jaredly:
bump? 🙏

jaredly:
@satya164 I'd really like to get this merged so I can stop depending on a fork -- is there anything I can do to help this along?

satya164:
@jaredly I'll try to review it this week. would you be able to write a sort usage example with the code so it's easier for me to see the API?

jaredly:
Fantastic, thanks! Here's a little demo repo I made to show how `retain` works together with PictureInPicture https://github.com/jaredly/rnsdemo/blob/main/App.js
We've been using this in production at Khan Academy for 6 months now, and it's been working great.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+ * supporting native iOS "Picture in Picture" mode, as the
+ * originating UIViewController must be retained in memory for the
+ * PiP overlay to continue running.
+ * Retained screens can also be "restored" to the top of the stack,
+ * for example when the user taps the button in the PiP overlay
+ * to return to the app.
+ */
+export const RetainContext = React.createContext<RetainContextT>({
+  retain() {
+    throw new Error(`Not in a native screen stack`);
```
osdnk, line 32:
```suggestion
    throw new Error(MISSING_CONTEXT_ERROR);
```

and then:
```
const MISSING_CONTEXT_ERROR = `Couldn't find a native screen context.`
```

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const hiddenRoutes: Route<any>[] = [];
+  const hiddenDescriptors: NativeStackDescriptorMap = {};
+  Object.keys(retainedScenes).forEach((key) => {
+    if (!routeKeys[key]) {
+      hiddenRoutes.push(retainedScenes[key].route);
+      hiddenDescriptors[key] = retainedScenes[key].descriptor;
+    }
+  });
+
+  return {
```
osdnk, line 132: is that supposed to be a public API?

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+          const { [key]: _, ...rest } = screens;
+          return rest;
+        });
+        return latestValues.current.retainedScenes[key] != null;
+      },
+      restore(key) {
+        const { retainedScenes, state } = latestValues.current;
+
+        const route = retainedScenes[key]?.route;
+        if (!route) {
```
osdnk, line 89: throw error

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+            descriptor: descriptors[route.key],
+          },
+        }));
+        return route.key;
+      },
+      release(key) {
+        setRetainedScenes((screens) => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const { [key]: _, ...rest } = screens;
+          return rest;
```
osdnk, line 81: check and warn if not present

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+) {
+  const [retainedScenes, setRetainedScenes] = React.useState<RetainedScenes>(
+    {}
+  );
+  const latestValues = React.useRef({ state, retainedScenes, descriptors });
+  React.useEffect(() => {
+    latestValues.current = { state, retainedScenes, descriptors };
+  });
+
+  const retainContext = React.useMemo<RetainContextT>(
```
osdnk, line 63:
```suggestion
  const retainContext = React.useMemo(
```

I thnk TS should be smart enough. Maybe not :)

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+        }));
+        return route.key;
+      },
+      release(key) {
+        setRetainedScenes((screens) => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const { [key]: _, ...rest } = screens;
+          return rest;
+        });
+        return latestValues.current.retainedScenes[key] != null;
```
osdnk, line 83: you should not rely on the ref here as concurrent mode can make it flaky. Why can't you rely on the previous state?

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const routeKeys = state.routes.reduce((map, route) => {
+    map[route.key] = true;
+    return map;
+  }, {} as { [key: string]: true });
+
+  // Routes that were marked as 'retain' and have been removed from the stack,
+  // for which we still want to keep the memory allocated.
+  const hiddenRoutes: Route<any>[] = [];
+  const hiddenDescriptors: NativeStackDescriptorMap = {};
+  Object.keys(retainedScenes).forEach((key) => {
```
osdnk, line 125: I'd prefer 
```
const {
  hiddenDescriptors: descriptors,
  hiddenRoutes: routes 
} = Object.values(retainedScreen).reduce(() => { ... }, { routes: [], descriptors: {} })
```

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+    if (!routeKeys[key]) {
```
osdnk, line 126: I would just check the `state.routes` here directly, even though it's a higher theoretical complexity. This is not a big deal as this array should be small anyway.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+  navigation: NativeStackNavigationHelpers,
+  descriptors: NativeStackDescriptorMap
+) {
+  const [retainedScenes, setRetainedScenes] = React.useState<RetainedScenes>(
+    {}
+  );
+  const latestValues = React.useRef({ state, retainedScenes, descriptors });
+  React.useEffect(() => {
+    latestValues.current = { state, retainedScenes, descriptors };
+  });
```
osdnk, line 61: I'm rather against patters like that. The new state should be the result of props and the previous state as react may batch those updates, we should not rely on what is currently in `useRef`.

osdnk: I believe this is not very needed here and it's possible to refactor the code without that.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const retainContext = React.useMemo<RetainContextT>(
+    () => ({
+      retain() {
+        const { state, descriptors } = latestValues.current;
```
osdnk, line 66: why not `state`? Also, please avoid shadowing variables as it gets harder to understand the code.

osdnk (changes requested):
@jaredly 
Thanks for your hard work. 
Few things:
- Please, specify what is the public API here. We prefer not to expose the contexts directly. 
- you should warn/throw if one wants to use that on Android/web 
- Avoid using values from refs in the state update. In general, I cannot see why `latestValue` is needed here. If it's very needed, please elaborate
- Avoid shadowing of variables
- Add some examples.

osdnk:
Actually, ignore my comment about adding the example. It's not possible now

packages/native-stack/src/views/RetainContext.tsx
satya164: I assume the intention is to memoize the `retain`, `release`, `restore` functions. In that case define those functions separately with `useLatestCallback` and then they can be safely added to dependency array for `useMemo`.

With that it's unnecessary to do such a ref.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+        const route = state.routes[state.routes.length - 1];
```
satya164, line 67: Instead of assuming the last screen, this hook should use the `useRoute` hook to get the correct route object

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+        setRetainedScenes((screens) => ({
+          ...screens,
+          [route.key]: {
+            route,
+            descriptor: descriptors[route.key],
+          },
```
satya164, line 73: This should only store the route key. Duplicating the `route` and `descriptor` is problematic as those could change after storing here making them out of sync. They should always be read from `state.routes` and `descriptors`.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+        });
+
+        const index = state.routes.findIndex((r) => r.key === key);
+        const routes =
+          index !== -1 ? state.routes.slice(0, index) : state.routes.slice();
+        routes.push(route);
+        navigation.dispatch({
+          type: 'RESET',
+          payload: { ...state, index: routes.length - 1, routes },
+        });
```
satya164, line 107: Don't refer to actions by string. Use the `CommonActions.reset` action creator instead. In this case it should be possible to just use `navigation.reset`.

Also, I assume it's moving the screen to the end. Is this necessary/desirable? If the screen wasn't the last one, seems strange to change the order after pip restore. I'd remove this, consumers can do this themselves if they want to.

packages/native-stack/src/views/RetainContext.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+        return true;
+      },
+      supported: true,
```
satya164, line 111: This boolean seems unnecessary if it's always `true`.

packages/native-stack/src/views/RetainContext.tsx
satya164: The code will be much simpler if there was just a `hiddenRouteKeys` array. Seems like the only place that needs it is when we set `hidden={true}`, and with such an array, it could just check `hidden={hiddenRoutes.includes(route.key)}` without changing descriptor and routes related logic.

satya164 (changes requested):
I think an API like the following would be simpler:

```js
const { retain, release, restore } = useRetain();

// ...

<Video
  // ...
  onPictureInPictureStatusChanged={isActive => {
    if (isActive) {
      retain();
    } else {
      release();
    }
  }}
>
```

It'd be simpler not to have to manage a `key`. If user wants to retain multiple things (is that even possible?), they can use the hook multiple times. Though looking at the code, the `key` seems to be just the route key which will always be the same and can be aquired from `useRoute` inside the hook itself.

Also why does `restoreUserInterfaceForPictureInPictureStopCompleted` need a timeout? Is it possible to make `restore` return a promise so consumer can wait until this function can be called?

satya164:
FYI we have added `retain` support to regular stack navigator via `navigation.retain` and would like to use the same method for native stack: https://github.com/react-navigation/react-navigation/pull/11765

jaredly:
Responding to your questions (thanks for giving it a thorough review!)

> It'd be simpler not to have to manage a `key`.

I realize now that my demo repository doesn't make use of the keys, but a full application (such as ours, the Khan Academy app) that needs to support picture-in-picture does need then to function correctly. Specifically, when viewing another screen with a video on it, the PIP overlay should be automatically dismissed, but when returning to the screen that originated the PIP overlay, it should not be automatically dismissed.

> Also why does restoreUserInterfaceForPictureInPictureStopCompleted need a timeout? Is it possible to make restore return a promise so consumer can wait until this function can be called?

In order for the PIP overlay to be restored fluidly (animating smoothly back to the location of the `<Video/>` element), the screen needs to have finished animating in. The `timeout` is a simple way to wait for the animation to complete. I couldn't see a way to have `restore` know the parameters of the screen animation, but if you know how that would be fantastic.

packages/native-stack/src/views/RetainContext.tsx
jaredly: Hmm it can't be read from `state.routes` if the screen has been popped off the stack; that's why I'm storing it here. In the case where the route hasn't been popped, I use the one from `state.routes`.

packages/native-stack/src/views/RetainContext.tsx
jaredly: As currently written, the `retain` function isn't a hook. I'll look into changing the API to a `useRetain` and see whether that can meet all our needs

packages/native-stack/src/views/RetainContext.tsx
jaredly: it's true here, but the default value for the `RetainContext` has `supported: false`

packages/native-stack/src/views/RetainContext.tsx
jaredly: > If the screen wasn't the last one, seems strange to change the order after pip restore.

I thought that in order for a screen to be visible on a screen stack, it needed to be the last one. Is that not the case? `restore()` is supposed to bring the screen back to visibility; either adding it to the end if it had been popped off, or popping off anything on top of it if the screen still exists in the stack.

satya164:
> we need to be able to "mark a screen as retained" without also removing it from the active routes

I think we can make that change.

danielchan303:
Any updates on this? I would like to keep on playing a video in background (e.g. MV), which would bring back to the top when the user click on a floating window.

satya164:
We have added support for retain in React Navigation 8.

Calling `retain` only marks the screen for retaining, and navigation actions like `goBack` will remove it from active routes and retain it.

https://reactnavigation.org/docs/8.x/stack-actions/#retain

For now, it's supported in Stack. For Native Stack, we need support in React Native Screens.

I'm closing the PR since it should now be good from React Navigation side. Let me know if the API looks good.

Issue #10219, opened 2021-12-26
support "maxFontSizeMultiplier" prop 


malacca:
### Current behavior

<img width="380" alt="header" src="https://user-images.githubusercontent.com/15194457/147416819-b7514ac9-2460-4ed3-ab35-061a474932eb.png">
<img width="365" alt="tabbar" src="https://user-images.githubusercontent.com/15194457/147416823-b4a383e4-5905-4069-8c98-d8e38b06c8f4.png">

If config `headerBackAllowFontScaling, headerTitleAllowFontScaling, tabBarAllowFontScaling` is `true` and user modifies the system font size, the situation shown above will appear

For `headerTitle` and `tabBarLabel`, I can also solve the problem by customizing components, but `headerBackTitle` and `tabBarBadge` not support customize components, even `tabBarBadge` does not `AllowFontScaling`

So it is particularly necessary to support `maxFontSizeMultiplier` in components that support `AllowFontScaling`

### Expected behavior

support "maxFontSizeMultiplier" prop 

### Reproduction

none

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack

### Environment

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |
| @react-navigation/bottom-tabs          |
| @react-navigation/drawer               |
| @react-navigation/material-bottom-tabs |
| @react-navigation/material-top-tabs    |
| @react-navigation/stack                |
| @react-navigation/native-stack         |
| react-native-safe-area-context         |
| react-native-screens                   |
| react-native-gesture-handler           |
| react-native-reanimated                |
| react-native-tab-view                  |
| react-native-pager-view                |
| react-native                           |
| expo                                   |
| node                                   |
| npm or yarn                            |

malacca:
hope `tabBarBadge` support ReactNode, and components support `maxFontSizeMultiplier`  that support `AllowFontScaling` 🙏

jpaas:
[...]

PR #11102, opened 2022-12-14
Prevent font scaling on tab bar item badge

jpaas:
**Motivation**

The layout on the tab bar badges breaks as soon as the user starts to change the OS settings for text size.
This is illustrated in #10219

**Test plan**

Set a badge on a tab bar item.
Increase the iOS or Android display text size.
Expect: The text in the tab bar badge doesn't change and the layout doesn't change,

### DIFF
--- a/packages/bottom-tabs/src/views/Badge.tsx
+++ b/packages/bottom-tabs/src/views/Badge.tsx
@@ -70,6 +70,7 @@ export default function Badge({
 
   return (
     <Animated.Text
+      allowFontScaling={false}
       numberOfLines={1}
       style={[
         {


StampixSMO:
Subscribing since we also experience the issue, but I believe we want to allow more flexibility compared to just globally disabling it?

satya164 (changes requested):
Thanks for the PR. However, I think disabling font scaling isn't the proper solution. If the user is scaling the fonts, then they need bigger fonts to be able to read it and disabling font scaling makes them unable to read it.

The appropriate solution would be to handle font scaling better so that the UI doesn't break.

Issue #11047, opened 2022-11-29
@react-navigation/material-top-tabs delay and lag on tab press


Jamal-ReachFirst:
### Current behavior

I am using topTab(scrollable) inside another screen.

Swipe b/w tabs are faster but facing issues when tabbar is pressed (delay and lag).

### Expected behavior

Swipe and tab press should perform the same way, without delays and lag.

### Reproduction

https://snack.expo.dev/@jamal_rn/tob-tabbar

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [X] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [ ] @react-navigation/​native-stack
- [ ] react-native-tab-view
- [ ] flipper-plugin-react-navigation

### Environment

    "@react-navigation/material-top-tabs": "^6.3.1",
    "@react-navigation/native": "^6.0.11",
    "@react-navigation/stack": "^6.2.2",
    "react": "17.0.2",
    "react-native": "0.67.4",
    "react-native-pager-view": "^6.1.1",
    "react-native-reanimated": "^2.9.1",
    "react-native-safe-area-context": "^4.3.1",
    "react-native-screens": "^3.15.0",
    "react-native-tab-view": "^3.3.0",

Jamal-ReachFirst:
> Couldn't find version numbers for the following packages in the issue:
> 
> * `flipper-plugin-react-navigation`
> 
> Can you update the issue to include version numbers for those packages? The version numbers must match the format 1.2.3.
> 
> The versions mentioned in the issue for the following packages differ from the latest versions on npm:
> 
> * `@react-navigation/native` (found: `6.0.11`, latest: `6.0.14`)
> * `@react-navigation/stack` (found: `6.2.2`, latest: `6.3.5`)
> 
> Can you verify that the issue still exists after upgrading to the latest versions of these packages?

Updated but still facing the same issue.

Jamal-ReachFirst:
> The versions mentioned in the issue for the following packages differ from the latest versions on npm:
> 
> * `@react-navigation/material-top-tabs` (found: `6.3.1`, latest: `6.5.1`)
> * `@react-navigation/native` (found: `6.0.14`, latest: `6.1.1`)
[...]

PR #11105, opened 2022-12-16
fix: add unstable_pressDelay to fix onPress delay on android

okwasniewski:
**Motivation**

This PR addresses: #11047

According to [this](https://github.com/facebook/react-native/issues/29376#issuecomment-689776027) comment there was an additional prop added called `unstable_pressDelay` which seems to fix the issues with the delay of Pressable component on android. 

There is also this open issue that states the Pressable component hasn't been fixed when embedded inside ScrollView (like in our case), and this unstable prop fixes it: https://github.com/facebook/react-native/issues/30563

**Test plan**

Run the example app on android and check if the delay on press still exists.


Jamal-ReachFirst (commented):
No it doesn't resolve the issue.

Can you please try with `swipeEnabled` having more tabs with list items.

I have shared a video, please have a look. [here](https://github.com/react-navigation/react-navigation/issues/11047#issuecomment-1354534536)

PR #11130, opened 2022-12-31
chore: remove macos hack

dcangulo:
**Motivation**
I added the hack before since `transform` is broken at that time. But it's now fixed in RN macOS `0.66+`, the current stable version of RN macOS is `0.68.X`.

We don't need this anymore.

Previous PR: https://github.com/satya164/react-native-tab-view/pull/1188

**Test plan**

Same test plan in https://github.com/satya164/react-native-tab-view/pull/1188

### DIFF
--- a/packages/react-native-tab-view/src/TabBarIndicator.tsx
+++ b/packages/react-native-tab-view/src/TabBarIndicator.tsx
@@ -3,7 +3,6 @@ import {
   Animated,
   Easing,
   I18nManager,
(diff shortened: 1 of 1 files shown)


satya164:
Thank you for the PR. I think it'd better to wait till next major to remove this since otherwise `react-native-tab-view` will break in older versions of `react-native-macos` which is a breaking change - unless the workaround breaks something in the new version.

PR #11133, opened 2023-01-02
feat: wrapper option, to wrap screen content with suspense and error boundaries

levibuzolic:
For apps that use `React.Suspense` ([Relay](https://relay.dev/docs/guided-tour/rendering/loading-states/), [React Query](https://react-query-v3.tanstack.com/guides/suspense) etc.) and/or error boundaries for each screen it can be awkard to wrap every screen with the required components in userland.

# Problem example

Given the following screen and suspense/error boundary wrappers:

```tsx
function HomeScreen() {
  const data = useQuery(); // A hook that will suspend
  return <Text>{data.text}</Text>;
}

function SuspenseBoundary({children}) {
  return (
    <ErrorBoundary>
      <React.Suspense fallback={<Text>Loading...</Text>}>
        {children}
      </React.Suspense>
    </ErrorBoundary>
  );
}
```

There are a few user-land options for handling wrapping every screen:

## Screen.children

This opts out of the `React.PureComponent` optimisations of React Navigation and becomes quite verbose with many routes.

<details>
<summary>Example</summary>

```tsx
function NavigatorRoot() {
  return (
    <Stack.Navigator>
      <Stack.Screen name="Home">
        {(props) => (
          <SuspenseBoundary>
            <HomeScreen {...props} />
          </SuspenseBoundary>
        )}
      </Stack.Screen>
    </Stack.Navigator>
  );
}
```
</details>

## Manually wrapping

This keeps the optimisations, but still painful with many routes.

<details>
<summary>Example</summary>

```tsx
function NavigatorRoot() {
  return (
    <Stack.Navigator>
      <Stack.Screen name="Home" component={HomeScreenWrapper} />
    </Stack.Navigator>
  );
}

function HomeScreenWrapper(props) {
  return (
    <SuspenseBoundary>
      <HomeScreen {...props} />
    </SuspenseBoundary>
  );
}
```
</details>

## Component factory

This is the currently the best user-land solution, but if we had support for providing a wrapper at the navigator level, this could be much simpler.

<details>
<summary>Example</summary>

```tsx
const HomeScreen = factoryToWrapWithSuspense(require('./HomeScreen'));

function NavigatorRoot() {
  return (
    <Stack.Navigator>
    </Stack.Navigator>
  );
}

function factoryToWrapWithSuspense(module) {
  const Component = module.default;

  return (props) => (
    <SuspenseBoundary>
      <Component {...props} />
    </SuspenseBoundary>
  );
}
```
</details>

# Proposed solution

A `wrapper` option added to each navigator which allows the user to supply a component to wrap the content of the screen. This allows consistent behaviour across an entire navigator in addition to flexibility to alter the behaviour for each screen should that be required.

```tsx
function NavigatorRoot() {
  return (
    <Stack.Navigator screenOptions={{wrapper: SuspenseBoundary}}>
      <Stack.Screen name="ScreenWithoutWrapper" component={AnotherScreen} options={{wrapper: null}} />
    </Stack.Navigator>
  );
}
```

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -254,6 +254,11 @@ export type BottomTabNavigationOptions = HeaderOptions & {
    * Only supported on iOS and Android.
    */
   freezeOnBlur?: boolean;
+
+  /**
+   * A component to wrap the screen content with, useful for suspense and error boundaries.
+   */
+  wrapper?: (props: { children: React.ReactNode }) => JSX.Element;
 };
 
 export type BottomTabDescriptor = Descriptor<
--- a/packages/drawer/src/types.tsx
+++ b/packages/drawer/src/types.tsx
@@ -220,6 +220,11 @@ export type DrawerNavigationOptions = HeaderOptions & {
    * Only supported on iOS and Android.
    */
   freezeOnBlur?: boolean;
+
+  /**
+   * A component to wrap the screen content with, useful for suspense and error boundaries.
+   */
+  wrapper?: (props: { children: React.ReactNode }) => JSX.Element;
 };
 
 export type DrawerContentComponentProps = {
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -246,6 +246,7 @@ function DrawerViewBase({
             headerStatusBarHeight,
             headerTransparent,
             sceneContainerStyle,
+            wrapper: Wrapper,
           } = descriptor.options;
 
           return (
@@ -272,7 +273,11 @@ function DrawerViewBase({
                 })}
                 style={sceneContainerStyle}
               >
-                {descriptor.render()}
+                {Wrapper ? (
+                  <Wrapper>{descriptor.render()}</Wrapper>
+                ) : (
+                  descriptor.render()
+                )}
               </Screen>
             </MaybeScreen>
           );
--- a/packages/material-bottom-tabs/src/types.tsx
+++ b/packages/material-bottom-tabs/src/types.tsx
@@ -87,6 +87,11 @@ export type MaterialBottomTabNavigationOptions = {
    * ID to locate this tab button in tests.
    */
   tabBarTestID?: string;
+
+  /**
+   * A component to wrap the screen content with, useful for suspense and error boundaries.
+   */
+  wrapper?: (props: { children: React.ReactNode }) => JSX.Element;
 };
 
 export type MaterialBottomTabDescriptor = Descriptor<
@@ -114,5 +119,6 @@ export type MaterialBottomTabNavigationConfig = Partial<
     | 'getColor'
     | 'getLabelText'
     | 'getTestID'
+    | 'wrapper'
   >
 >;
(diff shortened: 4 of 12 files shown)


xotahal:
Would this also help with a scenario where we have dynamic screens and we need to wrap `Header` and `Component` to the same and dynamic `React.Context` state? Or do you wrap only `component` by the provided `Wrapper` and `reanderHeader` component doesn't know about the `Wrapper`?

Let me explain the problem. Let's have this kind of wrapper (which maybe doesn't make sense but I am using it only to explain the problem):

```ts
const Wrapper = (props) => {
  const { itemId } = props.route.params
  
  return (
     <Suspense>
        <ItemProvider>
          {props.children}
        </ItemProvider>
     </Suspense> 
   )
}

const ItemProvider = (props) => {
   const item = fetchItem(props.itemId)
   return <ItemContext.Provider value={item}>{props.children}</ItemContext.Provider>
}
```
Now let's assume this configuration.
```ts
 <Stack.Screen
      name="item-detail"
      component={ItemDetail}
      getId={params => params.itemId}
      options={{
          header: () => <CustomHeader />,
      }}
  />
  ```
  Now I need `ItemDetail` and `CustomHeader` to get the same `ItemContext` provider. When I wrap only one of those component then only one of those component would get the Context. And I can't wrap the whole `Stack.Screen` because I don't know the `itemId`. So when I do this:
  
  ```ts
   <Stack.Screen
      name="item-detail"
      component={ItemDetail}
      getId={params => params.itemId}
      options={{
          header: () => <CustomHeader />,
          wrapper: Wrapper
      }}
  />
  ```
  
  Would this generate tree similar to this one?
  ```tsx
  <MaybeScreen>
     <View>
        ...
          <Suspense>
            <ItemProvider>
              <CustomHeader />
              <ItemDetail />
            </ItemProvider>
          </Suspense>
          ...
       </View>
   </MaybeScreen
   ```

satya164:
Thanks for the PR. I'd like to have deeper integration with suspense but I need more exploration to make sure that we're addressing different use cases such as @xotahal mentioned. Feel free to open an RFC in the discussion forum.

levibuzolic:
Thanks for the comments @xotahal, you're right in that this implementation would not wrap the header, so the context would not be available there. 

In our codebase we've often wanted a first-class solution for a similar problem, which is sharing context across a subset of screens -- we've often had to resort to moving that context to the root of our app, or wrapping the closest parent navigator.

For the Suspense use case it's likely not desireable to wrap the entire screen and header because we wouldn't want to remove the ability to navigate back while loading. So I think for an approach like this to effectively cater to more use cases we'd likely need to have the ability to inject wrapping components at multiple different levels in the component hierarchy.

Apreciale the response @satya164, I'll start an RFC and hopefully we'll be able to gather some more information to get a path forward. :+1:

levibuzolic:
I'll close this PR for now and move discussion to https://github.com/react-navigation/react-navigation/discussions/11152

This PR can be re-opened should this particular implementation be suitable.

@xotahal I'd appreciate your input on around [floating headers and context shared across multiple screens](https://github.com/react-navigation/react-navigation/discussions/11152#discussioncomment-4673222) and if that would work for your use case.

PR #11144, opened 2023-01-10
Prototype for static navigation config

satya164:
## Goal

React Navigation 5+ has a dynamic API which makes it possible to achieve advanced use cases. But it can result in additional boilerplate and duplication. This is a prototype for a static API similar to React Navigation 4 to provide a simpler API for people who don't need dynamic configuration.

This also intends to simplify 2 major issues:

- Deep link configuration
- TypeScript configuration

### Basic usage

```js
const HomeTabs = createBottomTabNavigator({
  screenOptions: {
    tabBarActiveTintColor: "tomato"
  },
  screens: {
    Groups: {
      screen: GroupsScreen,
      linking: "groups"
    },
    Chat: {
      screen: ChatScreen,
      linking: "chat/:id"
    }
  }
});

const RootStack = createStackNavigator({
  screens: {
    Home: HomeTabs,
    Profile: ProfileScreen,
    Feed: {
      screen: FeedScreen,
      options: {
        headerTitle: "My Feed"
      }
    },
    Settings: SettingsScreen
  }
});

const Navigation = createStaticNavigation(RootStack);

  return (
    <Navigation
      onStateChange={() => {
        // Do something
      }}
    />
  );
}
```

### Documentation

- [Getting started](https://github.com/react-navigation/react-navigation.github.io/blob/%40satya164/static-api/versioned_docs/version-7.x/static-configuration.md)
- [API reference](https://github.com/react-navigation/react-navigation.github.io/blob/%40satya164/static-api/versioned_docs/version-7.x/static-api-reference.md)
- [TypeScript configuration](https://github.com/react-navigation/react-navigation.github.io/blob/%40satya164/static-api/versioned_docs/version-7.x/static-typescript.md)
- [Authentication flow](https://github.com/react-navigation/react-navigation.github.io/blob/%40satya164/static-api/versioned_docs/version-7.x/static-authentication.md) - TODO
- [Combining static and dynamic APIs](https://github.com/react-navigation/react-navigation.github.io/blob/%40satya164/static-api/versioned_docs/version-7.x/static-combine-with-dynamic.md) - TODO

### Potential TODOs

- [ ] Add more checks to ensure types are working consistently
- [x] Add unit tests to ensure functionality
- [x] Generate linking config based on the `path` specified
- [ ] ~Support `getComponent` for lazy require~
- [ ] ~Add `layout`/`wrapper` option for custom UI and logic around the navigator~
- [ ] ~Add ability pass dynamic data to use for `options`~
- [x] Add `if` option to conditionally render screens, e.g. for authentication
- [ ] Prevent unknown properties in the config with TypeScript (probably not possible due since no `Exact` in TS)
- [ ] Add runtime validation for config properties
- [x] Add JSDoc to all types for static config for better DX
- [ ] ~Add helper to create a type for the navigation prop preserving navigator specific methods~
- [ ] Update warning in NavigationContainer to mention that StaticNavigation already uses it


mdjastrzebski:
@satya164 will users be able to mix static and dynamic navigators in their setup? 

I am thinking about scenario when users start with basic static setup, but as the app grows bigger they find that they need to use dynamic nav in one particular navigator.

satya164:
@mdjastrzebski it'd be possible to use dynamic navigators inside static navigators but not vice-versa.

ericvicenti:
Looks good to me!

Will we also support this options callback behavior (its a classic 😂)

```ts
const RootStack = createStackNavigator({
  screens: {
    Foo: {
      options: ({ navigation }) => ({ ... })
```

satya164:
@ericvicenti yup, it's the same as the dynamic version, so it can be an object or function that receives `navigation` or `route` props.

mdjastrzebski:
@satya164 I am curious:
1. What would be recommended for new apps/users of React Nav: static or dynamic API?
2. What criteria should users use to determine which one to use?c

satya164:
@mdjastrzebski 

1. Dynamic API will stay the default for now. It may change in the future where we document them side-by-side or make the static approach as default.
2. For now we'd document the problems this API solves and let devs decide for their project. We'd need to spend some time with alpha/beta users to evaluate this and document use cases where static API doesn't work, along with documentation on how to migrate away from the static API.

PR #11150, opened 2023-01-12
Add optional name to Group

milansusnjar:
Optional name property can make it easier to distinguish groups in the navigator. For example group of screens named onboarding, another group named profileTab, settingsTab...

Instead of writing a comment above the group we can write it as the name.
image


satya164:
Thanks for the PR, I don't really want to add a prop that doesn't do anything since it can cause confusion. Comments are my recommended way to add descriptions to groups, however, you can also use a static string for `navigationKey` if you really want a prop.

PR #11155, opened 2023-01-15
[react-native-tab-view] Do not change layout state on empty width or height

billouboq:
I have a stack navigator that has a page with a tab-navigator. If I push a page, the tab-navigator emit a layout event with width 0 and height 0 resulting in a graphical flicker when going back to that tab navigator.

By removing the empty layout event, it successfully remove my layout flickers.

### Structure :

- Stack Navigator
   - Home -> Tab Navigator
   - Settings
    
### Actions :
  Go to home page, then push Settings page, go back -> flicker occurs
  
I cannot do a video of it since it's on a work application unfortunately

**Motivation**

Having layout flicker on back is annoying for the user experience. This change fix it.

### DIFF
--- a/packages/react-native-tab-view/src/TabBar.tsx
+++ b/packages/react-native-tab-view/src/TabBar.tsx
@@ -330,11 +330,14 @@ export default function TabBar<T extends Route>({
   const handleLayout = (e: LayoutChangeEvent) => {
     const { height, width } = e.nativeEvent.layout;
 
-    setLayout((layout) =>
-      layout.width === width && layout.height === height
-        ? layout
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR, but the actual bug seems to be that `onLayout` event is emitted with an incorrect layout. Ignoring the event is not correct since it can ignore when the actual layout changes and would suppress the actual issue.

billouboq:
Oh right, was afraid of that 😬
Thanks for your answer, will dig deeper to find what’s going on, thank you

PR #11009, opened 2022-11-17
feat: add support testID in Drawer

awpogodin:
Added support for TestIDs for auto-tests

#7492

Issue #8906, opened 2020-09-26
DrawerItem should support testID


archfz:
**Current Behavior**
Can't set testID on the button. Additional props go to the view element.
https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/src/views/DrawerItem.tsx#L149

**Expected Behavior**
Can set testID.

Gamma169:
This looks to be a duplicate of 
https://github.com/react-navigation/react-navigation/issues/7492

ManAnRuck:
any updates?

Issue #7492, opened 2020-02-25
Support drawerTestIDs


mrozbarry:
**Current Behavior**

- I am testing with detox, and want a way to target a drawer option by testID

**Expected Behavior**

- I should be able to add `drawerTestID` as a navigationOptions setting that will set the testID appropriately on the touchable element.

**How to reproduce**

- N/A

**Your Environment**

| software                     | version |
| ---------------------------- | ------- |
| iOS or Android               | iPhone 11
| react-navigation             | 4.0.10
| react-navigation-drawer      | 2.3.3
| react-native-gesture-handler | 1.5.3
| react-native-reanimated      | 1.4.0
| react-native-screens         | 2.0.0-alpha.12
| react-native                 | 0.61.4
| expo                         | 36.02
| node                         | 10
| npm or yarn                  | yarn 1.21.1

nitaking:
me too.

nitaking:
and v5

satya164:
PR welcome

DallasCarraher:
Anybody already taken this on?

DallasCarraher:
I've created a PR in case it hasn't. https://github.com/react-navigation/react-navigation/pull/8720

Gamma169:
This looks fairly straightforward.   Looks like you just need to add a `testID` prop to [the componen](https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/src/views/DrawerItem.tsx)t, and pass it down to to [its wrapping View](https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/src/views/DrawerItem.tsx#L169).  I might look into it when I have some time.

In the mean time, I found that you can either wrap the `DrawerItem` in a `View` in order to add `testID` to that `View`.  But that was kinda clunky.  I found some success using the `icon` property and adding `testID` to the icon itself and that works.
```jsx
<DrawerItem
  icon={() => <FontAwesome5 testID="invite_friends" name="user-friends"  {...iconProps} />}
  label="Invite Friends"
  onPress={() => navigation.navigate(...)}
/>
```

If you don't have an icon, maybe you can pass in a temporary empty `View` to the `icon` prop with `testID` and that might work.  Haven't tried it out, though.

Also it [looks like testing for the entire package is pretty sparce.](https://github.com/react-navigation/react-navigation/tree/main/packages/drawer/src/__tests__)  What would mods recommend for adding tests for the functionality?

lindan4:
[...]

PR #11168, opened 2023-01-20
feat: add testID and accessibilityLabel to DrawerItem

andrewtremblay:
Added support for testIDs for the tappable DrawerItem and the DrawerItem label

Addresses https://github.com/react-navigation/react-navigation/issues/7492 and https://github.com/react-navigation/react-navigation/issues/8906

This was attempted before with https://github.com/react-navigation/react-navigation/pull/11009 but that PR was closed due to an update in the main branch.

### DIFF
--- a/packages/drawer/src/views/DrawerItem.tsx
+++ b/packages/drawer/src/views/DrawerItem.tsx
@@ -81,6 +81,25 @@ type Props = {
    * Whether label font should scale to respect Text Size accessibility settings.
    */
   allowFontScaling?: boolean;
+
+  /**
+   * Accessibility label for drawer item.
+   *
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR! Can you remove the `labeltestID`? Ideally users would use the test tool's methods to check the contained text.

Also remove the comments about `LinkPressable` since it's not relevant to users since it's an implementation detail.

andrewtremblay:
@satya164 your change requests have been addressed in [b19a2c5](https://github.com/react-navigation/react-navigation/pull/11168/commits/b19a2c56ffcd49c5d64327002167dbd47f30353f)

satya164 (approved):
Thank you!

PR #11170, opened 2023-01-22
feat: add tabBarInitialLayout to TabBar component

LukaRusadze:
**Motivation**

Just like the TabView component, the TabBar component depends on onLayout to set up different parts of the component, like the indicator, which causes a flicker when the TabBar component mounts.

I've implemented the same approach as the TabView component. 

Passing tabBarInitialLayout to the Navigator component, which gets passed to react-native-tab-view's TabBar component, eases the flicker that happens on the component mount.


satya164:
Can you provide more details about the issue such as some videos of before and after?

Issue #958, opened 2017-04-07
Tabbed Scenes to hide/show tabbar on listview scroll down/up?


MovingGifts:
**This is a very common use case:**

When you are on a scene with the bottom `tabBar` shown, as you scroll upwards with your finger, say `300` offset y, the `tabBar` hides for more real estate on the screen, and when you scroll down, again, say `300` offset y, it shows the `tabBar` again.

Using `react-native-router-flux`, I did it like this `Actions.refresh({ hideTabBar: true/false })`

How can this be implemented in `react-navigation`?

xiaokekeT:
this? https://github.com/react-community/react-navigation/issues/480

kevinold:
@ericvicenti 
Thanks for all the hard work in code and coordination with react-navigation!

I agree this would be a welcomed feature.  I've limited experience in this area, but wanted to ask for a bit of direction and/or if this is on the current roadmap for react-navigation?

MovingGifts:
bump

dayudodo:
waiting for answer...

cooperka:
This seems like a great idea! There's no way to do this currently as far as I know, but I'm sure a PR would be welcome.

cooperka:
It actually might be easier to implement this on a per-project basis by toggling `tabBarVisible`. Even in that case, we could add it to the demo app to show it off and make it easier for others to implement.

MovingGifts:
@cooperka How would you toggle the `tabBarVisible` to false when scrolling in a screen with a visible tab bar?

cooperka:
@MovingGifts it would probably look something like this:

```js
static navigationOptions = ({ navigation }) => ({
  tabBarVisible: navigation.state.params.tabBarVisible,
});
```

And to toggle it on scroll:

```js
onScroll() {
  const shouldShowTabBar = /* some logic to determine whether to show it */;
  this.props.navigation.setParams({ tabBarVisible: shouldShowTabBar });
}
```

Here are some relevant docs: https://reactnavigation.org/docs/navigators/tab#Screen-Navigation-Options

MovingGifts:
@cooperka Hmm... Doesn't work for me. Does it work for you?

cooperka:
@MovingGifts it's actually just pseudocode, I haven't implemented it myself. Can you post a link to a working demo app so I can take a look?

MovingGifts:
[...]

PR #11172, opened 2023-01-25
feat: add tabBarShown option to bottom tab navigation bar

acro5piano:
Add `tabBarShown` option to `BottomTabBar`.

**Motivation**

It is a common use case to hide `BottomTabBar` with smooth animation, e.g. scroll to hide like LinkedIn app.

This PR partially solves the issue: https://github.com/react-navigation/react-navigation/issues/958

Using `react-native-router-flux,` I did it like this `Actions.refresh({ hideTabBar: true/false })`.

**Test plan**

The following option will become available.

```tsx
const Tabs = createBottomTabNavigator();

<Tabs.Navigator tabBarShown={false}>
  // ...
```

https://user-images.githubusercontent.com/10719495/214465895-48dbf45a-4b8b-4737-8835-bdb6f3679bce.MOV

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -269,6 +269,10 @@ export type BottomTabNavigationConfig = {
    */
   tabBar?: (props: BottomTabBarProps) => React.ReactNode;
   /**
+   * Function that returns a React element to display as the tab bar.
+   */
+  tabBarShown?: boolean;
+  /**
    * Safe area insets for the tab bar. This is used to avoid elements like the navigation bar on Android and bottom safe area on iOS.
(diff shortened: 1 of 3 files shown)


satya164:
Thank you for the PR. The `tabBarVisible` option was removed in 6.x for few reasons:

- It is now possible to hide the tab bar with `tabBarStyle: { display: 'none' }`
- Showing/hiding the tab bar during the screen transition messes up animations, can cause jumps or bugs like leaving a blank space with native-stack, so we advise against it

This is the approach we recommend for hiding tab bar: https://reactnavigation.org/docs/hiding-tabbar-in-screens/

acro5piano:
@satya164 Okay thanks!

PR #10027, opened 2021-10-11
feat(drawer): add `accessibilityCloseDrawerLabel` option

grgr-dkrk:
fix #10026

Added `accessibilityCloseDrawerLabel` to the drawer options to read the appropriate message when TalkBack focuses on the drawer overlay.

TalkBack reads "Close the drawer navigation".

PR #11184, opened 2023-01-31
fix: added close drawer accessibility tap area

mikegarfinkle:
**Motivation**

Currently, a user can tap the `Overlay` next to an open drawer to close the drawer. This is useful in an Accessibility setting when swipe actions are not obvious to the user. Unfortunately, it is not obvious to someone with a visual deficit that this area can be tapped to close the drawer, as this overlay does not have an accessibilityLabel prop exposed. 

This PR exposes an `accessibilityLabel` prop on this overlay component and modifies existing accessibility props to allow for accessibility tap permissions on the overlay component. It then surfaces this prop at the `Drawer` level as `closeAccessibilityLabel`.
 
**Test plan**

Open up the example app with TalkBack (Android) or VoiceOver (iOS) enabled. Tap on the overlay next to the drawer and observe that it will now read out as "Close drawer".

### DIFF
--- a/packages/drawer/src/types.tsx
+++ b/packages/drawer/src/types.tsx
@@ -172,6 +172,11 @@ export type DrawerNavigationOptions = HeaderOptions & {
   overlayColor?: string;
 
   /**
+   * Screen readers will read this string when the user presses on the overlay adjacent to an opened drawer. Defaults to "Close drawer".
+   * **/
+  overlayAccessibilityLabel?: string;
+
+  /**
    * Style object for the component wrapping the screen content.
    */
   sceneContainerStyle?: StyleProp<ViewStyle>;
@@ -310,4 +315,5 @@ export type DrawerProps = {
   swipeEdgeWidth: number;
   swipeEnabled: boolean;
   swipeVelocityThreshold: number;
+  overlayAccessibilityLabel?: string;
 };
--- a/packages/drawer/src/views/modern/Drawer.tsx
+++ b/packages/drawer/src/views/modern/Drawer.tsx
@@ -61,6 +61,7 @@ export default function Drawer({
   swipeEdgeWidth,
   swipeEnabled,
   swipeVelocityThreshold,
+  overlayAccessibilityLabel,
 }: DrawerProps) {
   const getDrawerWidth = (): number => {
     const { width = DEFAULT_DRAWER_WIDTH } =
@@ -378,11 +379,11 @@ export default function Drawer({
                   toggleDrawer({ open: false, isUserInitiated: true })
                 }
                 style={overlayStyle}
+                accessibilityLabel={overlayAccessibilityLabel}
               />
             ) : null}
           </Animated.View>
           <Animated.View
-            accessibilityViewIsModal={isOpen && drawerType !== 'permanent'}
             removeClippedSubviews={Platform.OS !== 'ios'}
             style={[
               styles.container,
--- a/example/src/index.tsx
+++ b/example/src/index.tsx
@@ -230,6 +230,7 @@ export default function App() {
               <Drawer.Navigator
                 screenOptions={{
                   drawerType: isLargeScreen ? 'permanent' : undefined,
+                  overlayAccessibilityLabel: 'Close drawer',
                 }}
               >
                 <Drawer.Screen
--- a/packages/drawer/src/views/DrawerView.tsx
+++ b/packages/drawer/src/views/DrawerView.tsx
@@ -113,6 +113,7 @@ function DrawerViewBase({
       Platform.OS !== 'windows' &&
       Platform.OS !== 'macos',
     swipeMinDistance = 60,
+    overlayAccessibilityLabel,
   } = descriptors[focusedRouteKey].options;
 
   const [loaded, setLoaded] = React.useState([focusedRouteKey]);
@@ -296,6 +297,7 @@ function DrawerViewBase({
         statusBarAnimation={drawerStatusBarAnimation}
         keyboardDismissMode={keyboardDismissMode}
         drawerType={drawerType}
+        overlayAccessibilityLabel={overlayAccessibilityLabel}
         drawerPosition={drawerPosition}
         drawerStyle={[
           {
--- a/packages/drawer/src/views/legacy/Drawer.tsx
+++ b/packages/drawer/src/views/legacy/Drawer.tsx
@@ -492,6 +492,7 @@ export default class DrawerView extends React.Component<DrawerProps> {
       renderDrawerContent,
       renderSceneContent,
       gestureHandlerProps,
+      overlayAccessibilityLabel,
     } = this.props;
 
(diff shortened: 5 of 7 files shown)


packages/drawer/src/views/modern/Drawer.tsx
```diff
@@ -378,11 +379,12 @@ export default function Drawer({
                   toggleDrawer({ open: false, isUserInitiated: true })
                 }
                 style={overlayStyle}
+                accessibilityLabel={closeAccessibilityLabel}
               />
             ) : null}
           </Animated.View>
           <Animated.View
-            accessibilityViewIsModal={isOpen && drawerType !== 'permanent'}
+            accessibilityViewIsModal={!isOpen}
```
mikegarfinkle, line 387: now, for iOS (accessibilityViewIsModal is iOS only prop), no matter the type of drawer, it will always allow the overlay to be accessible tappable when the drawer is open.

before, it would only allow the overlay to be tappable if it wasn't open for the permanent type (which doesn't make much sense)

satya164: I'm not sure what do you mean since the overlay isn't (or shouldn't be) rendered for permanent drawer.

mikegarfinkle:
> Thanks for the PR. I have same comment as [this PR](https://github.com/react-navigation/react-navigation/pull/10027).
> 
> If this is not correct, can you share a video of how a native drawer works with talkback so and other accessibility tech so the drawer implementation can match it?

Hi! Thanks so much for taking a look at this. To address your question (I've used Gmail as the example app to show default native behaviour):

**Android setting:**

In Android, there is no Talkback accessibility for the overlay area in a drawer. This is because the Android ever-present back button can dismiss the drawer AND the overlay area is not tappable.

image

However, in react-navigation drawer, the overlay _is_ tappable and currently does not have an accessibility label to tell visually impaired users what tapping will do. 

image

As all tappable components should have an accessibility label, I see two ways forward: 1) eliminate tappable overlay (swipe would still work, but this would involve simply pressing on the overlay) in Android, or 2) keep the overlay and add accessibility label as I have done in this PR. 

**iOS setting:**

On iOS, Navigation Drawers are not a native design pattern. However, implementations on a drawer have been created, such as with the Gmail app.  In the Gmail implementation, Google has made the overlay as a tappable button (with an accessibility label that is read by iOS VoiceOver) to help visually impaired users dismiss the drawer (as there is no ever-present back button like there is in Android). 

image

I believe react-navigation drawer in iOS would benefit from a similar approach: ie. making the overlay tappable to dismiss (which is already done) _with_ an accessibility label (which this PR adds).

packages/drawer/src/views/modern/Drawer.tsx
mikegarfinkle: This logic does two things: 

1) Eliminates the dependancy on permanent drawerType. As overlay is not present when type == permanent, I don't see a reason to have a second check here. In other words, overlay specific logic can ignore the permanent drawerType altogether because permanent type does not even implement overlay.

2) Allows the overlay to be accessibility selectable (ie readable by VoiceOver) when the drawer is open in iOS.

satya164: The problem with this logic is that with permanent drawer, the open and closed state of the drawer doesn't matter. It's always present and is interactable. So it needs to be treated differently. The modal part of the drawer only matters when it's not permanent.

Though reading the description of this prop, I think it shouldn't be used here at all. The a11y of the overlay should already be taken care of the props on it so there are no siblings to hide. The closed drawer also doesn't appear like a modal so the prop feels like a misuse.

packages/drawer/src/types.tsx
```diff
@@ -220,6 +220,10 @@ export type DrawerNavigationOptions = HeaderOptions & {
    * Only supported on iOS and Android.
    */
   freezeOnBlur?: boolean;
+  /**
+   * If set, screen readers will read out this string when the user presses on the overlay adjacent to an opened drawer, which will close the drawer.
+   * **/
+  closeAccessibilityLabel?: string;
```
satya164, line 226: Can you change it to `overlayAccessibilityLabel`? Also, move it next to `overlayColor` so that related props are grouped

satya164:
> In Android, there is no Talkback accessibility for the overlay area in a drawer. This is because the Android ever-present back button can dismiss the drawer AND the overlay area is not tappable.

This isn't correct. I use Android as my daily driver and the drawer is always dismissable by tapping on the overlay area. Regarding the back button, there is no such button in recent versions of Android. There's a system back gesture on Android which does the same thing.

> Google has made the overlay as a tappable button (with an accessibility label that is read by iOS VoiceOver) to help visually impaired users dismiss the drawer

Makes sense. Then we can do the same. It may also be necessary to `accessibilityRole="button"`. And default the value of the accessibility label to `"Close drawer"`

packages/drawer/src/views/modern/Drawer.tsx
mikegarfinkle: Agree! Getting rid of this prop altogether would work too!

satya164 (changes requested):

mikegarfinkle:
> > In Android, there is no Talkback accessibility for the overlay area in a drawer. This is because the Android ever-present back button can dismiss the drawer AND the overlay area is not tappable.
> 
> This isn't correct. I use Android as my daily driver and the drawer is always dismissable by tapping on the overlay area. Regarding the back button, there is no such button in recent versions of Android. There's a system back gesture on Android which does the same thing.
> 

Ah let me clarify. Google, at least in the Gmail app, hasn't made it _accessible_. So they have made it tappable, but they didn't add an accessibility label on this tap functionality. This I think we can add though for react navigation!

packages/drawer/src/types.tsx
mikegarfinkle: Can do! Will add to next push.

mikegarfinkle:
> > Google has made the overlay as a tappable button (with an accessibility label that is read by iOS VoiceOver) to help visually impaired users dismiss the drawer
> 
> Makes sense. Then we can do the same. It may also be necessary to `accessibilityRole="button"`. And default the value of the accessibility label to `"Close drawer"`

Good call on explicitly adding the accessibilityRole, and will add default value too (being cognizant that will always be reading english if this is not set by the dev, regardless of local).

mikegarfinkle:
> **Android setting:**
> 
> In Android, there is no Talkback accessibility for the overlay area in a drawer. This is because the Android ever-present back button can dismiss the drawer AND the overlay area is not tappable.

I was wrong here (thanks for pointing this out!). The overlay IS tappable and dismisses the drawer in native Android. However, Google has not made this tap functionality _accessible_. I think this is a bit of an oversight, and we could make this accessible, as I have done in my PR. For react-navigation drawer Android.

satya164 (approved):
Thank you!

PR #11193, opened 2023-02-03
feat: Add useFocusedTabPressed

IMaz6:
**Motivation**
Sometimes you need to do an action when the currently focused bottom tab has been pressed. Loosely copied from `useScrollToTop`, this hook uses a callback parameter instead of a scroll ref, and invokes the callback when the currently focused tab bar is pressed.

Usage:
```
const someActionWhenFocusedTabPressedAgain = useCallback(() => { 
  // focused tab button was pressed again
}, []);

useFocusedTabPressed(someActionWhenFocusedTabPressedAgain);
```
Related:
https://react-navigation.canny.io/feature-requests/p/allow-function-in-usescrolltotop

### DIFF
--- a/packages/native/src/useFocusedTabPressed.tsx
+++ b/packages/native/src/useFocusedTabPressed.tsx
@@ -1,0 +1,52 @@
+import { useEffect } from "react";
+import {
+    EventArg,
+    NavigationProp,
+    useNavigation,
+    useRoute,
+  } from "@react-navigation/core";
+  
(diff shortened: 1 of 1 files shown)


satya164:
Thank you for the PR. However I don't want to add this hook because it's not a common use case and can already be achieved in userland by either copying this hook or providing a ref object with your own `scrollToTop` function. So I'm closing this for now.

PR #11194, opened 2023-02-03
feat: add support to override pager's overScrollMode

ouabing:
Add support to override pager's overScrollMode.

**Motivation**

The default value of `overScrollMode` in `react-native-pager-view` is set to `auto`, which makes the first and last scenes streched when over scroll the `TabView`. We should provide a way to change the default behavior.


satya164 (approved):
Thanks for the PR

Issue #11198, opened 2023-02-05
[react-native-tab-view] Setting TabBar background color to transparent causes interaction animation to overflow


acb:
### Current behavior

When you set `style={{backgroundColor: "transparent"}}` (or `style={{backgroundColor: undefined}}`) on the TabBar component on Android the interaction animation overflows the tab bar and looks quite jarring.

How it looks with a background color -- the interaction animation is confined to the tab bar element:
![Screenshot_1675574016](https://user-images.githubusercontent.com/4690898/216803396-0444b56c-7a79-4aff-b601-8ff9c8237040.png)

How it looks without a background color -- the interaction animation overflows both the tab bar and the content below:
![Screenshot_1675574030](https://user-images.githubusercontent.com/4690898/216803421-530a6e1a-7ad4-4e02-bef3-253212b69c0f.png)

### Expected behavior

With a transparent tab bar background the interaction animation should still be limited to the physical bounds of the TabBar component.

### Reproduction

https://snack.expo.dev/@taboobat/react-native-tab-view-bug

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-bottom-tabs
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [X] react-native-tab-view
- [ ] flipper-plugin-react-navigation

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native | 6.1.3
| react-native-tab-view                  | 3.3.4

acb:
> Can you verify that the issue still exists after upgrading to the latest versions of these packages?

Indeed it does my bot friend.

okwasniewski:
Hey @acb, 

We've released tab-view `3.4.0` it now allows you to pass `android_ripple={{borderless: false}}`. Let me know if this fixes your bug, so I can close this issue.

acb:
@okwasniewski sorry I missed your ping -- I just updated to 3.4.0 and confirm that the ripple is working as expected now!  Thank you for the quick turnaround!

PR #11203, opened 2023-02-06
fix: allow users to pass `android_ripple` config in TabView

okwasniewski:
**Motivation**

This PR solves an issue with setting transparent background color of tab-view on Android. It allows users to overwrite `android_ripple` config. (#11198) 

### Recording

https://user-images.githubusercontent.com/52801365/217030596-4a4fe2ce-c974-4c37-b3ee-84733bf1eb79.mp4

**Test plan**

Run the test app, passing `{ backgroundColor: transparent }` to tabBar.

### DIFF
--- a/packages/material-top-tabs/src/types.tsx
+++ b/packages/material-top-tabs/src/types.tsx
@@ -9,7 +9,12 @@ import type {
   TabNavigationState,
 } from '@react-navigation/native';
 import type React from 'react';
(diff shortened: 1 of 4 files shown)


satya164 (approved):
Thank you

frankusu:
would it be possible to add this to BottomTabBar as well ?

Issue #11086, opened 2022-12-08
Lots of updates useLinking in safari: SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds


AntonYushkevich:
### Current behavior

At every screen update history.replace is called that cause _SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds_ error in Safari. 

That peace of code trigging on each update inside components. 
<img width="736" alt="image" src="https://user-images.githubusercontent.com/25014691/206484785-d94854d5-e71f-4ded-a95b-96f5fde24503.png">

Maybe because that useEffect is calling without dependencies
<img width="786" alt="image" src="https://user-images.githubusercontent.com/25014691/206494934-56fb60e2-a828-47f3-b607-5e9c1c4a4ca2.png">

### Expected behavior

history.replace called on screen changes

### Reproduction

http://

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [X] @react-navigation/​native-stack
- [ ] react-native-tab-view
- [ ] flipper-plugin-react-navigation

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.10
| @react-navigation/bottom-tabs          | []
| @react-navigation/drawer               | []
| @react-navigation/material-bottom-tabs | []
| @react-navigation/material-top-tabs    | []
| @react-navigation/stack                | []
| @react-navigation/native-stack         | 6.2.5
| react-native-safe-area-context         | []
| react-native-screens                   | 3.10.0
| react-native-gesture-handler           | []
| react-native-reanimated                | []
| react-native-tab-view                  | []
| react-native-pager-view                | []
| react-native                           | 0.68.1
| expo                                   | []
| node                                   |  16.17.0
| npm or yarn                            |  1.22.18

jparkrr:
The issue still exists.

[...]

PR #11206, opened 2023-02-07
fix: multiple useEffect calls in useLinking hook

AntonYushkevich:
**Motivation**
[Issue](https://github.com/react-navigation/react-navigation/issues/11086). 

**Test plan**
Run test app in safari web and any native mobile and make sure everything works correctly, without [similar message](https://stackoverflow.com/questions/67038317/unhandled-rejection-securityerror-attempt-to-use-history-pushstate-more-tha).


satya164:
Thanks for the PR. I'm closing this since the diff doesn't contain any changes as this was already published.

PR #11233, opened 2023-02-19
refactor: use useContext in 'safeAreaProviderCompat'

Ne3l:
**Motivation**

I was profiling my App and found that using react-navigation adds a big stackStrace

image

Digging into the codebase i found the `SafeAreaProviderCompat` which appears 4 times on my stack.

Changed the consumer to the useContext which is the recommended approach according to the react docs and provide 1 less stack.

https://beta.reactjs.org/reference/react/createContext#consumer.

**Doubt**

One doubt i had doing this is that, on the react-navigation docs it's recommended to have a `NavigationContainer` at the root level.

https://reactnavigation.org/docs/getting-started/#wrapping-your-app-in-navigationcontainer

If this is the case, why not put the SafeAreaProvider there and you ensure it is present on all the children? I'm not sure if there is any case that react-navigation supports where the navigation don't respect the safeArea(Full screen apps maybe?) so i prefer to just changed this file.

**Test plan**

Tested on my app the change and didn't spot any issue. AFAIK, the useContext is equivalent as the consumer so no more impact.


satya164 (approved):
Thank you

billouboq:
Will this be available on version 6 ?

satya164:
@billouboq nope. only bug fixes will be released for version 6, and this is not a bug fix.

PR #11240, opened 2023-02-24
feat: gesture and transition events to drawer

BeeMargarida:
**Motivation**

Missing gesture and transition event handlers in the drawer navigation.

**Test plan**

I don't see any tests that deal with gestures. Should they be made for this PR?

The feature can be tested by adding callbacks similar to the ones below to the `DrawerView` example:

```
onGestureStart={() => {
  console.log('gesture start');
}}
onGestureEnd={() => {
  console.log('gesture end');
}}
onGestureCancel={() => {
  console.log('gesture cancel');
}}
onTransitionStart={(closing) => {
  console.log('transition start', closing);
}}
onTransitionEnd={(closing) => {
  console.log('transition end', closing);
}}
```


satya164 (changes requested):
Thanks a lot for the PR. I have a small comment. Can you replace the usage of `useCallback` with `use-latest-callback`? (they may not be in the deps of the relevant package so we'd need to add it).

satya164 (approved):
Thank you!

PR #11243, opened 2023-02-24
feat: add ability to customize the fonts with the theme

satya164:
**Motivation**

React Navigation 5 introduced a theming API. However, the theme only supported customizing colors. With this change, the theme can be used to customize font family as well - making it simpler to use a specific font across all of React Navigation's UI elements.

**Test plan**

Specify a custom font in the theme and check that React Navigation's UI elements use the new font. The following screenshots show the usage of the "Lato" font.

SCR-20230224-qy4
SCR-20230224-qyl
SCR-20230224-qz6
SCR-20230224-qzt

### DIFF
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -59,15 +59,22 @@ export function HeaderConfig({
   title,
   canGoBack,
 }: Props): JSX.Element {
-  const { colors } = useTheme();
+  const { colors, fonts } = useTheme();
   const tintColor =
     headerTintColor ?? (Platform.OS === 'ios' ? colors.primary : colors.text);
 
   const headerBackTitleStyleFlattened =
-    StyleSheet.flatten(headerBackTitleStyle) || {};
+    StyleSheet.flatten([headerBackTitleStyle, fonts.regular]) || {};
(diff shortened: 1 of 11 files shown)


packages/native-stack/src/views/HeaderConfig.tsx
```diff
@@ -73,16 +73,20 @@ export default function HeaderConfig({
 
   const [backTitleFontFamily, largeTitleFontFamily, titleFontFamily] =
     processFonts([
-      headerBackTitleStyleFlattened.fontFamily,
-      headerLargeTitleStyleFlattened.fontFamily,
-      headerTitleStyleFlattened.fontFamily,
+      headerBackTitleStyleFlattened.fontFamily ?? fonts.regular.fontFamily,
+      headerLargeTitleStyleFlattened.fontFamily ??
```
kacperkapusciak, line 77: `headerLargeTitleStyleFlattened` doesn't seem to be properly defaulted here for some reason

image


`headerTitleStyleFlattened` is also not getting defaulted in the next line but it's defaulted in next few lines in `titleFontWeight` variable so the end result for small title is OK

satya164: Fixed

kacperkapusciak (approved):

PR #11248, opened 2023-02-27
feat: add animation prop to stack

mdros:
PR with documentation: https://github.com/react-navigation/react-navigation.github.io/pull/1215

**Motivation**

This PR brings a new `animation` property to Stack, which adds a simpler way to animate a screen.

<details><summary>Code example</summary>
<p>

```javascript
import { createStackNavigator } from '@react-navigation/stack';

function FirstScreen({ navigation }: any) {
  return (
      <Text>First Screen</Text>
      <Button
        title="Go to Second Screen"
        onPress={() => navigation.navigate('SecondScreen')}
      />
    </View>
  );
}

function SecondScreen() {
  return (
      <Text>Second Screen</Text>
    </View>
  );
}

  return (
    <NavigationContainer>
      <Stack.Navigator screenOptions={{ animation: 'slide_from_right' }}>
        <Stack.Screen name="FirstScreen" component={FirstScreen} />
        <Stack.Screen name="SecondScreen" component={SecondScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}
```

</p>
</details>

### DIFF
--- a/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
+++ b/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
@@ -63,6 +63,13 @@ export function forHorizontalIOS({
   };
 }
 
+export function forHorizontalIOSInverted(
+  props: StackCardInterpolationProps
+): StackCardInterpolatedStyle {
+  props.inverted = new Animated.Value(-1);
+  return forHorizontalIOS(props);
+}
+
 /**
  * Standard iOS-style slide in from the bottom (used for modals).
  */
--- a/packages/stack/src/TransitionConfigs/TransitionPresets.tsx
+++ b/packages/stack/src/TransitionConfigs/TransitionPresets.tsx
@@ -1,6 +1,6 @@
 import { Platform } from 'react-native';
 
-import type { TransitionPreset } from '../types';
+import type { StackAnimationTypes, TransitionPreset } from '../types';
 import {
   forBottomSheetAndroid,
   forFadeFromBottomAndroid,
@@ -150,3 +150,19 @@ export const ModalTransition = Platform.select({
   ios: ModalPresentationIOS,
   default: BottomSheetAndroid,
 });
+
+type AnimationTransitionMap = {
+  [key in StackAnimationTypes]: TransitionPreset;
+};
+
+// TODO: Add proper transitions
+export const AnimationTransitions: AnimationTransitionMap = {
+  default: DefaultTransition,
+  fade: ModalFadeTransition,
+  fade_from_bottom: FadeFromBottomAndroid,
+  none: DefaultTransition,
+  simple_push: DefaultTransition,
+  slide_from_bottom: BottomSheetAndroid,
+  slide_from_left: DefaultTransition,
+  slide_from_right: SlideFromRightIOS,
+};
--- a/packages/stack/src/TransitionConfigs/TransitionPresets.tsx
+++ b/packages/stack/src/TransitionConfigs/TransitionPresets.tsx
@@ -6,6 +6,7 @@ import {
   forFadeFromBottomAndroid,
   forFadeFromCenter as forFadeCard,
   forHorizontalIOS,
+  forHorizontalIOSInverted,
   forModalPresentationIOS,
   forRevealFromBottomAndroid,
   forScaleFromCenterAndroid,
@@ -36,6 +37,11 @@ export const SlideFromRightIOS: TransitionPreset = {
   },
   cardStyleInterpolator: forHorizontalIOS,
   headerStyleInterpolator: forFade,
+};
+
+export const SlideFromLeftIOS: TransitionPreset = {
+  ...SlideFromRightIOS,
+  cardStyleInterpolator: forHorizontalIOSInverted,
 };
 
 /**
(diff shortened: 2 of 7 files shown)


packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -139,6 +139,11 @@ export default class StackView extends React.Component<Props, State> {
     const isAnimationEnabled = (key: string) => {
       const descriptor = props.descriptors[key] || state.descriptors[key];
 
+      const animation = descriptor?.options?.animation;
+      if (animation && animation === 'none') {
```
satya164, line 143: The `animation` check is unnecessary

```suggestion
      if (animation === 'none') {
```

packages/stack/src/views/Stack/CardStack.tsx
```diff
@@ -275,21 +276,30 @@ export default class CardStack extends React.Component<Props, State> {
...
-        headerStyleInterpolator = defaultTransitionPreset.headerStyleInterpolator,
         cardOverlayEnabled = (Platform.OS !== 'ios' &&
           optionsForTransitionConfig.presentation !== 'transparentModal') ||
           getIsModalPresentation(cardStyleInterpolator),
       } = optionsForTransitionConfig;
 
+      if (animation != null) {
+        cardStyleInterpolator =
+          AnimationTransitions[animation].cardStyleInterpolator;
+      }
```
satya164, line 301: Does `animation` only affect `cardStyleInterpolator`? It doesn't seem enough. Specifying an animation should also configure the correct gesture direction, animation curves etc. depending on what's appropriate for the animation. It's essentially a replacement for a transition preset, not only `cardStyleInterpolator`.

The `animation` option shouldn't override `cardStyleInterpolator`. It should be the other way around. More specific options override less specific ones, so user should be able to specify `animation` and override part of that configuration such as transition spec, card style interpolator etc.

packages/stack/src/views/Stack/Card.tsx
```diff
@@ -33,6 +34,7 @@ import ModalStatusBarManager from '../ModalStatusBarManager';
 import CardSheet, { CardSheetRef } from './CardSheet';
 
 type Props = ViewProps & {
+  animation?: StackAnimationTypes;
```
satya164, line 37: This prop doesn't seem to be used in this in this component. There are also both `animation` and `styleInterpolator` passed. Only one needs to be passed since `styleInterpolator` can be determined from the `animation` for whichever parent is passing it.

packages/stack/src/types.tsx
```diff
@@ -306,6 +315,21 @@ export type StackNavigationOptions = StackHeaderOptions &
...
+     * Supported values:
+     * - "default": use the platform default animation
+     * - "fade": fade screen in or out
+     * - "fade_from_bottom": fade screen in or out from bottom
+     * - "slide_from_bottom": slide in the new screen from bottom
+     * - "slide_from_right": slide in the new screen from right
+     * - "slide_from_left": slide in the new screen from left
+     * - "none": don't animate the screen
+     *
+     * Only supported on iOS and Android.
```
satya164, line 330: This isn't correct. Animations are supported on all platforms. They are only enabled by default on iOS and Android which can be configured with `animationEnabled: false`

packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
```diff
@@ -63,6 +63,13 @@ export function forHorizontalIOS({
   };
 }
 
+export function forHorizontalIOSInverted(
+  props: StackCardInterpolationProps
+): StackCardInterpolatedStyle {
+  props.inverted = new Animated.Value(-1);
```
satya164, line 69: Don't mutate objects. The mutation will alter the object in the parent too, not only here.

packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
```diff
@@ -63,6 +63,13 @@ export function forHorizontalIOS({
...
+  return forHorizontalIOS(props);
```
satya164, line 70: Hardcoding `inverted` to `new Animated.Value(-1)` means any `inverted` property passed here will be ignored. This needs to respect the existing `inverted` value:

```suggestion
 { inverted, ...rest }: StackCardInterpolationProps
): StackCardInterpolatedStyle {
  return forHorizontalIOS({
    ...rest,
    inverted: Animated.multiply(inverted, -1),
  });
```

satya164 (changes requested):

packages/stack/src/TransitionConfigs/TransitionPresets.tsx
```diff
@@ -150,3 +156,17 @@ export const ModalTransition = Platform.select({
...
+
+export const AnimationTransitions: AnimationTransitionMap = {
+  default: DefaultTransition,
+  fade: ModalFadeTransition,
+  fade_from_bottom: FadeFromBottomAndroid,
+  none: DefaultTransition,
+  slide_from_bottom: BottomSheetAndroid,
+  slide_from_left: SlideFromLeftIOS,
+  slide_from_right: SlideFromRightIOS,
+};
```
satya164, line 172: Instead of an annotation, it's simpler to let TS infer the type:

```suggestion
export const AnimationTransitions = {
  default: DefaultTransition,
  fade: ModalFadeTransition,
  fade_from_bottom: FadeFromBottomAndroid,
  none: DefaultTransition,
  slide_from_bottom: BottomSheetAndroid,
  slide_from_left: SlideFromLeftIOS,
  slide_from_right: SlideFromRightIOS,
} as const;
```

satya164:
Closing due to no response.

PR #11279, opened 2023-03-15
chore: rename type files from .tsx to .ts

liamjones:

**Motivation**

The type files are currently `.tsx`, this causes projects that import them to require the `--jsx` option even though there's no need for it (they contain no JSX tags). In our instance we have a helper function that needs the types but never renders JSX so doesn't have the option enabled in its tsconfig.json. This results in TS6142 errors:

image

**Test plan**

Automated tests pass, changing the extension shouldn't affect people importing the file unless they specify the full filename including the extension (unlikely)


satya164:
We use `.tsx` extension for all files as a convention in the repo so I won't merge this PR.

Here you're importing from internal files. This is not a supported use case and may break at any time. All our files are compiled when publishing so the file extensions in the repo don't matter. You should import anything you need from the public API from the `@react-navigation/x` packages.

liamjones:
Ah, I missed `Route` was in the umbrella export, thanks.

PR #11281, opened 2023-03-15
Update types.tsx

nikmantulenko:
I'm proposing to add better typing possibilities for custom header components

**Motivation**

using NativeStackHeaderProps type for 'options.header = (props: NativeStackHeaderProps) => ...
gives no type checking for route & navigation props


osdnk:
Hi, @nikmantulenko 

Thanks for the PR!

I think this might be a nice approach, but at this stage, this seems to be a very atomic change, whereas there are multiple places with a very similar approach implemented (to name a few, consider all components in the `stack`).

So if you are up to making the holistic migration, we will be happy to merge. Currently, we find this use case a bit rare (to have header parametrized by the route) so we will probably not spend the effort actively on implementing that. But, as stated earlier, we are happy to review the follow-up change.

Issue #11306, opened 2023-03-29
HeaderBackButton "label" prop not working properly


gkasdorf:
### Current behavior

Inside of `getState()`, I am attempting to modify the `onPress` event for the `HeaderBackButton`. I am using `setOptions()` to modify this under a certain condition.

```js
navigation.setOptions({
    headerLeft: () =>  <HeaderBackButton label="Back" onPress={navigation.popToTop} />
});
```

Although the `popToTop()` is firing `onPress`, the only thing that is rendering is the chevron for the back button, not the label specified. 
![Screenshot 2023-03-28 at 6 21 43 PM](https://user-images.githubusercontent.com/127079571/228410263-4da0a2d5-5873-4833-bed6-a557a93ec64b.png)

### Expected behavior

According to [the documentation](https://reactnavigation.org/docs/elements/#headerbackbutton) it appears that whatever the `label` prop is set to is what should be rendered as the "title" for the button. In this case, only the chevron is appearing.

### Reproduction

https://snack.expo.dev/@gkasdorf/headerbackbutton-label-prop

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^6.1.6
| @react-navigation/bottom-tabs          | ^6.5.7
| @react-navigation/native-stack         | ^6.9.12
| react-native-safe-area-context         | 4.5.0
| react-native-screens                   | ~3.20.0
| react-native                           | 0.71.4
| expo                                   | ~48.0.9
| node                                   | 19.8.1
| npm or yarn                            | npm 9.5.1

gkasdorf:
I should have tested this before, but it appears that `labelVisible` is not actually defaulted as true even on iOS, although the documentation says that it should be. Like I stated above, you are welcome to close this. However, I would recommend either updating the documentation or (probably more useful) just have it default to true on iOS.

gkasdorf:
https://github.com/react-navigation/react-navigation/pull/11308

PR #11308, opened 2023-03-29
Add default value to labelVisible in HeaderBackBox

gkasdorf:
Should default to "true" on iOS, "false" on Android. Currently no default value is given, contrary to documentation.

**Motivation**

The documentation states that `HeaderBackButton` should offer the default value of `true` for `labelVisible` on iOS. However, currently there is no default value given, and you must set this value to true manually. A default true value on iOS would be much more useful than a false value, as it is uncommon to see a back button without a label on iOS.

https://github.com/react-navigation/react-navigation/issues/11306

**Test plan**

https://snack.expo.dev/@gkasdorf/headerbackbutton-label-prop


Dhananjay-JSR (commented):
we should rather update documentation as it seems to be the hacky solution to get stuff done

gkasdorf:
> we should rather update documentation as it seems to be the hacky solution to get stuff done

While I understand that, it think it should be noted that there was probably a reason the documentation was written that way in the first place. As far as iOS development goes, the default with a back button is always to show the label. I personally cannot think of a time that I have seen a navigation header with a back button and no label. I would also say that it seems to be - without them explicitly saying it - Apple's considered best practice to show a label. 

Whichever you decide, it isn't a huge difference, just definitely update the documentation if you decide to not implement this. Thanks.

satya164 (approved):
Thank you

Issue #10989, opened 2022-11-09
`useHeaderHeight();` returning incorrect value on phones with Dynamic Island


objectivecosta:
### Current behavior

Currently, the library reports a value of 103pt, which appears correct when you use Xcode's View Hierarchy debugger, but does not match the required value when running the app (appears to be ~97pt).

![image](https://user-images.githubusercontent.com/753361/200795439-43c5df28-5aea-43b9-a855-3b3f400817fa.png)

(Observe the cyan gap between the Navigation Bar and the Yellow Rectangle. The Rectangle's .top position is set to the value of `useHeaderHeight()`

This behaviour does not occur on pre-Dynamic Island phones. The gap seems to be 6pt on iPhone 14 Pro Max, don't know if more on other devices.

### Expected behavior

It is expected that the library will be able to handle this case and report the correct value when the dynamic island is involved.

### Reproduction

https://github.com/objectivecosta/DynamicProblems

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/​bottom-tabs
- [ ] @react-navigation/​drawer
- [ ] @react-navigation/​material-bottom-tabs
- [ ] @react-navigation/​material-top-tabs
- [ ] @react-navigation/​stack
- [X] @react-navigation/​native-stack

### Environment

- [✅] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.13
| @react-navigation/native-stack         | 6.9.1
| react-native-safe-area-context         | 4.4.1
| react-native-screens                   | 3.18.2
| react-native                           | 0.70.5
| node                                   | v16.13.2
| npm                             | 8.1.2

objectivecosta:
Debugging further with Xcode, I found that pure UIViewController/UIView reports a safe area inset of 59.0, just as expected. However, when adding a UINavigationController, the OS recalculates the safe area insets to be 97, the actual expected value.

(Logging from [my sample app](https://github.com/objectivecosta/DynamicNavigationTest), you can change the entry point in Main.storyboard to see both cases)

```
With iOS native Navigation Stack
Safe Area Top: 97.66666666666667
Safe Area Top + NavigationBar inferred (44) height: 141.66666666666669
NavigationBar computed height: 44.0
```

```
Without iOS native Navigation Stack
Safe Area Top: 59.0
[...]

PR #11338, opened 2023-04-20
fix: headerHeight on phones with dynamic island

dylancom:
**Motivation**

The headerHeight is currently 5px too tall on phones with Dynamic Island.
(As the statusbar became taller)
Also fixes incorrect height returned from the `useHeaderHeight()` hook: #10989

Scherm­afbeelding 2023-04-20 om 11 05 13

Scherm­afbeelding 2023-04-20 om 11 05 22

**Test plan**

Compare the headers between using the Stack Navigator and the Native Stack navigator (on an iPhone 14 simulator). You will notice the headerHeight is 5px larger on the (non-native) Stack Navigator.


objectivecosta (approved):
Yes, please! 💯 

(My approval does not count, but take it anyway! Thanks for fixing the issue I reported)

outaTiME:
(to my understanding) is possibly something deeper, I think it is a problem that is introduced with some version of the iOS 16.x SDK or just something inside `react-native-screens`.

At the time, I reported the bug in your repository, but I had no response yet:
https://github.com/software-mansion/react-native-screens/issues/1671

But still and at the moment this PR is the workaround I am using.

objectivecosta:
@outaTiME when I investigated this, the problem was that iOS itself gave no way to programatically get the correct value. Using the same method as here, natively, I got the same additional spacing. AFAIK, the only way is to get the actual height of the UIView + some constant or to fix it in code like this PR ¯\_(ツ)_/¯

outaTiME:
I understand, but at some point some iOS version must have adjusted the size because this problem did not occur prematurely on devices with dynamic island.
My builds before Expo 47 worked perfectly on these types of devices (I assume it was because the build was done on another version of the iOS SDK).
Probably the there is a tweak to be made here as well:
https://github.com/software-mansion/react-native-screens/blob/main/src/native-stack/utils/getDefaultHeaderHeight.tsx

dylancom:
@outaTiME today I submitted a PR there aswel: https://github.com/software-mansion/react-native-screens/pull/1784

dylancom:
I did some more research and came to the following conclusion (improved the PR).
The actual header height should remain 44. On phones with dynamic island, the status bar height (54) is smaller than the safe area top inset (59). And thus the static height 54 has to be used instead.

Resources:
https://useyourloaf.com/blog/iphone-14-screen-sizes/#iphone-14-pro
https://github.com/xamarin/Xamarin.Forms/issues/15584
https://www.figma.com/community/file/1248375255495415511/Apple-Design-Resources-%E2%80%93-iOS-17-and-iPadOS-17

outaTiME:
Amazing work @dylancom, it would be great if your PRs are approved 🙌

newme616:
yes this should be approved!

dylancom:
You're welcome! I was confused myself too, until I found that example...

packages/elements/src/Header/Header.tsx
```diff
@@ -48,6 +48,10 @@ export function Header(props: Props) {
 
   const isParentHeaderShown = React.useContext(HeaderShownContext);
 
+  // On models with Dynamic Island the status bar height is smaller than the safe area top inset.
+  const hasDynamicIsland = Platform.OS === 'ios' && insets.top === 59;
+  const statusBarHeight = hasDynamicIsland ? 54 : insets.top;
```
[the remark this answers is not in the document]
dylancom: Good suggestion, added improvements.

satya164 (approved):

packages/elements/src/Header/Header.tsx
```diff
@@ -48,6 +48,10 @@ export function Header(props: Props) {
 
   const isParentHeaderShown = React.useContext(HeaderShownContext);
 
+  // On models with Dynamic Island the status bar height is smaller than the safe area top inset.
+  const hasDynamicIsland = Platform.OS === 'ios' && insets.top > 50;
+  const statusBarHeight = hasDynamicIsland ? insets.top - 5 : insets.top;
+
```
itsramiel, line 54: Is this value of 5 accurate? Because it **looks** 1 pixel short 
Screenshot 2023-10-24 at 8 56 40 AM

https://github.com/react-navigation/react-navigation/issues/11655

outaTiME: yup, it should probably be 5.33 instead of 5

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -220,9 +220,17 @@ const SceneView = ({
       ? 0
       : insets.top;
 
+  // On models with Dynamic Island the status bar height is smaller than the safe area top inset.
+  const hasDynamicIsland = Platform.OS === 'ios' && topInset > 50;
+  const statusBarHeight = hasDynamicIsland ? topInset - 5 : topInset;
```
outaTiME, line 225: (same as above) probably 5.33 should be used instead of 5

outaTiME (commented):
The offset values should be fine-tuned to have a perfect visualization

PR #11378, opened 2023-05-18
fix: correct supported react-native-reanimated version

nikgraf:
**Motivation**

After upgrading expo I got this error message:

> Invalid `RNReanimated.podspec` file: [react-native-reanimated] Multiple versions of Reanimated were detected. Only one instance of react-native-reanimated can be installed in a project. You need to resolve the conflict manually. Check out the documentation: https://docs.swmansion.com/react-native-reanimated/docs/next/fundamentals/troubleshooting#multiple-versions-of-reanimated-were-detected 

I noticed due the react-navigation peerDependency specification 1.x of reanimated was installed. After realizing that you do support reanimated 2 & 3 I investigated and found this example: https://stackoverflow.com/a/47310882/837709

**Test plan**

Everything should work as before.

### DIFF
--- a/packages/drawer/package.json
+++ b/packages/drawer/package.json
@@ -66,7 +66,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">= 1.0.0",
-    "react-native-reanimated": ">= 1.0.0",
+    "react-native-reanimated": ">= 1.0.0 || >= 2.0.0 || >= 3.0.0",
     "react-native-safe-area-context": ">= 3.0.0",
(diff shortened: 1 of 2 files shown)


nikgraf:
Now I'm not sure this actually adds value or would fix the issue. It was a false positive as described here: https://github.com/software-mansion/react-native-reanimated/issues/3509#issuecomment-1311094594

kacperkapusciak:
Hi, thanks for a PR but I can't merge this as the changes you've proposed aren't related to the problem

Cheers ✌️

PR #11380, opened 2023-05-19
chore: update example app to SDK 48

okwasniewski:
**Motivation**

Goal of this PR is to update example app dependencies to expo SDK 48. 

**Test plan**

Run example app.


packages/react-native-tab-view/package.json
```diff
@@ -43,10 +43,9 @@
   },
   "devDependencies": {
     "del-cli": "^5.0.0",
-    "react": "18.1.0",
```
kacperkapusciak, line 46: Was this on purpose? I'm not sure if we'd want to remove `react` from our dependencies

okwasniewski: It was removed by mistake. I've fixed it.

kacperkapusciak (approved):

PR #11385, opened 2023-05-26
fix: prevent material tabBarItem from re-rendering when using material-top-tabs

okwasniewski:
**Motivation**

This issue addresses performance problems when using material-top-tabs and having many tabs. The problem was with passed callbacks which were causing all TabBarItems to re-render on index change.

Before: 
CleanShot 2023-05-26 at 12 12 12@2x

After: 
CleanShot 2023-05-26 at 12 10 47@2x

This PR needs a change in `useLatestCallback` dependency which has an open PR here: https://github.com/satya164/use-latest-callback/pull/2 

**Test plan**

Launch Material Top Tabs example and check that only necessary tab items are re-rendering.


andonivianez:
When is this going to be published? Is urgent, as material-top-tabs are laggy when you touch the tabs.

okwasniewski:
> When is this going to be published? Is urgent, as material-top-tabs are laggy when you touch the tabs.

Unfortunately, this PR can't be merged as-is because the additional memoization breaks some stuff with rendering. I'm going to work today on some performance boosts for material-top-tabs however, there will be RFC with breaking API changes proposed to make this faster and released as a new major version of tab-view.

okwasniewski:
Closing this PR all issues with `material-top-tabs` performance should be addressed after this RFC: https://github.com/react-navigation/rfcs/pull/88

PR #11393, opened 2023-05-31
feat!: add a direction prop to NavigationContainer to specify rtl

satya164:
BREAKING CHANGE: Previously the navigators tried to detect RTL automatically and
adjust the UI. However this is problematic since we cannot detect RTL in all cases
(e.g. on Web).

This adds an optional `direction` prop to `NavigationContainer` instead
so that user can specify when React Navigation's UI needs to be adjusted for RTL. It defaults to the value from `I18nManager` on native platforms, however it needs to be explicitly passed for Web.

### DIFF
--- a/example/src/Restart.native.tsx
+++ b/example/src/Restart.native.tsx
@@ -1,5 +1,0 @@
-import * as Updates from 'expo-updates';
-
-export function restartApp() {
-  Updates.reloadAsync();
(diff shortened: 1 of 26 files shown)


a-eid:
@satya164 I can't find such `direction` prop on `NavigationContainer` ?

PR #11415, opened 2023-06-12
feat: add autoHideHomeIndicator prop for @react-navigation/stack

mortend:
An autoHideHomeIndicator prop exists in native-stack (added in 00b2c24), and this updates @react-navigation/stack to add an identical prop.

**Motivation**

We have an app using @react-navigation/stack that needs to hide the home indicator in some cases and switching to native-stack breaks existing functionality for us.

**Test plan**

Set `autoHideHomeIndicator` to `true` and use iPhone 14 Simulator (or similar) to observe that the home indicator bar automatically hides after a few seconds when there is no activity on the screen.


mortend:
Just rebased on `main` so this might be a good time to review :)

satya164 (approved):
Thanks

PR #11421, opened 2023-06-14
chore: upgrade dependencies on 6.x

kacperkapusciak:
Followed https://github.com/react-navigation/react-navigation/pull/11090 and https://github.com/react-navigation/react-navigation/pull/11380

## Test

- `yarn lint`
- `yarn build`
- `yarn test`
- `yarn example test:e2e`

Run example app


packages/native/src/Link.tsx
```diff
@@ -12,7 +12,10 @@ type Props<ParamList extends ReactNavigation.RootParamList> = {
   onPress?: (
     e: React.MouseEvent<HTMLAnchorElement, MouseEvent> | GestureResponderEvent
   ) => void;
-} & (TextProps & { children: React.ReactNode });
+} & (Omit<TextProps, 'disabled'> & {
+  disabled?: boolean | null;
```
okwasniewski, line 16: Why is this change needed? We are adding an option to pass `null`

packages/stack/src/types.tsx
```diff
@@ -396,15 +396,15 @@ export type StackCardInterpolationProps = {
...
   /**
    * Animated node representing whether the card is being swiped (1 - swiping, 0 - not swiping).
    */
-  swiping: Animated.AnimatedInterpolation;
+  swiping: Animated.AnimatedInterpolation<0 | 1>;
   /**
    * Animated node representing multiplier when direction is inverted (-1 - inverted, 1 - normal).
    */
-  inverted: Animated.AnimatedInterpolation;
+  inverted: Animated.AnimatedInterpolation<1 | -1>;
```
okwasniewski, line 407:
```suggestion
  inverted: Animated.AnimatedInterpolation<-1 | 1>;
```

okwasniewski: Just to be consistent with other types

okwasniewski (approved):

satya164 (approved):

packages/native/src/Link.tsx
kacperkapusciak: After the upgrade of `@types/react-native` Typescript started to whine about incompatible `disabled` props:

```
Types of property 'disabled' are incompatible.
    Type 'boolean | null | undefined' is not assignable to type 'boolean | undefined'.
      Type 'null' is not assignable to type 'boolean | undefined'.
``` 

So I've looked up on how it's figured out on `main` and simply did the same:
https://github.com/react-navigation/react-navigation/blob/main/packages/native/src/Link.tsx#L15

packages/stack/src/types.tsx
kacperkapusciak: Thanks! 🙌 
Fixed in [7d7b3f7](https://github.com/react-navigation/react-navigation/pull/11421/commits/7d7b3f7a23dc758866b3c501cb1868b7d09ee8ee)

Issue #10876, opened 2022-09-23
TypeScript types for LinkingOptions break with nested navigation.


phryneas:
### Current behavior

When typing a nested navigation as shown in the documentation with

```js
interface BreakingPathConfig {
    Sub_1: { that: 'works' };
    Sub_2: NavigatorScreenParams<SubPathConfig>;
}

interface SubPathConfig {
    Sub_2_1: { that: 'breaks' };
}

export const brokenLinking: LinkingOptions<BreakingPathConfig> 
```
TypeScript falls back to `any` instead of the correct type of `SubPathConfig` and allows `Sub_2_1` to have any kind of sub-screens as a result.

### Expected behavior

Sub-screens of sub-screens should be correctly restricted.

### Reproduction

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3nAMsAdgazwHMB5MGYCXAZwBo4A5AQwDdgimZoBlAYyhQpcABSZQmIanAC+cAGZQIIOAHIAAgKa8YAWlyt2nSrgD0+iixQqA3AChbeGCihytaAEKbCuIqJgALAGEqOXYMWzhIuG4AVwAjAH0ARgAuDAR-TjSVAHdofGoVGTso6PiEgCY05jYOLig+ASFRcUkAHli4vyCQ9gA+O2l7R2dXXjRO7uDcUKJw0s7K5LTMAKzVOK9C4tsh2xQAD0hYOF4qanhNiHwhHAJiNLvvUnJjajbPFCZnqd6iPrgAF55lEwAJQgcUNQ0gBtAC6tAiUTOM3YKyRpUi1H4gho6MxBLKiVS6TAnH82SK0kRhMxiyqINpBLJAUpcBpTIJ2KaePCJhMnMxAD0APy2fmC0oBKFoaj+CAxAA2ABM4LgIJc0AADJi4RBa+hxGLwLWTcnTWZajGSqL05aMm2lFkU1QqDmO0rc3HQh0eqIStQwag6Q6obQhqCKKDWv2RAAq-mcKG48qVyoYGvcyZxQh96CGsaiBdjxcdpaZ1JjRfdMkR0msQA

### Platform

- [x] TypeScript

### Packages

- [x] @react-navigation/​native

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.0.13
| TypeScript                            | 4.8.2

[...]

PR #11430, opened 2023-06-19
fix: remove NonNullable to fix recursive infer in PathConfigMap

johankasperi:
**Motivation**

Fixes https://github.com/react-navigation/react-navigation/issues/10876 and https://github.com/react-navigation/react-navigation/issues/9897

**Test plan**

This change breaks typescript in the example app because the linking config does not match how all the navigators is setup. This wasn't catched before because of this bug in the `PathConfigMap` type. For example we are configuring a link to the `Chat` screen inside `NativeStack` ([with the path "native-stack/chat"](https://github.com/react-navigation/react-navigation/blob/main/example/src/index.tsx#L178)) but the `Chat` route isn't configured in the `NativeStack` navigator: https://github.com/react-navigation/react-navigation/blob/main/example/src/Screens/NativeStack.tsx#L15

I will happily fix this but wanted some feedback on the actual change to `PathConfigMap` first before putting in the hour(s) to fix the example project. 

Also I would like some feedback on how we would like to change the example project. Should all the routes "Article", "Albums", "Chat", "Contacts", "NewsFeed" and "Dialog" exist in every navigator? Or should just the linking config match how each individual navigator is setup? This would probably mean that we cannot reduce all SCREEN_NAMES to create the config because all navigators have different param lists.

I also have a fix for the 6.x branch: https://github.com/react-navigation/react-navigation/compare/6.x...johankasperi:react-navigation:fix-pathconfig-recursive-infer?expand=1

### DIFF
--- a/example/src/screens.tsx
+++ b/example/src/screens.tsx
@@ -76,7 +76,7 @@ export type RootDrawerParamList = {
 };
 
 export type RootStackParamList = ExampleScreensParamList & {
-  Home: NavigatorScreenParams<RootDrawerParamList> | undefined;
+  Home: NavigatorScreenParams<RootDrawerParamList>;
   NotFound: undefined;
 };
 
--- a/example/__typechecks__/common.check.tsx
+++ b/example/__typechecks__/common.check.tsx
@@ -8,11 +8,12 @@ import type {
   DrawerNavigationOptions,
   DrawerScreenProps,
 } from '@react-navigation/drawer';
-import type {
-  CompositeScreenProps,
-  NavigationAction,
-  NavigationHelpers,
-  NavigatorScreenParams,
+import {
+  type CompositeScreenProps,
+  type LinkingOptions,
+  type NavigationAction,
+  type NavigationHelpers,
+  type NavigatorScreenParams,
 } from '@react-navigation/native';
 import {
   createStackNavigator,
@@ -498,3 +499,34 @@ const FourthStack = createStackNavigator<FourthParamList, 'MyID'>();
 expectTypeOf(FourthStack.Navigator).parameter(0).toMatchTypeOf<{
   id: 'MyID';
 }>();
+
+/**
+ * Infer correct LinkingOptions
+ */
+export const linkingOptions: LinkingOptions<RootStackParamList> = {
+  prefixes: ['reactnavigation://'],
+  config: {
+    screens: {
+      Home: {
+        screens: {
+          Feed: {
+            screens: {
+              Popular: 'popular',
+              Latest: 'latest',
+              // @ts-expect-error "NonExistingRoute" does not exist in the FeedTabParamList
+              NonExistingRoute: 'non-existing',
+            },
+          },
+          Account: 'account',
+        },
+      },
+      PostDetails: 'post/:id',
+      Login: {
+        path: 'login',
+        // @ts-expect-error The Login route is not a nested navigator and should not have a screens object
+        screens: {},
+      },
+      NotFound: '*',
+    },
+  },
+};
--- a/example/src/Screens/NotFound.tsx
+++ b/example/src/Screens/NotFound.tsx
@@ -14,7 +14,7 @@ export const NotFound = ({
       <Text style={styles.title}>404 Not Found ({route.path})</Text>
       <Button
         variant="filled"
-        onPress={() => navigation.navigate('Home')}
+        onPress={() => navigation.navigate('Home', { screen: 'Examples' })}
         style={styles.button}
       >
         Go to home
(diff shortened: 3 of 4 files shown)


pagkt2:
Before, I figured this out on my own by removing `NonNullable`, although I still left the `extends {}`

stereodenis:
Not working for me...

osdnk:
Hi @johankasperi,

Thanks for the PR!

Can you please rebase and fix the CI? Also, can you add tests (use `__typechecks__` folder).

example/src/screens.tsx
```diff
@@ -76,7 +76,7 @@ export type RootDrawerParamList = {
 };
 
 export type RootStackParamList = ExampleScreensParamList & {
-  Home: NavigatorScreenParams<RootDrawerParamList> | undefined;
+  Home: NavigatorScreenParams<RootDrawerParamList>;
```
satya164, line 79: Why did you change this? Please revert to how it was

satya164 (changes requested):

example/src/screens.tsx
johankasperi: Could you please help me with how linking options should be defined here if the `Home` route is  `NavigationScreenParams<RootDrawerParamList> | undefined`. I'm getting this typecheck error when it can be undefined:
```
Object literal may only specify known properties, and 'screens' does not exist in type 'Omit<PathConfig<{}>, "initialRouteName" | "screens">'
```
on this row https://github.com/react-navigation/react-navigation/blob/d0abdee67f5db8cf39112af535846ffededfb21d/example/src/index.tsx#L69

I have no experience from using `NavigatorScreenParams` in a union with `undefined` and really no idea how it should be handled by the lib. Would really appreciate some guidance. Thanks!

johankasperi: We are omitting `screens` from the type if it's not extending `NavigatorScreenParams` here: https://github.com/react-navigation/react-navigation/blob/d0abdee67f5db8cf39112af535846ffededfb21d/packages/core/src/types.tsx#L1012C37-L1012C46 

Since this PR removes `NonNullable` from that line (because `NonNullable` broke recursive infer) I think that makes it impossible for a route to be union of `NavigatorScreenParams` and `undefined`?

satya164: If it makes it impossible to have `undefined` then it'd be a breaking change. The change needs to support `| undefined` to be merged

johankasperi:
I checked the main branch of the lib and from my testing the original 

> Hi @johankasperi,
> 
> Thanks for the PR!
> 
> Can you please rebase and fix the CI? Also, can you add tests (use `__typechecks__` folder).

Thank you! I added a test in `common.check.tsx`. Please let me know if it was something like that you had in mind or if we should test it in some other way.

example/src/screens.tsx
johankasperi: Ok, but if a route has is defined as `NavigatorScreenParams<> | undefined` should you be able to config linking options for the nested navigator as in the example?

satya164: It wouldn't make sense to support `| undefined` if it only works with some of the things and not others.

ChrisYohann:
HI @johankasperi , did you manage to effectively resolve #9897 with your fix ? I copy pasted your changes, and it did not fix anything.
The issues says that it's not possible to type a screen that has its own parameters and a nested navigator together. 

I think, as soon as you define your type like this (from the example) :
```typescript
export type ParamList = {
  ScreenA: undefined
  ScreenB: NavigatorScreenParams<ParamListInnerNavigator> & {
    someparam?: number,
  }
}
```
Assuming `PathConfigMap` is defined as follows :
```typescript
export type PathConfigMap<ParamList = {}> = {
  [RouteName in keyof ParamList]?: ParamList[RouteName] extends NavigatorScreenParams<
    infer T
  >
    ? string | PathConfig<T>
    : string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>;
};
```

This check :
```typescript
ParamList[RouteName] extends NavigatorScreenParams<
    infer T
  >
    ? string | PathConfig<T>
    : string | Omit<PathConfig<{}>, 'screens' | 'initialRouteName'>;
```
resolves to false, so the linking config is not correct.
The solution should be not to enhance `NavigatorScreenParams` generic type to allow passing params to the screen ?
like : 

```typescript
export type NavigatorScreenParams<
  ParamList,
  State extends NavigationState = NavigationState,
  OwnScreenParams extends {} = {},
> = OwnScreenParams &
  (
    | {
        screen?: never;
        params?: never;
        initial?: never;
        path?: string;
        state: PartialState<State> | State | undefined;
      }
    | {
        [RouteName in keyof ParamList]: undefined extends ParamList[RouteName]
          ? {
              screen: RouteName;
              params?: ParamList[RouteName];
              initial?: boolean;
              path?: string;
              state?: never;
            }
          : {
              screen: RouteName;
              params: ParamList[RouteName];
              initial?: boolean;
              path?: string;
              state?: never;
            };
      }[keyof ParamList]
  );
```

Thanks for your help !

PR #11440, opened 2023-06-26
fix tab auto width

souzabrs:
**Motivation**

When using `width: "auto"` on `tabBarItemStyle`, the `TabBarIndicator` size is rendered with an unexpected width on chrome for linux and for android, and in opera for android (not tested in opera for linux), because there's a bug related to `scaleX` css transformation in these platforms. This doesn't happen in chrome for windows, nor in firefox for linux.

img

The larger is the tab label, the bigger is the difference in the indicator width and the expected one, and this difference varies when zooming the browser in and out.

In this pull request, the animation is changed from `transform: translateX(<offset_interpolation>) scaleX(<width_interpolation>);` to `width: <width_interpolation>; left: <offset_interpolation>;` only when the platform is web and the `width` is set as `auto` in `tabBarItemStyle`.

**Test plan**
It's possible to check the browser's bug by rendering the HTML below in chrome for linux or for android:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>translateX vs width</title>
    <style type="text/css">
      #container {
        padding: 30px 30px;
      }
      #translated {
        width: 1px;
        height: 10px;
        background: #f00; 
        transform: scaleX(400);
        margin: 0;
        padding: 0;
        position: relative;
        left: 200px;
      }
      #set-width {
        width: 400px;
        height: 10px;
        background: #00f; 
        margin: 0;
        padding: 0;
        position: relative;
        left: 0;
      }
    </style>
  </head>
  <body>
    <div id="container">
      <div id="translated"></div>
      <div id="set-width"></div>
    </div>
  </body>
</html>
```

To test the tab navigation, run the following code and open the app on chrome for linux or for android:

```javascript
import { Text } from 'react-native';
import { createMaterialTopTabNavigator } from '@react-navigation/material-top-tabs';
const Tab = createMaterialTopTabNavigator();

function Content() {
  return (
    <Text>Here goes the screen content</Text>
  );
}

export default function Another() {

  return (
    <Tab.Navigator
      screenOptions={{
        tabBarItemStyle: { 
          width: "auto",
        },
      }}>
      <Tab.Screen
        name="tab1"
        component={Content}
        options={{
          title: "myApp"
        }}
      />
      <Tab.Screen
        name="tab2"
        component={Content}
        options={{
          title: "this is a big label in my app"
        }}
      />
      <Tab.Screen
        name="tab3"
        component={Content}
        options={{
          title: "short"
        }}
      />
      <Tab.Screen
        name="tab4"
        component={Content}
        options={{
          title: "another big label in my app"
        }}
      />
    </Tab.Navigator>
  );
}
```


packages/react-native-tab-view/src/TabBarIndicator.tsx
```diff
@@ -120,19 +120,35 @@ export function TabBarIndicator<T extends Route>({
     );
   }
 
+  const styleList = [];
+
+  // scaleX doesn't work on chrome and opera for linux, including their
+  // versions for android
+  if (Platform.OS === 'web' && width === 'auto') {
```
okwasniewski, line 127: Maybe we could detect the operating system through `window.navigator.userAgent`?  Because it does not apply to MacOS / Windows.

packages/react-native-tab-view/src/TabBarIndicator.tsx
```diff
@@ -120,19 +120,35 @@ export function TabBarIndicator<T extends Route>({
...
+    styleList.push(
+      { width: transform[1].scaleX },
+      { left: transform[0].translateX },
+    );
```
okwasniewski, line 131:
```suggestion
    styleList.push({
      width: transform[1].scaleX,
      left: transform[0].translateX,
    });
```

okwasniewski (dismissed):
Thanks! I've checked on Chrome for Android and the issue looks fixed however, let's apply this only to affected platforms.

usedlife:
```Style property 'left' is not supported by native animated module```

left and width is not supported animated value, why would I report this wrong ?

File: packages/react-native-tab-view/src/TabBarIndicator.tsx (hunk 1 above)
Comment (osdnk) [reply packages/react-native-tab-view/src/TabBarIndicator.tsx:127]: this is not good for SSR

File: packages/react-native-tab-view/src/TabBarIndicator.tsx [hunk 3]
```diff
@@ -120,19 +120,33 @@ export function TabBarIndicator<T extends Route>({
...
-        { width: width === 'auto' ? 1 : width },
-        // If layout is not available, use `left` property for positioning the indicator
-        // This avoids rendering delay until we are able to calculate translateX
-        // If platform is macos use `left` property as `transform` is broken at the moment.
-        // See: https://github.com/microsoft/react-native-macos/issues/280
-        layout.width && Platform.OS !== 'macos'
-          ? { left: 0 }
-          : { left: `${(100 / routes.length) * navigationState.index}%` },
-        { transform },
+        ...styleList,
```
osdnk, line 149:
```suggestion
        styleList,
```

osdnk (approved):

satya164:
Thank you for the PR!

ngocdinhimpl:
android still get this error. does anyone have solution?

Issue #8606, opened 2020-07-18
InteractionManager.runAfterInteractions not running at all! (In Production/Release builds)


mrousavy:
**Current Behavior**

I am using a lot of `InteractionManager.runAfterInteractions(() => ...)` calls throughout my project.

For example, I have a **feed of multiple posts** in my HomeScreen (Tab Navigator) and each post has `useEffect` and `useFocusEffect` hooks which call `InteractionManager.runAfterInteractions` inside the hook to schedule loading extra post data after some animations have completed. _"Some animations"_ = mostly navigation transitions from react-navigation, but since the post items specifically are inside a Tab Navigator, no navigation transitions/animations are run so this should fire almost immediately.

In development/debug build everything works fine, but once I build for production/release, the extra post data never gets loaded, a test `Alert.alert(...)` I've put in there never gets shown, and my animations will never start, causing my app to be fully stuck.

I have added a `GlobalErrorHandler` callback to my entry point (`App.tsx`):

```ts
const defaultErrorHandler = ErrorUtils.getGlobalHandler();
ErrorUtils.setGlobalHandler((error, isFatal) => {
	Alert.alert(`Unexpected ${isFatal ? 'fatal error' : 'error'}!`, `Unexpected error occured! ${JSON.stringify(error)}`);
	Logger.logError(isFatal ? 'A fatal, unhandled error occured!' : 'An unhandled error occured!', error);
	if (defaultErrorHandler) defaultErrorHandler(error, isFatal);
});
```

but this doesn't get called either.

Unfortunately, since this only happens on production/release builds, I'm having a hard time debugging this. I will happily provide any extra information needed.

### Extra Details

* I'm using react-navigation, so this might cause a problem I suppose. My root view is a `NavigationContainer`, and inside of it, I conditionally render the `LoginStackNavigator` or the `HomeStackNavigator` depending if the user is logged in or not.
* I'm using react-native-screens (because of react-navigation), but I call the `enableScreens()` function at the top of my entry point (`App.tsx`):
    ```ts
    import { enableScreens } from 'react-native-screens';
    enableScreens();
    ```
* I'm using react-native-gesture-handler (because of react-navigation), but I import the library at the top of my entry point (`App.tsx`):
    ```ts
    import 'react-native-gesture-handler';
    ```
* I'm using react-native-vector-icons, but I call `loadFont()` at the top of my entry point (`App.tsx`):
     ```ts
[...]

PR #11465, opened 2023-07-11
fix(Stack): added failed gesture state to clear interaction

Iltimirov:
Added failed gesture state to clear interaction manager queue 

**Motivation**

My team ran into an issue on IOS where the user taps somewhere on the left edge of the screen without any swipe, and upon releasing their finger, the InteractionManager has one outstanding task in the queue. And this task can block user interactions wrapped into `InteractionManager.runAfterInteractions`

It could be related with https://github.com/react-navigation/react-navigation/issues/8606 but i'm not sure

**Test plan**

1. Create a stack with two screen with an ability to navigate by swipe
2. Navigate from screen (1) to screen (2)
3. Tap somewhere in the left edge of the screen
4. After that all actions wrapped into `InteractionManager.runAfterInteractions` will not work


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -336,6 +336,10 @@ export class Card extends React.Component<Props> {
         onGestureEnd?.();
         break;
       }
+      case GestureState.FAILED: {
+        this.handleEndInteraction();
```
satya164, line 340: I'm not sure if this is enough. Isn't `this.isSwiping.setValue(FALSE);` needed as well?

Iltimirov: Hmm, maybe. Thank you for review! I'll check it

Iltimirov: done

satya164 (approved):
Thank you

PR #11478, opened 2023-07-17
Allow headerShown to be a function

JonathanCallewaert:
I was going through the docs to find a way to hide my header based on the name of a nested route. I could only find a way to adapt the title, but not to hide the header (https://reactnavigation.org/docs/screen-options-resolution/) 
```js
import { getFocusedRouteNameFromRoute } from '@react-navigation/native';

function getHeaderTitle(route) {
  // If the focused route is not found, we need to assume it's the initial screen
  // This can happen during if there hasn't been any navigation inside the screen
  // In our case, it's "Feed" as that's the first screen inside the navigator
  const routeName = getFocusedRouteNameFromRoute(route) ?? 'Feed';

  switch (routeName) {
    case 'Feed':
      return 'News feed';
    case 'Profile':
      return 'My profile';
    case 'Account':
      return 'My account';
  }
}
```

However, in this PR, I want to achieve something similar but with the headerShown prop

### DIFF
--- a/example/src/Screens/TabView/AutoWidthTabBar.tsx
+++ b/example/src/Screens/TabView/AutoWidthTabBar.tsx
@@ -46,8 +46,10 @@ export const AutoWidthTabBar = () => {
       scrollEnabled
       indicatorStyle={styles.indicator}
       style={styles.tabbar}
+      contentContainerStyle={styles.tabbarContentContainer}
       labelStyle={styles.label}
       tabStyle={styles.tabStyle}
+      gap={20}
     />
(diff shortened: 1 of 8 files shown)


migueldaipre:
You could create the options dynamically for each screen, anyway it would be interesting to change the headerShown documentations.

satya164:
Thank you for the pull request. However, I’m not sure if the change makes sense. If the function is not receiving any new information, then making it a function doesn’t seem to add any value. So I'm closing the PR.

PR #11485, opened 2023-07-20
fix: replace series function used to queue async callbacks

karlsander:
**Motivation**

This function is used to ensure that all state change callbacks to sync the local browser history state run in series. The functions are async because `history.go` is not a synchronous function.

I noticed that in chrome, the URL was updating to the previous url instead of the new one in some navigations and tracked it down to the fact that the `history.go` navigation takes longer than in other browsers. This function is supposed to mitigate that, but I believe it has a bug. The function will not properly return early, because the finally clause will still always run and immediately reset the `isRunning` state. Even after fixing this, the function didn't quite work right for queuing more than two callbacks.

I have replaced the function with a simpler implementation that does the same task.

**Test plan**

### DIFF
--- a/packages/native/src/__tests__/NavigationContainer.test.tsx
+++ b/packages/native/src/__tests__/NavigationContainer.test.tsx
@@ -6,7 +6,7 @@ import {
   TabRouter,
   useNavigationBuilder,
 } from '@react-navigation/core';
-import { act, render } from '@testing-library/react-native';
+import { act, render, waitFor } from '@testing-library/react-native';
 import * as React from 'react';
 
 import { window } from '../__mocks__/window';
@@ -18,9 +18,7 @@ Object.assign(global, window);
 // eslint-disable-next-line import/extensions
 jest.mock('../useLinking', () => require('../useLinking.tsx'));
 
-it('integrates with the history API', () => {
-  jest.useFakeTimers();
-
+it('integrates with the history API', async () => {
   const createStackNavigator = createNavigatorFactory((props: any) => {
     const { state, descriptors, NavigationContent } = useNavigationBuilder(
       StackRouter,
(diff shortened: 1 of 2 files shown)


satya164 (approved):
Thanks for the PR

satya164:
This change seems to break unit tests. Can you take a look?

karlsander:
I'm not sure exactly, but the test is failing on `main` for me in the same way. I think it is addressed here? https://github.com/react-navigation/react-navigation/pull/11509

mountiny:
@karlsander Would you be able to merge main to your branch? The tests are passing on main

adamgrzybowski:
Hey, @karlsander I've tried wrapping `expect` with `waitFor` and tests seem to work fine. Here's how it looks. https://gist.github.com/adamgrzybowski/7839818fb1a9728052bab0a1d84464e4

There's one thing I'm not 100% sure of. I had to remove `is.useFakeTimers()` but the tests still passed.

adamgrzybowski:
@karlsander Hey I created a [PR](https://github.com/react-navigation/react-navigation/pull/11565) to test new tests for the series without distracting you if something went wrong 😄 

I've used `await` and it seems to work fine. I've also checked with the debugger and the value for `window.location.pathname` matches the expected value.

BTW we used your `series` implementation as a patch in our app and it seems to solve one problem that we had so I think you did a good job with this solution

BTW_2.0 the old implementation of `series` is probably broken because of the `try {...} finally {...}` blocks. If we remove them the result is the same as with your simpler implementation (the problem in our app is fixed and the tests fail)

The `finally` block is executed even if you return in `try` and that breaks the queue. At least that's my guess

karlsander:
> BTW_2.0 the old implementation of series is probably broken because of the try {...} finally {...} blocks. If we remove them the result is the same as with your simpler implementation (the problem in our app is fixed and the tests fail)

I don't remember exactly, but I think in that case two async functions at the same time are queued correctly, but not 3+ (a situation that doesn't really happen in this usage anyway).

karlsander:
@satya164 all looks green now, but I cannot merge because I don't have access

osdnk:
@adamgrzybowski @karlsander I guess that is safe to go?

mountiny:
I think so, we are using it in our App without any issues found so far 🤞 https://github.com/Expensify/App/pull/25989/files

karlsander:
We are also using the patch with success, would appreciate if we can get this merged.

PR #11487, opened 2023-07-21
Fix overlapping issue between status bar and tab bar in TabView.tsx o…

ASBirinci:
In iOS on phones, the tab bar overlaps with the status bar.


ASBirinci:
The tab bar overlaps with the status bar.
310ed568-dc02-42ba-834e-5ca939debebc

Solved Picture
7455e631-440a-42d8-ad26-09755eb77f81

okwasniewski:
Okay, I get it now. However user can add it at the root level of their app / wrap TabView. 

In order to keep TabView flexible for all use cases it should be up to the user of the library. Therefore I won't approve this PR.

PR #11501, opened 2023-07-26
fix: buggy behaviour of search bar / large title on Fabric with native stack v7

kkafar:
**Motivation**

`UIKit` requires `ScrollView` to be at index 0 in given view's subview array to enable it's interaction with navigation bar. On Fabric `DebugContainer` view got flattened, however it was not removed from hierarchy -- instead the view was attached as first child of the `RNSScreenView` disabling system interaction between navigation bar and a scroll view. 

See

* https://github.com/software-mansion/react-native-screens/pull/1825

for broader context and explanation.

**Test plan**

### Before

https://github.com/software-mansion/react-native-screens/assets/50801299/3b89dee8-a307-44ba-a9f4-02e4dfdd95ac

### After

https://github.com/software-mansion/react-native-screens/assets/50801299/205f24d4-590b-402e-8b0a-848fd4482d59

### DIFF
--- a/packages/native-stack/src/views/DebugContainer.native.tsx
+++ b/packages/native-stack/src/views/DebugContainer.native.tsx
@@ -9,12 +9,14 @@ type ContainerProps = ViewProps & {
   children: React.ReactNode;
 };
 
-// This view must *not* be flattened.
-// See https://github.com/software-mansion/react-native-screens/pull/1825
-// for detailed explanation.
-export function DebugContainer(props: ContainerProps) {
+/**
+ * This view must *not* be flattened.
+ * See https://github.com/software-mansion/react-native-screens/pull/1825
+ * for detailed explanation.
+ */
+export let DebugContainer = (props: ContainerProps) => {
   return <View {...props} collapsable={false} />;
-}
+};
 
 if (process.env.NODE_ENV !== 'production') {
   DebugContainer = (props: ContainerProps) => {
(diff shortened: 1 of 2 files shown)


kkafar:

packages/native-stack/src/views/DebugContainer.native.tsx
```diff
@@ -9,8 +9,12 @@ type ContainerProps = ViewProps & {
   children: React.ReactNode;
 };
 
-export let DebugContainer =
-  View as unknown as React.ComponentType<ContainerProps>;
+// This view must *not* be flattened.
+// See https://github.com/software-mansion/react-native-screens/pull/1825
+// for detailed explanation.
+export let DebugContainer = function (props: ContainerProps): JSX.Element {
```
satya164, line 15:
```suggestion
export function DebugContainer(props: ContainerProps): JSX.Element {
```

packages/native-stack/src/views/DebugContainer.native.tsx
```diff
@@ -9,8 +9,12 @@ type ContainerProps = ViewProps & {
...
 };
 
-export let DebugContainer =
-  View as unknown as React.ComponentType<ContainerProps>;
+// This view must *not* be flattened.
+// See https://github.com/software-mansion/react-native-screens/pull/1825
+// for detailed explanation.
+export function DebugContainer(props: ContainerProps): JSX.Element {
+  return <View {...props} collapsable={false} />;
+};
```
satya164, line 17:
```suggestion
export function DebugContainer(props: ContainerProps) {
  return <View {...props} collapsable={false} />;
}
```

PR #11502, opened 2023-07-26
fix: buggy behaviour of search bar / large title on Fabric with native stack v6

kkafar:
**Motivation**

`UIKit` requires `ScrollView` to be at index 0 in given view's subview array to enable it's interaction with navigation bar. On Fabric `DebugContainer` view got flattened, however it was not removed from hierarchy -- instead the view was attached as first child of the `RNSScreenView` disabling system interaction between navigation bar and a scroll view. 

See

* https://github.com/software-mansion/react-native-screens/pull/1825

for broader context and explanation.

**Test plan**

### Before

https://github.com/software-mansion/react-native-screens/assets/50801299/3b89dee8-a307-44ba-a9f4-02e4dfdd95ac

### After

https://github.com/software-mansion/react-native-screens/assets/50801299/205f24d4-590b-402e-8b0a-848fd4482d59

### DIFF
--- a/packages/native-stack/src/views/DebugContainer.native.tsx
+++ b/packages/native-stack/src/views/DebugContainer.native.tsx
@@ -14,12 +14,12 @@ type ContainerProps = ViewProps & {
  * See https://github.com/software-mansion/react-native-screens/pull/1825
  * for detailed explanation.
  */
-export let DebugContainer = (props: ContainerProps) => {
+let DebugContainer = (props: ContainerProps) => {
   return <View {...props} collapsable={false} />;
 };
 
 if (process.env.NODE_ENV !== 'production') {
-  const DebugContainer = (props: ContainerProps) => {
+  DebugContainer = (props: ContainerProps) => {
     const { stackPresentation, ...rest } = props;
 
     if (Platform.OS === 'ios' && stackPresentation !== 'push') {
@@ -33,8 +33,6 @@ if (process.env.NODE_ENV !== 'production') {
 
     return <View {...rest} collapsable={false} />;
   };
-
-  Container = DebugContainer;
 }
 
-export default Container;
+export default DebugContainer;
--- a/packages/native-stack/src/views/DebugContainer.native.tsx
+++ b/packages/native-stack/src/views/DebugContainer.native.tsx
@@ -12,9 +12,7 @@ type ContainerProps = ViewProps & {
 // This view must *not* be flattened.
 // See https://github.com/software-mansion/react-native-screens/pull/1825
 // for detailed explanation.
-let Container = function (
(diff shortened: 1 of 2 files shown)


kkafar:

packages/native-stack/src/views/DebugContainer.native.tsx
```diff
@@ -9,7 +9,12 @@ type ContainerProps = ViewProps & {
   children: React.ReactNode;
 };
 
-let Container = View as unknown as React.ComponentType<ContainerProps>;
+// This view must *not* be flattened.
+// See https://github.com/software-mansion/react-native-screens/pull/1825
+// for detailed explanation.
+let Container = function (props: ContainerProps): JSX.Element {
+  return <View {...props} collapsable={false} />;
+};
```
kkafar, line 17: I'm not a typescript chad by any means, so please pay attention to whether the type of the `Container` has changed. It was earlier `React.ComponentType<ContainerProps>`.

The same applies in mirror PR to v7

* https://github.com/react-navigation/react-navigation/pull/11501/

kkafar (commented):
Just a remark for future reviewer

packages/native-stack/src/views/DebugContainer.native.tsx
satya164, line 17:
```suggestion
/**
 * This view must *not* be flattened.
 * See https://github.com/software-mansion/react-native-screens/pull/1825
 * for detailed explanation.
 */
export let DebugContainer = (props: ContainerProps) => {
  return <View {...props} collapsable={false} />;
};
```

satya164 (approved):
Thanks for the PR

kacperkapusciak (approved):

Gregoirevda:
@satya164 @kkafar This PR which is the only code diff in @react-navigation/native-stack 6.9.14 totally breaks the android search when not used with Fabric. 6.9.13 works fine. The search icon doesn't appear in the header. iOS seems to be working fine.

kkafar:
@Gregoirevda, would you mind posting a snack / repo so that I might reproduce? I've got working Android searchbar in my example app.

Ideally create a new issue.

kkafar:
@Gregoirevda is is also likely that the issue has been already fixed with https://github.com/software-mansion/react-native-screens/pull/1883

New screens version will be released with this fix in couple of days.

Gregoirevda:
@kkafar Great work, will give an update when available

SamPoate:
FYI, this PR broke header Left/Right components that used absolute placed buttons on iOS. I've traced it back to the latest working package being `"@react-navigation/native-stack": "6.9.13"`.

PR #11509, opened 2023-07-28
fix: bug with out of sync index in createMemoryHistory

adamgrzybowski:

**Motivation**

The index in `createMemoryHistory` may be out of sync and that causes the items array to drop entries in some cases. 

**Test plan**

In example app, add this.log in onPopState callback, push many screens in simple stack, end up on Article screen, do one go back and one forward actions with browser buttons, click POP SCREEN . You will end up in the initial page instead of the previous screen you were at.

https://github.com/react-navigation/react-navigation/assets/67908363/64c8236c-8717-4519-b3b0-fbeeab58681d

**Note**

I've added a commit to log `createMemoryHistory` items to make it easier to see what's actually going on. It can be removed before merging. 

To see the logs you have to uncomment  the line 238 in `createMemoryHistory`

### DIFF
--- a/packages/native/src/createMemoryHistory.tsx
+++ b/packages/native/src/createMemoryHistory.tsx
@@ -14,6 +14,26 @@ export function createMemoryHistory() {
   let index = 0;
   let items: HistoryRecord[] = [];
 
+  // @ts-ignore no-unused-vars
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR. Can you add a test for this under `example/e2e/tests`?

PR #11522, opened 2023-08-04
fix: Drawer on macos is registering clicks despite being inactive

hsjoberg:
Hey.

**Motivation**

If you use the drawer on react-native-macos, it will register click events even if the drawer is inactive, resulting in everything underneath the drawer being unclickable.

I suspect it's because of some `zIndex` layering issues (we set zIndex to -1 when inactive) going on, but not sure.

This PR addresses this issue by setting `pointerEvents` to `"none"` for react-native-macos.

**Test plan**

Try @react-navigation/drawer@7.0.0-alpha.2 on react-native-macos old arch (could be the same on new arch, I don't know). Anything under where the drawer would be cannot be clicked on.

Note: react-reanimated broke its react-native-macos support >= 2.0, I have a pending fix for that here:
https://github.com/software-mansion/react-native-reanimated/pull/4669

So either use react-reanimated 1 or use my fork:
`"react-native-reanimated": "git+https://github.com/hsjoberg/react-native-reanimated.git#e998c740a083f075e6845a9ed1c81d5df1c220fc",`

### DIFF
--- a/packages/react-native-drawer-layout/src/views/modern/Drawer.tsx
+++ b/packages/react-native-drawer-layout/src/views/modern/Drawer.tsx
@@ -421,6 +421,11 @@ export function Drawer({
               drawerAnimatedStyle,
               drawerStyle as any,
             ]}
+            pointerEvents={
+              Platform.OS === 'macos' && drawerType !== 'permanent' && !isOpen
+                ? 'none'
+                : undefined
+            }
           >
             {renderDrawerContent()}
           </Animated.View>


osdnk (commented):
Thanks for the PR. I suggest we wait for the fix in Reanimated. Can you please attach some video?

satya164 (changes requested):
Thanks for the PR. Please link to the related bug report in react-native-macos repo to make sure we can track it.

Also the `pointerEvents` should be set based on similar logic as detecting when overlay can be tapped https://github.com/react-navigation/react-navigation/blob/85f6087116cd198d8ff408f4cbe1d239a741ba1d/packages/react-native-drawer-layout/src/views/Overlay.tsx#L36
Can you update the logic to match it?

PR #11523, opened 2023-08-05
fix: ensure latest values are used when `preventRemove` is triggered in `usePreventRemove` hook

andrejpavlovic:
**Motivation**

In `react-navigation/native@6.1.7` with `@react-navigation/core@6.4.9` `UNSTABLE_usePreventRemove`, when  `beforeRemove` listener is triggered it, it uses old value of `preventRemove` parameter that's passed into the hook. By using a reference to `preventRemove` (and `callback`) we can ensure the latest value is always used.

**Test plan**


andrejpavlovic:
Nevermind, looks like the issue is the lack of support of `useLatestCallback` in React Native.

Opened the issue here: https://github.com/satya164/use-latest-callback/issues/3

PR #11525, opened 2023-08-05
feat: makes NavigationContainerRef.getCurrentRoute type safe

lucasloisp:
**Motivation**

When doing analytics, it has been quite helpful to know the current screen and its params on the `onStateChange` callback directly, instead of deriving that from `state` by running through nested routes.

Even then, there are other cases that use `getCurrentRoute` and could use a more specific type than `Route<string>` based on the already present `ParamList` generic on `NavigationContainerRef`.

Currently, the typed returned is `Route<string>`, which provides no info on the valid "names" (as defined on the `ParamList`), nor on the `params`, whose type _could_ be restricted by checking for `name`.

**Test plan**

All changes presented are typescript-only, validated with tests on `example/__typechecks__` for both the cases (static or otherwise) with a defined `ParamList` as well as an empty type `{}` returning `Route<string>` which could be useful for projects adopting TS but that haven't yet typed their routes.


packages/core/src/types.tsx
```diff
@@ -648,6 +648,18 @@ export type NavigationContainerEventMap = {
...
 
+type ParamListRoute<ParamList extends ParamListBase> = {
+  [RouteName in keyof ParamList]: ParamList[RouteName] extends NavigatorScreenParams<
+    infer T extends ParamListBase
+  >
+    ? ParamListRoute<T>
+    : Route<Extract<RouteName, string>, ParamList[RouteName]>;
+}[keyof ParamList];
+
+type MaybeParamListRoute<ParamList extends {}> = ParamList extends ParamListBase
```
lucasloisp, line 659: Less than sure on the naming for this one. Need a second type to manage the case were `ParamList` does not extend `ParamListBase`, as the restriction on the ref is `extends {}`.

Would need the name to signify that it might be a "well typed" `Route` or it could be `Route<string>` depending on what you pass.

osdnk (commented):
Can you please check in tests if works for the nested routes?

lucasloisp:
So, I've been playing with that. It seems like `StaticParamList`  adds an ` | undefined` to nested navigators, which does not happen on the "common" navigator definitions.

```
const DemoStackToNest = createStackNavigator({
  screens: {
    Groups: () => null,
    Chat: (_: StaticScreenProps<{id: number}>) => null,
  }
})
const DemoStackWithNested = createStackNavigator({
  screens: {
    Home: DemoStackToNest
  }
})
type DemoParamListOfNest = StaticParamList<typeof DemoStackToNest>
type DemoParamList = StaticParamList<typeof DemoStackWithNested>
declare const pl: DemoParamList
expectTypeOf(pl['Home']).toEqualTypeOf<NavigatorScreenParams<DemoParamListOfNest> | undefined>()
expectTypeOf(pl['Home']).toEqualTypeOf<NavigatorScreenParams<DemoParamListOfNest>>() // Fails TS, missing undefined

type CommonDemoParamListOfNest = {
  Groups: undefined
  Chat: { id: number }
}
type CommonDemoParamList = {
  Home: NavigatorScreenParams<CommonDemoParamListOfNest>
}
declare const plc: CommonDemoParamList
expectTypeOf(plc['Home']).toEqualTypeOf<NavigatorScreenParams<DemoParamListOfNest> | undefined>() // Fails TS, extra undefined
expectTypeOf(plc['Home']).toEqualTypeOf<NavigatorScreenParams<DemoParamListOfNest>>()
```

Working on that one and will then add tests for the new Route type

lucasloisp:
Seems to stem from `ParamsForScreen` having an explicit `| undefined` on it:

https://github.com/react-navigation/react-navigation/blob/8e7ac4f18545887b905f921df469dbf69d7951c7/packages/core/src/StaticNavigation.tsx#L49

This type, ultimately, seems to only be used for defining `StaticParamList`.
Any reason not to make that no longer be optional? codebase itself passes typechecks after the change.

satya164:
> Any reason not to make that no longer be optional?

Because you can navigate to a screen without needing to specify any child screens - which renders the initial route defined in the navigator. The types need to handle `| undefined`. It's not specific to static config, the same can be done in dynamic config.

packages/core/src/StaticNavigation.tsx
```diff
@@ -46,7 +46,7 @@ type ParamsForScreenComponent<T> = T extends {
 type ParamsForScreen<T> = T extends { screen: StaticNavigation<any, any, any> }
   ? NavigatorScreenParams<StaticParamList<T['screen']>> | undefined
   : T extends StaticNavigation<any, any, any>
-  ? NavigatorScreenParams<StaticParamList<T>> | undefined
+  ? NavigatorScreenParams<StaticParamList<T>>
```
satya164, line 49:
```suggestion
  ? NavigatorScreenParams<StaticParamList<T>> | undefined
```

satya164 (changes requested):

lucasloisp:
> Because you can navigate to a screen without needing to specify any child screens - which renders the initial route defined in the navigator. The types need to handle `| undefined`. It's not specific to static config, the same can be done in dynamic config.

Got it! Missed that on `common` as those types are _defined_ as non-undefined on the test. Having a hard time figuring in out `ParamListRoute` with this though - kinda need to check `ParamList[RouteName] extends (NavigatorScreenParams<infer T extends ParamListBase> OR undefined)` in a sense, but then recursively put in those "undefined" values in the right place.

I'll post updates if I can make progress!

satya164 (approved):
Thank you!

PR #11547, opened 2023-08-19
fix: optimize route cache

okwasniewski:
Optimize route cache to prevent re-rendering.
 
**Motivation**

Routes shouldn't re-render, we are now using route key as the cache key


okwasniewski:
Make sure this works before merging: https://reactnavigation.org/docs/7.x/screen-options-resolution/#setting-parent-screen-options-based-on-child-navigators-state

satya164 (approved):

ranaavneet:
Shouldn't this be back-ported as well?

satya164:
no, it's a breaking change

Issue #9232, opened 2021-01-19
Can't define headers testIDs


paulatomasi:
I'm using detox to E2E tests, for that I need to inform testID props to the components I'm interacting with. Since I started using this new approach to stack navigator, I couldn't find a way to use testID.

```
const HomeStack = createStackNavigator<HomeStackParamsList>()

const HomeNavigator = (): ReactNode => {
  return (
    <HomeStack.Navigator
      initialRouteName={Screens.HomeScreen}
      screenOptions={({ navigation }) => ({
        ...TransitionPresets.SlideFromRightIOS,
        ...getHeaderOptions({ navigation }),
      })}
    >
      [...]
      <HomeStack.Screen
        name={Screens.FindSellerScreen}
        component={FindSellerScreenConnected}
        options={{
          title: 'Find a seller',
        }}
      />
      <HomeStack.Screen
        name={Screens.StoresListScreen}
        component={StoresListScreenConnected}
        options={{
          title: 'Stores list',
        }}
      />
    </HomeStack.Navigator>
  )
```
In the docs, I only find about it [here](https://wix.github.io/react-native-navigation/6.12.2/api/options-bottomTabs/#testid), but it's from bottom tab navigator, which is not the case. 

I thought about something like this: 
```
<HomeStack.Screen
  name={Screens.FindSellerScreen}
  component={FindSellerScreenConnected}
  options={{
    headerRight: () => null,
    title: 'Find a seller',
    testID: 'nav_bar.find_seller',
  }}
/>
```
Is it possible?

jerimiah797:
Still occurring with:

- `@react-navigation/bottom-tabs` 5.11.15
- `@react-navigation/native` 5.9.8
- `@react-navigation/stack` 5.14.9

arttumat:
Still an issue with:
```
"@react-navigation/drawer": "^6.5.0",
"@react-navigation/native": "^6.0.2",
"@react-navigation/native-stack": "^6.1.0",
```

I am trying to write e2e tests using appium, and it looks like Appium cannot see the header at all.

barrymichaeldoyle:
Still an issue 😢

jasongaare:
solved it this way

```tsx
// before
options={{
  title: 'Search'
}}
```

much more verbose - but adds an id for Appium, and looks the same (from my testing, YMMV)

```jsx
// after
import { HeaderTitle, type StackHeaderTitleProps } from '@react-navigation/stack';

options={{
[...]

PR #11559, opened 2023-08-25
feat: added headerTitleTestID prop to HeaderOptions and implemented it

barrymichaeldoyle:
**Motivation**

Solves this issue: https://github.com/react-navigation/react-navigation/issues/9232 (and many others like it)

When writing e2e tests with detox, there is no easy way to target the `headerTitle` other than getting the element by text. but in a lot of cases it is better to get the element by testID,

**Test plan**

This is quite a basic change, linting and existing tests pass. There aren't any existing tests setup for the Header component, if there were I would add a test case.


osdnk (changes requested):
While your proposal is solid, I'm thinking it might be best to put it on hold for now. I'm keen on avoiding a single-location implementation and potentially missing out on other suitable spots for it.

If you're up for it, we could explore a middle ground—perhaps something like adding testId to the entire header, including Drawer, TabBar, etc. This way, it becomes a versatile solution, and folks can write tests based on the text using the `withAncestor` annotation.

What do you think?

PR #11566, opened 2023-08-28
feat: add children prop to tab bar indicator component

grezxune:
**Motivation**

I want to be able to display a `<LinearGradient/>` component inside the `<TabBarIndicator/>`. While you can use the `renderIndicator` prop on `TabBar`, you cannot render a wrapper or child of the indicator and have it inherit the animation the library already provides. By providing a `children` option to `TabBarIndicator`, this problem is easily solved.

**Test plan**

Tests all still pass and to test this specific change, try something like this:

```
<TabBar
  renderIndicator={(props) => {
    return (
      <TabBarIndicator {...props}>
        {someChildComponent}
      </TabBarIndicator>
    )
  }}
/>
```

This allows more complex components to be animated and accurately denote the active tab. I have some screenshots of an example where I'm using a `<LinearGradient/>` container to display a gradient indicator - something not possible with just a style object.

Screenshot 2023-08-28 at 4 16 24 PM
Screenshot 2023-08-28 at 4 16 46 PM

### DIFF
--- a/packages/react-native-tab-view/src/TabBarIndicator.tsx
+++ b/packages/react-native-tab-view/src/TabBarIndicator.tsx
@@ -20,6 +20,7 @@ export type Props<T extends Route> = SceneRendererProps & {
   style?: StyleProp<ViewStyle>;
   getTabWidth: GetTabWidth;
   gap?: number;
+  children?: React.ReactNode;
 };
 
 const getTranslateX = (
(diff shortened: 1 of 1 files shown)


osdnk (approved):
looks nice! Thanks

I guess it wasn't supported earlier as there was no deep motivation, but it looks like you found one :)

satya164:
@grezxune what do you mean? the prop is still present on latest commits https://github.com/react-navigation/react-navigation/blob/7.x/packages/react-native-tab-view/src/TabBarIndicator.tsx#L176

PR #11568, opened 2023-08-30
feat: add useCardAnimation

adamgrzybowski:
**Motivation**

Make animation progress available inside a screen.

**Test plan**

### DIFF
--- a/example/src/Restart.native.tsx
+++ b/example/src/Restart.native.tsx
@@ -1,5 +1,0 @@
-import * as Updates from 'expo-updates';
-
-export function restartApp() {
-  Updates.reloadAsync();
-}
--- a/example/src/Restart.tsx
+++ b/example/src/Restart.tsx
@@ -1,1 +1,0 @@
-export function restartApp() {}
--- a/example/src/Screens/SimpleStack.tsx
+++ b/example/src/Screens/SimpleStack.tsx
@@ -1,6 +1,7 @@
 import type { ParamListBase } from '@react-navigation/native';
 import {
   createStackNavigator,
+  HeaderStyleInterpolators,
   StackNavigationOptions,
   StackScreenProps,
 } from '@react-navigation/stack';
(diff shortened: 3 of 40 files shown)


packages/material-top-tabs/src/utils/useTabAnimation.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
+import { TabAnimationContext } from './TabAnimationContext';
+
+export function useTabAnimation() {
+  const animation = React.useContext(TabAnimationContext);
+
+  if (animation === undefined) {
+    throw new Error(
+      "Couldn't find values for card animation. Are you inside a screen in Tab?"
```
osdnk, line 10: i believe it's worth clarifying that it's material-top-tabs (and not bottom-tabs).

osdnk (changes requested):
Thanks for those changes. I think I need to suggest a slightly different API. In my opinion, we should target a similar API as `useCardAnimation`.

Particularly, I believe that in the future we may support multiple animations. In the same way, as we recently merged #11323, we can also imagine a similar animation inside `material-top-tabs`. 

Therefore, I would suggest making the API more open for multiple animations and returning an object, instead of just a single animated value. The current value may live under the `position` field in the object. With this approach, we will not be forced to make a breaking change (or design a weird API).

`const { position } = useTabAnimation()`

packages/material-top-tabs/src/utils/useTabAnimation.tsx
osdnk: Also, `card animation`? 😅

adamgrzybowski:
Hey @osdnk, I made some changes. Please take a look when you have a moment

osdnk (changes requested):
I think maybe i was not clear about that. I think that the name `useCardAnimation` is not OK here - we don't have any card. `useTabAnimation` is ok.

`Couldn't find values for tab animation. Are you inside a screen in Material Top Tab navigator?` should be the message

adamgrzybowski:
@osdnk Yeah I can see it now 😄 idk why I understood it wrong. I pushed fixed version.

osdnk (approved):

Issue #11481, opened 2023-07-18
TypeScript Error: Type 'OpaqueColorValue' is not assignable to type 'string' when I use PlatformColor


retyui:
### Current behavior

```tsx
import {PlatformColor} from 'react-native';

import {DefaultTheme} from '@react-navigation/native';
import type {Theme} from '@react-navigation/native';

const CallstackTheme: Theme = {
  ...DefaultTheme,
  colors: {
    ...DefaultTheme.colors,
    primary: PlatformColor('systemBlue'), // TS2322: Type 'OpaqueColorValue' is not assignable to type 'string'.
  },
};
```

### Expected behavior

`Theme` type should allow to use `PlatformColor`

### Reproduction

https://snack.expo.dev/404_sorry

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [X] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 6.1.7
| @react-navigation/native-stack         | 6.9.13
| react-native-safe-area-context         |4.7.1
| react-native-screens                   | 3.22.1
| react-native                           | 0.72.x
| node                                   | 20
| npm or yarn                            | yarn v1

retyui:
@kkafar can you reopen this issue, as PR for **REACT-NAVIGATION** isn't merged yet https://github.com/react-navigation/react-navigation/pull/11495

kkafar:
Yep

Thanaen:
Hello !
I wanted to add that resolving this issue would also make it possible to exploit DynamicColorIOS, which allows light/dark theme management without re-render.
Note that this already works today with react-navigation, if you ignore the TypeScript error.

retyui:
@Thanaen it wont't work in case of using `Color` util 

https://github.com/react-navigation/react-navigation/blob/adbf863c148a6d389a69745b8a88c0c706ebfb4c/example/src/Shared/Chat.tsx#L76

osdnk:
This will partially work, but those manipulatation won't work unfortunately

Thanaen:
[...]

PR #11528, opened 2023-08-07
fix: incorrect url parsing (isssue: #10924)

retyui:
**Motivation**

fix: https://github.com/react-navigation/react-navigation/issues/10924

**Test plan**

```tsx
import extractPathFromURL from '@react-navigation/native/src/extractPathFromURL';

test('extractPathFromURL', () => {
  expect(extractPathFromURL(['https://mysite.com'], 'https://mysite.com/readPolicy?url=https://test.com')).toBe(
    '/readPolicy?url=https://test.com',
  );
});
// test fail ^^^ because `extractPathFromURL` removed all double slashes
// Expected: "/readPolicy?url=https://test.com"
// Received: "/readPolicy?url=https:/test.com"
```

PR #11570, opened 2023-08-31
fix: Allow to use `PlatformColor` in the theme

retyui:
Fix: https://github.com/react-navigation/react-navigation/issues/11481

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -53,7 +53,7 @@
     "react": "18.2.0",
     "react-native": "0.71.8",
     "react-native-builder-bob": "^0.21.0",
-    "react-native-screens": "~3.20.0",
+    "react-native-screens": "~3.25.0",
     "typescript": "^4.9.4"
   },
   "peerDependencies": {
--- a/example/package.json
+++ b/example/package.json
@@ -33,7 +33,7 @@
     "react-native-paper": "^4.12.5",
     "react-native-reanimated": "~2.14.4",
     "react-native-safe-area-context": "4.5.0",
-    "react-native-screens": "~3.20.0",
+    "react-native-screens": "~3.25.0",
     "react-native-web": "~0.18.7"
   },
   "devDependencies": {
--- a/example/src/Screens/BottomTabs.tsx
+++ b/example/src/Screens/BottomTabs.tsx
@@ -13,7 +13,7 @@ import {
 import type { StackScreenProps } from '@react-navigation/stack';
 import { BlurView } from 'expo-blur';
 import * as React from 'react';
-import { ScrollView, StatusBar, StyleSheet } from 'react-native';
+import { ColorValue, ScrollView, StatusBar, StyleSheet } from 'react-native';
 
 import { Albums } from '../Shared/Albums';
 import { Chat } from '../Shared/Chat';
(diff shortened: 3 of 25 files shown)


packages/native-stack/package.json
```diff
@@ -53,7 +53,7 @@
     "react": "18.2.0",
     "react-native": "0.71.8",
     "react-native-builder-bob": "^0.21.0",
-    "react-native-screens": "~3.20.0",
+    "react-native-screens": "~3.25.0",
```
retyui, line 56: need to use latest ver. with a types update: https://github.com/software-mansion/react-native-screens/pull/1846#issuecomment-1700507895

osdnk (approved):
Hi,
Thanks for your PR.  Please, next time provide a deeper description of your changes. I didn't understand initially how that change basically "propagates" from rn-screens :)

retyui:
here is more deeper description  https://github.com/react-navigation/react-navigation/pull/11528
but PR hasn't merged yet

osdnk:
Hey, @retyui 

I'm sorry I have to revert this PR #11576.

The concern, which I didn't notice earlier is that as some places (at example/src/Screens/NativeStackHeaderCustomization.tsx) you need to down-cast PlatformColor. Let me think about that. I wonder if this is the only place, but eventually, we shouldn't ask users to do it.

retyui:
right in time when I did a [tweet](https://twitter.com/tell_me_mur/status/1698272773939007627?s=20) lol) 

@osdnk  you don't need to worry about `example/` as theme isn't use a `PlatformColor`'s  there. (Otherwise, you must first resolve this problem: https://github.com/callstack/react-native-paper/issues/4012)

osdnk:
Hi @retyui 
I'm sorry for that, I can empathise it might be disappointing. Let me explain my reasoning here: 
1. It's not about, where is the bug, but more what's the final user experience. With this change in, we get a breaking change (with respect to typings). 
2. In react-navigation, we do color manipulation in many places to improve DX by automating some stuff (search for `Color()` and it only works with string. by supporting `PlatformColor` , those color manipulation are essentially pointless.

Is there a way we can support those kind of manipulations? Also, why do you need this change? Maybe understanding this motivation would help us to find out a good solution.

khagesh:
From our use case, we are trying to use `PlatformColor` for switching between light & dark mode. `PlatformColor` is nice because we don't need to add `if...else` everywhere for theme and then set colors.

The way we can support the manipulation using Color is by checking `PlatformColor` type and then using `resolveColorSync` from `@klarna/platform-colors`. In this way, it won't be a breaking change, and we can have `PlatformColor` as well. But it does add dependency on `@klarna/platform-colors`, and also add checks for `PlatformColor` type inside react-navigation codebase. Personally, I don't think these are huge issues for react-navigation, but you would know better.

khagesh:
I did this workaround for the issue to use `PlatformColor` with react-navigation theme.  Basically I did conversion to string before passing to `NavigationContainer`, and then made sure that NavigationContainer re-renders when color scheme on device changes using `useColorScheme` hook in App.tsx where we render `NavigationContainer`.

```typescript
import { resolveColorSync } from "@klarna/platform-colors";
......

const scheme = useColorScheme();
  const theme = useMemo(() => {
    const navigationThemeColors = {
      primary: resolveColorSync(Colors.primary),
      background: resolveColorSync(Colors.background),
      card: resolveColorSync(Colors.background),
      text: resolveColorSync(Colors.primary),
      border: resolveColorSync(Colors.background),
      notification: resolveColorSync(Colors.primary),
    };
    return {
      ...(scheme === "dark" ? DarkTheme : DefaultTheme),
      colors: {
        ...(scheme === "dark" ? DarkTheme.colors : DefaultTheme.colors),
        ...navigationThemeColors,
      }
    };
  }, [scheme]);

...........

// pass theme 
<NavigationContainer theme={theme}>

```

osdnk:
Alternatively, if you don't rely on our colors' manipulations, you can simply ignore TS error

PR #11572, opened 2023-08-31
fix: bottom bar animation tied with a screen

osdnk:
**Motivation**

Follow-up after the previous PR, when we used only the upper-most descriptor. Also, example has been adjusted to use either fully turned on or off animation., 

**Test plan**

press the button
image

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabView.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabView.tsx
@@ -71,7 +71,7 @@ export function BottomTabView(props: Props) {
           .map((route) => {
             const { animationEnabled, transitionSpec } =
               descriptors[route.key].options;
-            if (!animationEnabled) {
+            if (!animationEnabled || !transitionSpec) {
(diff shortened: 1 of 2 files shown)


packages/bottom-tabs/src/views/BottomTabView.tsx
```diff
@@ -68,35 +66,29 @@ export function BottomTabView(props: Props) {
...
-              ...transitionSpec?.config,
-              toValue: route.key === focusedRouteKey ? 0 : 1,
-              useNativeDriver: true,
+        state.routes
+          .map((route) => {
+            const { animationEnabled, transitionSpec } =
+              descriptors[route.key].options;
+            if (!animationEnabled) {
+              return false;
             }
```
satya164, line 76: I'd change this to:

```suggestion
            if (!animationEnabled || !transitionSpec) {
              return false;
            }
```

satya164 (changes requested):

PR #11578, opened 2023-09-04
feat: add option to show tabs on the side

satya164:
**Motivation**

This adds ability to render the tab bar on left/right as a sidebar. It useful on larger screens where we may want to take advantage of the available horizontal screen and show a sidebar instead of a bottom tab bar.

**Test plan**

Open the "Bottom Tabs" example and resize the screen to see the different modes of the tab bar. Tap on the "<<" button on the bottom left to switch between a smaller and larger sidebar.

**Video**

https://github.com/react-navigation/react-navigation/assets/1174278/224beaa9-8e26-4d94-a70d-11e54c37d944

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -1,4 +1,7 @@
-import { MissingIcon } from '@react-navigation/elements';
+import {
+  getDefaultSidebarWidth,
+  MissingIcon,
+} from '@react-navigation/elements';
 import {
   CommonActions,
   NavigationContext,
@@ -8,6 +11,7 @@ import {
   useLinkTools,
   useTheme,
 } from '@react-navigation/native';
+import Color from 'color';
 import React from 'react';
 import {
   Animated,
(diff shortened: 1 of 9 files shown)


packages/bottom-tabs/src/views/TabBarIcon.tsx
```diff
@@ -50,25 +50,21 @@ export function TabBarIcon({
...
       <Badge
         visible={badge != null}
-        style={[
-          styles.badge,
-          horizontal ? styles.badgeHorizontal : styles.badgeVertical,
-          badgeStyle,
-        ]}
-        size={(size * 3) / 4}
+        style={[styles.badge, badgeStyle]}
+        size={(ICON_SIZE * 3) / 4}
```
osdnk, line 67: * 0.75?

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -256,42 +270,67 @@ export function BottomTabBar({
...
+                          StyleSheet.hairlineWidth,
+                        0,
+                      ],
+                    }),
+                  },
                 ],
-              }),
+                // Absolutely position the tab bar so that the content is below it
+                // This is needed to avoid gap at bottom when the tab bar is hidden
+                position: isTabBarHidden ? 'absolute' : (null as any),
```
osdnk, line 301: Hm, I'm not sure how you like this pattern, but I typically use

```
   ...(isTabBarHidden && { position: 'hidden' })
```

osdnk (commented):
The code works and looks good in general. One question is if the name "bottom-tab" is still valid?

packages/bottom-tabs/src/views/BottomTabBar.tsx
satya164: Changed to `position: isTabBarHidden ? 'absolute' : undefined`

satya164:
> @osdnk: One question is if the name "bottom-tab" is still valid

You could make the argument is that the default one is still bottom, and you could position tabs at the top in material-top-tabs so it's like this already 😅

But yeah, the only real argument I have is to avoid another big breaking change (especially to keep older blog posts, examples etc. working) and to keep our download counts on npm.

Issue #8709, opened 2020-08-12
headerTitle can be a component, headerRight has to be a function


kbrandwijk:
**Current Behavior**
When specifying the options for a screen, there's a discrepancy between headerTitle accepting a component directly (`headerTitle: <MyTitleComponent />`), but headerRight _has_ to be a function (`headerRight: () => <MyRightComponent />`) or you'll get the `TypeError: right is not a function` error, that isn't really nicely handled (as in: not at all).

**Expected Behavior**
Consistent behavior allowing `headerRight` to accept a component directly as well.

**How to reproduce**

`<Stack.Screen name="MyScreen" component={MyScreen} options={{ headerTitle: <MyTitle />, headerRight: <Right /> }} />`

**Your Environment**

| software                       | version |
| ------------------------------ | ------- |
| iOS or Android                 | both
| @react-navigation/native       | 5.7.3
| @react-navigation/stack        | 5.9.0
| react-native-gesture-handler   | 1.7.0
| react-native-safe-area-context | 3.1.1
| react-native-screens           | 2.9.0
| react-native                   | 0.63.1
| expo                           | - 
| node                           | 12.18.2
| npm or yarn                    | yarn

kbrandwijk:
Version numbers updated after retesting on latest

mcsky:
Allowing to pass directly the component is a very good thing because otherwise React seems to recreate the DOM of right header at each render. See https://reactjs.org/docs/reconciliation.html#elements-of-different-types

migsc:
I ran into this today and I thankfully had a workaround but I went in and fixed it anyway because it seemed simple enough: https://github.com/react-navigation/react-navigation/pull/11592

satya164:
If you look at the [documentation](https://reactnavigation.org/docs/elements#headertitle) and the [types](https://github.com/react-navigation/react-navigation/blob/e432c32d1e346f7c40cb42772ae5b515c76ef297/packages/stack/src/types.tsx#L139), `headerTitle` only accepts a string. React Element works because you can often render an element where you can render a string, however this is not a documented feature and may break in the future.

kbrandwijk:
So the implementation is inconsistent, but because the documentation isn't, so the issue gets closed after 3 years 👏🏻

PR #11592, opened 2023-09-11
fix(HeaderSegment): handle headerRight/Left as a component/JSX element

migsc:
Motivation

Solves this issue: https://github.com/react-navigation/react-navigation/issues/8709

The headerRight and headerLeft properties can only be functions and not components/JSX elements. Reasons we'd want to fix this:
- It caused an error that was a little tricky to track down when I wrote code that dynamically updated the button in the top right with a JSX element. This previously worked in older versions of react navigation and broke when I upgraded. So this fix can make upgrades slightly smoother for people
- As [mcsky suggested in the ticket](https://github.com/react-navigation/react-navigation/issues/8709#issuecomment-1246925899), some may prefer to pass a component/element directly to avoid recreating DOM of the right header on render

Test plan

In `/react-navigation/packages/stack/src/__tests__/index.test.tsx` I've added a test case now that renders two screens: one with headerLeft/headerRight passed as functions and the other with elements. The test verifies that they are both rendered on each screen.

UPDATE: Apologies, I initially forgot to update types in some places and circled back to do that. I have fixed my test coverage as well. Please let me know what else I can do to improve this PR.

### DIFF
--- a/packages/elements/src/Header/Header.tsx
+++ b/packages/elements/src/Header/Header.tsx
@@ -185,22 +185,37 @@ export function Header(props: Props) {
     },
   ];
 
-  const leftButton = headerLeft
-    ? headerLeft({
-        tintColor: headerTintColor,
-        pressColor: headerPressColor,
-        pressOpacity: headerPressOpacity,
-        labelVisible: headerLeftLabelVisible,
-      })
-    : null;
+  const leftButton =
+    typeof headerLeft === 'function'
+      ? headerLeft({
+          tintColor: headerTintColor,
+          pressColor: headerPressColor,
+          pressOpacity: headerPressOpacity,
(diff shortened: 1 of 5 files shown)


migsc:
I added a test for this change and updated all the necessary types to get typescript to compile.

satya164:
Thanks for the PR. However, this is currently intentional.

All other `options` in React Navigation accept a function returning a react element and changing header options to accept an element will be inconsistent. In addition, `React.cloneElement` won't work with TypeScript and one of React Navigation's goals is to keep the API typesafe.

> some may prefer to pass a component/element directly to avoid recreating DOM of the right header on render

This is not correct. Passing a function returning a React element won't recreate the DOM.

PR #11602, opened 2023-09-17
Add `useUnhandledLinking`

osdnk:
This code is mostly moved changes from #10939.
Additionally, I propose a different name and add ability to handle the login in nested navigators 

**Motivation**

This PR implements API for handling deep links with conditional rendering. Refer to [Handle deep link + auth in React Navigation](https://gist.github.com/satya164/af3a6f25c03dfbae9ca72ede19d170bc).
**Test plan**

To check, open `index.ts` and change the flag `LINKING_EXAMPLE = true`

Then, use `xcrun simctl openurl booted exp://127.0.0.1:19000/--/outer/profile`


example/src/LinkingPlayground.tsx
```diff
@@ -0,0 +1,133 @@
+import {
+  NavigationContainer,
+  useDeferredLinking,
+} from '@react-navigation/native';
+import { createStackNavigator } from '@react-navigation/stack';
+import { createURL } from 'expo-linking';
+import React from 'react';
+import { Button, StyleSheet, Text, View } from 'react-native';
+
+const HomeScreen = ({ navigation }: any) => (
```
osdnk, line 10: move this to the main app.

packages/native/src/useDeferredLinking.tsx
```diff
@@ -0,0 +1,96 @@
...
+    let rootState = getStateFromPathHelper(path ?? '', config);
+    if (!rootState) {
+      // is that possible?
+      return;
+    }
+
+    const parent = navigation.getParent();
+    if (parent) {
+      // Then, we consider a portion of the state.
+      const parentState = navigation.getParent()!.getState();
```
osdnk, line 58:
```suggestion
      const parentState = parent.getState();
```

extract that to the a separate function

packages/native/src/useDeferredLinking.tsx
```diff
@@ -0,0 +1,96 @@
...
+      }
+      const innerState = state.routes[0].state;
+      if (!innerState) {
+        return;
+      }
+      rootState = innerState;
+    }
+
+    // Then we traverse the root state and find the part of the state that corresponds to this navigator
+    const routesWithKey =
```
osdnk, line 75: this is not needed, should be by the rehydration logic

packages/native/src/__tests__/useDeferredLinking.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+    async getInitialURL() {
+      return 'rn://outer/profile';
+    },
+  };
+
+  const navigation = createNavigationContainerRef<ParamListBase>();
+
+  const App = () => {
+    const [isSignedIn, setSignedIn] = React.useState(false);
+    // const [isSignedIn, setSignedIn] = [false, (_: boolean) => {}];
```
osdnk, line 197: remove that

packages/native/src/__tests__/useDeferredLinking.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+                )}
+              </Stack.Navigator>
+            )}
+          </OuterStack.Screen>
+        </OuterStack.Navigator>
+      </NavigationContainer>
+    );
+  };
+
+  const { queryByText } = await waitFor(() => render(<App />));
```
osdnk, line 242: fix tests not to show warning

packages/native/src/__tests__/useDeferredLinking.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+    async getInitialURL() {
+      return 'rn://profile';
+    },
+  };
+
+  const navigation = createNavigationContainerRef<ParamListBase>();
+
+  const App = () => {
+    const [isSignedIn, setSignedIn] = React.useState(false);
+    // const [isSignedIn, setSignedIn] = [false, (_: boolean) => {}];
```
osdnk, line 72: remove

packages/native/src/useLinking.tsx
```diff
@@ -80,7 +80,8 @@ export function useLinking(
     getStateFromPath = getStateFromPathDefault,
     getPathFromState = getPathFromStateDefault,
     getActionFromState = getActionFromStateDefault,
-  }: Options
+  }: Options,
+  _?: React.MutableRefObject<string | undefined> // native only prop
```
osdnk, line 84: add support to the browser

packages/native/src/useLinking.native.tsx
```diff
@@ -134,6 +135,30 @@ export function useLinking(
...
+    },
+    [ref]
+  );
+
+  const saveLastUnhandledUrl = React.useCallback(
+    (url: string | null | undefined) => {
+      // Save last unhandled url for later use in conditional rendering
+      lastUnhandledURL.current = url;
+
+      // TODO: how to silence warning when it may or may not be handled in conditional rendering?
```
osdnk, line 154: move it

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -432,6 +432,16 @@ export function useNavigationBuilder<
 
   const previousRouteKeyList = previousRouteKeyListRef.current;
 
+  // This state gets set on unhandled deep link
+  const stateForNextRouteNamesChange = React.useRef<State | null>(null);
```
osdnk, line 436: this should be state

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -447,18 +457,28 @@ export function useNavigationBuilder<
...
+          (name) =>
+            previousRouteKeyList.hasOwnProperty(name) &&
+            routeKeyList[name] !== previousRouteKeyList[name]
+        ),
+      });
+    // Clear scheduled state after it's handled
+    stateForNextRouteNamesChange.current = null;
   }
 
+  React.useEffect(() => {
```
osdnk, line 476: this doesn't make sense

packages/native/src/__tests__/useDeferredLinking.test.tsx
```diff
@@ -0,0 +1,379 @@
...
+  ParamListBase,
+  StackRouter,
+  useNavigationBuilder,
+} from '@react-navigation/core';
+import { act, fireEvent, render, waitFor } from '@testing-library/react-native';
+import * as React from 'react';
+import { Button, Text } from 'react-native';
+
+import { NavigationContainer } from '../NavigationContainer';
+import { useDeferredLinking } from '../useDeferredLinking';
```
osdnk, line 13: `useUnhandledLinking` + return an object. + `getUnhandledLink`.

packages/native/src/useDeferredLinking.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    const state = navigation.getState();
+
+    const nextState = {
+      ...state,
+      routes: routesWithKey,
+    };
+
+    // Once we have the state, we can tell React Navigation to use it for next route names change (conditional rendering logic change)
+    navigation.setStateForNextRouteNamesChange(nextState);
```
osdnk, line 89: test the function by itself.

osdnk: print a warning if something happened some action. (not routes change) but we store + clean.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -69,6 +70,14 @@ const getPartialState = (
   };
 };
 
+function usePrevious<T>(value: T): T | undefined {
```
osdnk, line 73: Move it

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,88 @@
...
+    // First, we parse the URL to get the desired state
+    const getStateFromPathHelper =
+      options?.getStateFromPath ?? getStateFromPath;
+
+    const rootState = getStateFromPathHelper(path ?? '', config);
+
+    if (!rootState) {
+      return;
+    }
+    const state = extractNavigatorSpecifcState(navigation, rootState);
```
osdnk, line 73: Return if undefined

packages/native/src/__tests__/useUnhandledLinking.test.tsx
```diff
@@ -0,0 +1,373 @@
...
+  expect(queryByText('SignIn')).not.toBeNull();
+
+  fireEvent.press(queryByText(/sign in/i));
+
+  expect(queryByText('SignIn')).toBeNull();
+
+  expect(queryByText('Profile')).not.toBeNull();
+});
+
+it('should schedule a state to be handled on conditional linking in nested stack', async () => {
```
satya164, line 248: For tests, I use the pattern "does something" instead of "should do something" since it's shorter and means the same thing.

e.g. - `it('schedules a state....`

packages/native/src/__tests__/useUnhandledLinking.test.tsx
```diff
@@ -0,0 +1,373 @@
...
+  useNavigationBuilder,
+} from '@react-navigation/core';
+import { act, fireEvent, render, waitFor } from '@testing-library/react-native';
+import * as React from 'react';
+import { Button, Text } from 'react-native';
+
+import { NavigationContainer } from '../NavigationContainer';
+import { useUnhandledLinking } from '../useUnhandledLinking';
+
+it('should schedule a state to be handled on conditional linking', async () => {
```
satya164, line 15: conditional rendering*

packages/native/src/__tests__/useUnhandledLinking.test.tsx
```diff
@@ -0,0 +1,373 @@
...
+        Home: 'home',
+        Profile: 'profile',
+      },
+    },
+    getInitialURL() {
+      return 'rn://profile';
+    },
+  };
+
+  const navigation = createNavigationContainerRef<ParamListBase>();
```
satya164, line 68: the ref here and in other tests in this file seem to be unused

packages/native/src/__tests__/useUnhandledLinking.test.tsx
```diff
@@ -0,0 +1,373 @@
...
+              )}
+            </Stack.Screen>
+          )}
+        </Stack.Navigator>
+      </NavigationContainer>
+    );
+  };
+
+  const app = render(<App />);
+  const { queryByText } = await waitFor(() => app);
```
satya164, line 108: not sure if this is correct, `waitFor` works by waiting until the code inside doesn't throw. for example,

```js
await waitFor(() => {
  expect(getByText('SignIn')).toBeOnTheScreen();
})
```

in this case, the callback passed to `waitFor` will be called again and again until the code inside (in this case `expect(getByText('SignIn')).toBeOnTheScreen();`) doesn't throw anymore, or it times out.

if you're just returning something without throwing any error then `waitFor` won't do anything. same comment for `waitFor` in other places in this file

packages/native/src/__tests__/useUnhandledLinking.test.tsx
```diff
@@ -0,0 +1,373 @@
...
+                />
+              )}
+            </Stack.Screen>
+          )}
+        </Stack.Navigator>
+      </NavigationContainer>
+    );
+  };
+
+  const consoleWarnMock = jest.spyOn(console, 'warn').mockImplementation();
```
satya164, line 352: What's the mock for?

packages/native/src/useLinking.native.tsx
```diff
@@ -134,6 +135,17 @@ export function useLinking(
     []
   );
 
+  const validateRoutesNotExistInRootState = React.useCallback(
```
satya164, line 138: Avoid `useCallback` for these cases as React doesn't guarantee that it won't be cleared. Use `use-latest-callback` instead. or better, define this function inside the effect/inline the logic as before.

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    // Finally, we clear unhandled link after it was handled
+    lastUnhandledLinking.current = undefined;
+  };
+
+  const getUnhandledLink = useLatestCallback(
+    () => lastUnhandledLinking.current
+  );
+
+  return { handleLastLinking, getUnhandledLink };
```
satya164, line 91: I'd call it `handleLastLink` since `Linking` refers to the feature as a whole.

```suggestion
  return { handleLastLink, getUnhandledLink };
```

Though probably `handleOnNextRouteNamesChange()` is more descriptive on what it does as it doesn't handle it immediately.

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (!state) {
+      return;
+    }
+
+    // Once we have the state, we can tell React Navigation to use it for next route names change (conditional rendering logic change)
+    // @ts-expect-error: this is ok
+    navigation.setStateForNextRouteNamesChange(state);
+
+    // Finally, we clear unhandled link after it was handled
+    lastUnhandledLinking.current = undefined;
```
satya164, line 84: We should have some tests to ensure that it gets cleared after handling

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,92 @@
...
+      return;
+    }
+
+    const { config, prefixes } = options;
+
+    // at web, the path is already extracted
+    const path = Platform.select({
+      default: extractPathFromURL(prefixes, lastUnhandledLinking.current),
+      web: lastUnhandledLinking.current,
+    });
```
satya164, line 62: Can we handle the inconsistency inside the `useLinking` hook itself to make it consistent?

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const linking = React.useContext(LinkingContext);
+
+  const { options, lastUnhandledLinking } = linking;
+
+  /*
+   * Function to handle last unhandled URL. This function has to be called when the conditional
+   * rendering of the navigator is about to happen e.g. in the `onPress` of a log in button.
+   */
+  const handleLastLinking = () => {
+    if (options == null || lastUnhandledLinking?.current == null) {
```
satya164, line 51: we still handle deep links when `options` isn't defined.

```suggestion
    if (lastUnhandledLinking?.current == null) {
```

packages/native/src/useUnhandledLinking.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const path = Platform.select({
+      default: extractPathFromURL(prefixes, lastUnhandledLinking.current),
+      web: lastUnhandledLinking.current,
+    });
+
+    // First, we parse the URL to get the desired state
+    const getStateFromPathHelper =
+      options?.getStateFromPath ?? getStateFromPath;
+
+    const rootState = getStateFromPathHelper(path ?? '', config);
```
satya164, line 68: if `path` is `undefined`, it shouldn't come here, right?

satya164 (changes requested):
Thanks for the hard work. I did a quick review and left some comments. I will do another pass and test it locally when I get some more time.

Also one other thing which I think is not implemented:

- If user calls this function and there was some other navigation before we could handle the deep link, then we should warn.
- If user calls this function without any link, maybe we should warn?

packages/native/src/__tests__/useUnhandledLinking.test.tsx
osdnk: forget to remove

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -447,16 +461,23 @@ export function useNavigationBuilder<
...
-      routeParamList,
-      routeGetIdList,
-      routeKeyChanges: Object.keys(routeKeyList).filter(
-        (name) =>
-          previousRouteKeyList.hasOwnProperty(name) &&
-          routeKeyList[name] !== previousRouteKeyList[name]
-      ),
-    });
+    nextState = navigatorStateForNextRouteNamesChange
+      ? // @ts-expect-error this is ok
```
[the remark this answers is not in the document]
osdnk: unfortunately

osdnk:
> If user calls this function without any link, maybe we should warn
I believe this is perfectly valid and there's no need ot warn in this case.

Otherwise, all of your feedback was applied

example/src/Screens/LinkingScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+  createStackNavigator,
+  StackScreenProps,
+} from '@react-navigation/stack';
+import React, { useContext } from 'react';
+import { Button, Platform, StyleSheet, Text, View } from 'react-native';
+
+const info = `
+\u2022 xcrun simctl openurl booted exp://127.0.0.1:19000/--/linking/profile
+\u2022 adb shell am start -W -a android.intent.action.VIEW -d "exp://127.0.0.1:19000/--/linking/profile" host.exp.exponent
+\u2022 http://localhost:19006/linking/profile
```
satya164, line 12: fyi you can do `npx uri-scheme open exp://127.0.0.1:19000/--/linking/profile` with `--ios`/`--android`

satya164 (approved):

PR #11640, opened 2023-10-12
print invalid routes names on incorrect deep link

Maxim-Filimonov:
**Motivation**

When dealing with complex navigation cofiguration it's very tricky to determine which path are invalid without any diagnostic information from react-navigation. 
This adds printing of invalid routes names to make it easier to understand what went wrong.

**Test plan**

Try to use invalid configuration for deep linking. Names of invalid configs should be printed comma separated.


osdnk (commented):
Hey,
After introducing the new API for unhandled linking, that working is not longer present. 

I'm sorry for that, but this PR is no longer valid against the updated main branch

PR #11651, opened 2023-10-21
fix: right drawer layouts display over content

brannonvann:
**Motivation**

Using the example project if you set the `drawerPosition="right"` and `drawerType="front"` (the default for the example) the drawer is positioned in the middle of the screen. Other combinations of these settings are also affected including at least `drawerType="slide"` with `drawerPosition="right"`. 

It seems to be platform independent. I'm able to reproduce on web, iOS, and Android. Below is a screenshot from web on mobile in landscape. 

Image of the issue: 
image

When using the RTL setting in the example app it became clear that It needs the same logic for `isRTL` that the Drawer Navigator has. 

I took the logic from the Drawer Navigator related to `I18nManager` from `react-native` and leveraged it in the react-native-drawer-layout. It appears these two extremely similar. Perhaps it's worth leveraging the react-native-drawer-layout package for drawer navigation and maintaining it in one place but that's beyond the scope of this pr/issue.

**Test plan**

To test  using the example, set `drawerPosition="right"` and `drawerType="front"` (example currently had front already set) for a Drawer from react-native-drawer-layout. The layout will should be off screen on the right. Not in the middle hovering above the content.

Image of resolved issue (with drawer open).
image


osdnk (changes requested):
Hey,
thanks for the PR!

Please, do not use the `isRTL` directly. You should have that information coming from the `drawerPosition` prop.

See that `Drawer.tsx` (from the outer folder) is setting `drawerPosition` accordingly

brannonvann:
Thanks @osdnk  for getting back to me on this one. I took a look at what you said. I believe it was initially using `drawerPosition` but not adding in the offset for drawerPosition`"right"` vs `"left"` and the `isRTL` logic I added was masking the problem. The solution was to simply add the offset based on the `drawerPosition` to the `translateX`. Please let me know what you think.

likeSo:
All of the "right" positioned have incorrect behavior. 
`drawerPosition: 'right'` + `drawerType: 'front'`: (drawer closed)
IMG_2BC28349FE43-1

`drawerPosition: 'right'` + `drawerType: 'back'`: (drawer opened)
IMG_43046EBD0062-1

`drawerPosition: 'right'` + `drawerType: 'permanent'`: (drawer closed)
IMG_132D3970E105-1

Does this PR contains these scenes?

stoickeyboard:
These changes worked for me. Thanks!

tmKnight01:
iis useful for me. many thanks !

tmKnight01:
i  very much doubt that the Drawer library has not been unit tested. common function and 100% reproducibility,  but this problem has not resolved so long. i`m disappointed with the react-navigation team.

satya164 (approved):
Thank you

satya164 (approved):
Thanks for the PR

PR #11652, opened 2023-10-21
feat: add overlayEnable option in drawer layout

brannonvann:

**Motivation**

I like the react-native-drawer-layout and want to use it to host some settings that will pull out in a drawer. In my case, the requirements are that the drawer open and close but the content remain interactive. Currently, when using `drawerType="front"`, `drawerType="slide"`, and `drawerType="back"` and the drawer is open the content in the drawer is covered in an overlay that close the drawer when one attempts to interact with the content (children). This works well for use in a navigation but in my case I need to be able to interact with the content like `drawerType="permanent"` but open and close like `drawerType="front"`. 

I added a boolean option available on Drawer props following the naming conventions called `overlayEnabled` to allow the overlay to be turned off. It defaults to true to maintain the original functionality but can be set to false to allow the users to interact with the drawer content and leave the drawer open. 

**Test plan**

1. Use the repo example project. 
2. In the DrawerView.tsx set the Drawer prop `overlayEnabled={false}` and notice that you can click on the button to close the drawer in the screen (from the Drawer children). 
3. In the DrawerView.tsx set the Drawer prop `overlayEnabled={true}` the overlay appears over the content (children). Clicking any part of the will close the drawer.
4. In the DrawerView.tsx omit the Drawer prop `overlayEnabled` the overlay appears over the content (children). Clicking any part of the will close the drawer.

### DIFF
--- a/packages/react-native-drawer-layout/src/types.tsx
+++ b/packages/react-native-drawer-layout/src/types.tsx
@@ -79,6 +79,13 @@ export type DrawerProps = {
   drawerStyle?: StyleProp<ViewStyle>;
 
   /**
+   * Whether the overlay is enabled.
+   * If false, interaction is possible with the drawer content while drawer is open.
+   * Defaults to `true`.
+   */
+  overlayEnabled?: boolean;
+
+  /**
    * Style object for the drawer overlay.
    */
   overlayStyle?: StyleProp<ViewStyle>;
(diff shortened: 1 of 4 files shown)


satya164 (changes requested):
Thanks for the PR. Is there any reason `overlayStyle={{ display: 'none' }}` isn't sufficient? I'd rather not add additional props unless necessary.

PR #11663, opened 2023-10-25
feat: add useAnimatedHeaderHeight hook

tboba:
**Motivation**

Currently when user wants to get header's height, its only option is to use `useHeaderHeight` hook. The main downside of using it is that it returns hardcoded values that may cause issues like wrong calculated header height (see #11338). That's why in react-native-screens we've allowed users to use the hook for calculating header's height dynamically with the use of `useAnimatedHeaderHeight` and `useReanimatedHeaderHeight` (available in `react-native-screens/reanimated`) hooks. 

This PR introduces this hook and adds a logic of putting the header's height to the Animated.Value in `NestedStackView.tsx`.

**Test plan**

The easiest possible way to test the hook is to create FullWindowOverlay.
First, go to example app -> `NativeStack.tsx`. Then, add the component that uses useAnimatedHeaderHeight (import `Animated` and `Text` from react-native as well):

```tsx
import { FullWindowOverlay } from 'react-native-screens';

function ExampleScreen() {
  const headerHeight = useAnimatedHeaderHeight();

  return (
    <FullWindowOverlay>
      <Animated.View
        style={{
          display: 'flex',
          justifyContent: 'flex-end',
          alignItems: 'center',

          backgroundColor: 'red',
          width: '100%',
          opacity: 0.5,
          height: 60,
          zIndex: 1,
          transform: [
            {
              translateY: headerHeight,
            },
          ],
        }}>
        <Text>I'm a header!</Text>
      </Animated.View>
    </FullWindowOverlay>
  );
}
```

Then, add this component to the view, e.g. to `ArticleScreen` from `NativeStack.tsx`. After that, try swiping the page and see how the red rectangle scrolls with the large header 🎉

https://github.com/react-navigation/react-navigation/assets/23281839/d7eb6658-767d-415e-a576-4d31873d2019


tboba:
Oh, I see I didn't attached test yet 🙈 I'll do that shortly!
One thing about failing `CI / dependencies` task - I think I shouldn't right now bump `react-native-screens` version in the example, as it could require bumping Expo. Am I right there, or should I bump it no matter of Expo version?

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -236,6 +237,13 @@ const SceneView = ({
   const [customHeaderHeight, setCustomHeaderHeight] =
     React.useState(defaultHeaderHeight);
 
+  const cachedAnimatedHeaderHeight = React.useRef(customHeaderHeight);
```
osdnk, line 240: we have `useAnimatedValue` hook for that

packages/native-stack/src/index.tsx
```diff
@@ -8,6 +8,12 @@ export { createNativeStackNavigator } from './navigators/createNativeStackNaviga
  */
 export { NativeStackView } from './views/NativeStackView';
 
+/**
+ * Hooks
+ */
+export { AnimatedHeaderHeightContext } from './utils/AnimatedHeaderHeightContext';
```
osdnk, line 14: Well, that's the context, not a hook

packages/native-stack/src/utils/AnimatedHeaderHeightContext.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
+import type { Animated } from 'react-native';
+
+export const AnimatedHeaderHeightContext = React.createContext<
```
osdnk, line 4: Personally, I'd put it next to the hook and not export that context at all

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -457,6 +475,21 @@ function NativeStackViewInner({ state, navigation, descriptors }: Props) {
...
+const handleHeaderHeightChangeEvent = (
+  height: number,
+  cachedHeaderHeight: React.MutableRefObject<number>,
+  animatedHeaderHeight: Animated.Value
+) => {
+  if (cachedHeaderHeight.current !== height) {
+    // Currently, we're setting value by Animated#setValue, because we want to cache animated value.
+    // Also, in React Native 0.72 there was a bug on Fabric causing a large delay between the screen transition,
+    // which should not occur.
+    // TODO: Check if it's possible to replace animated#setValue to Animated#event.
```
osdnk, line 487: I think that is quite important. If it doesn't work, leave a note why

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -236,6 +237,13 @@ const SceneView = ({
   const [customHeaderHeight, setCustomHeaderHeight] =
     React.useState(defaultHeaderHeight);
 
+  const cachedAnimatedHeaderHeight = React.useRef(customHeaderHeight);
+  const animatedHeaderHeight = React.useRef(
+    new Animated.Value(customHeaderHeight, {
+      useNativeDriver: true,
```
osdnk, line 243: ...and you don't need that in a value

osdnk (changes requested):
If the update is not happening via `Animated.event`, I don't a big value of that change.

I.e., one can just transform a value to an animated value pretty easily outside of the library. 

Did you try to make it an animated value? Wh

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: I'll try that, thanks!

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: I didn't know that 😄
Resolved in https://github.com/react-navigation/react-navigation/pull/11663/commits/f7947728c3bab59c1d88585d3c819d85bf540f44.

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: Sure, I rewrited todo in https://github.com/react-navigation/react-navigation/pull/11663/commits/e8f8bb1099524bc357b7bd1bbaf7d415be7628c9.

packages/native-stack/src/utils/AnimatedHeaderHeightContext.tsx
tboba: I thought about adding an ability for users to use Consumer from the context, but tbh I don't know any specific reason why it wouldn't be possible to just use the value from the hook 😛 

Resolved in https://github.com/react-navigation/react-navigation/pull/11663/commits/c1aa5fef1ef2d7cf15898baf611367da094f98a2, https://github.com/react-navigation/react-navigation/pull/11663/commits/e107aa10f940ba57a31f1a2806d3b1ebddccbf3d.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -457,6 +473,23 @@ function NativeStackViewInner({ state, navigation, descriptors }: Props) {
...
+  cachedHeaderHeight: React.MutableRefObject<number>,
+  animatedHeaderHeight: Animated.Value
+) => {
+  if (cachedHeaderHeight.current !== height) {
+    // Currently, we're setting value by Animated#setValue, because we want to cache animated value.
+    // Unfortunately, because of the bug that exists on Fabric, where native event drivers
+    // for Animated objects are being created after the first notifications about the header height
+    // from the native side, it's not possible to use Animated#event, as
+    // we won't receive initial values on appearing screens.
+    // TODO(tboba): Replace animated#setValue to Animated#event.
```
osdnk, line 487: Is there a way to set it only initially and then rely on the animated.event? 
Please, attach a link, etc.

packages/native-stack/src/index.tsx
tboba: You're right 😄 removed this export in https://github.com/react-navigation/react-navigation/pull/11663/commits/c1aa5fef1ef2d7cf15898baf611367da094f98a2.

packages/native-stack/src/views/NativeStackView.native.tsx
osdnk: Sorry, I linked to the wrong line.

The following:
```
const animatedHeaderHeight = React.useRef(
    new Animated.Value(customHeaderHeight, {
      useNativeDriver: true,
    })
  ).current;
````

can be replaces with `useAnimatedValue`

File: packages/native-stack/src/views/NativeStackView.native.tsx (hunk 6 above)
Comment (tboba) [reply packages/native-stack/src/views/NativeStackView.native.tsx:487]: Yeah it's possible (and currently we're [using](https://github.com/tboba/react-navigation/blob/%40tboba/react-native-screens-3.27.0/packages/native-stack/src/views/NativeStackView.native.tsx#L242) values from `useHeaderHeight` as initial value), but I think that we shouldn't really avoid this problem - this hook not only provide the values on the header height updates, but also guarantees that initial values are the real ones that comes from the native side - this could fix a potential problem that users are facing when there's a new iOS update that sometimes changes the status bar height (take a look on #11338, #11655 😄). I think this hook should provide as pure as possible experience of getting the values of header height dynamically.

Currently I know where the problem lies and I'll try to patch it in official react-native repo, but for now using the `setValue` should be sufficient (even if this works really flaky on Fabric). What do you think?

Also, thanks for spending the time on reviewing this PR! ❤️

Comment (osdnk) [reply packages/native-stack/src/views/NativeStackView.native.tsx:487]: As mentioned earlier in the comment concluding my first review, I fail to see the motivation for this PR if we don't use Animated.event.

Particularly, now the user can simply do 

```
const animatedHeight = useAnimatedValue()
const  height = useHeaderHeight()
animatedHeight.setValue(height) // maybe check if equal with ref, then avoid update 
```

Comment (tboba) [reply packages/native-stack/src/views/NativeStackView.native.tsx:487]: Wouldn't it be possible even if we use Animated.event? I believe we have no influence on whether the user sets the value on its own or not 🤷. There's only a difference on updating the value during the `onHeaderHeightChange` event. The main downside of using `setValue` is that we cannot use animated driver on setting the value.

Comment (osdnk) [reply packages/native-stack/src/views/NativeStackView.native.tsx:487]: I didn't mean I wanted to prevent the user from doing anything. My point is that the exposed hook functionality (`useAnimatedHeaderHeight`) proposed in this PR can be quickly achieved by the lines of code I pasted above. 

In other words, why do we need to add an extra hook for doing the functionality, which is already easily available right now?

Comment (tboba) [reply packages/native-stack/src/views/NativeStackView.native.tsx:487]: Oh, yeah I understand right now.
The main reason is that `useHeaderHeight` works by returning hardcoded values. This prevents the user to follow the header height updates resulted by, for example, using and swiping the large header on iOS. As I've mentioned above, it also guarantees that values returned by this hook will always be correct, no matter of status bar height changes from iOS, as it returns the values from the native side (check the example above how does it work in practice).

The main difference is that useHeaderHeight returns the static values by checking the conditions from JS, whereas useAnimatedHeaderHeight returns the values from the native side dynamically by receiving events. I agree that user can simply put the value from useHeaderHeight to useAnimatedValue, but it won't be updated dynamically, sadly.

Comment (osdnk) [reply packages/native-stack/src/views/NativeStackView.native.tsx:491]: Okay, I see. It was me not understanding how it works.
In this case, I think, I'd prefer to make it an `Animated.event` anyway and accept that it won't work for now. 
It is okay as eventually it will be fixed in the RN core. Furthermore, we would merge it to the main branch and the 7.x still has a long enough backlog so the issue will be probably fixed. Let's leave a TODO there.

Comment (tboba) [reply packages/native-stack/src/views/NativeStackView.native.tsx]: Sure! I've replaced that in https://github.com/react-navigation/react-navigation/pull/11663/commits/d202582e238288c04f5185449a739211c345e1ec.

File: packages/native-stack/src/views/NativeStackView.native.tsx [hunk 7]
```diff
@@ -294,6 +303,21 @@ const SceneView = ({
       onHeaderBackButtonClicked={onHeaderBackButtonClicked}
       preventNativeDismiss={isRemovePrevented} // on iOS
       onNativeDismissCancelled={onNativeDismissCancelled}
+      // @ts-expect-error this prop is available since rn-screens 3.26
+      // Unfortunately, because of the bug that exists on Fabric, where native event drivers
+      // for Animated objects are being created after the first notifications about the header height
+      // from the native side, `onHeaderHeightChange` event does not notify
+      // `animatedHeaderHeight` about initial values on appearing screens at the moment.
```
tboba, line 310: Just one note from me, regarding the comment - I've decided not to add a TODO here, as fixing Animated.event is not related to native stack. I'll remove this comment once the Animated.event on Fabric will be fixed 👍

osdnk: There is a TODO on our end to check if it works. I believe that before releasing we'll search for TODOs so I'd suggest something like this:
`TODO: before realising a stable version, check if the following works`

packages/native-stack/src/types.tsx
```diff
@@ -31,6 +31,10 @@ export type NativeStackNavigationEventMap = {
    * Event which fires when a transition animation ends.
    */
   transitionEnd: { data: { closing: boolean } };
+  /**
+   * Event which fires when a header height gets changed.
+   */
+  headerHeightChange: { data: { headerHeight: number } };
```
osdnk, line 37: So we're not using that event anymore?

osdnk (approved):
See the two final comments. Overall, that seems to be reasonable!

packages/native-stack/src/views/NativeStackView.native.tsx
satya164: @osdnk maybe add it to our board instead 🤔

packages/native-stack/src/types.tsx
satya164, line 37: Let's not add an event for this. IMO people shouldn't be using this directly. And with event, we'd be emitting values constantly during scroll even if no one is using it.

packages/native-stack/src/utils/useAnimatedHeaderHeight.tsx
```diff
@@ -0,0 +1,18 @@
...
+export const AnimatedHeaderHeightContext = React.createContext<
+  Animated.Value | undefined
+>(undefined);
+
+export function useAnimatedHeaderHeight() {
+  const animatedValue = React.useContext(AnimatedHeaderHeightContext);
+
+  if (animatedValue === undefined) {
+    throw new Error(
+      "Couldn't find the header height. Are you inside a screen in a navigator with a header?"
```
satya164, line 13: Let's say "screen in a native stack navigator"

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -294,6 +303,21 @@ const SceneView = ({
...
+      onHeaderHeightChange={Animated.event(
+        [
+          {
+            nativeEvent: {
+              headerHeight: animatedHeaderHeight,
+            },
+          },
+        ],
+        { useNativeDriver: true }
```
satya164, line 319: In cases other than large title, we should add a listener here so we can provide correct value in useHeaderHeight as well. But can be done in a follow up.

satya164 (changes requested):

packages/native-stack/src/types.tsx
tboba: Yep, I've exposed this event type if anyone would like to use it, but tbh I don't think if anyone would like to emit it 😅

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: Won't it break the purpose of the useHeaderHeight? For me `useHeaderHeight` hook is great when you need to have the values that are being updated synchronously and when your app has heavy components that cannot use `useAnimatedHeaderHeight` because of the amount of component updates that uses this hook. In react-native-screens we also have a note about it: `Please beware of using this hook in heavy components, as it may result in performance issues.`.

If user wants to have direct access to the dynamic header's height I would suggest to use `useReanimatedHeaderHeight`, as it uses SharedValue, rather than `useAnimatedHeaderHeight` and going for a listener there. We also have some plans for updating header's height synchronously, but everything will be in its time 😁 

What do you think about it?

packages/native-stack/src/types.tsx
tboba: Removed in https://github.com/react-navigation/react-navigation/pull/11663/commits/d075678ee60d13ec2d4f2615dd665566c7d37a03 👍

tboba: Removed in https://github.com/react-navigation/react-navigation/pull/11663/commits/d075678ee60d13ec2d4f2615dd665566c7d37a03 👍

packages/native-stack/src/utils/useAnimatedHeaderHeight.tsx
tboba: Done! https://github.com/react-navigation/react-navigation/pull/11663/commits/02ad6eb0028d1bdd83ecf520eb293c20b9fc392e 👍

packages/native-stack/src/views/NativeStackView.native.tsx
satya164: I'm talking about the header without large title, so there won't be too many updates.

The reason is that the value is hardcoded in JS side for useHeaderHeight and not the actual value. With a listener we should be able to get proper value.

tboba: Oh, okay I understand 😄 
Then I would suggest to do this as a followup. I'll need to check if `headerHeight` prop in NativeStackView should be changed in any way because of that and furthermore I'll do a refactor or getting status bar height that is being used by `getDefaultHeaderHeight` method 👍

satya164: Sounds good

satya164 (approved):

PR #11672, opened 2023-10-28
New API for unhandled linking

osdnk:
To make the API easier, we decided to make the hook attached directly to the navigator, instead of returning the function to handle the link.

Now, an unhandled link will be handled on the route change via the provided callback. In order to cancel unhandled links, we provide an additional function `clearUnhandledLink`


packages/routers/src/types.tsx
```diff
@@ -125,6 +125,13 @@ export type DefaultRouterOptions<RouteName extends string = string> = {
    * If not specified, usually the first route is used.
    */
   initialRouteName?: RouteName;
+  /**
+   A function returning a state, which may be set after modifying the routes name.
+   */
+  getStateForRouteNamesChange?: (
+    state: NavigationState
+  ) => PartialState<NavigationState> | undefined;
+  linking?: any;
```
satya164, line 134: These don't look right 🤔

`getStateForRouteNamesChange` is an option for `useNavigationBuilder` and `linking` is not for router.

satya164 (changes requested):

PR #11676, opened 2023-10-30
Update CardStyleInterpolators.tsx

ahmed-gameel:
Removed the inversion of transition when the axis is Y in RTL view since only the X axis should be changed

When using the an app in Arabic the views will be inverted in the X axis and will be from Right To Left instead of Left To Right, when using the prebuilt animations in CardStyleInterpolators the animations will be inverted and it's ok unless it's the Y axis transitions since there is no change in it. 
I have removed all the inversion of the animation in RTL and fixed the animation that come from the bottom and Vertical direction in general

### DIFF
--- a/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
+++ b/packages/stack/src/TransitionConfigs/CardStyleInterpolators.tsx
@@ -68,17 +68,13 @@ export function forHorizontalIOS({
  */
 export function forVerticalIOS({
   current,
-  inverted,
   layouts: { screen },
 }: StackCardInterpolationProps): StackCardInterpolatedStyle {
-  const translateY = multiply(
-    current.progress.interpolate({
+  const translateY = current.progress.interpolate({
       inputRange: [0, 1],
       outputRange: [screen.height, 0],
       extrapolate: 'clamp',
-    }),
-    inverted
-  );
+    });
 
   return {
     cardStyle: {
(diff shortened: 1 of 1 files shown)


osdnk:
I'm confused what does it have in common with RTL?

ahmed-gameel:
Y axis should not be inverted in RTL since only the X axis is mirrored. In
these animations even the Y axis is inverted and it shouldn't be.
For example when using the ios vertical transition on normal ltr it comes
from the bottom to the top but when the app is in RTL mode it's inverted
and it comes from the top of the screen. I have fixed this and made it only
occur on horizontal transitions

satya164:
The inverted value is set based on RTL only when it has a horizontal gesture https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/utils/getInvertedMultiplier.tsx

It is not inverted on vertical gestures. So this is only an issue if you are using horizontal gesture with a vertical animation. If this is the case then you need to fix your code to specify the gesture direction to match the animation direction.

The style interpolator needs to be inverted if user specified gestureDirection: 'vertical-inverted'. It's not correct just to ignore inverted entirely.

ahmed-gameel:
No it's not, it's also inverted in vertical transition you can review the
code and test it with enabling rtl

satya164:
> No it's not, it's also inverted in vertical transition you can review the

You haven't provided any test plan for the change or code to repro the issue. Besides you can clearly see in the code I linked that the inverted is set according to RTL only for horizontal direction.

PR #11680, opened 2023-11-01
fix!: don't use screen from params as initialRouteName

satya164:
BREAKING CHANGE:

Previously when using nested navigation API, we used that screen as `initialRouteName`.

e.g.

```js
navigation.navigate('MyScreen', { screen: 'NestedScreen' });
```

Here, the `'NestedScreen'` will be set as `initialRouteName`.

It has the result of rendering the navigator with that screen as focused screen. While this behaviour is expected, that screen was also used as the `initialRouteName` for every other operation ignoring the `initialRouteName` prop - which is not expected.

This change refactors the way the focused screen for initial render is set to avoid this behavior.


osdnk (approved):
works

PR #11685, opened 2023-11-03
refactor: rework how sync state is handled

### LINKED ISSUE #9739: Navigation state not always reset when navigators are unmounted
### Current behavior

I have a skeleton app with the following screen hierarchy:
```
StackNavigator
    LoginScreen
TabNavigator
    HomeScreen
    StackNavigator
        ProfileScreen
        SettingsScreen
``` 

Screens are conditional (as per [auth flow docs](https://reactnavigation.org/docs/auth-flow/)), so only the `LoginScreen` is mounted when the user is unauthenticated, and vice-versa for the remaining screens.

When logging out via the `ProfileScreen` and then logging back in, the user is returned to the `HomeScreen`, as expected.

When logging out via the `SettingsScreen` and then logging back in, the user is returned to the `SettingsScreen`, which is unexpected. I can't figure out a way to reset that navigation state.

I've created a Snack and a video (with audio explanation) of the issue, and put the code up on GitHub.

### Expected behavior

Given that the tabs and stacks are unmounted after logging out, I would expect that logging back in would reset the navigation state and return the user to the initial tab.

Thanks!

### Reproduction

Snack (v5): https://snack.expo.io/SCWaKO40u
Snack (v6): https://snack.expo.dev/gIbYd7EX0
GitHub (v5): https://github.com/chrisparton1991/react-navigation-issue
Video (with audio): https://imgur.com/a/8Og3Cjc

### Platform

- [X] Android
- [X] iOS
- [X] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/stack

### Environment

| @react-navigation/native               | 5.9.4, 6.0.0
| @react-navigation/bottom-tabs          | 5.11.11, 6.0.0
| @react-navigation/drawer               | 
| @react-navigation/material-bottom-tabs |
| @react-navigation/material-top-tabs    | 
| @react-navigation/stack                | 5.14.5, 6.0.0
| react-native-tab-view                  | 
| react-native-pager-view                | 
| react-native                           | 0.64.1
| expo                                   | 
| node                                   | 14.15.1
| npm or yarn                            | npm 6.14.8

satya164:
This simplifies the sync store for navigation state:

- Use a external store to store the state instead of `useState` - so the store can be updated synchronously
- Use `useSyncExternalStore` API to subscribe to the state changes
- Use an effect when a child navigator needs to update state due to list of screens changing etc.
- Removed timeout from state cleanup - fixes #9739

The changes should make the code easier to understand, as well as should make the code compatible with concurrent mode as we're not doing anything weird with the state.


packages/core/src/useNavigationBuilder.tsx
```diff
@@ -571,31 +566,21 @@ export function useNavigationBuilder<
...
-      // For example, if another navigator has started rendering, we should skip cleanup
-      // Otherwise, our cleanup step will cleanup state for the other navigator and re-initialize it
-      setTimeout(() => {
-        if (getCurrentState() !== undefined && getKey() === navigatorKey) {
-          cleanUpState();
-        }
-      }, 0);
+      if (getCurrentState() !== undefined && getKey() === navigatorKey) {
+        setCurrentState(undefined);
+        stateCleanedUp.current = true;
```
osdnk, line 571: this doesn't seem to be ideal, but it looks like it was already here

packages/core/src/useSyncState.tsx
```diff
@@ -1,74 +1,46 @@
...
+    state = newState;
+    listeners.forEach((listener) => listener());
+  };
 
-  const setState = React.useCallback((state: T) => {
-    if (state === stateRef.current || !isMountedRef.current) {
-      return;
-    }
+  const subscribe = (callback: () => void) => {
+    listeners.push(callback);
```
osdnk, line 16: personally, I'd prefer using objects at this will be used quite a lot (a bit better access and removing time)

packages/core/src/useSyncState.tsx
```diff
@@ -1,74 +1,46 @@
...
+  const store = React.useRef(createStore(getInitialState)).current;
 
-  const state = stateRef.current;
+  const state = React.useSyncExternalStore(
+    store.subscribe,
+    store.getState,
+    store.getState
+  );
 
   React.useDebugValue(state);
```
osdnk, line 43: I guess it should be `React.useDebugValue(store.getState());`

packages/core/src/useSyncState.tsx
```diff
@@ -1,74 +1,46 @@
...
-    setTrackingState(stateRef.current);
-  }
+export function useSyncState<T>(getInitialState: () => T) {
+  const store = React.useRef(createStore(getInitialState)).current;
 
-  const state = stateRef.current;
+  const state = React.useSyncExternalStore(
+    store.subscribe,
+    store.getState,
+    store.getState
```
osdnk, line 40: I guess we don't need the third arg if the same as the second

packages/core/src/useSyncState.tsx
osdnk: Actually, it's used directly only once so w/e

osdnk (commented):
That all looks very reasonable and I'm happy we have it a bit simpler. I'm curious about the following from the react documentation:

"It’s not recommended to suspend a render based on a store value returned by useSyncExternalStore. The reason is that mutations to the external store cannot be [marked as non-blocking transition updates](https://react.dev/reference/react/useTransition), so they will trigger the nearest [Suspense fallback](https://react.dev/reference/react/Suspense), replacing already-rendered content on screen with a loading spinner, which typically makes a poor UX."

I think, ideally, we'd prefer to have state mutations non-blocking.

packages/core/src/useSyncState.tsx
satya164: there's a warning during SSR if the third argument is not provided

packages/core/src/useSyncState.tsx
satya164: `state` should give the latest value. the `getState` method is only for stuff that's not called during render, e.g. callbacks, effects etc.

satya164:
> I think, ideally, we'd prefer to have state mutations non-blocking.

Maybe in the future. This keeps the behavior close to how the previous code was. Without synchronous state update, a lot of tests fail and we will likely need a bigger refactor to handle that.

PR #11694, opened 2023-11-10
Add full history type

JustJoostNL:

**Motivation**

I am creating an app where it is important to not ignore duplicate routes (for routes with parameters, etc.). So that's why I thought it would be nice to have a "fullHistory" option.


JustJoostNL:
Closing this since I forgot the conventional commit message.

PR #11695, opened 2023-11-10
feat: add new BackBehavior option, fullHistory

JustJoostNL:

**Motivation**

I am creating an app where it is important to not ignore duplicate routes (for routes with parameters, etc.). So that's why I thought it would be nice to have a "fullHistory" option.


JustJoostNL:
Is someone able to review this?

osdnk:
Hi, thanks for the PR!

The problem is that only one route is tracked for the given index (tab). Therefore, when you push the new route, the old one gets removed (from `routes` array) so it doesn't make much sense to track the old key as that key is no longer corresponding to any tracked route.
I cannot really see how it can work by the code you proposed.  You can try to write some tests, but I'm pretty convinced that cannot work. If you want to achieve that, then probably you need to use a stack or modify the existing tab router.,

JustJoostNL:
Ah I see. Thanks. I'll close this PR for now and have a look at it later.

PR #11702, opened 2023-11-15
Preloading / pt. 1 - routers

osdnk:
# Preloading screens

## Problem

We noticed that it is common to have some heavy screens, which will need to be loaded and some point. The obvious idea is to pre-load the screen so it can live already rendered under the current screen. 

## Action plan

This project is roughly divided into parts. 

First, the logic should include changes inside the routers (included in this PR). Second, those changes should be reflected inside navigators (or they should throw errors if unimplemented, e.g., in the native stack).

### Routers

The navigation state should hold, together with `routes`, also `preloadedRoutes: NavigationRoute<ParamList, keyof ParamList>[]` or `preloadedRoutesKeys: string[];`.

That is the shape of the `PRELOAD` action:

```tsx
{
  type: 'PRELOAD';
  payload: {
    name: string;
    params?: object;
  };
  source?: string;
  target?: string; 
} 
```

That list will indicate to the navigator that those routes should be readily available to show at any point. 

Roughly, routers are split into two categories:

- Tab and Drawer routers
    - If the route is not loaded yet, we should simply add a key to the `preloadedRoutesKeys` array.
    - If the route was already loaded with a matching ID, it should also get added to the `preloadedRoutesKeys` (consequently) and replace the route in `routes`.
    - If the route was already loaded (the route with the name), but with no matching ID, it should also get added to the `preloadedRoutesKeys` and we should change the route under the proper index in `routes`.
    - On navigating to that route, it should get removed from `preloadedRoutesKeys`.
    - If the route is already focused, we warn and do nothing.
    - if the route (with a matching ID) was already preloaded, calling `PRELOAD` again should not update the `preloadedRoutesKeys` and it should get reflected by updating the params in the route.
- Stack router
    - If the route is not loaded yet, we should simply add a new route to the `preloadedRoutes` array.
    - If the route was already loaded (the route with the name), but with no matching ID, it should also get added to the `preloadedRoutes` and not modify `routes`.
    - If the route was already loaded with a matching ID, it should also get added to the `preloadedRoutes` (to make it clear for the navigator to keep it loaded) and update params.
    - On navigating to that route, it should get removed from `preloadedRoutes`.
    - If the route is already focused, we warn and do nothing.
    - if the route (with a matching ID) was already preloaded, calling `PRELOAD` again should update the `preloadedRoutes` (params of the route) and it should get reflected by updating the params in the route if already in `routes`.

On top of that, we should have `DISMISS_PRELOAD` action, which (in both cases) removes the route from the array.

```tsx
{
  type: 'DISMISS_PRELOAD';
  payload: {
    name: string;
    params?: object; // To get ID
  };
  source?: string;
  target?: string;
}
```

This PR includes those changes enriched with tests.


packages/routers/src/TabRouter.tsx
```diff
@@ -41,6 +41,10 @@ export type TabNavigationState<ParamList extends ParamListBase> = Omit<
    * List of previously visited route keys.
    */
   history: { type: 'route'; key: string }[];
+  /**
+   * List of routes' key, which are supposed to be preloaded before navigating to.
+   */
+  preloadedRoutesKeys: string[];
```
osdnk, line 47:
```suggestion
  preloadedRouteKeys: string[];
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+        case 'PRELOAD': {
+          const routeIndex = state.routes.findIndex(
+            (route) => route.name === action.payload.name
+          );
+          if (routeIndex === state.index) {
+            console.warn('Cannot preload the focused page.');
+            return state;
+          }
+          if (routeIndex === -1) {
+            return state;
```
osdnk, line 413:
```suggestion
            return null;
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
           };
         }
 
+        case 'PRELOAD': {
+          const routeIndex = state.routes.findIndex(
+            (route) => route.name === action.payload.name
+          );
+          if (routeIndex === state.index) {
+            console.warn('Cannot preload the focused page.');
+            return state;
```
osdnk, line 410:
```suggestion
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+
+          let params;
+
+          if (currentId === nextId) {
+            params =
+              action.payload.params !== undefined ||
+              routeParamList[route.name] !== undefined
+                ? {
+                    ...routeParamList[route.name],
+                    ...route.params,
```
osdnk, line 433:
```suggestion
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+                    ...action.payload.params,
+                  }
+                : route.params;
```
osdnk, line 436:
```suggestion
                : undefined;
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+            params !== route.params ? { ...route, key, params } : route;
+
+          return {
+            ...state,
+            preloadedRoutesKeys: state.preloadedRoutesKeys
+              .filter((key) => key !== route.key)
+              .concat(newRoute.key),
+            routes: state.routes.map((route, index) =>
+              index === routeIndex ? newRoute : route
+            ),
```
osdnk, line 457: also, remove from the history

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+          };
+        }
+
+        case 'DISMISS_PRELOAD': {
```
osdnk, line 461:
```suggestion
        case 'REMOVE_PRELOAD': {
```

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,88 @@ export function TabRouter({
...
+          const routeIndex = state.routes.findIndex(
+            (route) => route.name === action.payload.name
+          );
+          if (routeIndex === -1) {
+            return state;
```
osdnk, line 466:
```suggestion
            return null;
```

packages/routers/src/StackRouter.tsx
```diff
@@ -209,6 +215,7 @@ export function StackRouter(options: StackRouterOptions) {
         index: routes.length - 1,
         routeNames,
         routes,
+        preloadedRoutes: [], // ?
```
osdnk, line 218: it is actually possible to have some things before rehydration. Screen names should be valid and keys matching (if the are already in the stack)

packages/routers/src/StackRouter.tsx
```diff
@@ -379,11 +394,19 @@ export function StackRouter(options: StackRouterOptions) {
...
             routes,
           };
         }
 
         case 'NAVIGATE_DEPRECATED': {
+          if (state.preloadedRoutes.length !== 0) {
+            console.warn(
+              'Preloading is not supported with navigateDeprecated.'
+            );
+          }
```
osdnk, line 409: return null

osdnk: make the check fine-grained

packages/routers/src/StackRouter.tsx
```diff
@@ -482,10 +505,17 @@ export function StackRouter(options: StackRouterOptions) {
...
+
             return {
               ...state,
               index: routes.length - 1,
               routes,
+              preloadedRoutes: state.preloadedRoutes.filter(
+                (route) =>
+                  removedRoutes.findIndex((r) => route.key === r.key) === -1 &&
+                  route.key !== routes[routes.length - 1].key
+              ),
```
osdnk, line 518: remove

packages/routers/src/StackRouter.tsx
```diff
@@ -606,6 +636,77 @@ export function StackRouter(options: StackRouterOptions) {
...
+              ...state,
+              preloadedRoutes: state.preloadedRoutes
+                .filter((r) => {
+                  if (r.name === action.payload.name) {
+                    if (id === undefined) {
+                      return false;
+                    }
+                    return id !== getId?.({ params: r.params });
+                  }
+                  return true;
```
osdnk, line 664: make sure that this is correct

packages/routers/src/StackRouter.tsx
```diff
@@ -606,6 +636,77 @@ export function StackRouter(options: StackRouterOptions) {
...
+            return {
+              ...state,
+              preloadedRoutes: state.preloadedRoutes
+                .filter((r) => r.key === route?.key)
+                .concat(
+                  action.payload.params
+                    ? {
+                        ...route,
+                        params: {
+                          ...route.params,
```
osdnk, line 688: not merge

packages/core/src/types.tsx
```diff
@@ -281,6 +281,76 @@ type NavigationHelpersCommon<
       : never
   ): void;
 
+  /**
+   * Preloads the route in current navigation tree.
+   *
+   * @param route Object with `name` for the route to navigate to, and a `params` object.
+   */
+  preload<RouteName extends keyof ParamList>(
```
osdnk, line 289: just name and params

packages/drawer/src/views/DrawerView.tsx
```diff
@@ -56,6 +56,11 @@ function DrawerViewBase({
 }: Props) {
   const { direction } = useLocale();
 
+  if (state.preloadedRoutesKeys.length !== 0) {
+    console.warn(
```
osdnk, line 60: throw

packages/routers/src/CommonActions.tsx
```diff
@@ -107,3 +125,38 @@ export function reset(state: ResetState | undefined): Action {
...
+      payload: { name: args[0], params: args[1] },
+    };
+  } else {
+    const payload = args[0] || {};
+
+    if (!('name' in payload)) {
+      throw new Error(
+        'You need to specify a name when calling navigate with an object as the argument. See https://reactnavigation.org/docs/navigation-actions#navigate for usage.'
+      );
+    }
```
osdnk, line 158: remove the `else`. Similar to `push`

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,74 @@ export function TabRouter({
...
+          return {
+            ...state,
+            preloadedRouteKeys: state.preloadedRouteKeys
+              .filter((key) => key !== route.key)
+              .concat(newRoute.key),
+            routes: state.routes.map((route, index) =>
+              index === routeIndex ? newRoute : route
+            ),
+            history: state.history.filter(
+              (record) => key === route.key || record.key !== route.key
```
osdnk, line 442: add a comment on what happens here

packages/routers/src/TabRouter.tsx
```diff
@@ -380,11 +393,74 @@ export function TabRouter({
...
+
+          return {
+            ...state,
+            preloadedRouteKeys: state.preloadedRouteKeys
+              .filter((key) => key !== route.key)
+              .concat(newRoute.key),
+            routes: state.routes.map((route, index) =>
+              index === routeIndex ? newRoute : route
+            ),
+            history: state.history.filter(
```
osdnk, line 441: make it a ternary

PR #11713, opened 2023-11-21
fix: change of width of tabBarIndicator due to change of leftPadding

g4ps:
**Motivation**

The issue manifests itself only on the android phones (as far as I know).

When we try to add paddingLeft and width: 'auto' to contentConteinerStyle in TabBar, we get a
certain problem: our TabBarIndicator's width goes completely out of whack. This problem is due to
a more general problem in react-native, which can't properly process large values of scaleX
transforms when position of the block is either 'relative' or 'absolute' and 'left' is defined
on android. There are two main ways to solve this problem: the first one is to wail
'till the folks in react-native community solve the underlying issue, and the second
(and way more attractive in my personal opinion) solution is to make width of the tabBarElement
on android larger so that we don't have to deal with scaleX's in triple digits. This results
with a tad bit more complicated maths in TabBarIndicator, particularly when it comes to calculating
translations after the scaling, but notihng too complicated.

**Test plan**

- Get yourself an android phone
- go to https://github.com/g4ps/testApp, download and install (yarn  && yarn run start) the app
- change the values of paddingLeft in contentContainerStyle in tabBarStyles (values between 20 and
28 should do)
- you should see something like this, where the width of the TabBarIndicator is gonna be completely
out of whack:

https://github.com/react-navigation/react-navigation/assets/70518602/a7c2954f-0ba1-42f8-90ca-c125753a0abf

- change "react-native-tab-view": "^4.0.0-alpha.2" to "react-native-tab-view": "g4ps/react-navigation.git#react-native-tab-view-v4.0.0-alpha.2-gitpkg" (which was build with this PR) in package.json
- reinstall the app
- change the values of paddingLeft to whatever value your heart desires, and everything's
going to be completely within the acceptable levels of whack


satya164 (changes requested):
Thank you for the PR. Seems the indicator is misaligned after this change, here are some screenshots from the tab view examples in this repo

**Before**

before

before-scroll

**After**

after


after-scroll

Can you check?

PR #11740, opened 2023-12-10
fix: Registered listener not found when mounting and unmounting

g1d5ny:

**Motivation**

On iOS, when you try to scroll back a stack to get back to the previous page, it can't find a listener to unmount the currently mounted stack.

1. I added code to register a listener for being mounted at the bottom line of componentDidMount().

2. I added code to delete the mounted listener at the bottom of componentWillUnmount().

**Test plan**


g1d5ny:

osdnk:
@g1d5ny Thanks for the PR.
Please elaborate on the issue you solve. This seems to be a bit cryptic in the current shape

khrulev:
> @g1d5ny Thanks for the PR. Please elaborate on the issue you solve. This seems to be a bit cryptic in the current shape

The warning occurs when you use gestures for moving between screens (in the navigation stack) on iOS. The ** removeAllListeners** function solves this problem, but to ensure it does not cause any side problems, we added 1 more empty listener in **componentDidMount()** function. And then it works fine.

g1d5ny:
> @g1d5ny홍보해주셔서 감사합니다. 해결한 문제에 대해 자세히 설명해주세요. 현재 모습으로는 좀 애매한 것 같습니다

hello. I will explain in detail the problem I fixed.

The problem I described is that when you try to go back to the previous page by scrolling the stack in iOS, you cannot find a listener to unmount the currently mounted stack. So the following warning appears.

"Animation Warning - Sending onAnimatedValueUpdate with no listeners registered"

So, we added a listener so that it can be recognized every time it mounts or unmounts. Then the warning disappears.

I am uploading a video comparing the before and after videos.


https://github.com/react-navigation/react-navigation/assets/55727475/d380fa19-255a-4463-8d40-688d4ef56804


https://github.com/react-navigation/react-navigation/assets/55727475/03d77f3e-1263-4129-860f-700985c7353b

g1d5ny:
> @g1d5ny Thanks for the PR. Please elaborate on the issue you solve. This seems to be a bit cryptic in the current shape


hello. I will explain in detail the problem I fixed.

The problem I will explain is that when you try to go back to the previous page by scrolling the stack in iOS, you cannot find a listener to unmount the currently mounted stack. Currently, the following warning appears.

"Sending onAnimatedValueUpdate with no listeners registered"

So, we added a listener so that it can be recognized every time it mounts or unmounts. Then the warning disappears.

I am uploading a video comparing the before and after.

https://github.com/react-navigation/react-navigation/assets/55727475/d380fa19-255a-4463-8d40-688d4ef56804
https://github.com/react-navigation/react-navigation/assets/55727475/03d77f3e-1263-4129-860f-700985c7353b

satya164:
Thanks for the pull request. However it's a workaround and doesn't address the actual issue.

The problem here is that it sends values when there are no listeners. Instead of adding listeners that don't do anything, the fix would be to identify why it is happening - which may either be a bug in React Navigation, Gesture Handler, or React Native itself, and then address it.

satya164:
Closing as per my response above.

PR #11741, opened 2023-12-11
feat: add layout and screenLayout props for screens

satya164:
**Motivation**

This adds a new `layout` prop that can be used to wrap screens.

It makes it easier to provide things such as a global error boundary and suspense fallback for a group of screens without having to manually add HOCs for every screen separately.

See https://github.com/react-navigation/react-navigation/discussions/11152

**Test plan**

This adds unit tests for testing, as well as a demo screen in the example app showing error boundary and suspense using the `layout` prop.

https://github.com/react-navigation/react-navigation/assets/1174278/545f7bef-e087-45e9-800a-ed9fbded5166

### DIFF
--- a/example/src/Screens/Layouts.tsx
+++ b/example/src/Screens/Layouts.tsx
@@ -1,0 +1,150 @@
+import { Button } from '@react-navigation/elements';
+import type { ParamListBase } from '@react-navigation/native';
+import {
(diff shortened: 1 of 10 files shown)


osdnk (approved):
looks reasonable

Issue #7963, opened 2020-04-06
Event order: focus from Screen2 execute before blur from Screen1


allandiego:
**Current Behavior**
Two <Drawer.Screen> that add a ScreenOrientation listener on focus event and remove it on blur. The problem is when Screen1 execute the blur event to remove the orientation listener its after Screen2 addListerner on focus, so is removes Screen2 listener too.

Time line for better understanding:
- Screen1 - focus > register orientation listener
- user navigate to Screen2 -  focus > register orientation listener
- Screen1 execute blur event > remove orientation Listeners including the newly registered by focus from Screen2

Screen1
```js
 useEffect(() => {
    navigation.addListener('focus', () => {
      function handleOrientationChange(orientation) {
        const isLandscape = orientation.orientationInfo.orientation > 2;
        setIsLandscapeLayout(isLandscape);
      }

      ScreenOrientation.addOrientationChangeListener(handleOrientationChange);
    });

     navigation.addListener('blur', () => {
       ScreenOrientation.removeOrientationChangeListeners();
       navigationListenerBlur();
       navigationListenerFocus();
     });
  }, [navigation]);
```

Screen2 have the same code with different handleOrientationChange function body

**Expected Behavior**

- Screen1 focus > register orientation listener
- user navigate to Screen2
- Screen1 blur > remove orientation listener
- Screen2 focus > register orientation listener

I have achieved that behavior using the useEffect return (component unmount) to remove the listener and adding the unmountOnBlur option, but that way i lose the component state and i need preserve it, 
Is there any way to execure blur before next screen focus?

```js
<Drawer.Screen options={{ unmountOnBlur: true }} />
```

```js
 useEffect(() => {
      function handleOrientationChange(orientation) {
        const isLandscape = orientation.orientationInfo.orientation > 2;
        setIsLandscapeLayout(isLandscape);
      }
      ScreenOrientation.addOrientationChangeListener(handleOrientationChange);

    return () => ScreenOrientation.removeOrientationChangeListeners();
  }, [navigation]);
```

i have tried useIsFocused too and its the same behavior as the focus / blur event its set to false after new screen focus is set to true

BooDzioo:
Did you managed to resolve this problem?

allandiego:
yes, they fixed the blur and focus order, now blur from screen1 execute before focus from screen2

SeveHo:
@allandiego how did you fix this?

allandiego:
@SeveHo it was fixed:
[...]

PR #11742, opened 2023-12-11
Fix focus and blur callbacks' order when using useFocusEffect() hook

KroosX4V:
**Motivation**

I noticed when using useFocusEffect() hook, the focus callback of a just mounted component runs before blur/cleanup callback of a previously or already mounted component. This caused issues with screens requiring transactional behavior.

For example, imagine you have three screens. Two screens are in full screen mode, one is not. The two screens in full screen mode use useFocusEffect() hook - entering full screen mode on focus and exiting it on blur. Now, say I'm currently in one of the two screens requiring full screen mode, and I navigate to the other screen in full screen mode, then, this happens: focus callback (second screen), blur callback (first screen). By the time I'm in the second screen, I won't be in full screen mode due to the error in the order of execution of focus/blur callbacks.

This issue is similar to #7963 which has been resolved, but it fixed only the execution order of navigation focus/blur events (i.e., `navigation.addListener`), not useFocusEffect() hook.

**Test plan**

Use useFocusEffect() hook in two screens and observe the order of execution of the focus/blur callbacks.

### DIFF
--- a/packages/core/src/useFocusEffect.tsx
+++ b/packages/core/src/useFocusEffect.tsx
@@ -72,10 +72,15 @@ export function useFocusEffect(effect: EffectCallback) {
     };
 
     // We need to run the effect on intial render/dep changes if the screen is focused
-    if (navigation.isFocused()) {
-      cleanup = callback();
-      isFocused = true;
-    }
+    let immediateId: NodeJS.Immediate | undefined = setImmediate(() => {
+      immediateId = undefined;
+
+      if (!isFocused && navigation.isFocused())
+      {
+        cleanup = callback();
+        isFocused = true;
+      }
+    });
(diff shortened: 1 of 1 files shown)


osdnk (changes requested):
Hi!

Thanks for the PR.
Unfortunately, we cannot accept the PR in this state.

First, CI is not passing.
Then, the solution seems to only delay the execution of the callback, but not the logic causing the effects to execute in the wrong order.
Eventually, those callbacks were not designed to run in the way you described. Of course, it would be preferable, but they were not tested for this behaviour, so it's not surprising that they are working this way. 

I suggest trying to design a less superficial solution as this is clearly not addressing the core of the problem.

PR #11744, opened 2023-12-12
chore: update transform style for web

EvanBacon:
**Motivation**

This API is "deprecated" according to `react-native-web` so we get the following warning whenever anyone attempts to render React Navigation: `Warning: "transform" style array value is deprecated. Use space-separated string functions, e.g., "scaleX(2) rotateX(15deg)".`

**Test plan**

### DIFF
--- a/packages/elements/src/Header/HeaderBackButton.tsx
+++ b/packages/elements/src/Header/HeaderBackButton.tsx
@@ -251,6 +251,6 @@ const styles = StyleSheet.create({
     alignSelf: 'center',
   },
   flip: {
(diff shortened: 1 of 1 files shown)


osdnk (commented):
Hi, thanks. I think you need to have it platform-specific, don't you?

satya164:
I guess it's supported in React Native now https://reactnative.dev/docs/transforms

Though it's unfortunate that we're forced to use non-typesafe string format instead of a typesafe format.

satya164 (approved):

PR #11746, opened 2023-12-13
`tabBarBadge` also accepts a bool (or more specifically true)

Workshed:
**Motivation**

The docs state that `tabBarBadge` accepts a boolean: https://reactnavigation.org/docs/material-bottom-tab-navigator/#tabbarbadge

If you want to display badge without a number in, this is how it's done. However the current code does not allow this and so there's currently not a supported method of displaying a badge with no value. 

**Test plan**

We've added a badge with no value to the `BottomTabs.tsx` in the example project.
image

### DIFF
--- a/example/src/Screens/BottomTabs.tsx
+++ b/example/src/Screens/BottomTabs.tsx
@@ -138,6 +138,7 @@ export function BottomTabs({
           options={{
             title: 'Article',
             tabBarIcon: getTabBarIcon('file-document'),
+            tabBarBadge: true,
           }}
         />
         <Tab.Screen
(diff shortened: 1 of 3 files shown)


satya164:
> The docs state that tabBarBadge accepts a boolean

They don't. That's the documentation for material bottom tabs. You're changing bottom tabs.

This is not currently supported and as you can see in the screenshot, it just looks the same size without content - which is not how it should look.

If you want to create a feature request, use https://react-navigation.canny.io/feature-requests

Workshed:
Thanks @satya164, I've created a feature request.

PR #11758, opened 2023-12-17
chore: move removePreload to the stack router and enable preloading in the material-top-tabs

osdnk:
Please provide enough information so that others can review your pull request.

**Motivation**

Explain the **motivation** for making this change. What existing problem does the pull request solve?

If this pull request addresses an existing issue, link to the issue. If an issue is not present, describe the issue here.

**Test plan**

Describe the **steps to test this change** so that a reviewer can verify it. Provide screenshots or videos if the change affects UI.

The change must pass lint, typescript and tests.

PR #11765, opened 2023-12-22
feat: add retaining of the screen to the stack navigator

osdnk:
The changes are pretty straightforward and should be expressive enough to understand by reading the codebase. This PR:
- Adds the new retaining feature including changes in the router and minor changes in the card stack (as temporarily scene can include both the preloaded route and the one getting detached ) 
- Moves around some leftovers from the previous PR (#11758 )
- Adds tests 

To test, try the example.

### DIFF
--- a/packages/routers/src/CommonActions.tsx
+++ b/packages/routers/src/CommonActions.tsx
@@ -54,15 +54,6 @@ export type Action =
       };
       source?: string;
       target?: string;
-    }
-  | {
-      type: 'REMOVE_PRELOAD';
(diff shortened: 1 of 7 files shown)


packages/routers/src/CommonActions.tsx
```diff
@@ -54,15 +54,6 @@ export type Action =
       };
       source?: string;
       target?: string;
-    }
-  | {
```
osdnk, line 58: lefover. Should be in the stack actions

PR #11770, opened 2024-01-05
fix: run focus effects after navigation transition

### LINKED ISSUE #11613: [bottom-tabs] Navigation delay with useFocusEffect, especially on low-end devices
### Current behavior

I have an app in production that has a bottom tab navigation and screens with `useFocusEffect` that refresh the data in a list. Recently some users, particularly with low-end devices, reported attempt to navigate between bottom tabs, upon selecting a screen, they expect to transition smoothly to the new tab. However, the current behavior is:

- The user initiates navigation from Screen 1 to Screen 2 by tapping the tab.
- Instead of a swift transition, the user remains on Screen 1 until the expensive operation within the `useFocusEffect` on Screen 2 completes.
- Only after the operation finishes, the user is finally able to access Screen 2.

Example of the problem:

https://github.com/react-navigation/react-navigation/assets/51209886/66c84923-7d9d-4443-a54b-75250c1f1f6e

I have already tried some approaches to mitigate this issue, including:

- Adjusting the `lazy`, `unmountOnBlur`, and `freezeOnBlur` properties of the navigation components.
- Implementing the `useEffect` hook with different configurations.
- Using `InteractionManager.runAfterInteractions` to try not to interfere with the transition between screens.

Please let me know if you need any further information or if there are specific actions or configurations I should try.

### Expected behavior

I think the expected behavior in this scenario would be to respond immediately to the user's interaction, taking them to the target tab and then triggering `useFocusEffect` to update the data.

I assumed that the problem is in React Navigation because it happens during tab navigation and with the use of `useFocusEffect`, but I could be wrong.

### Reproduction

https://snack.expo.dev/@rnqs/bottom-tabs-navigator-%7C-react-navigation

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| @react-navigation/native               | 6.1.1
| @react-navigation/bottom-tabs          | 6.5.2
| @react-navigation/drawer               | 6.5.7
| @react-navigation/native-stack         | 6.9.7
| react-native-reanimated                | 3.3.0
| expo                                   | 49.0.0
| node                                   | 18
| yarn                            | 1.22.19

oblador:
**Motivation**

I've been investigating why our app had turned slow and less responsive esp on older Android devices. Seemingly a navigation between two already mounted tabs should be near instant. I found that it only happened where we were doing side effects in `useFocusEffect` and I attribute it to both the introduction of automatic batching in React 18 and that the effects themselves might be taxing on the CPU and/or bridge. 

In this PR I change the behaviour of the hook when being _reactivated_ (mount behaviour remains the same) to run just after the render update being committed by wrapping it inside a `setTimeout`. I had originally used `requestIdleCallback` which seemed more correct, however it seems that function has had issues historically and might break on iOS for users on older versions of React Native, see https://github.com/facebook/react-native/issues/28602. Using `setImmediate` does not avoid the batching behaviour as IIRC the implementation on RN side will then not involve a native timer. 

Fixes https://github.com/react-navigation/react-navigation/issues/11613

**Test plan**

Verified on an enterprise app and updated unit tests to reflect the change.

### DIFF
--- a/packages/core/src/__tests__/useFocusEffect.test.tsx
+++ b/packages/core/src/__tests__/useFocusEffect.test.tsx
@@ -9,6 +9,7 @@ import { useNavigationBuilder } from '../useNavigationBuilder';
 import { MockRouter, MockRouterKey } from './__fixtures__/MockRouter';
 
 beforeEach(() => {
+  jest.useFakeTimers();
   MockRouterKey.current = 0;
 });
 
@@ -48,21 +49,25 @@ test('runs focus effect on focus change', () => {
 
   render(element);
 
+  jest.runAllTimers();
   expect(focusEffect).not.toHaveBeenCalled();
   expect(focusEffectCleanup).not.toHaveBeenCalled();
 
   act(() => navigation.current.navigate('second'));
 
+  jest.runAllTimers();
   expect(focusEffect).toHaveBeenCalledTimes(1);
   expect(focusEffectCleanup).not.toHaveBeenCalled();
 
   act(() => navigation.current.navigate('third'));
 
+  jest.runAllTimers();
   expect(focusEffect).toHaveBeenCalledTimes(1);
   expect(focusEffectCleanup).toHaveBeenCalledTimes(1);
 
   act(() => navigation.current.navigate('second'));
 
+  jest.runAllTimers();
   expect(focusEffect).toHaveBeenCalledTimes(2);
   expect(focusEffectCleanup).toHaveBeenCalledTimes(1);
 });
@@ -157,6 +162,7 @@ test('runs focus effect when initial state is given', () => {
 
   act(() => navigation.current.navigate('first'));
 
+  jest.runAllTimers();
   expect(focusEffect).toHaveBeenCalledTimes(1);
   expect(focusEffectCleanup).toHaveBeenCalledTimes(1);
 });
@@ -200,6 +206,7 @@ test('runs focus effect when only focused route is rendered', () => {
 
   act(() => navigation.current.navigate('second'));
 
+  jest.runAllTimers();
   expect(focusEffect).toHaveBeenCalledTimes(1);
   expect(focusEffectCleanup).toHaveBeenCalledTimes(1);
 });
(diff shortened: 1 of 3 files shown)


smartin-cian:
I was changing a state in a custom hook, however I had that deley when changing the state, I solved it by adding a setTimeout to change the state, Example: 

```
const [isActive, setIsActive] = useState(true);

const subscribe = () => {
  setTimeout(() => {
    setIsActive(true);
  }, 300);
};

const unsubscribe = () => {
  setTimeout(() => {
    setIsActive(false);
  }, 300);
};

useFocusEffect(
  useCallback(() => {
    subscribe();

    return () => unsubscribe();
  }, [])
);
```

santosdmg:
I was changing a state in a custom hook, however I had that deley when changing the state, I solved it by adding a setTimeout to change the state, Example: 

```
const [isActive, setIsActive] = useState(true);

const subscribe = () => {
  setTimeout(() => {
    setIsActive(true);
  }, 300);
};

const unsubscribe = () => {
  setTimeout(() => {
    setIsActive(false);
  }, 300);
};

useFocusEffect(
  useCallback(() => {
    subscribe();

    return () => unsubscribe();
  }, [])
);
```

sirusbaladi:
any update on this? this is needed.

debitan:
Came across this today. Worked around it by using setTimeout(() => ..., 0). Even zero timeout worked.

oblador:
@satya164 Mind having a look at this? Should be a small and safe change.

satya164:
Thanks for the PR. But this change feels risky to me as it will change the order of calls if you have multiple listeners in addition to the `useFocusEffect`. So it may introduce race conditions and break any existing code that relied on the synchronous behavior.

Requiring to run `jest.runAllTimers()` in tests for side effects is also breaking change, and not good for the ergonomics.

Normally, side effects don't update state - and setting state in `useEffect` is also generally considered an antipattern. So I'm not sure about adding timeout that could cause other issues to solve this use case. Imo it'd be more predictable to do it where you actually need it in your code.

PR #11771, opened 2024-01-05
feat: add option to set tab bar width

robertli93:

**Motivation**

Pull Request #11578 https://github.com/react-navigation/react-navigation/pull/11578 added an option to show the tab bar on the left or the right.

This PR adds ability to set the tab bar width. It useful on larger screens where tab bar is set to the side and allows customisation of tab bar width according to the ux ui need

**Test plan**

Open the "Bottom Tabs" example and resize the screen to show to tab bar on the left. 
Press the FAB on the bottom right to open the setting modal
Check the checkbox and select different tab bar width to show the effect

**Video**

https://github.com/react-navigation/react-navigation/assets/146067776/1dc09cde-0ea5-41aa-b4ed-04c189a42e00

### DIFF
--- a/example/src/Screens/BottomTabs.tsx
+++ b/example/src/Screens/BottomTabs.tsx
@@ -20,9 +20,18 @@ import {
   ScrollView,
   StatusBar,
   StyleSheet,
+  Text,
   useWindowDimensions,
+  View,
 } from 'react-native';
-import { Appbar, IconButton } from 'react-native-paper';
+import {
+  Appbar,
+  Button,
+  Checkbox,
+  FAB,
+  IconButton,
+  RadioButton,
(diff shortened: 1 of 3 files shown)


osdnk (changes requested):
Hi, @robertli93 
Thanks for the PR!

That looks great.

I believe I have a single minor comment. Can you please make it just a value, not a function? The argument of the function is something trivially obtained by the hook `useWindowDimensions()`. Therefore, a value should be enough and users can access the dimensions themselves.

osdnk:
Also, maybe it should warn if we are not using the tab bar in the "`sideBar` mode"?

osdnk:
Also, why can't you just set the width via `tabBarStyle`?

PR #11782, opened 2024-01-13
feat: [bottom-tabs] allow position = top

douglowder:

**Motivation**

Currently in `bottom-tabs`, the `tabBarPosition` prop can only have `bottom`, `left`, or `right` values.

For Apple TV and Android TV, it is helpful to allow `top` as well, so that there can be a top nav bar without absolute positioning (which causes problems for TV focus navigation).

**Test plan**

Test app that works correctly with this change: https://github.com/douglowder/IgniteTV/tree/top-navigation


satya164 (approved):
Thanks!

Issue #11655, opened 2023-10-24
useHeaderHeight() returns incorrect height for ios


itsramiel:
### Current behavior

I am trying to create a full screen vertical scroll window so I need to get the height of the window, using useWindowDimensions, minus the header height, using useHeaderHeight. Here is a simple reproducable:
```
// In App.js in a new project
import * as React from "react";
import { View, useWindowDimensions } from "react-native";
import { useHeaderHeight } from "@react-navigation/elements";
import { NavigationContainer, DefaultTheme } from "@react-navigation/native";
import { createNativeStackNavigator } from "@react-navigation/native-stack";

function HomeScreen() {
  const { height: windowHeight } = useWindowDimensions();
  const headerHeight = useHeaderHeight();
  return (
    <View
      style={{
        height: windowHeight - headerHeight,
        backgroundColor: "black",
      }}
    />
  );
}

const Stack = createNativeStackNavigator();

function App() {
  return (
    <NavigationContainer
      theme={{
        ...DefaultTheme,
        colors: { ...DefaultTheme.colors, background: "white" },
      }}
    >
      <Stack.Navigator>
        <Stack.Screen name="Home" component={HomeScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

export default App;
```

But this will leave a very small ~1px at the bottom:
<img width="1376" alt="Screenshot 2023-10-24 at 8 56 40 AM" src="https://github.com/react-navigation/react-navigation/assets/80689446/3595ace4-19dd-4efd-b99f-e35780d86aa9">

The issue is here is that when making a vertically scrolling list that takes the whole screen then each slide will be 1 pixel short and eventually the gap will increase with every slide. Also UI look not good when you have contrasting colors as you see the white line at the bottom

### Expected behavior

I expect `useHeaderHeight` to return the correct value

### Reproduction

https://github.com/itsramiel/HeaderHeightIos

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |  6.1.9
[...]

PR #11786, opened 2024-01-16
fix: status bar height for iOS devices with dynamic island

gianlucalippolis:
This PR addresses the issue https://github.com/react-navigation/react-navigation/issues/11655

### DIFF
--- a/packages/elements/src/Header/getDefaultHeaderHeight.tsx
+++ b/packages/elements/src/Header/getDefaultHeaderHeight.tsx
@@ -1,13 +1,19 @@
-import { Platform } from 'react-native';
+import { PixelRatio, Platform } from 'react-native';
 
 import type { Layout } from '../types';
 
 export function getDefaultHeaderHeight(
   layout: Layout,
   modalPresentation: boolean,
-  statusBarHeight: number
+  topInset: number
 ): number {
   let headerHeight;
+
+  // On models with Dynamic Island the status bar height is smaller than the safe area top inset.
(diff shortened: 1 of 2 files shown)


outaTiME:
Hi @gianlucalippolis, 
be careful, probably the value that should be used is 5.33 instead of 6 for a perfect fit.

https://github.com/react-navigation/react-navigation/pull/11338#pullrequestreview-1725831043

gianlucalippolis:
hi @outaTiME, you are right, by reducing that value from 6 to 5.33 the bug is fixed anyway. I tried reducing it further to 5.2, is there any particular reason why you suggest the value 5.33?

outaTiME:
Here is the original hot fix and report:

https://github.com/react-navigation/react-navigation/issues/10989#issuecomment-1396251186

Now I don't remember well, but I understand that the `UINavigationBar` returns `53.67` as a fixed value for devices with dynamic island and the `useHeaderHeight` returns `59`, from the difference of both values comes the number `5.33`

satya164:
Where does the `0.17` come from? As @outaTiME mentioned the `0.33` makes sense as it's like the height of the border that needs to be removed, which is `1 / PixelRatio.get()`.

gianlucalippolis:
I set that value because it covered the screen perfectly. I will still make the change by setting 0.33 as you suggest.  I will also try moving that logic to getDefaultHeaderHeight.

Thanks

gianlucalippolis:
@satya164 @outaTiME i updated the PR.

satya164 (approved):
Thanks!

Issue #6851, opened 2020-02-14
route.params is undefined, nicer if it was an empty Object {} by default


Jarred-Sumner:
React Navigation v5's API is so much better! This is a tiny thing.

**Current Behavior**
When there are no `params` passed to `navigator.navigate`, `route.params` is undefined.  Additionally, the type definition doesn't indicate that `params` can be `undefined`.

This means that the following code will raise an exception because `params` is not an `Object`:
```javascript
const {showStart, onChange} = useRoute().params
```
![image](https://user-images.githubusercontent.com/709451/74502593-d1d09e00-4ea2-11ea-88a5-cb1daf889c14.png)

This makes sense. There are no params, so `params` is never set.

**Expected Behavior**

From a developer experience perspective though, It'd be a little easier if it was an empty `Object` when undefined. That way, I can use destructuring without optional chaining.

**How to reproduce**

Navigate to a route but don't pass any params to it.

```javascript
const {showStart, onChange} = useRoute().params
```

**Your Environment**
```
├─ @react-navigation/core@5.1.2
│  ├─ @react-navigation/routers@^5.0.1
├─ @react-navigation/native-stack@5.0.3
├─ @react-navigation/native@5.0.3
│  └─ @react-navigation/core@^5.1.2
├─ @react-navigation/routers@5.0.1
├─ @react-navigation/stack@5.0.3
├─ react-navigation-animated-switch@0.4.1
├─ react-navigation-shared-element@0.5.0
```

--

In terms of how to implement this...maybe it could be an `Object.defineProperty` getter on the `route` object returned by `useRoute`?

Grohden:
I'm curious:

> Additionally, the type definition doesn't indicate that params can be undefined.

You mean that every TS dev should deal with `undefined` even when they know that it will never be undefined?

I would rather use this from [the guides](https://reactnavigation.org/docs/typescript 
):
```
type RootStackParamList = {
  Home: undefined;
  Profile: { userId: string };
  Feed: { sort: 'latest' | 'top' } | undefined;
};
```
And annotate the provided hooks properly (https://reactnavigation.org/docs/typescript#annotating-useroute)


If you type things correctly, you **probably** will not run into these kind of issues, alternatively you could use:
```
const { showStart, onChange } = useRoute().params ?? {}
```

andreasmischke:
> You mean that every TS dev should deal with `undefined` even when they know that it will never be undefined?

That's the current state and should *not* be the case.

[...]

PR #11795, opened 2024-01-23
fix: make params property optional in the return type of useRoute

intomyam:
**Motivation**

The return type of useRoute includes `params` as non-optional, but in fact `params` is optional. Therefore, the return type has been modified to match.

cf. https://github.com/react-navigation/react-navigation/issues/6851

**Test plan**


satya164:
The type of params depends on how you define it in the param list. If it's defined as `SomeType | undefined` then it'll be the same here. It's incorrect to make the params optional when `| undefined` isn't specified.

Issue #11797, opened 2024-01-23
White flash transitions with DarkTheme


ArthurLaheyne:
### Current behavior

Using the DarkTheme, my app have white flashes on the edges when I navigate between screens.

### Expected behavior

No white flashes when navigating between screens

I found a [workaround](https://stackoverflow.com/questions/59900898/white-background-flashing-when-switching-screens-react-navigation-v5) on stackoverflow for the moment : 

### Reproduction

https://github.com/ArthurLaheyne/react-navigation-white-flash

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |
| @react-navigation/native-stack         |
| react-native                           |
| expo                                   |
| node                                   |
| npm or yarn                            |

ArthurLaheyne:
I have updated the issue to include the version numbers of the packages.

ArthurLaheyne:
Thank you @gianlucalippolis, it works very well !

I've tried it by copying your changes directly into my node_modules folder. Is there another way to proceed ? I guess I have to wait for the PR to get merged and then update react-navigation version ?

gianlucalippolis:
@ArthurLaheyne you can wait for the release of the new version if the PR will be merged, or you can use patch-package to remedy the problem temporarily, before the update.

satya164:
Closing as this should be fixed in [1770c96](https://github.com/react-navigation/react-navigation/commit/1770c96834abf8af28aa0f3c6bffb618ea3dcce9). Please upgrade to `@react-navigation/native-stack@7.0.0` and give it a try.

PR #11800, opened 2024-01-24
fix: White flash transitions with DarkTheme

gianlucalippolis:
This PR fixes the issue #11797

Even with the DarkTheme during the transition ScreenStack comes up with a white background. I solved it by setting the backgroundColor as the theme color


osdnk:
Hi, @gianlucalippolis 
Thanks for the PR!

Can you apply the same change in the web version? https://github.com/react-navigation/react-navigation/blob/cee6eb49f12ccd12b5720243b0b375ce1e7875c3/packages/native-stack/src/views/NativeStackView.tsx

PR #11803, opened 2024-01-31
feat: introduce missing props from native-stack v5

tboba:
**Motivation**

For some time now, people are complaining that current native stack does not have some props that are included in v5, but not in v6.
I've deeply investigated the `props.tsx` file and I can make a conclusion some of the things that are missing include:
- no `hideKeyboardOnSwipe` `gestureResponseDistance` property,
- no props for managing the form sheet - `sheetX` properties,
- missing support for `onWillAppear` (is it intentional?) and `onGestureCancel` events.

This PR fixes all those things by adding them to `props.tsx` and their support in `NativeStackView.native.tsx` file.
One thing I also wanted to mention is that some of the events from the event map in `types.tsx` like `dismiss`, `appear` or `headerHeightChange` were not included in changes. That's because I think these ones should not be emitted directly from `navigation#emit` and/or they might lose the support in future react-native-screens versions. For example, `appear` event (which is currently deprecated) should not be used and you should replace it with `transitionEnd` event instead.

**Test plan**

You can test these props by checking how they affect screens after adjusting them in Native Stack example. Remember about setting `presentation` to `formSheet` while using `sheetX` props.

Some of the examples I've already prepared:

<details>
<summary>Using GestureResponseDistance on NewsFeed</summary>

This will enable changing the area of the gesture with a rectangle starting from the ~middle of the screen down to bottom.

`NativeStack.tsx:139`
```tsx
      <Stack.Screen
        name="NewsFeed"
        component={NewsFeedScreen}
        options={{
          title: 'Feed',
          fullScreenGestureEnabled: true,
          gestureResponseDistance: {
            start: 40,
            end: 353,
            top: 400,
            bottom: 852,
          },
        }}
      />
```
</details>

<details>
<summary>Using sheet props on NewsFeed</summary>

This will change the NewsFeed screen to the form sheet with rounded corners and maximum height to the middle.

`NativeStack.tsx:139`
```tsx
      <Stack.Screen
        name="NewsFeed"
        component={NewsFeedScreen}
        options={{
          title: 'Feed',
          presentation: 'formSheet',
          sheetAllowedDetents: 'medium',
          sheetCornerRadius: 32,
        }}
      />
```
</details>

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -20,6 +20,7 @@ import type {
   ScreenProps,
   ScreenStackHeaderConfigProps,
   SearchBarProps,
+  SheetDetentTypes,
 } from 'react-native-screens';
 
 export type NativeStackNavigationEventMap = {
@@ -31,6 +32,10 @@ export type NativeStackNavigationEventMap = {
    * Event which fires when a transition animation ends.
    */
   transitionEnd: { data: { closing: boolean } };
+  /**
+   * Event which fires when a swipe back is canceled on iOS.
+   */
+  gestureCancel: { data: undefined };
 };
 
 export type NativeStackNavigationProp<
@@ -331,6 +336,12 @@ export type NativeStackNavigationOptions = {
    */
   autoHideHomeIndicator?: boolean;
   /**
+   * Whether the keyboard should hide when swiping to the previous screen. Defaults to `false`.
+   *
+   * @platform ios
+   */
+  hideKeyboardOnSwipe?: boolean;
+  /**
    * Sets the navigation bar color. Defaults to initial navigation bar color.
    *
    * @platform android
@@ -423,6 +434,12 @@ export type NativeStackNavigationOptions = {
    */
   gestureEnabled?: boolean;
   /**
+   * Use it to restrict the distance from the edges of screen in which the gesture should be recognized. To be used alongside `fullScreenGestureEnabled`.
+   *
+   * @platform ios
+   */
+  gestureResponseDistance?: ScreenProps['gestureResponseDistance'];
+  /**
    * The type of animation to use when this screen replaces another screen. Defaults to `pop`.
    *
    * Supported values:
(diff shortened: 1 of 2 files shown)


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -104,7 +104,12 @@ const MaybeNestedStack = ({
   if (isHeaderInModal) {
     return (
       <ScreenStack style={styles.container}>
-        <Screen enabled style={StyleSheet.absoluteFill}>
+        <Screen
+          enabled
+          isNativeStack
+          hasLargeHeader={options.headerLargeTitle ?? false}
```
tboba, line 110: One comment about `isNativeStack` and `hasLargeHeader` props:
- `isNativeStack` is used internally by react-native-screens mainly for transition progress. It prevents non native-stack navigators from sending transition progress from their Screen components. It was already set on native-stack v5, and it should be also set there.
- hasLargeHeader is a prop that is used by our `ReanimatedNativeStackScreen` component and is used to determine if the screen has large header (since this information cannot be obtained from the sole screen).

satya164: > isNativeStack is used internally by react-native-screens mainly for transition progress. It prevents non native-stack navigators from sending transition progress from their Screen components. It was already set on native-stack v5, and it should be also set there.

Couldn't it automatically be detected (e.g. with context) when `Screen` is used inside `ScreenStack`?

packages/native-stack/src/types.tsx
```diff
@@ -330,6 +335,12 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   autoHideHomeIndicator?: boolean;
+  /**
+   * Whether the keyboard should hide when swiping to the previous screen. Defaults to `false`.
+   *
+   * @platform ios
+   */
+  hideKeyboardOnSwipe?: boolean;
```
satya164, line 343: JS stack has https://reactnavigation.org/docs/stack-navigator/#keyboardhandlingenabled so let's align that if possible

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: To be fair I wouldn't put a special context just to detect if screen is inside the ScreenStack. What do you mean by detecting the screen with `React.Children`? 🤔

packages/native-stack/src/types.tsx
tboba: Good idea 👍

packages/native-stack/src/views/NativeStackView.native.tsx
satya164: Something like `React.Children.map(children, (child) => React.cloneElement(child, { isNativeStack: true })` in `ScreenStack`.

> To be fair I wouldn't put a special context just to detect if screen is inside the ScreenStack

IMO if it's already possible to know if `isNativeStack` needs to be `true`, the responsibility shouldn't fall on the consumer. Otherwise it's to make a mistake and introduce a bug due to refactors later for something that could be handled automatically.

satya164 (approved):

a-eid:
I'm wondering why the `@react-navigation/native-stack` doesn't accept a `direction` prop the same way `react-native-screen/native-stack` does ?

tboba:
Hi @a-eid, from what I can see, `direction` prop is being controlled by `useLocale` hook. Hence, it's not necessary to manipulate it directly!
https://github.com/react-navigation/react-navigation/blob/90cfbf23bcc5259f3262691a9eec6c5b906e5262/packages/native-stack/src/views/HeaderConfig.tsx#L55

PR #11806, opened 2024-02-01
fix(Android): don't remount components during navigation on the new architecture

j-piasecki:
**Motivation**

This PR prevents the native view hierarchy built by `Stack` from re-mounting to prevent blurring focus - when a view is removed from the hierarchy it's blurred on Android, but it's not focused back when added again. Due to one of the views in the `Card` component being flattened, children of `Card` were removed from the hierarchy and attached under a different parent.

<details>
<summary>Before the change</summary>

https://github.com/react-navigation/react-navigation/assets/21055725/38ce6cbe-a717-4ca4-be04-0fdfe6e2e796

Native hierarchy on the first screen:
Screenshot 2024-02-01 at 11 46 27

Native hierarchy on the second screen:
Screenshot 2024-02-01 at 11 46 46

Notice the different nesting levels between the two - one of the views got collapsed during navigation.
</details>

<details>
<summary>After the change</summary>

https://github.com/react-navigation/react-navigation/assets/21055725/fd0b4623-1033-4ec8-aedd-0a66ae9adc19

Native hierarchy on the first screen:
Screenshot 2024-02-01 at 11 47 29

Native hierarchy on the second screen:
Screenshot 2024-02-01 at 11 47 36

Notice the same nesting levels between the two.
</details>

**Test plan**

<details>
<summary>I've tested the change on the following code</summary>

```jsx
import 'react-native-gesture-handler';
import {NavigationContainer, useFocusEffect} from '@react-navigation/native';
import React, {useRef} from 'react';
import {createStackNavigator} from '@react-navigation/stack';
import {Button, TextInput, View} from 'react-native';

function Thing({navigation}: any) {
  const ref = useRef(null);

  useFocusEffect(() => {
    setTimeout(() => {
      ref.current?.focus();
    }, 300);
  });

  return (
    <View style={{flex: 1, backgroundColor: 'wheat'}}>
      <TextInput
        ref={ref}
        style={{backgroundColor: 'white', borderWidth: 1, borderColor: 'black'}}
      />
      <Button
        title="navigate"
        onPress={() => {
          navigation.navigate('Second');
        }}
      />
    </View>
  );
}

function Second() {
  return <View style={{flex: 1, backgroundColor: 'red'}} />;
}

function App() {
  return (
    <NavigationContainer>
      <Stack.Navigator
        screenOptions={{headerShown: false}}
        detachInactiveScreens={false}>
        <Stack.Screen name="Home" component={Thing} />
        <Stack.Screen name="Second" component={Second} />
      </Stack.Navigator>
    </NavigationContainer>
  );
}

export default App;

```

</details>

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -539,6 +539,7 @@ export class Card extends React.Component<Props> {
           // Make sure that this view isn't removed. If this view is removed, our style with animated value won't apply
           collapsable={false}
         />
+        {/* Make sure this view is not removed on the new architecture, as it causes focus loss during navigation on Android. */}
         <View pointerEvents="box-none" collapsable={false} {...rest}>
           {overlayEnabled ? (
             <View pointerEvents="box-none" style={StyleSheet.absoluteFill}>
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -539,8 +539,13 @@ export class Card extends React.Component<Props> {
           // Make sure that this view isn't removed. If this view is removed, our style with animated value won't apply
           collapsable={false}
         />
-        {/* Make sure this view is not removed on the new architecture, as it causes focus loss during navigation on Android. */}
-        <View pointerEvents="box-none" collapsable={false} {...rest}>
+        <View
+          pointerEvents="box-none"
+          // Make sure this view is not removed on the new architecture, as it causes focus loss during navigation on Android.
+          // This can happen when the view flattening results in different trees - due to `overflow` style changing in a parent.
+          collapsable={false}
+          {...rest}
+        >
           {overlayEnabled ? (
             <View pointerEvents="box-none" style={StyleSheet.absoluteFill}>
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. Isn't this a bug in React Native if the view nesting level changes like this? Or is this specific interaction with view flattening and react-native-screens?

Also can you add a comment before the line in the code with the details?

j-piasecki:
It's caused by this line: https://github.com/react-navigation/react-navigation/blob/90cfbf23bcc5259f3262691a9eec6c5b906e5262/packages/stack/src/views/Stack/CardContainer.tsx#L274
I guess when `overflow` gets changed to hidden the criteria for collapsing the view are met, so it's removed.

It's reproducible without `react-native-screens`, the videos and the testing snipped from the PR are all with `detachInactiveScreens={false}` set on the navigator.

> Also can you add a comment before the line in the code with the details?

I've added a high-level explanation, is it good enough or should I go more in-depth?

satya164:
I have updated the comment. Will merge the workaround for now, tho I think it'd be good to open this issue in React Native repo as imo view flattening should be only an optimization and not observable like this, and this doesn't happen in the old architecture.

PR #11807, opened 2024-02-01
fix(Android): don't remount components during navigation on the new architecture for 6.x

j-piasecki:
The same as https://github.com/react-navigation/react-navigation/pull/11806 but targeting 6.x branch

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -517,7 +517,7 @@ export default class Card extends React.Component<Props> {
           // Make sure that this view isn't removed. If this view is removed, our style with animated value won't apply
           collapsable={false}
         />
-        <View pointerEvents="box-none" {...rest}>
+        <View pointerEvents="box-none" collapsable={false} {...rest}>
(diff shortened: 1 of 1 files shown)


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -517,7 +517,8 @@ export default class Card extends React.Component<Props> {
           // Make sure that this view isn't removed. If this view is removed, our style with animated value won't apply
           collapsable={false}
         />
-        <View pointerEvents="box-none" {...rest}>
+        {/* Make sure this view is not removed on the new architecture, as it causes focus loss during navigation on Android. */}
+        <View pointerEvents="box-none" collapsable={false} {...rest}>
```
satya164, line 521:
```suggestion
        <View
          pointerEvents="box-none"
          // Make sure this view is not removed on the new architecture, as it causes focus loss during navigation on Android.
          // This can happen when the view flattening results in different trees - due to `overflow` style changing in a parent.
          collapsable={false}
          {...rest}
        >
```

satya164 (approved):

Issue #11701, opened 2023-11-15
Types for `getState` are incorrect


josh-:
### Current behavior

Calling `getState` on the navigation object from `useNavigation` can return `undefined`, despite the types indicating that this is not possible.

For example, the following code:

```ts
const isStateUndefined = navigation.getState() === undefined;
```

returns the following ESLint error:

```
  29:28  error  Unnecessary conditional, the types have no overlap  @typescript-eslint/no-unnecessary-condition
```

since the return type of `getState` is defined as:

```ts
Readonly<{
    key: string;
    index: number;
    routeNames: never[];
    history?: unknown[] | undefined;
    routes: NavigationRoute<ReactNavigation.RootParamList, never>[];
    type: string;
    stale: false;
}>
```

however in my reproduction it's evident that `getState` can return `undefined`.

### Expected behavior

The types for `getState` indicate that the return type is `State | undefined`.

### Reproduction

https://github.com/josh-/react-navigation-type-repro

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |6.1.9
| @react-navigation/native-stack         |6.9.17
| react-native-safe-area-context         |4.6.3
| react-native-screens                   |3.22.1
| react-native                           |0.72.6
| expo                                   |49.0.18
| node                                   |18.18.2

goelbenj:
I am looking into this too, I will make a PR with what I consider a reasonable solution.

roryabraham:
We've encountered this [here](https://github.com/Expensify/App/issues/32515) and a similar issue [here](https://github.com/Expensify/App/issues/34383) where `navigation.getState().routes` is undefined.

MrRefactor:
Hey! It's Kuba from Callstack - expert contributor group - I'd like to take this issue.

roryabraham:
Haha hi @MrRefactor – this looks like the kind of comment we see a lot on issues in https://github.com/Expensify/App, and while I'd love to assign it to you you might have to talk to @satya164 about that 🙂

[...]

PR #11827, opened 2024-02-15
fix: type errors when getState used outside of a screen

MrRefactor:
This PR addresses the issue: [11701](https://github.com/react-navigation/react-navigation/issues/11701)

### DIFF
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -209,6 +209,12 @@ export const BaseNavigationContainer = React.forwardRef(
         getCurrentRoute,
         getCurrentOptions,
         isReady,
+        setParams: () => {
+          throw new Error('Cannot call setParams when not inside a screen');
+        },
+        setOptions: () => {
+          throw new Error('Cannot call setOptions when not inside a screen');
+        },
       }),
       [
         canGoBack,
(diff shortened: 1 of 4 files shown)


packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -209,6 +209,12 @@ export const BaseNavigationContainer = React.forwardRef(
         getCurrentRoute,
         getCurrentOptions,
         isReady,
+        setParams: () => {
+          throw new Error('Cannot call setParams when not inside a screen');
```
osdnk, line 213:
```suggestion
          throw new Error('Cannot call setParams when outside a screen');
```

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -209,6 +209,12 @@ export const BaseNavigationContainer = React.forwardRef(
...
+        },
+        setOptions: () => {
+          throw new Error('Cannot call setOptions when not inside a screen');
```
osdnk, line 216:
```suggestion
          throw new Error('Cannot call setOptions outside a screen');
```

packages/native/src/useScrollToTop.tsx
```diff
@@ -49,13 +50,18 @@ function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
 }
 
 export function useScrollToTop(ref: React.RefObject<ScrollableWrapper>) {
-  const navigation = useNavigation();
+  const navigation = React.useContext(NavigationContext);
   const route = useRoute();
 
+  if (navigation === undefined) {
+    throw new Error(
+      "Couldn't find a navigation object. Is your component inside NavigationContainer?"
```
osdnk, line 58:
```suggestion
      "Couldn't find a navigation object. Is your component inside the NavigationContainer?"
```

PR #11849, opened 2024-02-25
fix: conflicting nested path params in state for linking

rexfordessilfie:
**Motivation**
When a path configuration contains multiple path parameters with the same name, the last path parameter value wins when constructing the navigation state.

For example consider the following scenario:

```ts
const path = '/foo/42/bar/43';

  const config = {
    initialRouteName: 'Foo',
    screens: {
      Foo: {
        path: 'foo/:id',
        screens: {
          Bar: {
            path: 'bar/:id',
          },
        },
      },
    },
  };

  const state = {
    routes: [
      {
        name: 'Foo',
        params: { id: '43' }, // PROBLEM: This should be 42
        state: {
          routes: [
            {
              name: 'Bar',
              params: { id: '43' },
              path,
            },
          ],
        },
      },
    ],
  }
```

This PR updates the resolution logic to assign positions to each of the path segments, so that we can assign the correct values to the screen in the navigation state based on the position which it appears in the path, and not just the name.

**Test plan**
* Copy the test in this PR, checkout to main and run the tests
* You will notice that the test fails to resolve the right params to the route `'Foo'`
* Now checkout to this PR's branch and run all tests and notice that they pass

**Considerations**
* Another approach could be to include validation logic that prevents this scenario from happening by validating that there are no repeated path params with the same name.
  * This however, could be a breaking change, and might call for additional functionality as "aliasing" to be able to name path params in the config, and then translate them to a different name before being passed to the screen.
  * Alternatively, in the implementing screen, the developer could update the route param name to match what is in the new path parameter
  * OR, in the implementing screen, accept either the original param name or deep link param name and pick whichever one is available

Happy to adjust this PR to the validation approach if desired, or close in favor of the user-land mitigations. Thanks!

### DIFF
--- a/packages/core/src/getStateFromPath.tsx
+++ b/packages/core/src/getStateFromPath.tsx
@@ -331,28 +331,30 @@ const matchAgainstConfigs = (remaining: string, configs: RouteConfig[]) => {
         );
         const prefixPartsLength = prefix?.split('/').length;
 
-        const params = routeConfig?.normalizedPath
-          ?.split('/')
-          .map((p, index) => [p, index] as const)
-          .filter(([p]) => p.startsWith(':'))
-          .reduce(
-            (acc, [p, index]) => {
-              // Get the position of the param in the matched params based on the parent path segments count
-              index = prefixPartsLength ? index + prefixPartsLength - 1 : index;
-
-              const value = matchedParams[p]?.[index];
-
-              if (value) {
-                const key = p.replace(/^:/, '').replace(/\?$/, '');
-                acc[key] = routeConfig.parse?.[key]
-                  ? routeConfig.parse[key](value)
-                  : value;
-              }
-
+        const params = routeConfig?.normalizedPath?.split('/').reduce(
+          (acc, p, index) => {
+            if (!p.startsWith(':')) {
               return acc;
-            },
-            {} as Record<string, unknown>
-          );
+            }
+
+            // Get the position of the param in the matched params based on the parent path segments count
+            const pos = prefixPartsLength
+              ? index + prefixPartsLength - 1
+              : index;
+
+            const value = matchedParams[p]?.[pos];
+
+            if (value) {
+              const key = p.replace(/^:/, '').replace(/\?$/, '');
+              acc[key] = routeConfig.parse?.[key]
+                ? routeConfig.parse[key](value)
+                : value;
+            }
+
+            return acc;
+          },
+          {} as Record<string, unknown>
+        );
 
         if (params && Object.keys(params).length) {
           return { name, params };
--- a/packages/core/src/getStateFromPath.tsx
+++ b/packages/core/src/getStateFromPath.tsx
@@ -288,23 +288,35 @@ const matchAgainstConfigs = (remaining: string, configs: RouteConfig[]) => {
 
     // If our regex matches, we need to extract params from the path
     if (match) {
-      const matchedParams = config.pattern
-        ?.split('/')
-        .map((p, index) => [p, index] as const)
-        .filter(([p]) => p.startsWith(':'))
-        .reduce(
-          (acc, [p, index], i) => {
-            // The param segments appear every second item starting from 2 in the regex match result
-            const decodedParamSegment = decodeURIComponent(match![(i + 1) * 2]);
+      const matchResult = config.pattern?.split('/').reduce(
+        (acc, p, index) => {
+          if (!p.startsWith(':')) {
+            return acc;
+          }
 
-            return Object.assign(acc, {
-              [p]: Object.assign(acc[p] || {}, {
-                [index]: decodedParamSegment.replace(/\//, ''),
-              }),
-            });
-          },
-          {} as Record<string, Record<string, string>>
-        );
+          // Next match so increment position for current param. (e.g in `/:id/:name`, `:id` is 0 and `:name` is 1)
+          acc.paramIdx += 1;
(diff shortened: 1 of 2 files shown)


packages/core/src/getStateFromPath.tsx
```diff
@@ -289,23 +290,34 @@ const matchAgainstConfigs = (remaining: string, configs: RouteConfig[]) => {
...
+        .map((p, x) => [p, x] as const)
+        .filter(([p]) => p.startsWith(':'))
         .reduce<Record<string, any>>(
-          (acc, p, i) =>
+          (acc, [p, x], i) =>
             Object.assign(acc, {
               // The param segments appear every second item starting from 2 in the regex match result
-              [p]: match![(i + 1) * 2].replace(/\//, ''),
+              // Also include position in the key to handle the same param appearing multiple times
+              [p + x]: match![(i + 1) * 2].replace(/\//, ''),
```
satya164, line 300: Including the index in the key seems problematic.

If path is `foo1` and index is `1`, how would it be distinguished from `foo` and index `11`? While it's an unlikely case, it's still a problematic implementation.

Something like `[p]: { [index]: match![(i + 1) * 2].replace(/\//, '') }` may be more appropriate.

Also use `index` for index name since `x` isn't obvious.

The filter can probably be removed by adding this inside the `reduce` itself, and by adjusting the code accordingly:

```js
if (!p.startsWith(':')) {
  return acc;
}
```

packages/core/src/getStateFromPath.tsx
```diff
@@ -289,23 +290,34 @@ const matchAgainstConfigs = (remaining: string, configs: RouteConfig[]) => {
...
+          ?.split('/').length;
+
+        const params = config?.normalizedPath
           ?.split('/')
-          .filter((p) => p.startsWith(':'))
-          .reduce<Record<string, any>>((acc, p) => {
-            const value = matchedParams[p];
+          .map((p, x) => [p, x] as const)
+          .filter(([p]) => p.startsWith(':'))
+          .reduce<Record<string, any>>((acc, [p, x]) => {
```
satya164, line 317: Don't use `any`

packages/core/src/getStateFromPath.tsx
```diff
@@ -441,6 +454,7 @@ const createConfigItem = (
     regex,
     pattern,
     path,
+    normalizedPath,
```
satya164, line 457: The `normalizedPath` can be calculated from `path` where needed, so there's no need to include it additionally.

Also is it necessary to have both `path` and `normalizedPath`?

packages/core/src/getStateFromPath.tsx
```diff
@@ -289,23 +290,34 @@ const matchAgainstConfigs = (remaining: string, configs: RouteConfig[]) => {
...
         );
 
       routes = config.routeNames.map((name) => {
         const config = configs.find((c) => c.screen === name);
-        const params = config?.path
+
+        // Get the number of segments in the initial pattern
+        const initialPartsLength = config?.pattern
+          .replace(new RegExp(`${escape(config.normalizedPath)}$`), '')
+          ?.split('/').length;
```
satya164, line 311: This isn't obvious to me. What's "initialParts" referring to here?

satya164 (changes requested):
Thanks for the PR

packages/core/src/getStateFromPath.tsx
rexfordessilfie: Thanks for the question! The solution this PR proposes relies on the path being a suffix of the pattern.

Normalizing it in a similar way to pattern helps to make sure that path is always a suffix of the pattern (even if the user entered double or leading/trailing slashes in their configuration). 

What do you think? Is there a better approach? Should we mutate path itself here instead, or clean it up at point of usage?

packages/core/src/getStateFromPath.tsx
rexfordessilfie: Got it! I'll update you use unknown instead if that's okay.

packages/core/src/getStateFromPath.tsx
rexfordessilfie: Many this is better named something as `prefixParts`. 

For example if the normalized path is:

'car/:id'

And pattern is:

'user/:tag/car/:id'


Then initialParts is:

'[user, :tag]'

packages/core/src/getStateFromPath.tsx
rexfordessilfie: I love this suggestion! I will update accordingly.

rexfordessilfie: I have made the changes for this. After removing the `filter()`, I noticed we could also get rid of the `map()` as well, and so I took care of that for us!

Thanks for the suggestion. With this approach we are able to do all the processing in one pass 🙌

packages/core/src/getStateFromPath.tsx
rexfordessilfie: I changed this to `numInitialSegments`, but I'm not 100% committed to it. Let me know if we could use a better name for this.

packages/core/src/getStateFromPath.tsx
satya164: I'm just saying that there's no need to include it in the object when it can be calculated when needed. I have updated the PR with the change.

satya164 (approved):
Thank you!

PR #11852, opened 2024-02-26
fix: Eliminates key errors in notes

ramen2020:
**Motivation**

"While reviewing the internal implementation, I noticed several typos stood out. Since these could lead to misunderstandings, especially for those not fluent in English, I took the opportunity to correct them as well."

**Test plan**

It's a comment-out modification, so there's no scope of impact.


satya164 (approved):
Thanks

PR #11854, opened 2024-02-27
feat: add animation property for js stack

groot007:
**Motivation**

The simple stack of navigation doesn't have an animation prop for easy configuration transition between screens 

**What was done**
Now we can pass the option "animation" to the screen options and have default animation 
And here are the basic transitions that could be used:
  | 'default'
  | 'fade'
  | 'fade_from_bottom'
  | 'none'
  | 'reveal_from_bottom'
  | 'scale_from_center'
  | 'slide_from_bottom'
  | 'slide_from_right'
  | 'slide_from_left'


packages/stack/src/TransitionConfigs/TransitionPresets.tsx
```diff
@@ -150,3 +151,27 @@ export const ModalTransition = Platform.select({
...
+  fade: ModalFadeTransition,
+  fade_from_bottom: FadeFromBottomAndroid,
+  none: DefaultTransition,
+  modal_slide_from_bottom: ModalSlideFromBottomIOS,
+  reveal_from_bottom: RevealFromBottomAndroid,
+  scale_from_center: ScaleFromCenterAndroid,
+  slide_from_bottom: BottomSheetAndroid,
+  slide_from_left: SlideFromLeftIOS,
+  slide_from_right: SlideFromRightIOS,
+} as const;
```
satya164, line 177: I think this will ensure that all the animation names are kept in sync.

```suggestion
as const satisfies Record<StackAnimationTypes, TransitionPreset>;
```

packages/stack/src/types.tsx
```diff
@@ -78,8 +78,19 @@ export type GestureDirection =
   | 'vertical'
   | 'vertical-inverted';
 
+export type StackAnimationTypes =
```
satya164, line 81:
```suggestion
export type StackAnimationName =
```

packages/stack/src/TransitionConfigs/TransitionPresets.tsx
```diff
@@ -150,3 +151,27 @@ export const ModalTransition = Platform.select({
...
+ */
+export const AnimationTransitions = {
+  default: DefaultTransition,
+  fade: ModalFadeTransition,
+  fade_from_bottom: FadeFromBottomAndroid,
+  none: DefaultTransition,
+  modal_slide_from_bottom: ModalSlideFromBottomIOS,
+  reveal_from_bottom: RevealFromBottomAndroid,
+  scale_from_center: ScaleFromCenterAndroid,
+  slide_from_bottom: BottomSheetAndroid,
```
satya164, line 174: I'd change it to:

```suggestion
  bottom_sheet: BottomSheetAndroid,
  reveal_from_bottom: RevealFromBottomAndroid,
  scale_from_center: ScaleFromCenterAndroid,
  slide_from_bottom: ModalSlideFromBottomIOS,
```

satya164 (changes requested):

packages/stack/src/types.tsx
```diff
@@ -326,11 +338,19 @@ export type StackNavigationOptions = StackHeaderOptions &
...
+     * How the screen should animate when pushed or popped.
+     *
+     * Supported values:
+     * - "none": don't animate the screen
+     * - "default": use the platform default animation
+     * - "fade": fade screen in or out
+     * - "fade_from_bottom": fade screen in or out from bottom
+     * - "slide_from_bottom": slide in the new screen from bottom
+     * - "slide_from_right": slide in the new screen from right
+     * - "slide_from_left": slide in the new screen from left
```
satya164, line 350: Also remember to update this list

packages/stack/src/types.tsx
```diff
@@ -78,8 +78,19 @@ export type GestureDirection =
...
+export type StackAnimationName =
+  | 'default'
+  | 'fade'
+  | 'fade_from_bottom'
+  | 'none'
+  | 'reveal_from_bottom'
+  | 'scale_from_center'
+  | 'slide_from_bottom'
+  | 'slide_from_right'
+  | 'slide_from_left';
```
groot007, line 90: @satya164 I've removed the 'bottom_sheet' cause it's the same as 'slide_from_bottom' - diffs only in few miliseconds

satya164 (changes requested):
After replacing `TransitionPresets` with `animation` in `MixedHeaderMode` example, seems it's not working quite the same.

When using the `animation: 'slide_from_bottom'` option and opening and closing the "Albums" screen, you can notice the closing animation is faster than what it is when using `TransitionPresets.ModalSlideFromBottomIOS`. It seems that somehow the default `transitionSpec` is being used for the animation.

They should both behave functionally identical. So this is something we need to investigate before merging.

groot007:
> After replacing `TransitionPresets` with `animation` in `MixedHeaderMode` example, seems it's not working quite the same.
> 
> When using the `animation: 'slide_from_bottom'` option and opening and closing the "Albums" screen, you can notice the closing animation is faster than what it is when using `TransitionPresets.ModalSlideFromBottomIOS`. It seems that somehow the default `transitionSpec` is being used for the animation.
> 
> They should both behave functionally identical. So this is something we need to investigate before merging.

looks like the problem is in animation itself - i've added a new animation for closing modal screen - but not sure about the values cause now I just made it intuitive - do we have some specific animation config we should be strict to?

groot007:
> > looks like the problem is in animation itself
> 
> I'm not sure what you mean. It's working as I would expect when using `...TransitionPresets.ModalSlideFromBottomIOS`, and `slide_from_bottom` is the same as `TransitionPresets.ModalSlideFromBottomIOS`. The issue is that these 2 things are behaving differently when they shouldn't.
> 
> > i've added a new animation for closing modal screen - but not sure about the values cause now I just made it intuitive
> 
> The current animations are based on native animations. Unless you have a fix to existing animations to match the native ones more closely, please don't change them.

ah got it thanks - just digged in wrong direction
fixed an issue with correct gestureDirection - it was wrong calculation for the animation distance and that's why animation was fasted

satya164 (approved):
Thanks

PR #11859, opened 2024-02-28
fix: fixing set params and set options methods on BaseNavigationContainer

alexandre1921:
**Motivation**
The latest update [@react-navigation/core@6.4.11](https://github.com/react-navigation/react-navigation/releases/tag/%40react-navigation%2Fcore%406.4.11) broke react navigation and subsequently expo router

**Test plan**

- Create an app
- Create the router
- Try to setParams on a screen
- Try to setOptions on a screen

### DIFF
--- a/packages/core/src/BaseNavigationContainer.tsx
+++ b/packages/core/src/BaseNavigationContainer.tsx
@@ -209,12 +209,6 @@ export const BaseNavigationContainer = React.forwardRef(
         getCurrentRoute,
         getCurrentOptions,
         isReady,
-        setParams: () => {
-          throw new Error('Cannot call setParams outside a screen');
-        },
-        setOptions: () => {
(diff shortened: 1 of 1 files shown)


satya164:
You can't call `setParams` on the navigation ref. Please provide a repro showing the issue if there's a bug.

satya164:
Fixed in https://github.com/react-navigation/react-navigation/commit/b99d3c3a94d7f4efe7fb0309d7a41654c21406ea

alexandre1921:
Thanks for ignoring the PR and fixing this bug yourself.

satya164:
@alexandre1921 so you're upset that I fixed the issue you were having when there was no repro for me to actually understand the issue to know what's the proper fix? people are really entitled.

alexandre1921:
I was just trying to contribute, but again thank you for fixing the issue.

PR #11860, opened 2024-02-29
fix: Make all screens after presentation: 'modal' as modal unless specified

groot007:
Make all screens after presentation: 'modal' as modal unless specified

### DIFF
--- a/packages/native-stack/src/utils/getModalRoutesKeys.ts
+++ b/packages/native-stack/src/utils/getModalRoutesKeys.ts
@@ -4,15 +4,15 @@ import type { NativeStackDescriptorMap } from '../types';
 
 export const getModalRouteKeys = (
   routes: Route<string>[],
-  descriptors: NativeStackDescriptorMap,
-  modalTypes: string[] = ['modal']
+  descriptors: NativeStackDescriptorMap
 ) =>
   routes.reduce<string[]>((acc, route) => {
     const { presentation } = descriptors[route.key]?.options ?? {};
 
     if (
       (acc.length && !presentation) ||
-      modalTypes.includes(presentation ?? '')
+      presentation === 'modal' ||
+      presentation === 'transparentModal'
     ) {
       acc.push(route.key);
     }
--- a/packages/native-stack/src/utils/getModalRoutesKeys.ts
+++ b/packages/native-stack/src/utils/getModalRoutesKeys.ts
@@ -1,0 +1,21 @@
+import type { Route } from '@react-navigation/native';
+
+import type { NativeStackDescriptorMap } from '../types';
+
+export const getModalRouteKeys = (
+  routes: Route<string>[],
+  descriptors: NativeStackDescriptorMap,
+  modalTypes: string[] = ['modal']
+) =>
+  routes.reduce<string[]>((acc, route) => {
+    const { presentation } = descriptors[route.key]?.options ?? {};
+
+    if (
+      (acc.length && !presentation) ||
+      modalTypes.includes(presentation ?? '')
+    ) {
+      acc.push(route.key);
+    }
+
+    return acc;
+  }, []);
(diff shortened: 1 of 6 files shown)


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -159,12 +159,17 @@ const SceneView = ({
...
-    presentation = 'card',
-  } = options;
+  let { animation, animationMatchesGesture, fullScreenGestureEnabled } =
+    options;
+
+  if (
+    !options.presentation &&
+    previousDescriptor?.options.presentation === 'modal'
+  ) {
+    options.presentation = 'modal';
```
satya164, line 169: We shouldn't mutate props/options.

satya164 (changes requested):

packages/native-stack/src/views/NativeStackView.native.tsx
groot007: good point, thanks
I've changed to another approach with an additional flag

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -449,6 +451,10 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
 
   useInvalidPreventRemoveError(descriptors);
 
+  // if the is a one screen with modal 'presentation'
+  // then all next screens also should be with the same presentation
+  let isForsedModalStack = false;
```
satya164, line 456: I'd change the naming from `isForcedModalStack` to something like `isModal`, since it can either be an automatic modal or specified by the user as a modal, and this is for a screen and not the whole stack.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -461,6 +467,12 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
             ? descriptors[previousKey]
             : undefined;
           const nextDescriptor = nextKey ? descriptors[nextKey] : undefined;
+          if (
+            !descriptor.options.presentation &&
+            previousDescriptor?.options.presentation === 'modal'
+          ) {
+            isForsedModalStack = true;
```
satya164, line 474: Changing an outside variable and using it in a map may be unexpected when reading the code since the function passed to map is usually expected to be a pure function.

It might be better to do something like:

```js
const modalRouteKeys = state.routes.reduce((modals, route) => {
  const { presentation } = descriptors[route.key].options;

  // Assume modal if there are already modal screens in the stack
  // or current screen is a modal when no presentation is specified
  if ((modals.length && !presentation) || presentation === 'modal') {
    acc.push(route.key);
  }

  return acc;
}, []);
```

And later check if something is modal by `const isModal = modalRouteKeys.includes(route.key)`

satya164 (changes requested):

satya164 (commented):
The other thing to check is if we also need to consider `transparentModal` along with `modal`

packages/native-stack/src/utils/getModalRoutesKeys.ts
```diff
@@ -0,0 +1,21 @@
+import type { Route } from '@react-navigation/native';
+
+import type { NativeStackDescriptorMap } from '../types';
+
+export const getModalRouteKeys = (
```
groot007, line 5: @satya164 can we place this util somewhere to share between the packages?
cause now I made a dublicate for native and simple stack :(

satya164: It's fine to duplicate. There's not really much logic here.

groot007:
> The other thing to check is if we also need to consider `transparentModal` along with `modal`

from my perspective automatically stack for 'modal' looks logical but 'transparentModal' can be a little mess with all these transparent screens stack

satya164:
> from my perspective automatically stack for 'modal' looks logical but 'transparentModal' can be a little mess with all these transparent screens stack

We shouldn't mark all screens after `transparentModal` as `transparentModal`, but as regular modal. The idea is to avoid the issue of screens appearing behind the transparent modal even when pushed.

satya164 (approved):

Issue #11861, opened 2024-02-29
Error from useBottomTabBarHeight() limits its usage.


therealkh:
### Current behavior

When you call `useBottomTabBarHeight` hook not inside tab you will get thrown Error.
This limits usage of this hook. For example you have some similar component with small differences. This components have some exactly same UI calculations. You decide to move this calculations logic to a reusable hook. But surprise surprise you will get thrown error as soon as you use this hook not inside tab. I have the same case. And I would like to `useBottomTabBarHeight` return 0 and maybe throw some warning instead of error.
![Simulator Screenshot - iPhone 14 Pro - 2024-02-29 at 15 59 42](https://github.com/react-navigation/react-navigation/assets/65399389/527beff8-4ef2-418f-9b6b-d72bbce99991)

### Expected behavior

useBottomTabBarHeight shouldn't throw error.

### Reproduction

Call `useBottomTabBarHeight()` in any screen without bottom tab bar.

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^6.1.6
| @react-navigation/bottom-tabs          | ^6.5.8
| @react-navigation/stack                | ^6.3.17
| @react-navigation/native-stack         | ^6.9.12
| react-native-safe-area-context         | ^4.7.4
| react-native-screens                   | ^3.20.0
| react-native-gesture-handler           | ^2.10.1
| react-native-reanimated                | ^3.5.4
| react-native                           | 0.72.6
| node                                   | 21.1.0
| yarn                            | 3.6.3

rexfordessilfie:
Hey @therealkh, you can also try this if using the context directly avoids the error, but does not give you the correct heights:
* From inside a screen/component/hook that is rendered inside the `BottomTabBarHeight` synchronize the height with an external store that is accessible from where you want

**Example (with React Context)**
NB: I have left out the definition of the Context + Provider, but it should be simple to setup!

```ts

// Use this inside some screen that is rendered inside BottomTabBarHeightContext
function useSynchronizeTabBarHeight(){
[...]

PR #11862, opened 2024-02-29
Remove throwing error from useBottomTabBarHeight

therealkh:
**Motivation**

Motivation of this PR is described on [this issue](https://github.com/react-navigation/react-navigation/issues/11861)


satya164:
Closing since the use case is already possible https://github.com/react-navigation/react-navigation/issues/11861#issuecomment-1970917545

PR #11866, opened 2024-03-05
feat: added support for switch native driver usage in react native ta…

vibhor-d11:
The following PR addresses the functionality to enable/disable the use of native driver for running animations for react-native-tab-view. For some specific devices the app crashes when native driver is enabled for android version 12 and above. So we have leveraged the `useNativeDriver`  prop to make it configurable.

For more info please refer to the issue : [Link](https://github.com/react-navigation/react-navigation/pull/11866)


satya164 (changes requested):
Thanks for the PR. Have you opened an issue in `react-native-pager-view` and `react-native` repos?

And please provide more details about the problem such as which devices have such issues, error message or stack trace etc.

vibhor-d11:
> Thanks for the PR. Have you opened an issue in `react-native-pager-view` and `react-native` repos?
> 
> And please provide more details about the problem such as which devices have such issues, error message or stack trace etc.

You can refer to the [issue](https://github.com/react-navigation/react-navigation/issues/11890) for the stacktrace and more info. The issue was mainly on android devices with android version 12.

lucasoliveiraw00:
> Obrigado pelo PR. Você abriu um problema `react-native-pager-view`e `react-native`repositórios?
> 
> E forneça mais detalhes sobre o problema, como quais dispositivos apresentam esses problemas, mensagem de erro ou rastreamento de pilha, etc.

https://github.com/react-navigation/react-navigation/issues/11890#issuecomment-2085264157

lucasoliveiraMM:
https://github.com/react-navigation/react-navigation/issues/11890#issuecomment-2090294675

anindya2791:
@satya164 are we planning to take this PR or are we planning some alternative? This exact crash is impacting a lot of Android 12 users of our app. We are looking forward to this being taken in the upcoming release

Screenshot 2024-07-16 at 10 29 43 AM

satya164:
@anindya2791 the bug is in react-native, not in this library. there are already patches in the linked issue that you can try with https://www.npmjs.com/package/patch-package

please spend your energy to get the actual issue fixed.

douglasjunior:
@satya164 , regardless of the bug in React Native, this PR just opens up an option for customizing animations; in my opinion, it shouldn't be a barrier to acceptance.

vibhor-d11:
> @satya164 , regardless of the bug in React Native, this PR just opens up an option for customizing animations; in my opinion, it shouldn't be a barrier to acceptance.
> 
> We know that fixes in the framework take much longer than in the libraries (at least they should).

yes can we look into this  @satya164 ? as it helps in enabling/disabling  the use of useNativeDriver for animations

satya164:
It's not a customization option. It's a workaround for a bug in React Native. The only reason you want to disable it is because of the bug. Otherwise disabling native driver isn't desirable as it can result choppy animation.

You can already use `patch-package` for the workaround, or there are other patches in the linked issue in React Native repo which also workaround this without disabling native driver. Choose whichever workaround is more appropriate for your case. A workaround doesn't need to be a public API that we can't remove in a next version once it's not needed.

Yes, fixes in framework take much longer, but it's not even going anywhere now as the [original issue is closed](https://github.com/facebook/react-native/issues/33375). It'll be better to prepare a repro or anything else React Native maintainers need to look into this bug to make sure that it actually gets fixed. Without that it's not gonna just take longer, it'll never be fixed.

If this bug affects your production app, you should be able to spend some time on that.

satya164:
What do you expect to change if this option is added?

Adding an option to disable native driver isn't going to fix it. Unless native driver is always disabled - people are still going to run into crashes and they'll arrive on a related issue which contains a patch they can use for the workaround.

Adding this option only means there's a public API for what's supposed to be a workaround that we cannot remove or risk breaking apps.

If patching being difficult is a concern, you can post a patch that people can easily copy/paste to their projects to make adding a patch easier. 

> Regarding patch-package, we know it's an option and we're already using it, but now what? Do we just sit back and do nothing?

If you want to do something then you help fix the root of the issue. Trading one workaround for another doesn't change anything.

> So I think it would be fair that, at the very least, they help us reproduce and relay the issue to the maintainers of React Native.

I would help if I could, but no, it's not "fair" to expect it.

If you're facing the crash in your production app, you are the best person to reproduce it. And you get paid to fix bugs in your app so you should be able to spend time on it.  If there's anything specific you need from me to be able to create a repro, then let me know.

> Relay the issue to the maintainers of React Native.

And I don't have any special privileges to relay this to React Native maintainers.

>  Today, we're in the dark.

The issue doesn't even mention which devices, just says "some devices" with Android 12. I know even less details than you do. I'm not in any less dark than you are.

douglasjunior:
I'm sorry man, I have no intention of offending you. I have also been working with open source for over a decade, I know how things works, I know that in the most the jobs are done by volunteers. 
My only goal here is to ask for help and highlight a pain point of the community.

You speak as if we are sitting back and waiting for others to solve our problems, but that is not true.

To reach the point of entering a discussion with you, rest assured that everything within my power has been done and the issue with react-navigation has indeed been mitigated.

Rest assured, this also applies to most of the community members who come asking for help.

I'm sorry, but if the maintainers of the largest navigation library don't have an open line with the React Native core team, then we are all adrift.

satya164:
> You speak as if we are sitting back and waiting for others to solve our problems, but that is not true.

I'm not saying that you aren't doing anything, but rather focusing on the wrong thing. Merging this PR may feel like we're doing something but the core issue still remains. We only trade using patch-package in favor of being able to disable native driver with an API - both are workarounds.

I don't expect anyone outside React Native core team to be able to trivially solve the problem. What you can do to help is provide as much as info you can and hope for the best. In the end, the more people are facing the issue, the better the chance that React Native team will prioritize it.

Adding a public API for disabling native driver isn't something I want to do as it will result in bad animation performance - and I can't remove it in future without breaking changes. The linked issue contains different kind of workarounds that doesn't involve disabling native driver.

With patch-package, people can choose whichever workaround they prefer. I'd encourage to try the patches mentioned in the linked issue that address this without disabling native driver and see if those help.

douglasjunior:
The `patch-package` suggested in that issue doesn't apply to the most recent version of React Native, because the Java code is now shipped as an `aar` in Maven Central.

Therefore, we have to disable the native driver via JS for Android 12-only devices. We temporarily solved this via `patch-package` on the `react-navigation` JS side (as submitted by this PR), as it is better to have lower performance than a crash.

anindya2791:
> The `patch-package` suggested in that issue doesn't apply to the most recent version of React Native, because the Java code is now shipped as an `aar` in Maven Central.
> 
> Therefore, we have to disable the native driver via JS for Android 12-only devices. We temporarily solved this via `patch-package` on the `react-navigation` JS side (as submitted by this PR), as it is better to have lower performance than a crash.

so you changed the transpiled files too using `patch-package`?

PR #11867, opened 2024-03-05
fix: Adjust URL encoding to avoid encoding unnecessarily

groot007:
**Motivation**

sometimes encodes characters unnecessarily, e.g. @ is converted to %40 even though it's valid in a URL.
valid characters - A–Z, a–z, 0–9, -, ., _, ~, !, $, &, ', (, ), *, +, ,, ;, =, :, @

### DIFF
--- a/packages/core/utils/encodeParams.ts
+++ b/packages/core/utils/encodeParams.ts
@@ -1,0 +1,7 @@
+export const encodeParams = (param: string) =>
+  String(param).replace(
+    /[^A-Za-z0-9\-._~!$&'()*+,;=:@]/g,
+    function (char: string) {
+      return '%' + char.charCodeAt(0).toString(16).toUpperCase();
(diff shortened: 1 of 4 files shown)


packages/core/utils/encodeParams.ts
```diff
@@ -0,0 +1,7 @@
+export const encodeParams = (param: string) =>
+  String(param).replace(
+    /[^A-Za-z0-9\-._~!$&'()*+,;=:@]/g,
```
satya164, line 3:
```suggestion
    // Valid characters according to https://datatracker.ietf.org/doc/html/rfc3986#section-3.3 (see pchar definition)
    /[^A-Za-z0-9\-._~!$&'()*+,;=:@]/g,
```

satya164 (approved):
Thanks

PR #11869, opened 2024-03-05
fix: [6.x] - decode params for state creation 

groot007:
**Motivation**

params was not decoded before put to the state - it can lead to different path


packages/core/src/__tests__/getStateFromPath.test.tsx
```diff
@@ -48,6 +48,44 @@ it('converts path string to initial state', () => {
   );
 });
 
+it('decode path params correctly', () => {
```
satya164, line 51:
```suggestion
it('decodes encoded params in path', () => {
```

satya164 (approved):
Thanks

PR #11872, opened 2024-03-06
fix: add drawer and header styles according to material design

groot007:
correct some styles according to material design


packages/elements/src/Header/Header.tsx
```diff
@@ -268,6 +268,7 @@ export function Header(props: Props) {
                       (rightButton ? 72 : 16) +
                       insets.left -
                       insets.right),
+              marginLeft: headerTitleAlign === 'center' ? 12 : 4,
```
satya164, line 271: Why the different left margin for the centered title? It makes it slightly off-center.

satya164 (approved):

PR #11873, opened 2024-03-06
fix: add links href to material top tabs

groot007:
**Motivation**

Material tabs on web had no links for better UX


satya164 (approved):
Thanks

PR #11876, opened 2024-03-07
fix: preserve # in the URL if the screen hasn't changed - partially done

groot007:
**Motivation**

when updating some params on the page, the hash shouldn’t be removed

### DIFF
--- a/packages/native/src/useLinking.tsx
+++ b/packages/native/src/useLinking.tsx
@@ -300,8 +300,19 @@ export function useLinking(
           }
         }
       }
+
+      const [previousFocusedState] = findMatchingState(
+        previousStateRef.current,
+        state
+      );
+
+      const lastScreen =
+        previousFocusedState?.routes?.[
+          previousFocusedState?.routes?.length - 1
+        ];
+      // @ts-expect-error key is ommited for focused route
+      const isSameScreen = lastScreen?.key === route?.key;
       const path = getPathFromStateRef.current(state, configRef.current);
-      const isSameScreen = path.split('?')[0] === route?.path?.split('?')?.[0];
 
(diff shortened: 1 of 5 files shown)


packages/native/src/useLinking.tsx
```diff
@@ -296,12 +296,14 @@ export function useLinking(
             focusedRoute.name === route.name &&
             isEqual(focusedRoute.params, route.params)
           ) {
-            return route.path;
+            return route.path + location.hash;
           }
         }
       }
+      const path = getPathFromStateRef.current(state, configRef.current);
+      const isSameScreen = path.split('?')[0] === route?.path?.split('?')?.[0];
```
satya164, line 304: `path` being the same doesn't always mean it's the same screen. the same screen could have 2 different paths if the params have changed for params in path segment, e.g. `/users/:userId`, 

To know if it's the same screen, you'd need to get the key of the current route (`route.key`) and compare it with the previous key (`route.key`).

satya164 (changes requested):

satya164:
Also let's add some tests as well. You can add them under `example/e2e`.

packages/native/src/useLinking.tsx
```diff
@@ -296,12 +296,25 @@ export function useLinking(
...
+      const [previousFocusedState] = findMatchingState(
+        previousStateRef.current,
+        state
+      );
+
+      const lastScreen =
+        previousFocusedState?.routes?.[
+          previousFocusedState?.routes?.length - 1
+        ];
+      // @ts-expect-error key is ommited for focused route
```
groot007, line 313: can not get it why do we have omitted 'key' prop in return type findFocusedRoute - but it present there

satya164 (approved):

Issue #11429, opened 2023-06-19
HeaderTitleStyle not setting Font Family in IOS


harshitsharmanagarro:
### Current behavior

I tried setting font family for header title using below code but it is not working specifically for IOS. for android it is working fine. this fontFamily is setting correctly at all other places for IOS but only in header it is not working. please check.

<Stack.Navigator
      screenOptions={{
        headerStyle: {backgroundColor: Colors.title_blue},
        headerBackTitleVisible: false,
        headerTitleStyle: {fontSize: 24, fontFamily: 'Rajdhani-Bold'},
        headerTintColor: Colors.light_blue,
        contentStyle: {backgroundColor: Colors.dark_blue},
        animation: 'slide_from_right',
        headerTitleAlign: 'center',
      }}>
      <Stack.Screen name="Home" component={Home} />
      <Stack.Screen name="Playlist" component={Playlist} />
      <Stack.Screen name="CreatePlaylist" component={CreatePlaylist} />
      <Stack.Screen name="ViewPlaylist" component={ViewPlaylist} />
      <Stack.Screen name="Settings" component={Settings} />
    </Stack.Navigator>

### Expected behavior

it should set header title style correctly for IOS as well.

### Reproduction

https://snack.expo.dev/yfdBLvQ6K

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^6.1.6
| @react-navigation/native-stack         | ^6.9.12
| react-native                           | 0.71.8
| node                                   | 16.16.0
| npm                         | 8.11.0

kyawthura-gg:
Can you please provide a valid repo? It will save alot of time for us. Thanks!

PR #11880, opened 2024-03-10
fix: support custom fontFamily configurations in headerTitleStyle

Kasendwa:
Passing a custom font for the `headerTitleStyle` configuration is ignored because the user set values always get overridden by the `fonts` values from the `useTheme()` hook. 

By changing the order of the values passed to `Stylesheet.flatten()`, any user passed `fontFamily` configurations then get retained.

I believe this fixes the issue raised in #11429. It's not just iOS specific; it cuts across all platforms.


satya164 (approved):
Thanks for the PR

PR #11887, opened 2024-03-12
feat: added integration for `InteractionManager` with `native-stack`

kirillzyusko:
**Motivation**

The integration with `interactionManager` was added in react-navigation@5 I believe (https://reactnavigation.org/blog/2020/02/06/react-navigation-5.0/#other-improvements).

And since then it was a convenient way to run tasks when animation gets finished.

Since `react-navigation@6` it's recommended to use `native-stack` instead of `stack`. However it seems like `InteractionManager` doesn't take into consideration animations driven by `native-stack`. As a result all code that was relying on execution `InteractionManager.runAfterInteractions` callback after animation is broken when we switch to `native-stack`.

**Test plan**

The way to test the integration would be to add a code, like:

```tsx
const [color, setColor] = useState("#000000");

const goToNextScreen = () => {
  navigation.navigate("NextScreenInStack");

  InteractionManager.runAfterInteractions(() => {
    setColor("#ff0000");
  });
};
```

- Without these change the color would be changed immediately and you could see a change during the animation;
- With these changes the callback will be executed only when animation gets completed.

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -264,18 +264,24 @@ const SceneView = ({
         InteractionManager.createInteractionHandle();
     }
   }, [focused]);
+  const finishInteraction = React.useCallback(() => {
+    if (interactionHandleRef.current !== undefined) {
+      InteractionManager.clearInteractionHandle(interactionHandleRef.current);
+      interactionHandleRef.current = undefined;
+    }
+  }, []);
+  // in case if screen is unmounted faster than transition finishes, then `onAppear` will not be fired
+  // so we clean up an interaction here
+  React.useEffect(() => finishInteraction, [finishInteraction]);
   const onAppearCallback = React.useCallback<
     NonNullable<ScreenProps['onAppear']>
   >(
     (e) => {
       onAppear?.(e);
 
-      if (interactionHandleRef.current !== undefined) {
-        InteractionManager.clearInteractionHandle(interactionHandleRef.current);
-        interactionHandleRef.current = undefined;
-      }
+      finishInteraction();
     },
-    [onAppear]
+    [onAppear, finishInteraction]
   );
 
   const [customHeaderHeight, setCustomHeaderHeight] =
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -19,6 +19,7 @@ import {
 import * as React from 'react';
 import {
   Animated,
+  InteractionManager,
   Platform,
   StyleSheet,
   useAnimatedValue,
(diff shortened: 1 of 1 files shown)


kirillzyusko:
@satya164 may I kindly ask you to review these changes? 🙏 👀

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -136,7 +137,7 @@ type SceneViewProps = {
   isPresentationModal?: boolean;
   onWillDisappear: () => void;
   onWillAppear: () => void;
-  onAppear: () => void;
+  onAppear: ScreenProps['onAppear'];
   onDisappear: () => void;
```
tboba, line 141: Just for clarity, can we propagate this change onto other events?

kirillzyusko: @tboba yes, sure. I'll make these changes and push new changes 👍

kirillzyusko:
@tboba since you reviewed this PR - I'd like to ask you one question 👀 

I found one bug in this implementation. Basically if we have a simple Navigation with bottom-tabs and nested stacks inside bottom tabs, then when you open a screen in nested stack - the `focused` property becomes `true` (for BottomTab navigator) but we never get `onAppear` event 😲 

I don't remember all details, like whether it happened on Android or iOS (or both). But I remember that I workarounded this problem by adding `setTimeout(finishInteraction, 500)` straight after `InteractionManager.createInteractionHandle();` 🙃 

Do you have any ideas what would be better to do here? I wouldn't like to add these changes to this PR. If needed I can prepare a reproduction example 👍

satya164:
Thanks for your PR. The reason `InteractionManager` is not integrated with `native-stack` is because the purpose of InteractionManager is to make animations run smoothly by deferring work - and it doesn't appear to be the case for native stack unless I'm missing something.

`InteractionManager` isn't a React Navigation specific API, but a generic API intended to improve animation perf by deferring work. So if we're deferring work even when not necessary, it can cause unnecessary delays.

It's important to note that `InteractionManager.runAfterInteractions` isn't a replacement for animation end events, but a tool to optimize animation and gesture performance. If I'm mistaken and it actually improves animation performance, then let me know.

I'm closing this PR for now.

kirillzyusko:
> If I'm mistaken and it actually improves animation performance, then let me know.

@satya164 It doesn't improve the animation performance. The reason why I added this code is because many `react-native` apps were using plain `@react-navigation/stack` and a lot of people used `InteractionManager` to run effects when a transitions has completed.

People also could register their own "interactions" and use `InteractionManager` and it was a pretty convenient way to run a code.

Since `@react-navigation/native-stack` is now a recommended navigator - missing integration with `InteractionManager` is kind of a big blocker, especially if you already have a pretty complex logic that heavily depends on `InteractionManager`.

So this PR was mostly for simplifying a migration from `stack` to `native-stack`.

PR #11896, opened 2024-03-15
refactor: drop unmountOnBlur in favor of popToTopOnBlur

satya164:
This drops the `unmountOnBlur` option of bottom tabs and drawer in favor of a `popToTopOnBlur` option.

In many cases, the requirement is to return to the first screen of the stack nested in tabs. However using `unmountOnBlur` in those scenarios would also destroy any state of the screen. It's also slower to remount a nested navigator on tab navigation.

The `popToTopOnBlur` option provides an alternative approach - it pops the screens on a nested stack to go back to the first screen in the stack. It's much faster than remounting the nested navigator, and also doesn't destroy the state of the first screen in the stack.

It's still possible to achieve the old behavior of `unmountOnBlur` by using the `useIsFocused` hook in the screen:

```js
const isFocused = useIsFocused();

if (!isFocused) {
  return null;
}
```

### DIFF
--- a/example/src/Screens/BottomTabs.tsx
+++ b/example/src/Screens/BottomTabs.tsx
@@ -145,6 +145,7 @@ export function BottomTabs() {
           name="TabStack"
           component={SimpleStack}
           options={{
+            popToTopOnBlur: true,
             title: 'Article',
             headerShown: false,
             tabBarIcon: getTabBarIcon('file-document'),
@@ -154,7 +155,7 @@ export function BottomTabs() {
           name="TabChat"
           component={Chat}
           options={{
-            tabBarLabel: 'Chat',
+            title: 'Chat',
             tabBarIcon: getTabBarIcon('message-reply'),
             tabBarBadge: 2,
           }}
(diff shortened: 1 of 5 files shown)


evoactivity:
I use `unmountOnBlur` because my screens have a semi-transparent background (with BlurView on iOS) to allow a background video to be seen through my screen at all times. Without `unmountOnBlur` I can see my previous tabs through my current screen. Would this change preserve the behaviour of removing the blurred tabs from view?

satya164:
@evoactivity `unmountOnBlur` can easily be implemented in userland:

```js
const isFocused = useIsFocused();

if (!isFocused) {
  return null;
}
```

satya164:
@mysport12 https://reactnavigation.org/docs/7.x/upgrading-from-6.x#the-unmountonblur-option-is-removed-in-favor-of-poptotoponblur-in-bottom-tab-navigator-and-drawer-navigator

satya164:
Screens are regular components. Wrap it in another component, write an HOC, use the `layout` prop to wrap, there are so many ways to do it using the same approaches that you'd use with regular components. Not everything needs an option in the navigator.

mysport12:
No I know it’s possible to do it other ways. Just wanted to pose the question and have the quick discussion since this behavior has been embedded in the behavior of the drawer navigator since its inception and may impact numerous other apps. I’ll explore the new layout prop as I don’t want to write wrappers for every screen in my navigator. That or I’ll just use my own fork/write a quick patch to restore the functionality. Appreciate the responses and thanks for the work on the library!

PR #11937, opened 2024-04-14
fix: setting headerBackTitleVisible to false not working on iOS

zetavg:
**Motivation**

According to [the docs](https://reactnavigation.org/docs/7.x/native-stack-navigator#headerbacktitlevisible), setting `headerBackTitleVisible` to `false` should hide the back title on iOS, while this is not the case with `react-native-screens` v3.29+.

**Cause**

With the current code, `backTitle={headerBackTitleVisible ? headerBackTitle : ' '}`, the back title is set to a blank string if `headerBackTitleVisible` is false. However, in newer versions of `react-native-screens`, [the `backTitle` will be considered blank if it only contains white spaces](https://github.com/software-mansion/react-native-screens/blob/d54a19a/ios/RNSScreenStackHeaderConfig.mm#L506), making it fallback to the default title - which makes the back title not hidden at all.

**Fix**

Assign [the `backTitleVisible` property of `RNSScreenStackHeaderConfig` native component](https://github.com/software-mansion/react-native-screens/blob/d54a19a9789b799566da16a89e2cd8d8f1ad0ba7/ios/RNSScreenStackHeaderConfig.h#L42), which [seems to be a legit way to hide the back title](https://github.com/software-mansion/react-native-screens/blob/d54a19a/ios/RNSScreenStackHeaderConfig.mm#L514-L539).

**Test plan**

1. Create a native stack navigator with the [`backTitleVisible` screen option set to `false`](https://reactnavigation.org/docs/7.x/native-stack-navigator#headerbacktitlevisible).
2. See if the back title is hidden on iOS when navigating to another screen.

| Expected | Not Expected |
| --- | --- |
|  |  |

* [x] May need to test if assigning the new property will break older versions of `react-native-screens`.
  * Tested with `react-native-screens` `3.20.0` and `3.21.0`, app does not crash and the back title remains hidden-able.

### DIFF
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -182,6 +182,7 @@ export function HeaderConfig({
       backButtonInCustomView={backButtonInCustomView}
       backgroundColor={headerBackgroundColor}
       backTitle={headerBackTitleVisible ? headerBackTitle : ' '}
+      backTitleVisible={headerBackTitleVisible}
       backTitleFontFamily={backTitleFontFamily}
       backTitleFontSize={backTitleFontSize}
       blurEffect={headerBlurEffect}
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -181,7 +181,11 @@ export function HeaderConfig({
     <ScreenStackHeaderConfig
       backButtonInCustomView={backButtonInCustomView}
       backgroundColor={headerBackgroundColor}
-      backTitle={headerBackTitleVisible ? headerBackTitle : ' '}
+      backTitle={
+        headerBackTitleVisible
+          ? headerBackTitle
+          : ' ' /* For backward compatibility with react-native-screens versions <3.21.0, where `backTitleVisible` is not available */
+      }
       backTitleVisible={headerBackTitleVisible}
       backTitleFontFamily={backTitleFontFamily}
       backTitleFontSize={backTitleFontSize}


packages/native-stack/src/views/HeaderConfig.tsx
```diff
@@ -182,6 +182,7 @@ export function HeaderConfig({
       backButtonInCustomView={backButtonInCustomView}
       backgroundColor={headerBackgroundColor}
       backTitle={headerBackTitleVisible ? headerBackTitle : ' '}
```
tboba, line 184: Should we still consider `headerBackTitleVisible` in backTitle, if we're passing `backTitleVisible` below? I believe the button will be hidden anyway.

zetavg: I think it's necessary for people who are using older versions of `react-native-screens`, since `backTitleVisible` [is introduced in `react-native-screens` `3.21.0`](https://github.com/software-mansion/react-native-screens/commit/b5612bfefc27150b81d7f22c6b03a587acd44ac6#diff-28f909d0bd17e9a01ef7f575ba575e3726751335724532573627b503aa377508R42) and `@react-navigation/native-stack` has [peer dependency of `react-native-screens` as `>= 3.0.0`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/native-stack%407.0.0-alpha.20/packages/native-stack/package.json#L63).

(However, I haven't tested if only setting `backTitleVisible` will actually not work with `react-native-screens` `<3.21.0` for now.)

zetavg: I have now tested and can confirm that this is necessary for backward compatibility with `react-native-screens` versions `<3.21.0`. Without this, the back title will not be hidden while using `react-native-screens` `3.20.0`.

Comment added to explain this.

tboba: That convinces me, thanks!

PR #11943, opened 2024-04-16
feat: add support for custom screen transitions with native-stack v7

tboba:
**Motivation**

This PR intents to add implementation of custom screen transitions, recently added in react-native-screens. You can find more about this feature [here](https://github.com/software-mansion/react-native-screens/pull/1913).

**Test plan**

Go to the code of the native stack in example (NativeStack.tsx) and wrap whole navigator with <GestureDetectorProvider> (import from `react-native-screens/gesture-handler`). After that, use `goBackGesture` prop on the desired screen.

<details>
<summary>Ready to paste code 🍝</summary>

```tsx
import { Button, Text, useHeaderHeight } from '@react-navigation/elements';
import { type PathConfigMap, useTheme } from '@react-navigation/native';
import {
  createNativeStackNavigator,
  type NativeStackScreenProps,
  useAnimatedHeaderHeight,
} from '@react-navigation/native-stack';
import { Animated, Platform, ScrollView, StyleSheet, View } from 'react-native';

import { COMMON_LINKING_CONFIG } from '../constants';
import { Albums } from '../Shared/Albums';
import { Article } from '../Shared/Article';
import { NewsFeed } from '../Shared/NewsFeed';
import { GestureDetectorProvider } from 'react-native-screens/gesture-handler';

export type NativeStackParams = {
  Article: { author: string } | undefined;
  NewsFeed: { date: number };
  Albums: undefined;
};

const linking: PathConfigMap<NativeStackParams> = {
  Article: COMMON_LINKING_CONFIG.Article,
  NewsFeed: COMMON_LINKING_CONFIG.NewsFeed,
  Albums: 'albums',
};

const scrollEnabled = Platform.select({ web: true, default: false });

const ArticleScreen = ({
  navigation,
  route,
}: NativeStackScreenProps<NativeStackParams, 'Article'>) => {
  return (
    <View>
      <ScrollView contentInsetAdjustmentBehavior="automatic">
        <View style={styles.buttons}>
          <Button
            variant="filled"
            onPress={() => navigation.push('NewsFeed', { date: Date.now() })}
          >
            Push feed
          </Button>
          <Button
            variant="filled"
            onPress={() => navigation.replace('NewsFeed', { date: Date.now() })}
          >
            Replace with feed
          </Button>
          <Button variant="filled" onPress={() => navigation.popTo('Albums')}>
            Pop to Albums
          </Button>
          <Button variant="tinted" onPress={() => navigation.pop()}>
            Pop screen
          </Button>
        </View>
        <Article
          author={{ name: route.params?.author ?? 'Unknown' }}
          scrollEnabled={scrollEnabled}
        />
      </ScrollView>
      <HeaderHeightView />
    </View>
  );
};

const NewsFeedScreen = ({
  route,
  navigation,
}: NativeStackScreenProps<NativeStackParams, 'NewsFeed'>) => {
  React.useLayoutEffect(() => {
    navigation.setOptions({
      headerSearchBarOptions: {
        placeholder: 'Search',
      },
    });
  }, [navigation]);

  return (
    <View>
      <ScrollView contentInsetAdjustmentBehavior="automatic">
        <View style={styles.buttons}>
          <Button variant="filled" onPress={() => navigation.push('Albums')}>
            Push Albums
          </Button>
          <Button variant="tinted" onPress={() => navigation.goBack()}>
            Go back
          </Button>
        </View>
        <NewsFeed scrollEnabled={scrollEnabled} date={route.params.date} />
      </ScrollView>
      <HeaderHeightView />
    </View>
  );
};

const AlbumsScreen = ({
  navigation,
}: NativeStackScreenProps<NativeStackParams, 'Albums'>) => {
  const headerHeight = useHeaderHeight();

  return (
    <View>
      <ScrollView contentContainerStyle={{ paddingTop: headerHeight }}>
        <View style={styles.buttons}>
          <Button
            variant="filled"
            onPress={() =>
              navigation.navigate('Article', { author: 'Babel fish' })
            }
          >
            Navigate to article
          </Button>
          <Button variant="tinted" onPress={() => navigation.pop(2)}>
            Pop by 2
          </Button>
        </View>
        <Albums scrollEnabled={scrollEnabled} />
      </ScrollView>
      <HeaderHeightView hasOffset />
    </View>
  );
};

const HeaderHeightView = ({
  hasOffset = Platform.OS === 'ios',
}: {
  hasOffset?: boolean;
}) => {
  const { colors } = useTheme();

  const animatedHeaderHeight = useAnimatedHeaderHeight();
  const headerHeight = useHeaderHeight();

  return (
    <Animated.View
      style={[
        styles.headerHeight,
        {
          backgroundColor: colors.card,
          borderColor: colors.border,
          shadowColor: colors.border,
        },
        hasOffset && {
          transform: [{ translateY: animatedHeaderHeight }],
        },
      ]}
    >
      <Text>{headerHeight.toFixed(2)}</Text>
    </Animated.View>
  );
};

const Stack = createNativeStackNavigator<NativeStackParams>();

export function NativeStack() {
  const { colors } = useTheme();

  return (
    <GestureDetectorProvider>
      <Stack.Navigator>
        <Stack.Screen
          name="Article"
          component={ArticleScreen}
          options={({ route }) => ({
            title: `Article by ${route.params?.author ?? 'Unknown'}`,
            headerLargeTitle: true,
            headerLargeTitleShadowVisible: false,
          })}
          initialParams={{ author: 'Gandalf' }}
        />
        <Stack.Screen
          name="NewsFeed"
          component={NewsFeedScreen}
          options={{
            title: 'Feed',
            fullScreenGestureEnabled: true,
            goBackGesture: 'verticalSwipe',
          }}
        />
        <Stack.Screen
          name="Albums"
          component={AlbumsScreen}
          options={{
            title: 'Albums',
            presentation: 'modal',
            headerTransparent: true,
            headerBlurEffect: 'light',
            headerStyle: {
              // Add a background color since Android doesn't support blur effect
              backgroundColor: colors.card,
            },
          }}
        />
      </Stack.Navigator>
    </GestureDetectorProvider>
  );
}

NativeStack.title = 'Native Stack';
NativeStack.linking = linking;
NativeStack.options = {
  gestureEnabled: false,
};

  buttons: {
    flexDirection: 'row',
    flexWrap: 'wrap',
    gap: 12,
    padding: 12,
  },
  headerHeight: {
    position: 'absolute',
    top: 0,
    right: 0,
    padding: 12,
    borderWidth: StyleSheet.hairlineWidth,
    borderRightWidth: 0,
    borderTopWidth: 0,
    borderBottomLeftRadius: 3,
  },
});
```

</details>

**Presentation**

https://github.com/react-navigation/react-navigation/assets/23281839/cae36411-2715-4d0a-ab30-95db370fdc45


tboba:
Looks like CIs are failing 🤔 lemme fix that quickly

piaskowyk:
@satya164 any chance to review it? 🙏

hanwenbo:
When to merge?

packages/native-stack/src/types.tsx
```diff
@@ -452,6 +454,43 @@ export type NativeStackNavigationOptions = {
...
+   * - ScreenTransition.SwipeDown
+   * - ScreenTransition.Horizontal
+   * - ScreenTransition.Vertical
+   * - ScreenTransition.TwoDimensional
+   * - ScreenTransition.SwipeRightFade
+   *
+   * You can also construct custom screen transition by providing an object with `topScreenFrame` and `belowTopScreenFrame` properties.
+   * Only supported on iOS and Android.
+   */
+  transitionAnimation?: AnimatedScreenTransition;
```
satya164, line 489: Similar comment as above, for animation related options, we use the `animation` prefix. So it'd be better to match the convention.

I'm also curious how this interacts with the `animation` option. Does it override it?

packages/native-stack/src/types.tsx
```diff
@@ -452,6 +454,43 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * Supported values:
+   * - ScreenTransition.SwipeRight
+   * - ScreenTransition.SwipeLeft
+   * - ScreenTransition.SwipeUp
+   * - ScreenTransition.SwipeDown
+   * - ScreenTransition.Horizontal
+   * - ScreenTransition.Vertical
+   * - ScreenTransition.TwoDimensional
+   * - ScreenTransition.SwipeRightFade
```
satya164, line 484: Have you considered a string union type instead? So we don't need an additional import.

packages/native-stack/src/types.tsx
```diff
@@ -452,6 +454,43 @@ export type NativeStackNavigationOptions = {
...
+   * - "verticalSwipe": swipe up or down to dismiss the screen
+   * - "horizontalSwipe": swipe left or right to dismiss the screen
+   * - "twoDimensionalSwipe": swipe in any direction to dismiss the screen
+   *
+   * Only supported on iOS and Android.
+   */
+  goBackGesture?: GoBackGesture;
+  /**
+   * Changes the animation for dismissing the screen when making a swipe gesture.
+   * You can choose one of the built-in transition presets, by using `ScreenTransition` from `react-native-reanimated`.
```
satya164, line 474: Probably should re-export these from `/native-stack`. In the future if we implement transition animations on the Web, we can make the same code work on the Web. And generally cleaner not to leak the abstraction.

packages/native-stack/src/types.tsx
```diff
@@ -452,6 +454,43 @@ export type NativeStackNavigationOptions = {
...
+   * - ScreenTransition.SwipeRightFade
+   *
+   * You can also construct custom screen transition by providing an object with `topScreenFrame` and `belowTopScreenFrame` properties.
+   * Only supported on iOS and Android.
+   */
+  transitionAnimation?: AnimatedScreenTransition;
+  /**
+   * Whether the screen should allow for making a swipe gesture from screen edges. Defaults to `true`.
+   */
+  screenEdgeGesture?: boolean;
```
satya164, line 493: I think something like `screenEdgeGestureEnabled` would be better following the convention for boolean options that we normally use.

Or maybe `gestureFromEdgeEnabled` to also follow the convention of starting gesture related options with `gesture` prefix.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -200,6 +206,17 @@ const SceneView = ({
     freezeOnBlur,
   } = options;
 
+  const screenRef = React.useRef(null);
+  React.useEffect(() => {
```
satya164, line 210: Layout effect would be better since ref is assigned earlier.

```suggestion
  React.useLayoutEffect(() => {
```

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -513,95 +531,122 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
 
   const { colors } = useTheme();
 
+  const currentRouteKey = state.routes[state.index].key;
+  const topScreenOptions = descriptors[currentRouteKey].options;
+  const gestureDetectorBridge = React.useRef<GestureDetectorBridge>({
+    stackUseEffectCallback: (_stackRef) => {
+      // this method will be overriden in GestureDetector
+    },
+  });
```
satya164, line 540: is it necessary to specify object with empty method here instead of `null`?

```suggestion
  const gestureDetectorBridge = React.useRef<GestureDetectorBridge>({
    stackUseEffectCallback: (_stackRef) => {
      // this method will be overriden in GestureDetector
    },
  });
```

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -513,95 +531,122 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
...
+  type RefHolder = Record<
+    string,
+    React.MutableRefObject<React.Ref<NativeStackNavigatorProps>>
+  >;
```
satya164, line 544: Normally we define types at the top level of the file. Though seems this type is also defined in `types.tsx` and could be imported from there.

satya164 (changes requested):

packages/native-stack/src/types.tsx
```diff
@@ -452,6 +454,43 @@ export type NativeStackNavigationOptions = {
...
+   * - "swipeLeft": swipe left to dismiss the screen
+   * - "swipeUp": swipe up to dismiss the screen
+   * - "swipeDown": swipe down to dismiss the screen
+   * - "verticalSwipe": swipe up or down to dismiss the screen
+   * - "horizontalSwipe": swipe left or right to dismiss the screen
+   * - "twoDimensionalSwipe": swipe in any direction to dismiss the screen
+   *
+   * Only supported on iOS and Android.
+   */
+  goBackGesture?: GoBackGesture;
```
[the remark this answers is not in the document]
tboba: I'm curious how we should name it then. `gestureToTraverseBack`? `gestureToPreviousPath`? @satya164 @piaskowyk do you have any ideas?

packages/native-stack/src/types.tsx
tboba: This prop does not override it, but rather it chooses the animation while doing the gesture to go back, while animation is only for classic transitioning - by clicking on the element (or performing navigation actions) that navigates to given screen.

packages/native-stack/src/types.tsx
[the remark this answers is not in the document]
satya164: @tboba how about `gestureType`? since it specifies the type of gesture to use. or something similar.

I think `TraverseBack`, `PreviousPath` etc. aren't necessary since that's what all of the gesture related options are for, so it'd be redundant.

packages/native-stack/src/types.tsx
satya164: Maybe `animationForGesture`. The `Back` seems redundant since all gesture related options are for back gesture.

packages/native-stack/src/types.tsx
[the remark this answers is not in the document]
satya164: > @tboba What about gestureToGoBack / gestureBack

But all of the gesture related props are for going back as I mentioned in both of my replies. Is there a gesture related prop that's not for going back? If we're not using the "back" terminology anywhere else and it's not confusing there, why would it be confusing here?

> Because you know, gestureType for what

You can ask the same question for all of the other `gestureX` options.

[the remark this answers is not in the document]
satya164: Another name could be `gestureDirection` since all of these props seem to be about direction of the gesture (and we already have such a prop on JS stack):

```ts
type GestureDirection = 'left' | 'right' | 'up' | 'down' | 'horizontal' | 'vertical' | 'any';
```

packages/native-stack/src/types.tsx
tboba: It may be problematic for now, since we're only [exporting](https://github.com/software-mansion/react-native-screens/blob/546e72f56359eacade1eb35700dcb19aa8434f66/src/native-stack/index.tsx#L23) some of the types in `native-stack` package, so changing this export could be a blocker, potentially.
Also, I wouldn't re-export this type from reanimated (obviously we'd make it a strict dependency).

packages/native-stack/src/types.tsx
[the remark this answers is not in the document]
tboba: Sure, you've convinced me 👍 However, I'm afraid about the inconsistency of passed values for both `gestureDirection` on JS stack and `goBackGesture` (for now) on native-stack, since they both could accept different props, even after making that prop accept strings, as of the idea you've mentioned below.

packages/native-stack/src/types.tsx
tboba: As above, this could be a problem here aswell, since `ScreenTransition` object isn't exported from react-native-screens. Should we wait for the release that exports that object then?

satya164: Missed that this was from reanimated. Then it seems fine.

packages/native-stack/src/types.tsx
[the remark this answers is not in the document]
satya164: @tboba v7 of stack is not stable yet, so there's still a chance to align it

Tho I'm also curious, what happens to `left` and `right` in RTL mode? are they flipped?

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: It must stay with defined object, since we're referring it directly in ScreenGestureDetector/ScreenStack: https://github.com/software-mansion/react-native-screens/blob/546e72f56359eacade1eb35700dcb19aa8434f66/src/gesture-handler/ScreenGestureDetector.tsx#L63.

satya164: @tboba I see, it's a bit strange.

That code also looks problematic. Refs shouldn't be assigned during render https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs

Though that's out of scope of this PR.

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: Done in https://github.com/react-navigation/react-navigation/pull/11943/commits/301cd8dace25c46ed8b0f5931333176b07dcc99c.

packages/native-stack/src/views/NativeStackView.native.tsx
tboba: Done in https://github.com/react-navigation/react-navigation/pull/11943/commits/301cd8dace25c46ed8b0f5931333176b07dcc99c.

packages/native-stack/src/types.tsx
tboba: Done in https://github.com/react-navigation/react-navigation/pull/11943/commits/301cd8dace25c46ed8b0f5931333176b07dcc99c.

packages/native-stack/src/types.tsx
tboba: Done in https://github.com/react-navigation/react-navigation/pull/11943/commits/301cd8dace25c46ed8b0f5931333176b07dcc99c.

packages/native-stack/src/types.tsx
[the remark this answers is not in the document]
tboba: @satya164 for now we can't align accepted values to these given from JS stack, since this type is taken from reanimated to avoid strong dependency. This would require us to wait for a pretty long chain from releasing a new version of reanimated (new values would be also a breaking change there) to have that version accepted on Expo Go in new release. 

Also, the gesture on RTL mode is not reverted, but we can fix this out of this PR.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -201,6 +207,19 @@ const SceneView = ({
     freezeOnBlur,
   } = options;
 
+  const Screen = React.useContext(ScreenContext);
```
tboba, line 210: This line has been added here, because functional components cannot have ref assigned. That's why we're doing a pretty tricky thing here and we're using a `ScreenContext,` which uses `InnerScreen` by default, where `InnerScreen` uses forwardRef under the hood.

tboba:
@satya164 I've processed all suggestions! Please let me know if you have still any questions.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -525,99 +544,132 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
...
   );
 }
 
+const ScreenWithRef = React.forwardRef<InnerScreen & View, ScreenProps>(
+  (props, ref) => {
+    const ScreenWrapper = React.useContext(ScreenContext) || InnerScreen;
+
+    return <ScreenWrapper {...props} ref={ref} />;
+  }
+);
```
satya164, line 671: We shouldn't be duplicating this logic into native stack code. Adding `React.forwardRef` here should be enough: https://github.com/software-mansion/react-native-screens/blob/c2ac40f5b3b1cc1c152126a1dd1c42b6834b8f74/src/components/Screen.tsx#L207

Exposing `InnerScreen` and requiring consumer code to use the context because of the logic inside screens leaks internal implementation details - which makes harder for you to change how things work in the future. It also makes the code for consumers much more verbose (e.g. a lot of code here instead of one line change in `react-native-screens`).

satya164 (changes requested):

maciekstosio:
Closing in favour of: https://github.com/react-navigation/react-navigation/pull/12204

PR #11976, opened 2024-05-05
fix: remove global in favor of globalThis for better compatibility

natew:

**Motivation**

I'm using react-navigation with Vite (browser-loaded esm) and this is actually really tricky to fix because imports are always hoisted above code, so my polyfill attempts get pushed below the react-navigation import unless I literally add a script tag above the app entrypoint. 

Moving to globalThis should be safe and supported in most node/react-native/browsers from the last few years.

**Test plan**

I haven't tested thoroughly.


satya164 (approved):
Thanks for the PR

PR #11983, opened 2024-05-09
Bugs + (partial) Fixes: Non-responsiveness of "tabBarLabelStyle" and "tabBarLabel" in MaterialTopTabBar

DrStoop:
**Motivation**

Bug fix: The props tabBarLabelStyle and tabBarLabel are ignored by MaterialTopTabBar. This pull requests solves the problem full for tabBarLabelStyle. For tabBarLabel this PR only fixes the problem for tabBarLabel(s) of type string.

**Test plan**

Create a MaterialTopTabBar navigator with customized tab bar label style(s) (prop: "tabBarLabelStyle") which is currently ignored. After merging the commit, this should not be the case anymore. The same applies to the "tabBarLabel".

### DIFF
--- a/packages/react-native-tab-view/src/TabBar.tsx
+++ b/packages/react-native-tab-view/src/TabBar.tsx
@@ -525,7 +525,6 @@ export function TabBar<T extends Route>({
         : undefined;
 
       const props = {
-        ...rest,
         key: route.key,
         position,
         route,
@@ -545,6 +544,7 @@ export function TabBar<T extends Route>({
         style: tabStyle,
         defaultTabWidth,
         android_ripple,
+        ...rest,
       } satisfies TabBarItemProps<T> & { key: string };
 
       return (
--- a/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
+++ b/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
@@ -76,7 +76,9 @@ export function MaterialTopTabBar({
           labelText:
             options.tabBarShowLabel === false
               ? undefined
-              : options.title !== undefined
+              : options.tabBarLabel !== undefined && typeof options.tabBarLabel === "string"
+                ? options.tabBarLabel
+                : options.title !== undefined
                 ? options.title
                 : route.name,
         },
--- a/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
+++ b/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
@@ -16,19 +16,19 @@ import type { MaterialTopTabBarProps } from '../types';
 type MaterialLabelType = {
   color: string;
   label?: string;
-  labelStyle?: StyleProp<ViewStyle>;
+  style?: StyleProp<ViewStyle>;
   allowScaling?: boolean;
 };
 
 const MaterialLabel = ({
   color,
   label,
-  labelStyle,
+  style,
   allowScaling,
 }: MaterialLabelType) => {
   return (
     <Text
-      style={[{ color }, styles.label, labelStyle]}
+      style={[{ color }, styles.label, style]}
       allowFontScaling={allowScaling}
     >
       {label}
--- a/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
+++ b/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx
@@ -76,11 +76,12 @@ export function MaterialTopTabBar({
           labelText:
             options.tabBarShowLabel === false
               ? undefined
-              : options.tabBarLabel !== undefined && typeof options.tabBarLabel === "string"
+              : options.tabBarLabel !== undefined &&
+                  typeof options.tabBarLabel === 'string'
                 ? options.tabBarLabel
                 : options.title !== undefined
-                ? options.title
-                : route.name,
+                  ? options.title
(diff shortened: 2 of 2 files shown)


martin-desimone:
we need this.

packages/react-native-tab-view/src/TabBar.tsx
```diff
@@ -545,6 +544,7 @@ export function TabBar<T extends Route>({
         style: tabStyle,
         defaultTabWidth,
         android_ripple,
+        ...rest,
```
satya164, line 547: This should stay at the top. It shouldn't be possible to override the other properties due to any options that user may specify as it could break the component.

If there are specific properties that need to be at the end, then those properties should be explicitly specified, though there's probably a different issue if the other properties specified here aren't correct.

satya164 (changes requested):
Thanks for the PR. Can you link to an existing issue? If not, please share a repro and if possible, before/after screenshots/videos.

DrStoop:
Hi @satya164 , 
sorry for the delay. I have not found an issue to link to, possible because I am using version 7.0.0-rc.2, as the latest version 6.6.13 confronted me with other issues elsewhere. But the 7.0.0-rc.2 is working fine despite the fixes I made. So maybe this is a fix before the bug actually happens...
As requested the [simply example repository with BEFORE- & AFTER-bugfix-commits](https://github.com/DrStoop/MaterialTopTabBarFix/tree/main). It's an Expo app. Unfortunately, I'm a javascript dummy so I could not manage to link a local material-top-tabs (yarn)-package to my npm project. As a workaround I simply added the `node_modules/@react-navigation/` to this repository. Hope that works for you.

## BEFORE bugfix
Before

- The [`Home`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L35) tab label is supposed to be customized with [`tabBarLabel`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L40) set to string "Custom Home" and the font size should be increased with [`tabBarLabelStyle`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L42) to 20, but none of that happens
- The [`Settings`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L45) default <Text>-component label is supposed to be customized with the [`TabBarLabel`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L6) component provided to [`tabBarLabel`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/6928afce7cd1a0b8e6dda6cac2306d54ac391a9f/App.tsx#L50), but it's non-responsive.

## AFTER bugfix
After_1
After_2

- To customize the label with a string a new prop [tabBarLabelText](https://github.com/DrStoop/MaterialTopTabBarFix/blob/0491d1a0eb33ae2f71257438c73e7f820eb36ec4/node_modules/%40react-navigation/material-top-tabs/src/types.tsx#L97) is required as [`TabDescriptor`](https://github.com/react-navigation/react-navigation/blob/c840aaeeccc8b5fb4ebc317e3f658bc76e64a7ea/packages/react-native-tab-view/src/types.tsx#L4) includes a purely component typed [`label`](https://github.com/react-navigation/react-navigation/blob/c840aaeeccc8b5fb4ebc317e3f658bc76e64a7ea/packages/react-native-tab-view/src/types.tsx#L11) corresponding to [`tabBarLabel`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/0491d1a0eb33ae2f71257438c73e7f820eb36ec4/node_modules/%40react-navigation/material-top-tabs/src/types.tsx#L104) and a [`labelText`](https://github.com/react-navigation/react-navigation/blob/c840aaeeccc8b5fb4ebc317e3f658bc76e64a7ea/packages/react-native-tab-view/src/types.tsx#L8) corresponding to the new [`tabBarLabelText`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/0491d1a0eb33ae2f71257438c73e7f820eb36ec4/node_modules/%40react-navigation/material-top-tabs/src/types.tsx#L97).
- With above changes in the material-top-tabs [`types.tsx`](https://github.com/DrStoop/MaterialTopTabBarFix/blob/0491d1a0eb33ae2f71257438c73e7f820eb36ec4/node_modules/%40react-navigation/material-top-tabs/src/types.tsx) and the redirecting within [MaterialTopTabBar.tsx](https://github.com/DrStoop/MaterialTopTabBarFix/blob/0491d1a0eb33ae2f71257438c73e7f820eb36ec4/node_modules/%40react-navigation/material-top-tabs/src/views/MaterialTopTabBar.tsx), the prop `tabBarLabelText` customizes the label text, `tabBarLabelStyle` customizes the label text style and `tabBarLabel` replaces the entire default label `<Text>`-component, as seen in the screenshots above.

satya164:
Thanks @DrStoop. Since this is no longer an issue, I'm closing the PR.

Issue #11606, opened 2023-09-19
App crashing in production in some devices when hardware back button is pressed in android


androidappteam:
### Current behavior

App crashing in production in some devices when hardware back button is pressed in android

**Crash Log from crashlytics** 

Fatal Exception: t4.c: TypeError: Cannot read property 'key' of undefined, js engine: hermes, stack:
getStateForAction@1:2324604
canGoBack@1:2384972
canGoBack@1:2385008
anonymous@1:2316781
anonymous@1:2382577
anonymous@1:2382518
anonymous@1:2382518
anonymous@1:2316742
anonymous@1:2400742
anonymous@1:328049
emit@1:128838
__callFunction@1:134928
anonymous@1:133434
__guard@1:134374
callFunctionReturnFlushedQueue@1:133392

       at com.facebook.react.modules.core.ExceptionsManagerModule.reportException(ExceptionsManagerModule.java:72)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
       at com.facebook.jni.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:984)
       at android.os.Handler.dispatchMessage(Handler.java:104)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
       at android.os.Looper.loopOnce(Looper.java:238)
       at android.os.Looper.loop(Looper.java:357)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:228)
       at java.lang.Thread.run(Thread.java:1012)

**react-native info**
System:
    OS: macOS 12.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 121.75 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.16.0 - /var/folders/c_/kxvkl4wd69z1xdflf_ncr0hjj_bqf_/T/yarn--1695121660525-0.2994264769809307/node
[...]

PR #11990, opened 2024-05-16
fix: Invalid Property Access on non existent entry

braandl:
There is an possibly invalid array access to the state.history in TabRouter.txt (see https://github.com/react-navigation/react-navigation/issues/11606)

In order to avoid seldom crashes on App closing via Hardware Buttons, I did add a nullcheck before accessing the property `key` on the array lookup result.


satya164 (approved):
Thank you

PR #12019, opened 2024-06-13
Remove 1.5x back-icon.png images. Ensures the back icon renders in iOS production builds, fixes #11792

### LINKED ISSUE #11792: Back button image not showing in Stack navigator (although button is there and pressable) - iOS and Android
### Current behavior

I'm having an issue where the arrow image of the back button does not show on Android (Expo Go - simulator and Device), despite the button being there, and being pressable. 

The image shows on iOS, and it's only an issue on a nested navigator, not the Root stack navigator
```
AppNav (Stack) -- Back button visible
   Drawer Nav (Drawer)
       User Settings Screen
       Project Navigator (Stack)   
         Tab navigator
             Screen 1
             Screen 2
         Stack Screen 1 -- Back button there, but not visible
```

* Because the button is visible on the Root stack I don't think this is an issue with the Assets imports / loading
*  It also doesn't' seem to be an issue with the header styles, since if I move `Stack screen 1` to the root stack, the back button starts showing.
* Inspecting the elements shows correctly the touchables, and the correct elements. The element is clickable, and does go back if pressed. It's just the arrow that is not visible. 
* I have removed all header styles, and all other headers are `headerShown: false`.  Nothing helped
* Because the button is there and is showing, it is not a problem of the stack being empty (with nothing to go back to)
* Adding a `headerBackImage` bypasses the issue

Screenshot 2024-01-18 at 17 30 04

iOs (left), Android (right)

NOTE: Cannot reproduce in Expo snack because of Reanimated issues

### Expected behavior

Back button showing.

### Reproduction

https://snack.expo.dev/@smaisidoro/talking-violet-coffee

### Platform

- [X] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [X] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

Expo Go

```
    "@react-navigation/bottom-tabs": "^6.5.11",
    "@react-navigation/drawer": "^6.6.6",
    "@react-navigation/native": "~6.1.9",
    "@react-navigation/native-stack": "^6.9.17",
    "@react-navigation/stack": "~6.3.20",
```

Similar issues found:
https://github.com/react-navigation/react-navigation/issues/8609
https://github.com/react-navigation/react-navigation/issues/11657

Worked around the issue by providing a custom back button image as a `headerBackImage` component

sparkertime:

**Motivation**

I've been seeing the issue described by https://github.com/react-navigation/react-navigation/issues/11792 where the back button is not shown on iOS. It's a frustrating problem since the issue – back buttons on a Stack Navigator not showing on iOS – only happens in production builds.

**Test plan**

I ported the React Native code for `renderBackImage` in `HeaderBackButton` along with all assets and used that as a starting point to troubleshoot the issue. Removing the `1.5x` variants fixes the problem for me. I'm enough of a React Native dunce that I don't totally understand why those two files cause issues, but removing those files seems fine since that scale doesn't seem to be supported anyway according to [the Image docs](https://reactnative.dev/docs/images).

> Note - it could also be the case that the `1x` variants are superfluous and should be removed, but I'll leave that to the maintainers to decide

I'd submit a test plan but as the others have reported in the issue linked above, this is a production-only bug for me.


satya164:
Thanks for the PR. Isn't this a bug in Expo's bundler that should be fixed? Have you opened a bug report there?

sparkertime:
> Thanks for the PR. Isn't this a bug in Expo's bundler that should be fixed? Have you opened a bug report there?

As I mentioned, I'm a React Native dunce so I have no idea and it's a bit beyond my capabilities to find out. Feel free to close this if this is the wrong solution and thanks for the quick response!

satya164:
@sparkertime I'll merge the PR, however I'd also request to open an issue in Expo regarding this.

satya164 (approved):
Thanks

PR #12030, opened 2024-06-21
fix: center align TabBarIndicator when width is specified

JAE-GYU:

**Motivation**

When width is set for the TabBarIndicator, it does not align correctly.
before

By allowing the getTranslateX function to accept a width parameter, it will now adjust the calculation of the outputRange to ensure the TabBarIndicator is properly centered when width is provided as a number.

```
<TabBar
renderIndicator={(props) => <TabBarIndicator {...props} width={16} />}
....
/>
```

after

**Test plan**

1. Apply a fixed width to the TabBarIndicator.
2. Verify that the TabBarIndicator is centered within its container.


satya164 (approved):
Thanks for the PR

BugfreeYeti:
May I ask, what property should I set in <Tab.Navigator tabBarIndicatorStyle={{width:16}}> ... <Navigator /> to center the custom tabBarIndicator? I look forward to your reply. Thank you!

PR #11516, opened 2023-08-01
feat(iOS,v6): support for form sheet medium detent

kkafar:
**Motivation**

WIP WIP

TODO:

* [ ] Consult prop naming & API
* [ ] Prepare PR for v7
* [ ] Discuss upgrade of `react-native-screens` to `3.23.0` across repo

If this pull request addresses an existing issue, link to the issue. If an issue is not present, describe the issue here.

**Test plan**

WIP WIP

The change must pass lint, typescript and tests.

PR #12032, opened 2024-06-24
feat: support for Android sheets & iOS custom detents

kkafar:
## Description 

This PR introduces support for `formSheet` presentation on Android and brings support for custom detents on iOS. 

I've ensured that this PR is backward compatible, i. g. I've brought back support for `medium`, `large` and `all` options, which were supposed to be dropped.

Please note this this changes will work with react-native-screens 4.0.0 onwards. 

There is plan to land this feature on v6 line (will require react-native-screens 3.35.0 or 3.36.0 or newer), however it is deferred for now. 

Associated PR:

* https://github.com/react-navigation/react-navigation/pull/11516

## Disclaimers

The extended functionalities of `formSheet` stack presentation do come with its limitations.

* Android supports only up to three possible detent values. 
* `fitToContents` detent level can not be yet passed together with other values (this is work in progress and if sensible API is feasible we'll release it later).
* there are limitations on flex layout inside the sheet's content on iOS. To mitigate other, more staggering issues, we do not notify React's layout system on sheet height's changes. This causes e.g. to cause issues when trying to center elements vertically depending on current sheet height. If that's your use case - form sheet is not eligible yet. 

If you happen to notice any issue related to the formsheet API consider forwarding it to the [`react-native-screens` repository](https://github.com/software-mansion/react-native-screens). Thanks!

**Test plan**

Since this change involves native-stack solely I've tested this on Example & FabricExample apps in [`react-native-screens` repository](https://github.com/software-mansion/react-native-screens/tree/main/Example) (e.g. `Test1649` can be used).


packages/native-stack/src/types.tsx
```diff
@@ -609,6 +631,16 @@ export type NativeStackNavigationOptions = {
...
+  /**
+   * Footer component that can be used alongside form sheet stack presentation style.
+   *
+   * This option is provided, because due to implementation details it might be problematic
+   * to implement such layout with JS-only code.
+   *
+   * Please note that this prop is marked as unstable and might be subject of breaking changes,
+   * including removal.
+   */
+  unstable_footerComponent?: React.ReactNode;
```
satya164, line 643: I wouldn't call it `footerComponent` coz you're not accepting a React Component (`MyComponent` - component, `<MyComponent />` - element).

Though also in React Navigation, generally we accept render callbacks:

```js
unstable_footer?: () => React.ReactNode
```

This way we can pass additional information if we need in future. So maybe worth aligning it like that.

Maclay74:
What's need to be done to merge it? This UX patter became very popular and we need to keep up.
Anything besides lint and tests?

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -488,38 +496,49 @@ const SceneView = ({
...
+                    style={headerTransparent ? styles.absolute : null}
+                  >
+                    {header({
+                      back: headerBack,
+                      options,
+                      route,
+                      navigation,
+                    })}
+                  </View>
+                ) : null}
```
kkafar, line 516: I don't know why this code was moved down in my PR, most likely some merge artifact. Let me know what should I do here in review please.

packages/native-stack/src/types.tsx
```diff
@@ -37,6 +36,15 @@ export type NativeStackNavigationEventMap = {
...
   gestureCancel: { data: undefined };
+  /**
+   * Event which fires when screen is in sheet presentation & it's detent changes.
+   *
+   * In payload it caries two fields:
+   *
+   * * index - current detent index in the `sheetAllowedDetents` array,
+   * * isStable - on Android `false` value means that the user is dragging the sheet or it is settling; on iOS it is always `true`.
+   */
+  sheetDetentChange: { data: { index: number; isStable: boolean } };
```
kkafar, line 47: This event is fired not so often, thus it's okay to not use reanimated for it and leave it be as navigation event.

packages/native-stack/src/types.tsx
```diff
@@ -439,6 +447,13 @@ export type NativeStackNavigationOptions = {
    * Style object for the scene content.
    */
   contentStyle?: StyleProp<ViewStyle>;
+  /**
+   * Style object for the screen native component. This might help to workaround
+   * some issues when using `formSheet` presentation.
+   *
+   * Only `backgroundColor` is accepted.
+   */
+  unstable_screenStyle?: ScreenProps['unstable_screenStyle'];
```
kkafar, line 456: This is needed right now to workaround "truncated formsheet" issue on iOS, where due to asynchronous nature of react's layout we do not trigger the layout when sheet height changes, causing view height to not be updated and effectively truncated in case its height is less than sheet height.

To workaround this, the user can set background color of the screen component itself. If I recall correctly it can not be set through `contentStyle`, cause the view the `contentStyle` is set on is also truncated. 

> [!note]
It might be worth considering to just hijack the value of `contentStyle.backgroundColor` and pass it to Screen component in case formSheet presentation is in use. Idk whether that is feasible, but if so, that might be the best solution.

packages/native-stack/src/types.tsx
```diff
@@ -532,17 +547,32 @@ export type NativeStackNavigationOptions = {
...
+   * There are also legacy & **deprecated** options available:
+   *
+   * * 'medium' - corresponds to `[0.5]` detent value, around half of the screen height,
+   * * 'large' - corresponds to `[1.0]` detent value, maximum height,
+   * * 'all' - corresponds to `[0.5, 1.0]` value, the name is deceiving due to compatibility reasons.
+   *
+   * Defaults to `[1.0]` literal.
    */
-  sheetAllowedDetents?: SheetDetentTypes;
+  sheetAllowedDetents?: number[] | 'fitToContents' | 'medium' | 'large' | 'all';
```
kkafar, line 565: Few points to notice here:

1. I've brought back the old-style-API values support, these now are resolved internally by `react-native-screens` (whole compatibility logic is hidden from native-stack code). This was done to mitigate the breaking changes. I've marked them as deprecated and have intention of removing them in the closer unspecified future.
2. I've moved from using `ScreenProps['sheetAllowedDetents']` etc. for a few props, to just manually specifying the type. I guess using `ScreenProps['propertyName']` has its own advantages, but I was frustrated with typescript errors throughout the native-stack code, since it does not see the changes from unpublished screens versions. **Let me know whether thats fine or not** in review, please.

packages/native-stack/src/types.tsx
```diff
@@ -609,6 +642,17 @@ export type NativeStackNavigationOptions = {
...
+   * Footer component that can be used alongside formSheet stack presentation style.
+   *
+   * This option is provided, because due to implementation details it might be problematic
+   * to implement such layout with JS-only code.
+   *
+   * Please note that this prop is marked as unstable and might be subject of breaking changes,
+   * including removal, in particular when we find solution that will make implementing it with JS
+   * straightforward.
+   */
+  unstable_sheetFooter?: () => React.ReactNode;
```
kkafar, line 655: Went for your suggestion here. Thanks.

packages/native-stack/src/views/DebugContainer.native.tsx
```diff
@@ -15,22 +18,26 @@ type ContainerProps = ViewProps & {
  * for detailed explanation.
  */
 export let DebugContainer = (props: ContainerProps) => {
-  return <View {...props} collapsable={false} />;
+  return <ScreenContentWrapper {...props} />;
```
kkafar, line 21: It is crucial for formsheet presentation to work with native-stack that the children of the Screen component are wrapped inside `ScreenContentWrapper`. This, and another change few lines below assures this.

packages/native-stack/src/views/FooterComponent.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ScreenFooter } from 'react-native-screens';
+
+type FooterProps = {
+  children?: React.ReactNode;
+};
+
+export default function FooterComponent({ children }: FooterProps) {
+  return <ScreenFooter collapsable={false}>{children}</ScreenFooter>;
+}
```
kkafar, line 10: I don't want this component to be publicly exported from react navigation. I guess it's fine to leave it as it is, right? If someone really wants, he could import it from this file directly but it won't be accessible from the `native-stack` module.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -92,7 +98,11 @@ const MaybeNestedStack = ({
   const content = (
     <DebugContainer
       style={[
-        styles.container,
+        presentation === 'formSheet'
+          ? Platform.OS === 'ios'
+            ? styles.absolute
+            : null
+          : styles.container,
```
kkafar, line 105: This is crucial for sheets to layout correctly on iOS. Just noting it.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -190,19 +204,31 @@ const SceneView = ({
...
 
+  // This is workaround for... find the commit here
+  // We want to allow only backgroundColor setting for now
+  unstable_screenStyle = unstable_screenStyle
+    ? { backgroundColor: unstable_screenStyle.backgroundColor }
+    : null;
+
+  if (sheetAllowedDetents === 'fitToContents') {
+    sheetAllowedDetents = [-1];
+  }
```
kkafar, line 230: This won't be necessary anymore - it will be now handled internally within react-native-screens once https://github.com/software-mansion/react-native-screens/pull/2356 is merged.

kkafar (commented):
Notes for reviewers.

kkafar:
I still have fixing the CI left, on it right now

kkafar:
@satya164, the state is as follows:

1. there is unsolved problem of mocking `ScreenContentWrapper` for jest unit tests,
2. linters do fail, because this PR still uses react-native-screens version for which these components such as `ScreenFooter`, `ScreenComponentWrapper` or props, e.g. `unstable_screenStyle` do not exist. Should I bump screens version in this PR? 
3. When porting this to v6 I think I should add some kind of future discovery, like it is already done e.g. in regular `stack` when there is detected presence of screens, and import these components (`ScreenContentWrapper` & `ScreenFooter`) only when available, same as you do with importing `Screen`.

packages/native-stack/src/types.tsx
```diff
@@ -532,17 +547,32 @@ export type NativeStackNavigationOptions = {
...
+   * * 'large' - corresponds to `[1.0]` detent value, maximum height,
+   * * 'all' - corresponds to `[0.5, 1.0]` value, the name is deceiving due to compatibility reasons.
+   *
+   * Defaults to `[1.0]` literal.
+   */
+  sheetAllowedDetents?: number[] | 'fitToContents' | 'medium' | 'large' | 'all';
+  /**
+   * Integer value describing elevation of the sheet, impacting shadow on the top edge of the sheet.
+   *
+   * Not dynamic - changing it after the component is rendered won't have an effect.
```
satya164, line 569: If it cannot be changed, does screens print a warning when changing this?

packages/native-stack/src/types.tsx
satya164, line 47: The `isStable` naming seems confusing. If this is about dragging, then wouldn't `dragging: boolean` be better?

packages/native-stack/src/types.tsx
```diff
@@ -532,17 +547,32 @@ export type NativeStackNavigationOptions = {
...
-   * - `medium` - only medium detent level will be allowed
-   * - `all` - all detent levels will be allowed
+   * Please note that the array **must** be sorted in ascending order.
    *
-   * @platform ios
+   * There are also legacy & **deprecated** options available:
+   *
+   * * 'medium' - corresponds to `[0.5]` detent value, around half of the screen height,
+   * * 'large' - corresponds to `[1.0]` detent value, maximum height,
+   * * 'all' - corresponds to `[0.5, 1.0]` value, the name is deceiving due to compatibility reasons.
```
satya164, line 561: > the name is deceiving due to compatibility reasons

Since this is a new API, maybe `all` should be removed from these values? It seems confusing.

satya164, line 561: Description for `fitToContents` is missing on JS doc.

We should also add description for what tuple values do before these descriptions so it's clearer.

packages/native-stack/src/types.tsx
```diff
@@ -532,17 +547,32 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * Defaults to `[1.0]` literal.
```
satya164, line 563:
```suggestion
   * Defaults to `[1.0]`.
```

packages/native-stack/src/types.tsx
```diff
@@ -572,19 +607,28 @@ export type NativeStackNavigationOptions = {
   sheetGrabberVisible?: boolean;
   /**
    * The largest sheet detent for which a view underneath won't be dimmed.
-   * Works only when `presentation` is se tto `formSheet`.
+   * Works only when `stackPresentation` is set to `formSheet`.
```
satya164, line 610:
```suggestion
   * Works only when `presentation` is set to `formSheet`.
```

packages/native-stack/src/types.tsx
```diff
@@ -572,19 +607,28 @@ export type NativeStackNavigationOptions = {
...
-   * - `all` - the view underneath will be dimmed for any detent level
+   * Additionaly there are following options available:
    *
-   * Defaults to `all`.
+   * * `none` - there will be dimming view for all detents levels,
+   * * `largest` - there won't be a dimming view for any detent level.
    *
-   * @platform ios
+   * There also legacy & **deprecated** prop values available: `medium`, `large` (don't confuse with `largest`), `all`, which work in tandem with
+   * corresponding legacy prop values for `sheetAllowedDetents` prop.
```
satya164, line 621: We shouldn't add deprecated values for new options.

packages/native-stack/src/types.tsx
```diff
@@ -558,10 +588,15 @@ export type NativeStackNavigationOptions = {
...
-   * @platform ios
    */
   sheetCornerRadius?: number;
+  /**
+   * Index of the detent the sheet should expand to after being opened.
+   * Works only when `stackPresentation` is set to `formSheet`.
+   *
+   * Defaults to `0` - which represents first detent in the detents array.
+   */
+  sheetInitialDetent?: number;
```
satya164, line 599: I think `sheetInitialDetentIndex` would make more sense as you're not specifying the detent, but index of the detent.

```suggestion
  sheetInitialDetentIndex?: number;
```

Also what happens when you specify an index which is out of bounds? Does it throw an error? If not then it should.

packages/native-stack/src/views/DebugContainer.native.tsx
```diff
@@ -15,22 +18,26 @@ type ContainerProps = ViewProps & {
...
-          <View {...rest} collapsable={false} />
+          <ScreenContentWrapper {...rest} />
         </AppContainer>
       );
     }
 
-    return <View {...rest} collapsable={false} />;
+    return <ScreenContentWrapper {...rest} />;
   };
 }
```
satya164, line 43: Is there a reason this code is not inside of `react-native-screens`? Seems problematic to leave this upto the consumer.

Since there's going to be major version of screens, I think this is a good time to move this code into screens.

satya164 (changes requested):

packages/native-stack/src/types.tsx
kkafar: Done, thanks

packages/native-stack/src/types.tsx
kkafar: 'fitToContents` is described in paragraph just above the fragment you highlighted.

packages/native-stack/src/types.tsx
kkafar: +1

packages/native-stack/src/types.tsx
kkafar: Yeah, I'll remove the legacy options from this PR. I'll keep them for v6 though, since they were already released (also on the side of `react-navigation/native-stack`).

packages/native-stack/src/types.tsx
kkafar: When I'm thinking of it right now - `dragging` might be misleading here, e.g. if someone wanted to dim sheet's contents on drag gesture. `isStable` is also `false`, when user is no longer dragging the sheet, but it is still settling (animating) to target detent.

packages/native-stack/src/types.tsx
kkafar: As to the naming? Won't it leave out the `sheetLargestUndimmedDetent`? It is also primarily an index (it has also additional options).

kkafar: I've aligned the naming. Both are now named `sheetInitialDetentIndex` and `sheetLargestUndimmedDetentIndex` accordingly. They both now accept `last` option instead of `largest`.

packages/native-stack/src/views/DebugContainer.native.tsx
kkafar: Let's do it in separate PR. I'll create a task on board for this. Is that ok?

kkafar: Added this to navigation's board.

packages/native-stack/src/types.tsx
kkafar: beta.6 will have this warning

packages/native-stack/src/types.tsx
satya164: @kkafar ok, lets change the property name to `stable`. we're not using `is` prefix in other places in such properties. e.g. stack has `closing`

satya164 (changes requested):

satya164 (approved):

Maclay74:
Wooohooo! Thanks for your work!

PR #12053, opened 2024-07-09
feat: add fullScreenSwipeShadowEnabled prop to NativeStackView

maksg:
This is a PR to add a prop for enabling shadow during full screen swipe transition on iOS.
Required for https://github.com/software-mansion/react-native-screens/pull/2239.


packages/native-stack/src/types.tsx
```diff
@@ -438,6 +438,16 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   fullScreenGestureEnabled?: boolean;
+  /**
+   * Whether the full screen dismiss gesture has shadow under view during transition. The gesture uses custom transition and thus
+   * doesn't have a shadow by default. When enabled a custom shadow view is added during the transition which tries to mimick the
```
tboba, line 443:
```suggestion
   * doesn't have a shadow by default. When enabled, a custom shadow view is added during the transition which tries to mimic the
```

packages/native-stack/src/types.tsx
```diff
@@ -438,6 +438,16 @@ export type NativeStackNavigationOptions = {
...
+   * default iOS shadow. Defaults to `false`.
+   *
+   * Doesn't affect the behavior of transitions not using this gesture.
```
tboba, line 446:
```suggestion
   * This does not affect the behavior of transitions that don't use gestures, enabled by `fullScreenGestureEnabled` prop.
```

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -330,6 +335,7 @@ const SceneView = ({
       hasLargeHeader={options.headerLargeTitle ?? false}
       customAnimationOnSwipe={animationMatchesGesture}
       fullScreenSwipeEnabled={fullScreenGestureEnabled}
+      fullScreenSwipeShadowEnabled={fullScreenGestureShadowEnabled}
```
tboba, line 338: Remember, that you need to ignore this file with `@ts-expect-error`, since this prop is not available yet 😄

```suggestion
      // @ts-expect-error prop supported from react-native-screens 3.33.0 onwards
      fullScreenSwipeShadowEnabled={fullScreenGestureShadowEnabled}
```

tboba (changes requested):
Just left some comments, remember to align them in native-stack v5 aswell!
You can also create separate PR for adding this change to `6.x` branch (since `6.x` version is still supported).

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -220,6 +221,10 @@ const SceneView = ({
     }
   }
 
+  if (fullScreenGestureShadowEnabled === undefined) {
+    fullScreenGestureShadowEnabled = false;
+  }
+
```
satya164, line 227: Specify the default during destructuring `options` instead

tboba (approved):

maksg:
@satya164 I'm getting an `Unused '@ts-expect-error' directive.` error for `navigationBarTranslucent`. After discussing this with @tboba we decided to remove this line so the CI can pass but should we leave it be or do you know if there is something else we can do to fix it?

satya164:
> I'm getting an Unused '@ts-expect-error' directive.

Where are you getting the error?

> After discussing this with @tboba we decided to remove this line so the CI can pass

I see it's there in the diff and CI is passing 🤔

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -341,7 +344,6 @@ const SceneView = ({
       homeIndicatorHidden={autoHideHomeIndicator}
       hideKeyboardOnSwipe={keyboardHandlingEnabled}
       navigationBarColor={navigationBarColor}
-      // @ts-expect-error prop supported from react-native-screens 3.32.0 onwards
```
maksg, line 344: @satya164 This one caused the error. Not the one I added

maksg:
> Where are you getting the error?
Line 347 in `NativeStackView.native.tsx`. I mentioned you in the comment with this now.

> I see it's there in the diff and CI is passing 🤔
You're thinking about the one I added for `fullScreenSwipeShadowEnabled`. The one giving errors was lower for `navigationBarTranslucent` and was removed so the CI could pass.

packages/native-stack/src/views/NativeStackView.native.tsx
satya164: @maksg this is because the `@ts-expect-error` affects all of the props after it and all other errors are suppressed. so there can be only one `@ts-expect-error` for the props.

maksg: Ah, that explains it then. Thanks!

PR #12083, opened 2024-08-02
feat: enable preload in NativeStackView and add `hiddenFromStack` prop

maksg:
Adds support for using `CommonActions.preload` with `native-stack` and adds `hiddenFromStack` prop.
Required for https://github.com/software-mansion/react-native-screens/pull/2278.


packages/native-stack/src/types.tsx
```diff
@@ -596,6 +596,17 @@ export type NativeStackNavigationOptions = {
...
+   * If a screen is marked `hidden`, the resources will be retained, but it will not be rendered
+   * in the stack. This is used when preloading screen but can also be used as a way to support
+   * iOS picture-in-picture, where you have a video overlay originating on a screen, and even
+   * when popping the screen off the stack the resources need to be retained (not garbage collected),
+   * so that the video can continue playing. Additionally, this allows the screen to be "un-hidden"
+   * (visually, pushed back onto the stack) when the user taps on the overlay video.
+   *
+   * Only supported on iOS and Android.
+   */
+  hiddenFromStack?: boolean;
```
satya164, line 609: Let's not expose this option. We have different plans for how the functionality will be exposed and need to align between native stack and JS https://github.com/orgs/react-navigation/projects/1?pane=issue&itemId=48590954

satya164 (changes requested):
Thanks for the PR.

1. Let's also update the web implementation to support preload. It should be straightforward by rendering the component with the style `display: none`.
2. We need to update our docs for 7.x as well.

packages/native-stack/src/types.tsx
maksg: All right, I will remove that. Also I don't have access to this link so I can't check what the plans are.

satya164: The current idea is to have an action called retain:

1. Let's say you're on a screen where you want to have PiP enabled if the user pops it, you call `navigation.retain(true)`
2. After this, if the screen is removed from the stack with etc., it will still keep it rendered and hidden.
3. And when you navigate back to this screen, it'll be added to the stack and focused
4. Calling `navigation.retain(false)` will disable this behavior, or if the screen is currently hidden, it'll completely remove it

maksg: Sounds good actually! Thanks for explaining

karanguptaneo (approved):

satya164 (approved):

satya164:
Is there a reason it's still in draft?

maksg:
I'm waiting for a `screens` PR review to make sure that there is nothing more to fix here.
CC @kkafar

kkafar (commented):
I'm looking into it right now.

I'll change the prop name in native-stack to `retainOnly` or something similar.

Moreover, the relation between the new prop and `activityState` is yet unclear & it must be discussed here. I'll write a post later in the day. 

Stay tuned.

Thanks @maksg for awesome job 🎉

packages/native-stack/src/views/NativeStackView.tsx
```diff
@@ -86,6 +93,10 @@ export function NativeStackView({ state, descriptors }: Props) {
 
         const nextPresentation = nextDescriptor?.options.presentation;
 
+        const shouldBePreloaded =
```
satya164, line 96: I think `isPreloaded` is better. The screen is already preloaded in the router, the view layer needs to handle it.

```suggestion
        const isPreloaded =
```

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -565,6 +589,7 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
               previousDescriptor={previousDescriptor}
               nextDescriptor={nextDescriptor}
               isPresentationModal={isModal}
+              shouldBePreloaded={shouldBePreloaded}
```
satya164, line 592: Should be `isPreloaded` to match other prop names.

PR #11980, opened 2024-05-09
feat(iOS): Handle new backButtonDisplayMode functionality

maciekstosio:
Please provide enough information so that others can review your pull request.

**Motivation**

Adds corresponding functionality from https://github.com/software-mansion/react-native-screens/pull/2123. 

**Test plan**

Describe the **steps to test this change** so that a reviewer can verify it. Provide screenshots or videos if the change affects UI.

The change must pass lint, typescript and tests.

PR #12089, opened 2024-08-03
feat: add new backButtonDisplayMode functionality

dylancom:
**Motivation**

Adds corresponding functionality from https://github.com/software-mansion/react-native-screens/pull/2123.

This PR continues the work of the draft PR: #11980
Thanks to: @maciekstosio


satya164 (changes requested):
Hey @dylancom, thanks for the PR.

- How does this differ from `headerBackTitleVisible`? Seems to be doing similar things. Do we really need 2 options that do the same thing? Can this reuse the existing `headerBackTitleVisible` to pass appropriate combination of options instead?
- What's the difference between `generic` and `default`? Is the only difference that it'll ignore explicit back button label? In that case it should be possible to determine that based on `headerBackTitleVisible` option.

The linked PR description seems to be mostly technical details which don't seem to answer these questions.

dylancom:
Hey @satya164. headerBackTitleVisible is a hack. it sets the backbutton title to "nil". Leading to weird animations. Such as the arrow moving, which it shouldn't do...

The official way is to use backButtonDisplayMode.
See: https://developer.apple.com/documentation/uikit/uinavigationitem/backbuttondisplaymode/minimal

satya164:
> Hey @satya164. headerBackTitleVisible is a hack. it sets the backbutton title to "nil". Leading to weird animations. Such as the arrow moving, which it shouldn't do...

Seems you're talking about the `backTitleVisible` prop of screens and the implementation detail. I'm asking about how does this new option differ from the existing `headerBackTitleVisible` option of native stack?

From your response seems like there's no difference in the intention of the option other than it being the correct implementation. In that case, either only the internal implementation should be changed to use the `backButtonDisplayMode` instead of `backTitleVisible`, or the old option should be removed in favor of the new option. Is there a reason for having both?

> Default: The navigation item attempts to display a specific title, a generic title, or no title for the Back button, depending on the space available.
> Generic: The navigation item attempts to display a generic title or no title for the Back button, depending on the space available.

It's unclear to me. What does "specific title" vs a "generic title" mean in this context exactly? Do you mean the screen's title vs something like "Back"?

packages/native-stack/src/types.tsx
```diff
@@ -342,6 +342,15 @@ export type NativeStackNavigationOptions = {
...
   headerBackButtonMenuEnabled?: boolean;
+  /**
+   * How the back button behaves by default (when not customized). Available on iOS>=14.
+   * The following values are currently supported (they correspond to https://developer.apple.com/documentation/uikit/uinavigationitembackbuttondisplaymode?language=objc):
+   * - "default" – shows given back button title/previous controller title, system default or just icon based on available space
+   * - "generic" – shows given system default or just icon based on available space
+   * - "minimal" – shows just an icon
+   * @platform ios
+   */
+  backButtonDisplayMode?: ScreenStackHeaderConfigProps['backButtonDisplayMode'];
```
satya164, line 353: All header related options are prefix with `header` so this should be changed to:

```suggestion
  headerBackButtonDisplayMode?: ScreenStackHeaderConfigProps['backButtonDisplayMode'];
```

Also if this new option is necessary, we need to remove the existing `headerBackTitleVisible` option and align JS stack to have this new option as well.

satya164 (changes requested):

dylancom:
To be honest I don't understand what the benefits of the other options are and my intention is indeed to use the correct implementation of hiding a title.

packages/native-stack/src/types.tsx
```diff
@@ -342,6 +342,15 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   headerBackButtonMenuEnabled?: boolean;
+  /**
+   * How the back button behaves by default (when not customized). Available on iOS>=14.
+   * The following values are currently supported (they correspond to https://developer.apple.com/documentation/uikit/uinavigationitembackbuttondisplaymode?language=objc):
+   * - "default" – shows given back button title/previous controller title, system default or just icon based on available space
+   * - "generic" – shows given system default or just icon based on available space
+   * - "minimal" – shows just an icon
```
satya164, line 350: We also need to make the descriptions more clear.

- The terminology should be updated (e.g. screen title instead of controller title)
- Not sure what "given" means in system default context, maybe a mistake?
- And add some example app what system default means in this case
- There being a "default" option and the "system default" also seem a bit confusing since they seem like different things

packages/native-stack/src/types.tsx
```diff
@@ -342,6 +342,15 @@ export type NativeStackNavigationOptions = {
...
   headerBackButtonMenuEnabled?: boolean;
+  /**
+   * How the back button behaves by default (when not customized). Available on iOS>=14.
+   * The following values are currently supported (they correspond to https://developer.apple.com/documentation/uikit/uinavigationitembackbuttondisplaymode?language=objc):
+   * - "default" – The navigation item attempts to display a specific title, a generic title, or no title for the Back button, depending on the space available.
+   * - "generic" – The navigation item attempts to display a generic title or no title for the Back button, depending on the space available.
+   * - "minimal" – The navigation item displays the Back button indicator instead of a title.
+   * @platform ios
+   */
+  headerBackButtonDisplayMode?: ScreenStackHeaderConfigProps['backButtonDisplayMode'];
```
satya164, line 353: Let's remove "the navigation item" as it's redundant, and the terminology doesn't match anything else in react navigation, so it'll be confusing unless someone iOS navigation system. I'd also remove "for the Back button" since it's implied since this option is for back button from the option name and the description.

I'd also remove the link to UIKit docs. Our options aren't intended to match native APIs 1:1 and we may change them to what feels more natural.

Though these options are still unclear to me,

- I think "icon" clearer than "indicator" - and matched the existing terminology
- Still unsure what "specific title" and "generic title" mean

Generally we need to match the terminology we already use in react navigation and think in terms of users who don't know anything about native iOS APIs.

packages/native-stack/src/types.tsx
dylancom: I am unsure if we can safely remove headerBackTitleVisible or replace it. It is also kept within the react-native-screens/native-stack lib.

satya164 (changes requested):

packages/native-stack/src/types.tsx
satya164: The options in `react-native-screens/native-stack` are not relevant as it mostly exposes the underlying options verbatim. In our case we need to match it with other navigators, think about how different options work together and how it fits with overall system. Often we don't want to expose screen's props verbatim.

packages/native-stack/src/types.tsx
dylancom: Thanks for the feedback, I also did some more research and improved the explanations.

packages/native-stack/src/types.tsx
```diff
@@ -342,6 +342,15 @@ export type NativeStackNavigationOptions = {
    * @platform ios
    */
   headerBackButtonMenuEnabled?: boolean;
+  /**
+   * How the back button behaves by default (when not customized). Available on iOS>=14.
+   * The following values are currently supported:
+   * - "default" – Attempts to display an appropriate title, followed by 'Back' or no title, depending on the space available.
+   * - "generic" – Attempts to display 'Back' or no title, depending on the space available.
+   * - "minimal" – Displays the Back button icon instead of a title.
+   * @platform ios
```
satya164, line 351:
```suggestion
  /**
   * How the back button title behaves. Available on iOS >= 14.
   * 
   * Supported values:
   * - "default" - Displays one of the following depending on the available space: previous screen's title, generic title (e.g. 'Back') or no title (only icon).
   * - "generic" – Displays one of the following depending on the available space: generic title (e.g. 'Back') or no title (only icon).
   * - "minimal" – Always displays only the icon without a title.
   *
   * @platform ios
```

dylancom:
Could also probably refactor it to:
`backButtonDisplayMode={headerBackTitleVisible ? "default" : "minimal"}`

satya164:
Seems this new option is only available on iOS >= 14, in that case I think we should check iOS version and fallback to `backTitleVisible` option.

So I guess the generic option will be available only for iOS >= 14.

Otherwise, "default" should set `backTitleVisible` to `true` and "minimal" should set it to `false`.

satya164:
> Could also probably refactor it to, if not caring about all display modes:
backButtonDisplayMode={headerBackTitleVisible ? "default" : "minimal"}

I think it'd be better to remove `headerBackTitleVisible` since the new option has more flexibility.

satya164:
I'll wait for @kkafar or @tboba's reviews, and if my assumptions are correct, before merging, we'd need to:

- Remove `headerBackTitleVisible` option
- Fallback to `backTitleVisible` prop on iOS < 14
- Update the description to mention that `generic` is only available on iOS >= 14

satya164 (approved):

kkafar:
Damn, I was just writing a post 😅 Anyway then. @maciekstosio, can you test these changes?

kkafar:
The `backButtonDisplayMode` is a prop directly exposing [native API](https://developer.apple.com/documentation/uikit/uinavigationitem/3656350-backbuttondisplaymode?language=objc), which had various caveats described [here](https://github.com/software-mansion/react-native-screens/pull/2123), notably not working with custom back items on iOS. Hence I'm not convinced that this will properly replace current logic.

dylancom:
I was afraid of this aswel, that's why my initial PR was just adding backButtonDisplayMode as an extra prop. @satya164 @kkafar

satya164:
We don't directly expose the native API in native-stack unlike screens. We can use whatever internal implementation based on the scenario. Having 2 different options doing the same thing will only cause confusion.

PR #12094, opened 2024-08-06
fix: add more fallbacks for headerBackButtonDisplayMode

satya164:
We can't use `backButtonDisplayMode` prop from `react-native-screens` in the following scenarios:

- The iOS version is lower than 14
- Custom back title is set
- Custom back title style is set (`fontFamily` or `fontSize`)
- Back button menu is disabled

This change falls back to `backTitleVisible` in those scenarios.

### DIFF
--- a/packages/native-stack/src/views/HeaderConfig.tsx
+++ b/packages/native-stack/src/views/HeaderConfig.tsx
@@ -176,8 +176,17 @@ export function HeaderConfig({
       Platform.OS === 'ios' &&
       headerTransparent !== false);
 
-  const isBackButtonDisplayModeAvailableForCurrentPlatform =
-    Platform.OS === 'ios' && parseInt(Platform.Version, 10) >= 14;
+  const isBackButtonDisplayModeAvailable =
+    // On iOS 14+
+    Platform.OS === 'ios' &&
+    parseInt(Platform.Version, 10) >= 14 &&
+    // Doesn't have custom back title
+    headerBackTitle == null &&
+    // Doesn't have custom styling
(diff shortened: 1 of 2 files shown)


maciekstosio (approved):

packages/native-stack/src/views/HeaderConfig.tsx
```diff
@@ -176,21 +176,30 @@ export function HeaderConfig({
...
+    // On iOS 14+
+    Platform.OS === 'ios' &&
+    parseInt(Platform.Version, 10) >= 14 &&
+    // Doesn't have custom back title
+    headerBackTitle == null &&
+    // Doesn't have custom styling
+    backTitleFontFamily == null &&
+    backTitleFontSize == null &&
+    // Back button menu is not disabled
+    headerBackButtonMenuEnabled === false;
```
satya164, line 189:
```suggestion
    headerBackButtonMenuEnabled !== false;
```

asadwan:
It's no longer possible to hide back button title and back button menu at the same time :/

maciekstosio:
@asadwan what flags are you using? 

```
      headerBackButtonDisplayMode: 'minimal',
      headerBackButtonMenuEnabled: false,
```

Works for me

asadwan:
> @asadwan what props are you using?
> 
> ```
>       headerBackButtonDisplayMode: 'minimal',
>       headerBackButtonMenuEnabled: false,
> ```
> 
> Works for me

I use the same props but I wasn't on the latest release, it's all good now. Thanks!

PR #12117, opened 2024-08-27
Reduce number of dependencies for `@react-navigation/devtools` package

sairus2k:
Switch to `fast-deep-equal` from `deep-equal` for `@react-navigation/devtools` package. By the way, `fast-deep-equal` already used in `@react-navigation/native` package.

**Motivation**

`deep-equal` is a terribly heavy library: [https://npmgraph.js.org/?q=%40react-navigation%2Fdevtools](https://npmgraph.js.org/?q=%40react-navigation%2Fdevtools)

**Test plan**

Lint, typescript and tests are passed.

### DIFF
--- a/packages/devtools/package.json
+++ b/packages/devtools/package.json
@@ -47,13 +47,12 @@
     "clean": "del lib"
   },
   "dependencies": {
-    "deep-equal": "^2.2.3",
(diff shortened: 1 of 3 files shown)


satya164 (approved):
Thanks for the PR

PR #12120, opened 2024-08-28
refactor: getStateFromPath config calculations caching

kacper-mikolajczak:
**_Problem_**

When `getStateFromPath` receives big _linking config_ ([example](https://github.com/Expensify/App/blob/071f11ce1012db7d07eb750d74f58b552eba6144/src/libs/Navigation/linkingConfig/config.ts#L1152)) as an option parameter, it tends to slow down quite drastically.

**_Investigations_** 

The config-related data is created every time`getStateFromPath` is called, resulting in noticeable overhead.

Often times, configs provided to `getStateFromPath` are going to be static (e.g. initialised once at the app setup and remained untouched throughout its lifetime) and the data derived from it could also be static.

**_Solution_**

PR tries to improve `getStateFromPath` helper performance by extracting and caching calculations related to last linking config provided (due to its static nature, keeping reference to the latest value seems an OK heuristic).

Here is an example of potential gains for Exfy app and its config: https://github.com/Expensify/App/issues/48150

_**Test plan**_

No manual testing required.

_**PR introduces**_

- [x] `getStateFromPath` config data caching
- [x] `setTimeout` typings fixes

### DIFF
--- a/packages/native/src/createStaticNavigation.tsx
+++ b/packages/native/src/createStaticNavigation.tsx
@@ -51,32 +51,46 @@ export function createStaticNavigation(tree: StaticNavigation<any, any, any>) {
     { linking, ...rest }: Props,
     ref: React.Ref<NavigationContainerRef<ParamListBase>>
   ) {
-    const screens = React.useMemo(() => {
-      if (tree.config.screens) {
-        return createPathConfigForStaticNavigation(
-          tree,
-          { initialRouteName: linking?.config?.initialRouteName },
-          linking?.enabled === 'auto'
-        );
+    const linkingConfig = React.useMemo(() => {
+      if (!tree.config.screens) return;
+
+      const screens = createPathConfigForStaticNavigation(
+        tree,
+        { initialRouteName: linking?.config?.initialRouteName },
+        linking?.enabled === 'auto'
+      );
+
+      if (!screens) return;
+
+      return {
+        path: linking?.config?.path,
+        initialRouteName: linking?.config?.initialRouteName,
+        screens,
+      };
+    }, [
+      linking?.enabled,
+      linking?.config?.path,
+      linking?.config?.initialRouteName,
+    ]);
+
+    const memoizedLinking = React.useMemo(() => {
+      if (!linking) {
+        return undefined;
       }
 
-      return undefined;
-    }, [linking?.config, linking?.enabled]);
-
-    const transformedLinking = React.useMemo(() => {
-      if (!linking) return undefined;
+      const enabled =
+        typeof linking.enabled === 'boolean'
+          ? linking.enabled
+          : linkingConfig?.screens != null;
 
       return {
         ...linking,
-        enabled:
-          typeof linking.enabled === 'boolean'
-            ? linking.enabled
-            : screens != null,
-        config: screens ? { ...linking.config, screens } : undefined,
+        enabled,
+        config: linkingConfig,
       };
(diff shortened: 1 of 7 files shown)


kacper-mikolajczak:
Could not define helpers as inline functions (`const f = () => {}`) due to the fact that they were crashing tests as (_assumption_) they are called in the module and hoisting did not work for them.

It would be cool to change that if possible!

satya164 (changes requested):
We should also add `useMemo` for `linking.config` in `createStaticNavigation.tsx` so the config is always the same reference when using static navigation.

Currently, we have `useMemo` for the `linking.config.screens` only, but we can change it to `linking.config` - and tweak the dep array to be more granular.

kacper-mikolajczak:
Sure thing @satya164!

[In this commit](https://github.com/react-navigation/react-navigation/pull/12120/commits/e94d45a7f48938ded3ec4aa519bbb782bf2c5999), the entire `linking` property is memoized, so we should keep stable ref is user decided to memoize it as well.

packages/native/src/createStaticNavigation.tsx
```diff
@@ -63,6 +63,19 @@ export function createStaticNavigation(tree: StaticNavigation<any, any, any>) {
...
+
+      return {
+        ...linking,
+        enabled:
+          typeof linking.enabled === 'boolean'
+            ? linking.enabled
+            : screens != null,
+        config: screens ? { ...linking.config, screens } : undefined,
+      };
+    }, [linking, screens]);
```
satya164, line 77: Looking at a bit more, we should be able to always keep same reference for `linking.config` with static config as it only uses 2 string properties from `linking.config`:

```js
const linkingConfig = React.useMemo(() => {
  const screens = tree.config.screens
    ? createPathConfigForStaticNavigation(
        tree,
        { initialRouteName: linking?.config?.initialRouteName },
        linking?.enabled === 'auto'
      )
    : undefined;

  const config = screens
    ? {
        path: linking?.config?.path,
        initialRouteName: linking?.config?.initialRouteName,
        screens,
      }
    : undefined;

  return config;
}, [
  linking?.enabled,
  linking?.config?.path,
  linking?.config?.initialRouteName,
]);

const memoizedLinking = React.useMemo(() => {
  if (!linking) {
    return undefined;
  }

  return {
    ...linking,
    enabled:
      typeof linking.enabled === 'boolean'
        ? linking.enabled
        : linkingConfig.screens != null,
    config: linkingConfig,
  };
}, [linking, screens]);
```

kacper-mikolajczak: @satya164 Thanks, suggestion applied [here](https://github.com/react-navigation/react-navigation/pull/12120/commits/74a71ad7d49ea3b9c83012cc13273d584086335b) ✅ 

I've refactored code a bit to take advantage of early returns and boost up readability. Let me know what you think!

kacper-mikolajczak:
Hey @satya164, addressing your suggestions about checking `getPathFromState` perf, I've added the caching on normalized configs calculation [here](https://github.com/react-navigation/react-navigation/pull/12120/commits/69438a86c4f40ae18def08f26833f394d8d5a721). 

The results can be checked here: https://github.com/Expensify/App/issues/48150#issuecomment-2317690321

tl;dr; It gets down to under 1ms from couple of ms of calculations 🥳

satya164 (approved):
Thanks

Issue #11989, opened 2024-05-15
A props object containing a "key" prop is being spread into JSX


KrisLau:
### Current behavior

Every time I load my app I get this error: 
```js
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, position: ..., route: ..., navigationState: ..., getAccessibilityLabel: ..., getAccessible: ..., getLabelText: ..., getTestID: ..., renderBadge: ..., renderIcon: ..., renderLabel: ..., activeColor: ..., inactiveColor: ..., pressColor: ..., pressOpacity: ..., onLayout: ..., onPress: ..., onLongPress: ..., labelStyle: ..., style: ..., defaultTabWidth: ..., android_ripple: ...};
  <TabBarItem {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {position: ..., route: ..., navigationState: ..., getAccessibilityLabel: ..., getAccessible: ..., getLabelText: ..., getTestID: ..., renderBadge: ..., renderIcon: ..., renderLabel: ..., activeColor: ..., inactiveColor: ..., pressColor: ..., pressOpacity: ..., onLayout: ..., onPress: ..., onLongPress: ..., labelStyle: ..., style: ..., defaultTabWidth: ..., android_ripple: ...};
  <TabBarItem key={someKey} {...props} />
    in CellRenderer (created by VirtualizedList)
    in RCTScrollContentView (created by ScrollView)
    in RCTScrollView (created by ScrollView)
    in ScrollView (created by ScrollView)
    in ScrollView (created by VirtualizedList)
    in VirtualizedListContextProvider (created by VirtualizedList)
    in VirtualizedList (created by FlatList)
    in FlatList (created by Animated(Anonymous))
    in Animated(Anonymous) (created by TabBar)
    in RCTView (created by View)
    in View (created by TabBar)
    in RCTView (created by View)
    in View (created by Animated(View))
    in Animated(View) (created by TabBar)
    in TabBar (created by PagerViewAdapter)
    in PagerViewAdapter (created by TabView)
    in RCTView (created by View)
    in View (created by TabView)
    in TabView (created by TeamHome)
    in TeamHome (created by withObservables[teamTypeID,roles] { teamType, resources, roleAccess })
    in withObservables[teamTypeID,roles] { teamType, resources, roleAccess }
    in Unknown
    in Unknown (created by SceneView)
    in StaticContainer
    in EnsureSingleNavigator (created by SceneView)
[...]

PR #12142, opened 2024-09-11
fix: spread key into JSX warning

jewhyena:
**Motivation**

After bumping up versions of `react` from `18.2.0` to `18.3.1` as well as `react-native` from `0.74.5` to `0.75.2` in one of my projects, the application started to give me a warning `A props object containing a "key" prop is being spread into JSX`.

It happens on the screens which use `@react-navigation/material-top-tabs`.

Addresses the existing issue https://github.com/react-navigation/react-navigation/issues/11989.

**Test plan**

1. Initialize a new project with the latest versions of `react` and `react-native`
2. Make use of `@react-navigation/material-top-tabs`
3. See the warning in the console

**Solution**

Passing the key straight to the component instead of spreading it.

I've also removed the `renderTabBarItem` prop since it doesn't seem to be used anywhere and could've caused a more verbose fix. 🤔

### DIFF
--- a/packages/react-native-tab-view/src/TabBar.tsx
+++ b/packages/react-native-tab-view/src/TabBar.tsx
@@ -45,9 +45,6 @@ export type Props<T extends Route> = SceneRendererProps & {
   options?: Record<string, TabDescriptor<T>>;
   commonOptions?: TabDescriptor<T>;
   renderIndicator?: (props: IndicatorProps<T>) => React.ReactNode;
-  renderTabBarItem?: (
-    props: TabBarItemProps<T> & { key: string }
-  ) => React.ReactElement;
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
> I've also removed the renderTabBarItem prop since it doesn't seem to be used anywhere and could've caused a more verbose fix.

Just because it's not used anywhere in the example in this repo doesn't mean that nobody else is using it. Please don't remove any features in your PR.

It doesn't affect anything in the fix. It's up to the user of that prop to handle the `key` in their own code.

Mibou:
Hello, would it possible to merge this fix soon please? 😇🙏

dacitto (approved):

satya164 (approved):
Thanks!

PR #12145, opened 2024-09-13
feat: export LocaleDirContext and UNSTABLE_UnhandledLinkingContext

marklawlor:
**Motivation**

Expo Router is a library built on top of React Navigation. It customises the linking behaviour of React Navigation by providing a custom `<NavigationContainer />` and changes the behaviour how it performs linking.

In React Navigation 7, the packages were changed to use `package.json#exports` which only expose the public APIs and not the internal contexts/hooks that are needed by to create a custom `<NavigationContainer />`.

Instead of adding these values to the public exports, this PR proposes that a new submodule is exposed `@react-navigation/native/extend` which exports public API and the required hooks/contexts needed to create a custom `<NavigationContainer />`

### DIFF
--- a/packages/native/src/index.tsx
+++ b/packages/native/src/index.tsx
@@ -1,11 +1,13 @@
 export { createStaticNavigation } from './createStaticNavigation';
 export { Link } from './Link';
 export { LinkingContext } from './LinkingContext';
+export { LocaleDirContext } from './LocaleDirContext';
 export { NavigationContainer } from './NavigationContainer';
 export { ServerContainer } from './ServerContainer';
 export { DarkTheme } from './theming/DarkTheme';
 export { DefaultTheme } from './theming/DefaultTheme';
 export * from './types';
+export { UnhandledLinkingContext as UNSTABLE_UnhandledLinkingContext } from './UnhandledLinkingContext';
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Both are public APIs. I don't think it makes sense to have exports only for libraries. Regardless of whether they are exported "for library" or for users have the same effect - internal files are now public APIs and we can't change them without breaking other code.

If there's a functionality lacking, then it should be part of the public API - not by exposing internal files. If Expo Router needs anything specific to be customized, you need to discuss how to do that with public APIs instead of forking `NavigationContainer`.

For what to do with this PR:

`LocaleDirContext` should be exported as we generally export contexts for use with class components etc. `UnhandledLinkingContext` is an experimental API and if it's exported then it should be prefixed with `unstable_`. Both of these should be exported in the main `index.tsx` and not a separate file.

As for `useBackButton`, `useDocumentTitle` and `useThenable`, these are internal implementation details and if you need these then you should vendor them.

marklawlor:
* Updated `index.ts`
* Exposed `UnhandledLinkingContext` as `UNSTABLE_UnhandledLinkingContext`
* Removed `extend.ts`

satya164 (approved):
Thanks

PR #12161, opened 2024-09-22
refactor: move web-only checks inside the Platform.OS web condition

bytehala:
Prevents react-native-tab-view from throwing an error on fireEvent.press when running Jest tests

**Motivation**

Apps that use react-native-tab-view can't test tab switching using only `fireEvent.press(component)` and devs have to delve into the code to realize that they have to use `fireEvent.press(component, eventObject)` [reproduction proof](https://github.com/bytehala/rntv-issue/pull/1)

When using `fireEvent.press(component)`, Jest throws an error `TypeError: Cannot read properties of undefined (reading 'metaKey')`. This is because of 2 main reasons:

1. The React Native Testing Library fireEvent API [does not send an event](https://callstack.github.io/react-native-testing-library/docs/api/events/fire-event)
2. The code was checking for web-only attributes outside of the `Platform.OS === web` condition, like metaKey

Screenshot 2024-09-21 at 10 04 14 PM

**Test plan**

Proof of the error:
1. Check out this branch
2. Revert the react-native-tab-view PlatformPressable.tsx to the "before" revision
3. Run the new tests `yarn test ./packages/react-native-tab-view/src/__tests__/PlatformPressable.test.tsx`
- [ ] The fireEvent.press test should throw an error

Proof that the refactor works for fireEvent.press and UserEventInstance.press:
1. Check out this branch
2. Run the new tests `yarn test ./packages/react-native-tab-view/src/__tests__/PlatformPressable.test.tsx`
- [ ] All 4 tests should pass

### DIFF
--- a/packages/react-native-tab-view/src/PlatformPressable.tsx
+++ b/packages/react-native-tab-view/src/PlatformPressable.tsx
@@ -30,16 +30,16 @@ export function PlatformPressable({
   ...rest
 }: Props) {
   const handlePress = (e: GestureResponderEvent) => {
-    // @ts-expect-error: these properties exist on web, but not in React Native
-    const hasModifierKey = e.metaKey || e.altKey || e.ctrlKey || e.shiftKey; // ignore clicks with modifier keys
-    // @ts-expect-error: these properties exist on web, but not in React Native
-    const isLeftClick = e.button == null || e.button === 0; // only handle left clicks
-    const isSelfTarget = [undefined, null, '', 'self'].includes(
-      // @ts-expect-error: these properties exist on web, but not in React Native
-      e.currentTarget?.target
-    ); // let browser handle "target=_blank" etc.
-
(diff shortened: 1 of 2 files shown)


packages/react-native-tab-view/src/PlatformPressable.tsx
```diff
@@ -30,16 +30,16 @@ export function PlatformPressable({
   ...rest
 }: Props) {
   const handlePress = (e: GestureResponderEvent) => {
-    // @ts-expect-error: these properties exist on web, but not in React Native
-    const hasModifierKey = e.metaKey || e.altKey || e.ctrlKey || e.shiftKey; // ignore clicks with modifier keys
```
bytehala, line 34: Since these values are only ever checked inside the first if-condition, move all of them inside that if block.

satya164 (approved):
Thanks

Issue #11493, opened 2023-07-25
[react-native-tab-view] TabView testability issues


rafael-rollo:
### Current behavior

I am currently facing an issue testing components that depend on TabView. I'm migrating from the deprecated tab-view package to the current one under the react-navigation repo. Tests that used to work with the previous package/version - like the example below - now fail.

``` js
import React from 'react';

import {render, fireEvent} from '@testing-library/react-native';
import App from './App';

test('Should present the route title on scene content', () => {
  const {getByLabelText, getByText} = render(<App />);

  expect(getByText('First route')).toBeDefined();

  fireEvent.press(getByLabelText('Second'));

  expect(getByText('Second route')).toBeDefined();
});
```

It seems the event of `fireEvent.press` just don't be triggered anymore. I noticed the absence of an `onPress` handler in the rendered node (content below) and tried something like `fireEvent(getByLabelText('Second'), 'click')` but doesn't work too. I tried to query for some custom components rendered under a custom tab bar also.

``` jsx
rendered content

...
<View
  accessibilityLabel="Second"
  accessibilityRole="tab"
  ...
  onBlur={[Function]}
  onClick={[Function]}
  onFocus={[Function]}
  onLayout={[Function]}
  ...
>...</>
```

Does anyone see some workaround on that so we don't need to skip tests like these?

A link to a repro is provided.

Thank you.

### Expected behavior

The press event on the tab should be triggered and the scene for the selected tab should be rendered to make the test pass.

### Reproduction

https://github.com/rafael-rollo/rntv-issue

### Platform

- [ ] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [X] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| react-native-tab-view                  | 3.5.2
| react-native-pager-view                | 6.2.0
| react-native                           | 0.72.1
| node                                   | v18.16.1
| npm or yarn                            | 1.22.19
| npm or yarn                            | 1.22.19
| @testing-library/react-native | 12.1.3 (also tested on ^11.5.4)
| jest | ^29.2.1 
| react-test-renderer | 18.2.0

[...]

PR #12162, opened 2024-09-22
test: react-native-tab-view

bytehala:
**Motivation**

`react-native-tab-view` doesn't have Jest tests yet. This provides the setup as well as initial component and interaction tests for rendering and switching tabs.

Closely related to https://github.com/react-navigation/react-navigation/issues/11493

**Test plan**

- [ ] Optional because CI runs it: Run `yarn test` All tests, including the newly added tests, must pass
- [x] The change must pass lint, typescript and tests.

### DIFF
--- a/packages/react-native-tab-view/package.json
+++ b/packages/react-native-tab-view/package.json
@@ -53,6 +53,8 @@
     "use-latest-callback": "^0.2.1"
   },
   "devDependencies": {
+    "@jest/globals": "^29.7.0",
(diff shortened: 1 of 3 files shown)


satya164 (changes requested):
Thanks for the PR. We generally don't `describe` when the file name already says what the test is for. Can you remove the redundant `describe('react-native-tab-view'`?

satya164 (approved):
Thanks

PR #12169, opened 2024-09-28
test: press tab instead of button-in-scene

bytehala:
**Motivation**

The existing test was pressing a button inside the scene, which was calling `navigation.navigate`.
This can cause the tests to pass even if the `BottomTabBar` is not rendered.

**Test plan**


packages/bottom-tabs/src/__tests__/index.test.tsx
```diff
@@ -45,7 +40,12 @@ test('renders a bottom tab navigator with screens', async () => {
   expect(queryByText('Screen A')).not.toBeNull();
   expect(queryByText('Screen B')).toBeNull();
 
-  fireEvent.press(await findByText('Go to B'));
+  // The tab bar should be rendered
```
satya164, line 43:
```suggestion
```

satya164 (approved):
Thanks

PR #12179, opened 2024-10-01
chore: bump react-native-screens version to 4-beta for v7

kkafar:
**Motivation**

`react-navigation@v7` brings support for `screens@v4` stable version of which will be released soon. This PR bumps used version of react native screens in all the packages and example application.

**Test plan**

Example app.

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -21,7 +21,6 @@ import type {
   ScreenProps,
   ScreenStackHeaderConfigProps,
   SearchBarProps,
-  SheetDetentTypes,
 } from 'react-native-screens';
 
 export type NativeStackNavigationEventMap = {
@@ -545,7 +544,7 @@ export type NativeStackNavigationOptions = {
    *
    * @platform ios
    */
-  sheetAllowedDetents?: SheetDetentTypes;
+  sheetAllowedDetents?: number[] | 'fitToContents';
   /**
    * Whether the sheet should expand to larger detent when scrolling.
    * Works only when `presentation` is set to `formSheet`.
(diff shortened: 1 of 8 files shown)


packages/native-stack/src/types.tsx
```diff
@@ -545,7 +544,7 @@ export type NativeStackNavigationOptions = {
    *
    * @platform ios
    */
-  sheetAllowedDetents?: SheetDetentTypes;
+  sheetAllowedDetents?: number[] | 'fitToContents';
```
kkafar, line 547: This is done just to satisfy typechecker temporality before we merge: https://github.com/react-navigation/react-navigation/pull/12032

If you're not fine with that, let me know.

packages/native-stack/src/types.tsx
```diff
@@ -587,7 +586,7 @@ export type NativeStackNavigationOptions = {
    *
    * @platform ios
    */
-  sheetLargestUndimmedDetent?: SheetDetentTypes;
+  sheetLargestUndimmedDetent?: number;
```
kkafar, line 589: Same here

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -423,12 +425,9 @@ const SceneView = ({
           },
         }
       )}
-      freezeOnBlur={freezeOnBlur}
       // When ts-expect-error is added, it affects all the props below it
       // So we keep any props that need it at the end
       // Otherwise invalid props may not be caught by TypeScript
```
kkafar, line 430: Keep / remove this comment? 

I'm personally for keeping it in, so that we don't need to bring it back once new props emerge.

kkafar (commented):
Notes for reviewers.

kkafar:
Bypassing the branch protection, since all unresolved conversations are mine notes.

PR #12180, opened 2024-10-02
refactor: move web-only checks inside web condition

bytehala:
**Motivation**

When writing tests that have `fireEvent.press(somePlatformPressable)` the test throws an error because event is undefined.

This change is similar to https://github.com/react-navigation/react-navigation/pull/12161 

**Test plan**


satya164 (approved):
Thanks

PR #12181, opened 2024-10-02
fix: handlePress type coercion

bytehala:
**Motivation**

In a previous PR https://github.com/react-navigation/react-navigation/pull/12161 I mistook the loose equality operators as typo errors. By using strict equality, the logic got changed on `undefined` values.

```typescript
// pseudocode
  if (rest.href != null) {
    // LOGIC A
  } else {
    // LOGIC B
  }
```

When `rest.href` is `undefined`
- Old Behavior --> LOGIC B
- New Behavior --> LOGIC A ❌

This PR reverts the changes to the equality operators to keep the old behavior.

**Test plan**

- [ ] Optional: temporarily change the `isLeftClick` to have `e.button === null`, the last test should fail

### DIFF
--- a/packages/react-native-tab-view/src/PlatformPressable.tsx
+++ b/packages/react-native-tab-view/src/PlatformPressable.tsx
@@ -30,11 +30,11 @@ export function PlatformPressable({
   ...rest
 }: Props) {
   const handlePress = (e: GestureResponderEvent) => {
-    if (Platform.OS === 'web' && rest.href !== null) {
+    if (Platform.OS === 'web' && rest.href != null) {
       // @ts-expect-error: these properties exist on web, but not in React Native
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. However, it's fine to treat `null` the same as `undefined` in this case.

bytehala:
> Thanks for the PR. However, it's fine to treat `null` the same as `undefined` in this case.

@satya164 Hi, yes I understand. My previous PR makes `null` and `undefined` distinct, as I added strict equality/inequality and thus changed the behavior.

Previous behavior: `null` and `undefined` are treated the same
Current behavior: `undefined` is not treated the same as `null` and goes to the else block

This was my attempt at reinstating the previous behavior. As long as the maintainer team are aware that the behavior got changed, then I'll leave this as is for now.

Thanks!

PR #12188, opened 2024-10-05
Export NativeStackHeaderRightProps and NativeStackHeaderLeftProps

gorbypark:
**Motivation**

I was building a custom header right component, and noticed the props are typed as NativeStackHeaderRightProps, but I couldn't import that type from anywhere.  Of course it was trivial to just make my own type, but it would be nice to be able to just import the type from the package since it's actually available in the types, just not exported from the barrel file.  I also am exporting NativeStackHeaderLeftProps for the same reasons.

**Test plan**

It's just exporting a type from the barrel file, so not quite sure if/how it could be tested.  No actual types have been changed.


satya164 (approved):
Thanks for the PR

PR #12189, opened 2024-10-07
feat: preload using activityState prop

maciekstosio:
Preload feature implemented in https://github.com/react-navigation/react-navigation/pull/12083, with the difference of renaming prop `shouldBePreloaded` into `isPreloaded` and reusing `activityState` instead of new `hiddenFromStack`. Matching PR in react-native-screens repo: https://github.com/software-mansion/react-native-screens/pull/2389


satya164 (approved):
Thank you!

kkafar (approved):
Looks good to me!

PR #12199, opened 2024-10-21
refactor: replace unstable_screenStyle with backgroundColor from contentStyle

maciekstosio:
Corresponding PR in react-native-screens: https://github.com/software-mansion/react-native-screens/pull/2419. 

## Change

Instead of adding `unstable_screenStyle` that directly adds background on `Screen` component, we can pick the `backgroundColor` from `contentStyle` and apply it on `Screen` component, without exposing unstable prop.

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -445,6 +445,8 @@ export type NativeStackNavigationOptions = {
   gestureDirection?: ScreenProps['swipeDirection'];
   /**
    * Style object for the scene content.
+   *
+   * As a workaround to truncated sheet content, formSheet uses backgroundColor from contentStyle and applies it on Screen.
    */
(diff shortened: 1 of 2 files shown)


kkafar (commented):
Okay, this looks solid. 

One last thing to do would be to update description of `contentStyle` in types so that it is known that you can set it to workaround the "truncated sheet content" issue.

kkafar:
I think you're right @satya164, but we'll keep the order here. Merge this & later migrate with your after your PR lands (because we want to include this in todays beta).

kkafar:
Hey, https://github.com/software-mansion/react-native-screens/pull/2433 has been merged - we should be able to proceed now

maciekstosio:
Closing in favor of: https://github.com/react-navigation/react-navigation/pull/12218/files

PR #12204, opened 2024-10-25
feat: add custom screen transitions

maciekstosio:
Continuation of https://github.com/react-navigation/react-navigation/pull/11943 after moving most of the logic to react-native-screens (https://github.com/software-mansion/react-native-screens/pull/2449/files).

The CI will fail until I changes on react-native-screens will be merged and we bump the version in react-navigation.


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -544,118 +542,107 @@ export function NativeStackView({ state, navigation, descriptors }: Props) {
...
-      <ScreenGestureDetector
-        gestureDetectorBridge={gestureDetectorBridge}
+      <ScreenStack
+        style={styles.container}
+        /** Props for custom screen transitions */
         goBackGesture={topScreenOptions?.gestureType}
         transitionAnimation={topScreenOptions?.animationForGesture}
         screenEdgeGesture={topScreenOptions?.gestureFromEdgeEnabled ?? false}
         screensRefs={screensRefs}
         currentRouteKey={currentRouteKey}
```
satya164, line 561: maybe better to name this prop as `currentId` or something like that since `route.key` is a react navigation specific concept, so may not be appropriate for screens

maciekstosio: changed in: https://github.com/react-navigation/react-navigation/pull/12204/commits/057e143e691c16f7a99edc622cf764a5ba3769a6

maciekstosio:
closed in favor of: https://github.com/react-navigation/react-navigation/pull/12225

PR #12217, opened 2024-10-28
feat: Detect edge to edge and bypass statusBarTranslucent / navigationBarTranslucent

zoontek:
**Motivation**

The future of Android is [edge-to-edge](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge), and to make the React Native developer experience seamless in this regard, the ecosystem needs to transition from “opaque system bars by default” to “edge-to-edge by default.”

To prevent library authors from implementing their own edge-to-edge solutions—which could interfere with other libraries—and because it’s not possible to reliably detect if edge-to-edge is already enabled on Android, we have collaborated with [Expo](https://expo.dev/) to create a library that handles this functionality and is detectable using a simple helper: [`react-native-is-edge-to-edge`](https://github.com/zoontek/react-native-edge-to-edge/tree/main/react-native-is-edge-to-edge).

This approach allows you to bypass certain options and props (in this case, `SceneView` `statusBarTranslucent` and `navigationBarTranslucent`), helping to avoid frustrating issues like, “My app starts in edge-to-edge, but the system bars reappear immediately.” (because `react-navigation` causes them to reappear)

This PR is intentionally lightweight, as `react-navigation` doesn’t seem to interfere with edge-to-edge when `statusBarTranslucent` and `navigationBarTranslucent` are set to `true`.

> [!NOTE]  
> I didn't update `statusBarBackgroundColor` / `navigationBarColor`, but note that those are [deprecated](https://developer.android.com/about/versions/15/behavior-changes-15#deprecated-apis) and should not be used anymore (and will likely to have no impact on Android 15+, targetting SDK 35)

**Test plan**

- Install [react-native-edge-to-edge](https://github.com/zoontek/react-native-edge-to-edge) on the example app.
- Replace the `StatusBar` call in `example/src/index.tsx` with `<SystemBars style={theme.dark ? 'light' : 'dark'} />` and the one in `example/src/screens/BottomTabs.tsx` with `<SystemBars style="light" />`
- Create an Expo build (react-native-edge-to-edge doesn't work on Expo go yet)

Screenshot 2024-10-28 at 21 50 12

### DIFF
--- a/packages/native-stack/package.json
+++ b/packages/native-stack/package.json
@@ -53,6 +53,7 @@
   },
   "dependencies": {
     "@react-navigation/elements": "^2.0.0-rc.26",
+    "react-native-is-edge-to-edge": "^1.1.5",
     "warn-once": "^0.1.1"
   },
   "devDependencies": {
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -136,10 +136,8 @@ const SceneView = ({
 
   if (__DEV__) {
     controlEdgeToEdgeValues({
-      navigationBarColor,
       navigationBarTranslucent,
       statusBarTranslucent,
-      statusBarBackgroundColor,
     });
   }
 
@@ -314,7 +312,7 @@ const SceneView = ({
       }
       homeIndicatorHidden={autoHideHomeIndicator}
       hideKeyboardOnSwipe={keyboardHandlingEnabled}
-      navigationBarColor={IS_EDGE_TO_EDGE ? undefined : navigationBarColor}
+      navigationBarColor={navigationBarColor}
       navigationBarTranslucent={IS_EDGE_TO_EDGE || navigationBarTranslucent}
       navigationBarHidden={navigationBarHidden}
       replaceAnimation={animationTypeForReplace}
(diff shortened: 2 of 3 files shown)


satya164:
Thank you for the PR. My aim is to keep native-stack as a thin layer that provides an API for React Navigation, and have as much of the logic and functionality we can in `react-native-screens` itself. So seems like it would be better to move this logic there?

It also mean if someone builds another native stack API around `react-native-screens`, they'll also benefit from these changes.

zoontek:
@satya164 I agree with you. I started by working on `react-native-screens`, but the contributing guide mention that it should be done on this repo. I guess this is an exception, will open a PR there.

satya164:
@zoontek but this PR is not adding a new API that needs to be exposed in `native-stack` package, so it's not necessary to make changes here

zoontek:
@satya164 There's at least one necessary change here: `headerTopInsetEnabled` should be `true` when edge-to-edge is enabled. For the rest, this can be delegated to `react-native-screens`

satya164:
> @satya164 There's at least one necessary change here: headerTopInsetEnabled should be true when edge-to-edge is enabled

Shouldn't it already happen since `topInset` will be non-zero when using edge to edge? If that's not happening then there are different fixes needed.

satya164:
But also that peace of code is working around incorrect value returned from safe area context on first render iirc, so when edge to edge is enabled by default and all apps have translucent statusbar, i hope such workarounds won't be needed.

satya164:
also cc @kkafar, i mentioned this before, but is there anyway native-stack can automatically set `headerTopInsetEnabled` natively without needing to pass it?

zoontek:
PR on the `react-native-screens` repo: https://github.com/software-mansion/react-native-screens/pull/2464

kkafar:
I've responded with my thoughts [under PR in `react-native-screens`](https://github.com/software-mansion/react-native-screens/pull/2464#issuecomment-2446317330`).

zoontek:
Closing this at it will be handled at `react-native-screens` level.

PR #12218, opened 2024-10-29
refactor: remove unstable screen style prop

maciekstosio:
Corresponding PR in react-native-screens: https://github.com/software-mansion/react-native-screens/pull/2419. 
It's a new PR created based on a suggestion from https://github.com/react-navigation/react-navigation/pull/12199 to move logic from react-navigation to react-native-screens. 

⚠️ https://github.com/react-navigation/react-navigation/pull/12214 needs to be merged first 

## Change

Instead of adding `unstable_screenStyle` that directly adds background on `Screen` component, we can pick the `backgroundColor` from `contentStyle` and apply it on `Screen` component, without exposing unstable prop.


satya164 (approved):
Thanks

PR #12232, opened 2024-11-06
feat(native-stack,iOS)!: change default value of `fullScreenGestureShadowEnabled` to `true`

kkafar:
**Motivation**

`simple_push` is intended to mimic the default transition, which uses the shadow.

In v3 we exposed this as a prop to allow opt-in, but now it should become the default.
At some point later on we might even consider deprecating this prop.

**Test plan**

`Test1072` in `FabricExample` in `react-native-screens` repo.


kkafar:
@satya164 I'll proceed here, let me know in case you have some remarks I'll address them later in separate PR.

PR #12233, opened 2024-11-06
docs(native-stack): update JSDoc on default animation duration

kkafar:
**Motivation**

In `react-native-screens` v4 we change default animation on iOS duration to 500ms.

See: 

* https://github.com/software-mansion/react-native-screens/pull/2477


kkafar:
We landed it on screens line. Last build tests & releasing

satya164 (approved):

PR #12246, opened 2024-11-10
fix: add missing screenLayout prop to the bottom tabbar navigator

dimonnwc3:

**Motivation**

`screenLayout` prop was missing in  BottomTabNavigator


satya164 (approved):
Thanks for the PR!

PR #12248, opened 2024-11-11
Fix: cannot remove child at index

r0b0t3d:
**Motivation**

In Android, when new arch enabled, I got the error **cannot remove child at index** when `replace` or `go back` from a screen with drawer layout.
Screenshot_1731308895

**Test plan**

Applied the change locally on my project and it won't throw above error anymore.


satya164:
Thanks for the PR. However, this is a bug in React Native Screens, so it needs to be fixed there instead of disabling this optimization.

https://github.com/software-mansion/react-native-screens/issues/2341

PR #12258, opened 2024-11-14
refactor: move unstable_sheetFooter rendering to `react-native-screens`

kkafar:
**Motivation**

The footer is currently broken, as in the effect of recent refactors it erroneously ended up being rendered as child as `ScreenStackItem`, thus due to internal implementation it is rendered under `ScreenContentWrapper`. However it has to be rendered as direct child of `Screen` component (the same as `HeaderConfig`). 
 
Under usual circumstances this would be a breaking change, however the prop is marked as `unstable`, therefore I'm proceeding. 

**Test plan**

`Test1649` in `FabricTestExample`.


kkafar:
Note for future:

> [!note]
> We can most likely remove `FooterComponent.tsx` file completely now

PR #12261, opened 2024-11-15
fix: the warning about key prop with spread operator

mirik999:
The warning:
(NOBRIDGE) ERROR  Warning: A props object containing a "key" prop is being spread into JSX: React keys must be passed directly to JSX without using spread.

key prop should come before the spread operator.


satya164:
> key prop should come before the spread operator.

Why?

Have you actually tested this?

Finetuned8:
I was trying to fix it and messed up more

dacitto (approved):

PR #12265, opened 2024-11-15
fix: `displayMode` description

KirillTregubov:
**Motivation**

I noticed while migrating from v6 to v7 that `labelVisible` was changed to `displayMode`. The migration guide and the elements documentation pages did not include this information, so I went ahead and [updated them](https://github.com/react-navigation/react-navigation.github.io/pull/1384) (the PR just landed).

However, I noticed while hovering in my IDE (VS Code) that the description for `displayMode` was not updated. I went ahead and updated it in the types to match the `headerBackButtonDisplayMode` description (it's the same property).

**Test plan**

There are no code changes, this is strictly related to TypeScript DX.

Before:
Screenshot of IDE Before

After:
Screenshot of IDE After

### DIFF
--- a/packages/elements/src/types.tsx
+++ b/packages/elements/src/types.tsx
@@ -273,8 +273,14 @@ export type HeaderBackButtonProps = Omit<HeaderButtonProps, 'children'> & {
    */
   truncatedLabel?: string;
   /**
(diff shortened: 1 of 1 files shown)


satya164 (approved):
Thank you!

PR #12290, opened 2024-11-23
feat(bottom-tabs): support large content viewer on iOS for better accessibility

bacarybruno:
**Motivation**
The [iOS Large Content Viewer](https://developer.apple.com/videos/play/wwdc2019/261/) provides enhanced accessibility by allowing elements like Tab Bars to remain readable for users with low vision. This feature is especially important for improving accessibility.

Currently, React Navigation does not expose properties that support implementing the iOS [UILargeContentViewerItem](https://developer.apple.com/documentation/uikit/uilargecontentvieweritem) protocol, which enables this functionality.

This PR introduces uses the tabBar label as large content viewer title and disables the fontScaling by default on iOS 13+ as the large content title will be displayed.

For reference, see the related PR in the React Native repository: [facebook/react-native#45903](https://github.com/facebook/react-native/pull/45903).

**Next steps (if this PR gets merged)**
- Implement the large content viewer on `@react-navigation/material-top-tabs`
- Support large content image (needs a PR on react-native repo first)

**Test plan**
Manual tests on the example project.

https://github.com/user-attachments/assets/ffe445c0-2708-4797-810f-4280d4baef7b


satya164 (changes requested):
Thanks for the PR!

I think instead of requiring the user to specify `accessibilityLargeContentTitle` explicitly, we should show the tab bar's label as the large content title by default.

The next question is whether this should be exposed in `options`. Does the native UIKit tab bar support overriding the large content title or does it always show the tab's label?

We should match the native behavior. So if it can't be overriden natively, we shouldn't expose an option for it either. If there a use case for the large content title to be something else other than the tab's label - it's always possible to acheive it by using the `tabBarButton` option.

bacarybruno:
@satya164 thanks for your review 🙏 I initially introduced a new prop to avoid applying this behaviour by default on all apps. But, since the native behavior is to display a large content title based on the tab bar title, I'm updating the PR to align with that 👍

bacarybruno:
I have few questions @satya164:
1️⃣ Since `tabBarLabel` can be empty or a function, would you consider using `title` as a fallback?
2️⃣ Setting `allowFontScaling` to false by default when largeContentTitle is available is feasible, but we would also need to adjust the allowFontScaling property for the badge. Would that work for you?

satya164:
> Since tabBarLabel can be empty or a function, would you consider using title as a fallback?

There's usage of `getLabel` function inside `renderLabel`. You can move it to the component's scope to something like:

```js
const labelString = typeof label === 'string' ? label : getLabel(...);
```

And use it for both `renderLabel` and large content title.

> adjust the allowFontScaling property for the badge.

Sure

bacarybruno:
@satya164 thanks for your help 🙏

I updated the PR to:
- Use the tabbar label as large content title
- Set `allowFontScaling` to false on iOS 13+ as the large content title will be displayed

satya164 (approved):
Thank you

PR #12303, opened 2024-11-26
fix: labelStyle prop removed from TabBar component (react-native-tab-view)

bradstiff:
The labelStyle prop was dropped from the TabBar component in the 4.0 version. It appears to be accidental as it is still defined in the Prop spec. 

This PR fixes that.

### DIFF
--- a/packages/react-native-tab-view/src/TabBar.tsx
+++ b/packages/react-native-tab-view/src/TabBar.tsx
@@ -354,6 +354,7 @@ export function TabBar<T extends Route>({
   renderTabBarItem,
   style,
   tabStyle,
+  labelStyle,
   layout: propLayout,
   testID,
   android_ripple,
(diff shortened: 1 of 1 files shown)


bradstiff:
The labelStyle prop is also still mentioned in the documentation, so really hoping it was accidentally removed. I need it!

satya164:
Thanks for the PR. `labelStyle` has been moved to `options`/`commonOptions`. The `labelStyle` in tab bar's props is a leftover.

xchwarze:
@satya164 I don't see any commonOptions implemented in the NPM sources that were installed, and I can't get options to really work either.

satya164:
@xchwarze if it doesn't work as documented then please open an issue
https://reactnavigation.org/docs/tab-view/#options

PR #12305, opened 2024-11-27
feat: add badgeStyle to TabBarItem

enomotodev:
**Motivation**

When specifying the badge prop for TabBarItem, I wanted to also customize the style (e.g., change the position from absolute to something else). However, this is not currently possible. To address this, I added a new badgeStyle prop to TabBarItem, allowing for custom styling of the badge.

### DIFF
--- a/packages/react-native-tab-view/src/TabBarItem.tsx
+++ b/packages/react-native-tab-view/src/TabBarItem.tsx
@@ -105,6 +105,7 @@ const TabBarItemInternal = <T extends Route>({
   defaultTabWidth,
   icon: customIcon,
   badge: customBadge,
+  badgeStyle,
   href,
   labelText,
   routesLength,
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR!

> I wanted to also customize the style (e.g., change the position from absolute to something else)

Please be more specific. What's your exact use case?

Changing the position from `absolute` would break positioning of other elements in the tab bar so that's not something I want to expose. For any other styling, I don't see anything preventing that - but hard to say without specific.

enomotodev:
@satya164 

Thank you for reviewing the PR.

スクリーンショット 2024-11-29 8 58 13

As shown in the image above, there's a requirement to place badges right next to the labels.

Luizrebelatto (approved):

satya164:
If you want the badge to show as part of the label I think it'll be simpler for you to render the dot in the label itself instead. Changing the positioning of the badge could easily break in a future update to the library if we do any refactors.

Finetuned8:
You can change it back I thought I was on my personal page still a little fresh I could probably find a badge somewhere

Finetuned8:
You can change it back or I will

PR #12307, opened 2024-11-27
fix: add pointerEvents box-none to Header container

hirbod:
**Motivation**

This fixes a problem on web where headerTransparent might overlap other elements and make it unresponsive. By adding box-none, this problem can easily be fixed (all the children containers already have that prop)

**Test plan**

In the browser

Screenshot 2024-11-27 at 21 37 27

The screenshot shows a site with `statusbarTranslucent`. The header overlaps `Edit Token`, even though there is no title or HeaderRight.


satya164 (approved):
Thanks for the PR

PR #12326, opened 2024-12-07
feat: provide gesture in DrawerGestureContext

gaearon:
**Motivation**

The drawer gesture can conflict with other gestures in the app. To coordinate multiple gestures, the app code can use APIs like `requireExternalGestureToFail`. However, since the gesture is hidden in the library, this is not currently possible without patching. This exposes the gesture to the app code so you can use `requireExternalGestureToFail` with it.

```js
import { DrawerGestureContext } from 'react-native-drawer-layout';

// ...
const drawerGesture = useContext(DrawerGestureContext)
const nativeGesture = Gesture.Native()
  .requireExternalGestureToFail(drawerGesture)

// ...

<GestureDetector gesture={nativeGesture}>
  <PagerView
    // ...
  />
</GestureDetector>
```

Note that the existing `configureGestureHandler` is not enough because we need it in the other direction: *other* gestures (deeper in the app) need access to it. It would not be possible to capture it in a closure because by the point `configureGestureHandler` runs, the owner component has already executed and passed the data down.

<s>I opted to use a render prop instead of actually creating a context in order to keep this addition zero-cost. If this was a context exposed by the library, we'd have to either memoize the `pan` gesture (to avoid it getting redefined on drawer re-renders) or we'd have to pay the price of React doing a context search traversal every time it changes. Seems unfortunate if the majority of this library's users don't care about the gesture. Render prop lets the app decide whether it cares.</s>

Re-done as context as requested in review. Added some memoization to avoid it changing on every render.

**Test plan**

Verify you're able to access `gesture` on mobile. On web, it should be `undefined`.

Verify you can use it with RNGH APIs as usual. For example, https://github.com/bluesky-social/social-app/pull/7007 fixes a longstanding problem with the Bluesky app — the drawer swipe barely worked on iOS. The reason it has been broken for the last two years is because this library did not expose the gesture, and we did not know that it needs to be coordinated.

### DIFF
--- a/packages/react-native-drawer-layout/src/views/Drawer.native.tsx
+++ b/packages/react-native-drawer-layout/src/views/Drawer.native.tsx
@@ -21,6 +21,7 @@ import Animated, {
 import useLatestCallback from 'use-latest-callback';
 
 import type { DrawerProps } from '../types';
+import { DrawerGestureContext } from '../utils/DrawerGestureContext';
 import { DrawerProgressContext } from '../utils/DrawerProgressContext';
 import { getDrawerWidth } from '../utils/getDrawerWidth';
 import {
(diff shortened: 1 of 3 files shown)


satya164 (changes requested):
Thanks for the PR.

I'd prefer to expose it in a context instead, similar to what we do for stack for consistency. In addition, wrapping it in context will make it easy to access it in both the sidebar content as-well-as children, but more importantly it is necessary to access it when using with the drawer navigator since there's no way to pass it in render prop there.

Memoizing seems fine to me. Does it make enough impact to worry about cost?

gaearon:
Ok, I’ll need to look closer at what it depends on. We just wouldn’t want to send a new value down on every re-render of this component. Since this component is usually at the top, and context search traversal is not free.

gaearon:
Pushed a context implementation.

packages/react-native-drawer-layout/src/views/Drawer.native.tsx
```diff
@@ -217,53 +222,76 @@ export function Drawer({
...
+
+        toggleDrawer(nextOpen, event.velocityX);
+        runOnJS(onGestureFinish)();
+      })
+      .activeOffsetX([-SWIPE_MIN_OFFSET, SWIPE_MIN_OFFSET])
+      .failOffsetY([-SWIPE_MIN_OFFSET, SWIPE_MIN_OFFSET])
+      .hitSlop(hitSlop)
+      .enabled(drawerType !== 'permanent' && swipeEnabled);
+
+    if (panGesture && configureGestureHandler) {
```
gaearon, line 272: It's important for this to be within the useMemo block since it could be mutating.

satya164 (approved):
Thank you

PR #12327, opened 2024-12-10
feat: export *NavigatorProps for each navigator

marklawlor:
**Motivation**

Fix the TypeScript error when trying to use inferred types with React Navigation navigators

```
 The inferred type of 'NativeStackNavigator' cannot be named without a reference to '../node_modules/@react-navigation/native-stack/lib/typescript/module/src/types'. This is likely not portable. A type annotation is necessary.
```

The inferred type of `NativeStackNavigator` references `NativeStackNavigatorProps` which is located in `@react-navigation/native-stack/lib/typescript/module/src/types` and not exposed by `package.json#exports`. 

This issue only affects apps using `moduleResolution: nodenext/bundler`

I have added additional types to support type inference for `NativeStack`, `Tab` & `Drawer`

**Issue Reproduction**

`npx create-expo-app@latest --template react-navigation/template`

- Ensure `moduleResolution: 'bundler'` is set in `tsconfig.json`
- Add these lines to `App.tsx`

```
const NativeStackNavigator = createNativeStackNavigator().Navigator;

export type A = typeof NativeStackNavigator & {
  foo: "bar";
};
```

- Run `npx tsx` in the root directory and observe error

### DIFF
--- a/packages/bottom-tabs/src/index.tsx
+++ b/packages/bottom-tabs/src/index.tsx
@@ -35,6 +35,7 @@ export type {
   BottomTabNavigationEventMap,
   BottomTabNavigationOptions,
   BottomTabNavigationProp,
+  BottomTabNavigatorProps,
   BottomTabOptionsArgs,
   BottomTabScreenProps,
 } from './types';
--- a/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx
+++ b/packages/bottom-tabs/src/navigators/createBottomTabNavigator.tsx
@@ -1,6 +1,5 @@
 import {
   createNavigatorFactory,
-  type DefaultNavigatorOptions,
   type NavigatorTypeBagBase,
   type ParamListBase,
   type StaticConfig,
@@ -13,23 +12,12 @@ import {
 } from '@react-navigation/native';
 
 import type {
-  BottomTabNavigationConfig,
   BottomTabNavigationEventMap,
   BottomTabNavigationOptions,
   BottomTabNavigationProp,
+  BottomTabNavigatorProps,
 } from '../types';
 import { BottomTabView } from '../views/BottomTabView';
-
-type Props = DefaultNavigatorOptions<
-  ParamListBase,
-  string | undefined,
-  TabNavigationState<ParamListBase>,
-  BottomTabNavigationOptions,
-  BottomTabNavigationEventMap,
-  BottomTabNavigationProp<ParamListBase>
-> &
-  TabRouterOptions &
-  BottomTabNavigationConfig;
 
 function BottomTabNavigator({
   id,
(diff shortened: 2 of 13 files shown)


Bartek532 (approved):

satya164 (changes requested):
The `XNavigationConfig` types are internal - and . What's exposed to the user is the `XView` components (e.g. `BottomTabView`).

So the exports should be for the props of those components (e.g. `BottomTabViewProps`, `NativeStackViewProps` etc.). If for some reason you need the `XConfig` type, you can exclude what you don't need from the props type.

Though the props should already be accessible to consumers with `React.ComponentProps<typeof BottomTabView>`, explicit export is fine.

You can change the `type Props =` to `type BottomTabViewProps =` etc. and then export in `index.tsx` along with the view.

packages/routers/src/index.tsx
```diff
@@ -24,5 +24,5 @@ export type {
   TabNavigationState,
   TabRouterOptions,
 } from './TabRouter';
-export { TabActions, TabRouter } from './TabRouter';
+export { type BackBehavior, TabActions, TabRouter } from './TabRouter';
```
satya164, line 27: This makes it seem like a generic type for all navigators, however this is specific to tab & drawer and we may add different back behavior for other navigators in future.

This should already be accessible with `TabRouterOptions['backBehavior']` which makes it more obvious that it's for tab router. So this export seems unnecessary. Please revert this change.

marklawlor:
My apologies, I missed a step in the reproduction stating that your `tsconfig.json` needs to be updated to emit declarations. 

This issue will not affect apps using React Navigation, only libraries that are reusing the navigators. The issue isn't with the libraries usage, rather the TypeScript compiler is unable to generate the declaration file due to inaccessible types. While I'm sure there are work-arounds to explicitly type this, your creating a workaround for the TypeScript compiler (the user isn't choosing to use these types) which seems like an anti-pattern.

Here is a reproduction repro https://github.com/marklawlor/react-navigation-navigator-props

You can close the repro and run `npx tsc` to observe the errors.

packages/routers/src/index.tsx
marklawlor: I believe you have misunderstood the issue. The user isn't accessing these types, they are being automatically generated by the TypeScript compiler. I cannot change how the compiler writes these types, hence options like `TabRouterOptions['backBehavior']` are not suitable

satya164 (changes requested):
According to the typescript error, it needs the type of `Navigator` components.

So it'd be best to export the type of the navigator component's props for each navigator (e.g. `BottomTabsimilar to native stack's `NativeStackNavigatorProps`).

The `XConfig` type is an internal type and and while exporting small parts of the props like `XConfig` and `BackBehavior`, it doesn't account for any future changes in the navigator's props that may not be exported. Exporting the navigator's props will avoid any more such error.

satya164 (approved):
Thanks!

PR #12332, opened 2024-12-11
feat: move freezing logic in native-stack to navigation

WoLewicki:
**Motivation**

PR connected to https://github.com/software-mansion/react-native-screens/pull/2566 moving the freezing logic to `react-navigation` since we have control over `preloaded` screens here.

**Test plan**

`TestFreeze` in `RNScreens` repo.


packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -500,11 +509,16 @@ export function NativeStackView({
             preloadedDescriptors[route.key] !== undefined &&
             descriptors[route.key] === undefined;
 
+          const shouldFreeze = isFabric()
+            ? !isPreloaded && !isFocused && !isBelowFocused
```
satya164, line 513: Interesting why we can't freeze the previous screen on Fabric. Maybe we can add a comment saying why?

WoLewicki:
Ok I updated PR in RNScreens and added same logic for rest of navigators. Would be nice if it was tested in all of them.

satya164 (approved):
It seems to work so I'm merging this

packages/native-stack/src/views/NativeStackView.native.tsx
hirbod: Maybe we should do the same for Paper as well? I noticed that many of NativeWind's active classes are also frozen while navigating. Watch the video, where we scale the image on press and it stays scaled until we return. The effect is cool sometimes, as it reminds me of cupertino lists keep the active item active until you return, but it's sometimes a bit weird too. 

https://github.com/user-attachments/assets/45e595c0-0b9a-412e-a619-a8ea72952357

and another example, what the Deposit button.

https://github.com/user-attachments/assets/c1c690cb-7a3f-43a9-9ef4-5c682b1a8a7a

satya164: I think the way it's currently described (freeze inactive screens), it's strange not freeze an inactive screen. Maybe this freeze feature needs a rethinking - though I think that's out of the scope of this PR.

hirbod: It is out of the scope, true, but that line reminded me of "this issue" and I just wanted to share it with you

hirbod: I've created a [discussion for it ](https://github.com/react-navigation/react-navigation/discussions/12337)

PR #12335, opened 2024-12-12
fix: flush state updates before `setPreventRemove` called

andrejpavlovic:
**Motivation**

**Test plan**

By flushing state updates inside `useLayoutEffect` we ensure `setPreventRemove` has the latest state before called inside `useEffect`.

### DIFF
--- a/packages/core/src/useScheduleUpdate.tsx
+++ b/packages/core/src/useScheduleUpdate.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { NavigationBuilderContext } from './NavigationBuilderContext';
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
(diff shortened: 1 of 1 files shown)


packages/core/src/useScheduleUpdate.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import { NavigationBuilderContext } from './NavigationBuilderContext';
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
```
satya164, line 4:
```suggestion
import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';

```

satya164 (approved):
Thanks for the PR. Guess we'll see if it breaks anything

PR #12341, opened 2024-12-13
ci: configure maestro tests on CI for iOS & Android

satya164:


szymonrybczak:
So to give an update here, tests are executing on CI for Android & iOS & Web. Native tests are failing because of actual tests (as of testing they're failing locally too).

PR #12387, opened 2025-01-11
feat: animatable font size and color

likeSo:

**Motivation**

This commit uses the `interpolate` method to support animated gradients of tab color and font size in `material-top-tab` and `react-native-tab-view`.

Active and inactive tabs usually have different colors and font sizes to show their active state, and I added a gradient process to make it smoother and more natural.

Changed:
`tabBarActiveTintColor`, `tabBarInactiveTintColor` now supports gradients.

Added:
`tabBarActiveFontSize`, `tabBarInactiveFontSize`, animatable.
`tabBarActiveFontWeight`, `tabBarInactiveFontWeight`, not animatable.

**Test plan**

On native platforms, the `fontSize` property does not support `useNativeDriver`, so currently if `tabBarActiveFontSize` or `tabBarInactiveFontSize` is passed, the `useNativeDriver` value will be automatically set to `false`. This change may cause differences between web and native platforms and may cause performance loss, so you can focus on testing this. The latest code effects can be seen in the "Material Top Tab" and "Tab View" in the example project.

Example videos:

https://github.com/user-attachments/assets/cc928743-58c8-4073-b645-27d78b9186e6

https://github.com/user-attachments/assets/0aeea75e-445b-461a-833d-1c15e1f4f69a

### DIFF
--- a/example/src/Screens/MaterialTopTabs.tsx
+++ b/example/src/Screens/MaterialTopTabs.tsx
@@ -23,7 +23,17 @@ const ChatScreen = () => <Chat bottom />;
 
 export function MaterialTopTabsScreen() {
   return (
-    <MaterialTopTabs.Navigator>
+    <MaterialTopTabs.Navigator
+      screenOptions={{
+        tabBarActiveFontSize: 18,
+        tabBarInactiveFontSize: 14,
+        tabBarActiveTintColor: '#FF0000',
+        tabBarInactiveTintColor: '#000000',
+        tabBarActiveFontWeight: 900,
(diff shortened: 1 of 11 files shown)


likeSo:
I'll fix these type errors later

likeSo:
`PanResponderAdapter` is missing type `useNativeDriver`, I'll fix this next week

likeSo:
So far, I have tested it on both web and Android iOS platforms. As for the above failed checks, it seems that the npm installation package is wrong, which seems to have nothing to do with my changes this time?

PR #12448, opened 2025-02-19
chore: add FormSheet example

kligarski:
## Description

Added formSheet example screen.

## Changes

- added formSheet example screen with buttons that open formSheets with different props:
  - Fit To Contents:
    - shows `fitToContents` prop,
  - Heights Steps:
    - shows `sheetAllowedDetents` prop,
  - No Rounded Corners:
    - shows `sheetCornerRadius` prop,
    - other examples use `sheetCornerRadius: 8` to show the difference (on default Android also uses no rounded corners),
  - Grabber Visible:
    - shows `sheetGrabberVisible` prop,
  - Custom Dimming:
    - shows `sheetLargestUndimmedDetentIndex` prop by disabling dimming for the first detent,
  - Initial Detent:
    - shows `sheetInitialDetentIndex` prop by showing last detent when opened,
  - Lock Expanding With Scrolling:
    - shows `sheetExpandsWhenScrolledToEdge`,
  - Sheet Elevation: 0, Sheet Elevation: 48
    - show different values for `sheetElevation` prop,
    - use no dimming to highlight elevation effect

There is a problem with TS, when adding a new screen - see comment below: https://github.com/react-navigation/react-navigation/pull/12448#issuecomment-2677830258.


kligarski:
There is a problem with TS I am unable to fix - when adding a new screen, after importing it and adding it to `SCREENS` in example/src/screens.tsx, `navigation.navigate(name)` in example/src/index.tsx reports an issue with the type of `name`.

satya164 (changes requested):
Does the `ScrollView` need to be imported from gesture handler?

Currently, the PR breaks due to this, as we have a bundler check to prevent gesture handler imports on the web.

Since sheets are not supported on web, probably makes sense to do `FormSheet.native.tsx` regardless, and have the web implementation be this:

```js
return (
    <Text>Not supported on this platform</Text>
    <Button action={CommonActions.goBack()}>Go back</Button>
  </View>
);
```

PR #12461, opened 2025-02-28
fix: header search not taking remaining space when title centered

ferretwithaberet:
**Motivation**
This problem occurs when using `headerTitleAlign` set to `"center"` when using the simple stack. The search should span the entire header.

**Test plan**
1. Open example app;
2. Go to Simple Stack > Navigate to feed > Replace with contacts;
3. Change `headerTitleAlign` to `"center"` in the code;
4. Check result.

Without fix:

With fix:


satya164 (approved):
Thanks for the PR

PR #12462, opened 2025-02-28
feat: add typing for onSubmitEditing

ferretwithaberet:
Allows sending `onSubmitEditing` to the underlying `TextInput` element. Useful when searching should happen after user hits the keyboard submit button.


satya164 (approved):
Thanks for the PR

PR #12463, opened 2025-02-28
feat: add onOpen handler

ferretwithaberet:
As the package already had `onClose`, I thought it would be useful to also have an `onOpen` hook.


satya164 (approved):
Thanks for the PR

PR #12464, opened 2025-02-28
feat: add enterKeyHint

ferretwithaberet:
Adds and passes `enterKeyHint` to header search's text input. Defaults to `"search"`.


satya164 (approved):
Thanks for the PR

PR #12479, opened 2025-03-13
feat: pass all TextInput props to search

ferretwithaberet:
**Motivation**
I had a need to pass `onSubmitEditing` to the stack navigator to get the input value after the user submits it using the phone's keyboard. This allow me to trigger my search logic after the user finished inputing his search string.

**Description**
Allows passing all `TextInput` props, except `value`, to the `TextInput` used by `HeaderSearchBarInternal`. The `onChange` and `onChangeText` props are reversed for compatibility with `react-native-screens`. This also defaults `enterKeyHint` to `"search"` to properly reflect the keyboard action.

**Test plan**
1. Open example app;
2. Go to Simple Stack > Navigate to feed > Replace with contacts;
3. Edit `headerSearchBarOptions` property in the source code to pass `TextInput` props;
4. Check result.

In the example app, the `onSubmitEditing` prop is passed for testing purposes. It launches an alert with the inputed string after submitting from the keyboard.

**Additional info**
Supersedes #12462 and  #12464.  If merged, close the other PRs.

### DIFF
--- a/example/src/Screens/SimpleStack.tsx
+++ b/example/src/Screens/SimpleStack.tsx
@@ -6,7 +6,7 @@ import {
   type StackScreenProps,
 } from '@react-navigation/stack';
 import * as React from 'react';
-import { Platform, ScrollView, StyleSheet, View } from 'react-native';
+import { Alert, Platform, ScrollView, StyleSheet, View } from 'react-native';
 
 import { COMMON_LINKING_CONFIG } from '../constants';
 import { Albums } from '../Shared/Albums';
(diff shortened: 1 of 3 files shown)


satya164 (changes requested):
Thanks for the pull request. I want to avoid passing all text input props to the search bar because it would be really hard not to break things people might be relying on if we made any changes.

It also changes the API surface greatly compared to Native Stack's search bar and we want them to make it easy to migrate.

Could you update the PR to include a smaller subset of props relevant to the search bar that you need, and other people may need instead?

ferretwithaberet:
> Thanks for the pull request. I want to avoid passing all text input props to the search bar because it would be really hard not to break things people might be relying on if we made any changes.
> 
> It also changes the API surface greatly compared to Native Stack's search bar and we want them to make it easy to migrate.
> 
> Could you update the PR to include a smaller subset of props relevant to the search bar that you need, and other people may need instead?

Then I will close this PR, as the other 2 PRs I made related to this were merged and those were the features I had a need for. I will open another PR with the subset of props that I find useful if needed.

PR #12483, opened 2025-03-17
fix: stop navigate and push from rearranging stack history

marklawlor:
**Motivation**

Fix the behaviour of `navigate` and `push` within the StackRouter to never rearrange history. The goal is to match the desired behaviour of `react-native-screens`, which does not allow a stack to be rearranged.

In the docs and CHANGELOG, it is stated that `navigate(screenName)` will stay on the current screen if the screen is already focused, otherwise push a new screen to the stack. This is later clarified that if a `getID()` is provided, then it will match on the result of `getID()`, not the name.

StackRouter does not seem to follow this logic. The current logic check if *any* route in the history matches, and if so *move* it to the current route with updated params. Otherwise push the new route. 

For `push` the history rearranging logic does not appear to be documented. 

**Test plan**

Test suite already covers the test scenarious. Updated tests with the new behaviour and added comments explaining the changes.

### DIFF
--- a/example/e2e/maestro/simple-stack.yml
+++ b/example/e2e/maestro/simple-stack.yml
@@ -39,4 +39,4 @@ name: Simple Stack
     text: 'Pop by 2'
     retryTapIfNoChange: false
 - assertVisible:
-    text: 'Contacts'
+    text: 'Article by Babel fish'
--- a/packages/routers/src/__tests__/StackRouter.test.tsx
+++ b/packages/routers/src/__tests__/StackRouter.test.tsx
@@ -540,10 +540,11 @@ test('ensures unique ID for navigate', () => {
     stale: false,
     type: 'stack',
     key: 'root',
-    index: 1,
+    index: 2,
     preloadedRoutes: [],
     routeNames: ['baz', 'bar', 'qux'],
     routes: [
+      { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
       { key: 'bar', name: 'bar' },
       { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
     ],
@@ -1727,11 +1728,12 @@ test('ensures unique ID for push', () => {
     stale: false,
     type: 'stack',
     key: 'root',
-    index: 1,
+    index: 2,
     preloadedRoutes: [],
     routeNames: ['baz', 'bar', 'qux'],
     routes: [
       { key: 'bar', name: 'bar' },
+      { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
       { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
     ],
   });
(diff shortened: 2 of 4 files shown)


packages/routers/src/__tests__/StackRouter.test.tsx
```diff
@@ -540,10 +540,11 @@ test('ensures unique ID for navigate', () => {
...
     key: 'root',
-    index: 1,
+    index: 2,
     preloadedRoutes: [],
     routeNames: ['baz', 'bar', 'qux'],
     routes: [
+      { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
       { key: 'bar', name: 'bar' },
       { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
     ],
```
marklawlor, line 550: The incoming route will have an id of `bar-a` which does not batch the current route's id: `bar`. Instead of the route at index 0 being moved, a new route should be pushed

packages/routers/src/__tests__/StackRouter.test.tsx
```diff
@@ -1727,12 +1728,13 @@ test('ensures unique ID for push', () => {
...
-    index: 1,
+    index: 2,
     preloadedRoutes: [],
     routeNames: ['baz', 'bar', 'qux'],
     routes: [
       { key: 'bar', name: 'bar' },
       { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
+      { key: 'bar-test', name: 'bar', params: { foo: 'a' } },
     ],
   });
```
marklawlor, line 1739: The `StackAction.push` should push a new route, even if the IDs match.

packages/routers/src/__tests__/StackRouter.test.tsx
```diff
@@ -2497,6 +2499,7 @@ test('handles screen preloading', () => {
     ],
     routeNames: ['baz', 'bar', 'qux'],
     routes: [
+      { key: 'qux-some', name: 'qux' },
       {
         key: 'bar-test',
```
marklawlor, line 2504: The `StackAction.push` should push a new route, even if the IDs match with a previous route in history.

example/e2e/maestro/simple-stack.yml
```diff
@@ -39,4 +39,4 @@ name: Simple Stack
     text: 'Pop by 2'
     retryTapIfNoChange: false
 - assertVisible:
-    text: 'Contacts'
+    text: 'Article by Babel fish'
```
marklawlor, line 42: Previously step 11 

```
- tapOn:
    text: 'Push article'
    retryTapIfNoChange: false
```

Would change the `article` screen from index 0 to index 2.

After this change, `article` would remain at index 0 and a new route will be pushed. It will have an index of 3 and be the current screen

PR #12492, opened 2025-03-18
feat: drawerToggleButton image option

prateeksure:

**Motivation**

Want a feature which could change the togglebutton image.

**Test plan**

You can now give props to the drawertogglebutton so that its image can change. I checked myself in my app through importing the edited file.


satya164 (approved):
Thanks for the PR

Issue #12468, opened 2025-03-03
TypeScript error after updating to RN 0.78


LA-Johan:
### Current behavior

```
@react-navigation/core/src/types.tsx:547:13 - error TS2503: Cannot find namespace 'JSX'.

547   render(): JSX.Element;
```

Coming from this line:

https://github.com/react-navigation/react-navigation/blob/2dc0ec3eaca7aa010d9934cabaa64562b265271a/packages/core/src/types.tsx#L547

In addition, since useRef now returns a nullable type it's unclear how to use some other methods in the library. 

https://github.com/react-navigation/react-navigation/blob/2dc0ec3eaca7aa010d9934cabaa64562b265271a/packages/native/src/useScrollToTop.tsx#L52

and 

https://github.com/react-navigation/react-navigation/blob/2dc0ec3eaca7aa010d9934cabaa64562b265271a/packages/elements/src/types.tsx#L33

### Expected behavior

Typescript should compile. 

### Reproduction

Use the above functions and call yarn tsc

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [x] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 7.0.14
| @react-navigation/bottom-tabs          | 7.2.0
| @react-navigation/native-stack         | 7.20
| react-native                           | 0.78
| npm or yarn                            | 1.22.22

oliverdolgener:
we also face the problem with nullable refs + useScrollToTop:

`const scrollView = useRef(null);`
`useScrollToTop(scrollView);`

is giving:
`Argument of type 'RefObject<null>' is not assignable to parameter of type 'RefObject<ScrollableWrapper>'.
  Type 'null' is not assignable to type 'ScrollableWrapper'.ts(2345)`

we are on the latest versions:
`"@react-navigation/bottom-tabs": "^7.2.1",`
`"@react-navigation/native": "^7.0.15",`

PR #12501, opened 2025-03-20
fix: fix JSX and useScrollToTop types for React 19 compatibility

teaualune:
**Motivation**

This PR addresses some TypeScript errors under React Native 0.78.x, namely React 19. They are also discussed in #12468 .

Specifically, 2 parts are modified in this PR:

1. Rewrite `JSX` into `React.JSX`: since React 19 `JSX` global namespace is removed ([ref](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript)). We need to use `React.JSX` instead.
2. Add `null` type into `ScrollableWrapper` type definition: React 19 removes `MutableRefObject` and makes all refs mutable ([ref](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument)). That makes the `ref` argument of `useScrollToTop` become `RefObject<ScrollView | null>` (using `ScrollView` as an example here) under React 19. In order to take this change into account, we can add `null` type into `ScrollableWrapper` def directly to solve this issue.

**Test plan**

Run `yarn typecheck` and ensure modified codes pass the test. Since those modifications are compatible with older React versions as well, we don't need to update `react` dependencies to 19.

### DIFF
--- a/packages/core/src/StaticNavigation.tsx
+++ b/packages/core/src/StaticNavigation.tsx
@@ -403,7 +403,7 @@ export function createComponentForStaticNavigation(
     );
   }
 
-  const items: (() => JSX.Element | null)[] = [];
+  const items: (() => React.JSX.Element | null)[] = [];
 
   // Loop through the config to find screens and groups
   // So we add the screens and groups in the same order as they are defined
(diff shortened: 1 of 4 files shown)


satya164 (approved):
Thanks for the PR

PR #12503, opened 2025-03-21
feat: add pageSheet presentation to native stack

kligarski:
**Motivation**

Since the release of iOS 18, screens with`presentation: 'modal'` changed size on devices with a bigger screen, such as an iPad - they became much smaller, which impacted applications of the library's users (see issue https://github.com/software-mansion/react-native-screens/issues/2549). 

To address the issue, `react-native-screens` will introduce new presentation type for native stack (`pageSheet`) that will allow users to use previous modal behavior (see PR https://github.com/software-mansion/react-native-screens/pull/2793). The behavior of `pageSheet` on Android will be the same as `modal`.

This PR contains changes required to support `pageSheet` in `react-navigation`. It will be also necessary to bump `react-native-screens` version.

**Test plan**

Open Stack Presentation and Modals example screens in example apps from `react-native-screens` and open a `pageSheet`.


kligarski:
@kkafar

Gregoirevda:
What's the status on this?

kligarski:
@kkafar It's ready but I think we need to bump `react-native-screens` version in order to fix TS warnings. `pageSheet` was added to `react-native-screens` in [this release](https://github.com/software-mansion/react-native-screens/releases/tag/4.11.0-beta.0) but 4.11.0 is still in beta - I guess that we want to wait for stable release.

kligarski:
@kkafar

PR #12521, opened 2025-03-30
fix: fix `navigationRef.getRootState()` returning an incomplete state during render

zetavg:
[`navigationRef.getRootState()`](https://reactnavigation.org/docs/navigation-container/#getrootstate) may return an incomplete state during the first render of nested navigators.

## Repro

https://github.com/zetavg/react-navigation-getRootState-bug-repro-202503

1. Run the app (`yarn start`), open web (`http://localhost:8081`).
2. Press the “Go to SubStack/Page via RESET” button.
3. See the displayed root state is incomplete: `routes[1].state` is missing.
4. Press “Re-render Page” button, see `routes[1].state` is there now.

### Context

`navigationRef.getRootState()` is:
* Provided [via `React.useImperativeHandle`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/BaseNavigationContainer.tsx#L230) in `BaseNavigationContainer`.
* [Defined as `keyedListeners.getState.root?.()` in `BaseNavigationContainer`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/BaseNavigationContainer.tsx#L172-L174).
* Where `keyedListeners.getState.root?.()` should be [registered](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L43) by a navigator which [isn’t nested under a route](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L19), by `useOnGetState()`, in `useNavigationBuilder()`.
* With is a function named [`getRehydratedState`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L21-L40), also defined in `useOnGetState()`, 

A thing that `getRehydratedState` will do, besides just returning the state from `getState()` (which in root should be provided by `BaseNavigationContainer`, with `useSyncState` which manages the store), is it will,
* As commented, [“avoid returning new route objects if it don't need to”](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L24-L33). 
  * (Also, in my understanding, that will make the function return the “derived” state that is actually being used to render instead just the state stored in the store, which [may be different](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useNavigationBuilder.tsx#L477-L483).)
  * That is done by going through all the routes, trying to call `getStateListeners[route.key]?.()`, and replacing the state (used for nested navigator) of the route with the state returned by `getStateListeners[...]()`.
    * `getStateListeners` is `keyedListeners.getState`, here `keyedListeners` is “scoped” since each `useNavigationBuilder` will have it’s own one for nested navigators under the current navigator.
      * Children navigators will register their keyedListeners.getState also with their `useOnGetState`, in [a `React.useEffect`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L42-L44).
    * What it achieves, basically, is making `getRehydratedState` a recursive function: for every route that contains a navigator, call the navigator’s `getRehydratedState` to get the state from that navigator until the leaf navigator - where all its routes don’t have a navigator.

### What causes the bug

* [`addKeyedListener?.('getState', ...)` is called inside a `React.useEffect`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.2.0/packages/core/src/useOnGetState.tsx#L42-L44) - which means that it will only be registered **after** the first render.
* So **during** the first render, `getStateListeners[route.key]?.()` will be `undefined`. [And we will end up with `{ ...route, state: undefined }`](https://github.com/react-navigation/react-navigation/blob/%40react-navigation/core%407.7.0/packages/core/src/useOnGetState.tsx#L26-L32).

## Test plan

Can use the [repro](https://github.com/zetavg/react-navigation-getRootState-bug-repro-202503) to verify this fixes the issue.

Can add a test case for this perhaps, but I'm not sure how to do it for now.

### DIFF
--- a/packages/core/src/useOnGetState.tsx
+++ b/packages/core/src/useOnGetState.tsx
@@ -25,7 +25,7 @@ export function useOnGetState({ getState, getStateListeners }: Options) {
     const routes = state.routes.map((route) => {
       const childState = getStateListeners[route.key]?.();
 
-      if (route.state === childState) {
+      if (!childState || route.state === childState) {
         return route;
       }
 


satya164 (changes requested):
Thanks for the PR.

However, this change is not correct.

Assuming `route.state` is always correct even though `childState` doesn't match can result in bugs. If `childState` exists but doesn't match `route.state`, it can mean the previous `route.state` was incorrect and got updated or removed - in which case returning `route.state` will return an incorrect value.

For example, change the nested state in your `RESET` to remove `stale: false` - then it becomes an incomplete state object - you can remove properties or have the wrong name for the screen (e.g. pass `Page2` instead of `Page`) that later gets fixed by rehydrate. This is by design - so you can persist state and restore it without breaking if navigation structure changed, or don't have to pass things like keys and index when you reset state etc.

But in this case `getRootState` returns this invalid state object. It's supposed to always return a valid state object so the contract is broken.

To get back to the original issue - in broad terms:

- Navigation is not complete until all the relevant screens and navigators are rendered
- In React, the child's render is called first, then the parent, so when the screen's render function is called, the navigator hasn't finished rendering etc.
- React Navigation updates state "after" the render finishes - this makes sure that any invalid state gets corrected, any new child navigators that are rendered get tracked, any navigators that didn't render get removed etc.

With the above in mind - it's impossible to do what you're doing - you can't get the final state before components finish rendering (i.e. the `useEffect` at the top-most component is called) - because the final state is the result of react navigation observing what got rendered during the navigation. React doesn't provide a mechanism to do it in any other way.

In addition, functions like `getRootState()` (or any `getX` functions in React Navigation) aren't meant to be used during render (unless you pair them with a listener, e.g. with `useSyncExternalStore`).

Not sure what your use case is, but if you need to get the final state inside of a screen, then you can add a state listener where you'll get notified when the state changes:

```js
useEffect(() => {
  rootNavigationRef.addListener('state', (e) => {
    console.log('state change', rootNavigationRef?.getRootState());
  });
}, [rootNavigationRef]);
```

I'm closing the PR because this behavior isn't something that can be changed - and it's due to how React itself works.

zetavg:
Thank you for the detailed and clear reply!

Got your point that `getRootState` should always return a valid state that complies with the current config.

## Context

I discovered this while debugging an issue in One and Expo Router, where the [`usePathname()`](https://github.com/expo/expo/blob/315240b9b05757c4cd7d63fbec64110f6ddceb6c/packages/expo-router/src/hooks.ts#L153-L155) hook ([Expo doc](https://docs.expo.dev/versions/latest/sdk/router/#usepathname)) may return an incorrect path on the first render.

Taking Expo Router as an example: internally, it’s [using `navigationRef.getRootState()` to sync state from the React Navigation store to another store managed by Expo Router](https://github.com/expo/expo/blob/315240b9b05757c4cd7d63fbec64110f6ddceb6c/packages/expo-router/src/global-state/router-store.tsx#L251-L259). When `usePathname` is called during render, it immediately syncs the state 0 but since `navigationRef.getRootState()` returns an incomplete state at that point in time, it may result in an entirely incorrect pathname. For example, on the `/sub-nav/blackberry` page, it may incorrectly return `/sub-nav/apple` instead of `/sub-nav/blackberry`, due to a missing sub-state and `getPathFromState` falling back to the first avaliable route as the default.

I’ve observed this causing bugs in apps built with One or Expo Router when users navigate back, since both frameworks dispatch a `RESET` in such cases.

So for the implemention of these hooks, it might be preferable to access a complete state - even if stale - so that we can have enough information to return a value that won’t be totally incorrect. If we got a stale state, at least we can return something that is closer to correct.

While using `addListener` is a possible solution, it does mean the hooks can’t return a value on the initial render, requiring components to re-render immediately afterward to receive the value, which may not be ideal.

## Further Questions

So it seems that this should be fixed in the upper framework level instead of in React Nativation.

To better understand the proper approach and future direction, I was hoping to ask - if you have the time to share your thoughts:

* Is there any API for accessing the internal store directly - such as `getState` and `subscribe` from `useSyncState`? (if that’s a reasonable approach for frameworks built on top of React Navigation)
* I get the sense that some of the limitations (like not being able to determine the final state before rendering all navigators in React) are due to the dynamic nature defining navigators. I’m curious if there are any plans to evolve the static configuration to address these constraints, or if it’s expected to remain a simplified wrapper around dynamic configuration - mainly for ease of use or for enabling compiler optimizations in the future?

satya164:
> it might be preferable to access a complete state - even if stale - so that we can have enough information to return a value that won’t be totally incorrect.
> If we got a stale state, at least we can return something that is closer to correct.

The thing is that there's no guarantee that it'll be closer to correct. It can be wildly incorrect - the most common scenario is state persistence, where the state could be completely different if navigation structure changes.

But closer to correct isn't correct. And I'd not recommend relying on such cases. It can break other assumptions and introduce subtle bugs.

> While using addListener is a possible solution, it does mean the hooks can’t return a value on the initial render, requiring components to re-render immediately afterward to receive the value, which may not be ideal.

Yes, but this is how the core architecture of the component-based API works. There's no way around that.

Everything in React Navigation is designed around these limitations. And I'd highly recommend that if you're building a framework on top of React Navigation, it should embrace the limitations. Otherwise, it's going to be a battle with workarounds, which will be fragile and hard to maintain.

> I’m curious if there are any plans to evolve the static configuration to address these constraints

The static configuration is for better DX, but it still builds on top of the core architecture. So it won't change the limitations of the core architecture. And even with the static configuration, there's some dynamism (such as `if` property).

> Is there any API for accessing the internal store directly

You can access it by calling `getState` with `navigation` in the root navigator. Here is a possible hook:

```js
const useRootState = () => {
  const navigation = useNavigation();

  let root = navigation;

  while (root) {
    const parent = root.getParent();

    if (parent) {
      root = parent;
    } else {
      break;
    }
  }

  const state = React.useSyncExternalStore(
    (onStoreChange) => {
      const unsubscribe = root.addListener('state', onStoreChange);

      return () => {
        unsubscribe();
      };
    },
    root.getState,
    root.getState
  );

  return state;
};
```

But I still want to emphasize that this is not a good idea.

In React Navigation, the way screens are intended to be used is they only deal with their own state (params), which is in the `route` object, or the navigator's state (which can be accessed with `navigation.getState()` or `useNavigationState`).

Apart from correctness, the primary design of these APIs is for performance. Accessing only their state means they don't re-render unnecessarily. But if you access the global state in every screen (or even some of the screens), any navigation in the app will re-render all of those screens - even if we assume this works.

> So for the implemention of these hooks

The only reliable way I can see a `usePathname` hook working is if it's only concerned with own and parent state - and not the child navigator state. A very rough implementation could look like this:

```js
const usePathname = () => {
  const root = useRootNavigation();
  const parentRouteKey = useContext(ParentRouteKeyContext);
  const route = useRoute();
  const { options } = React.useContext(LinkingContext);

  const replaceRouteState = (state) => {
    return {
      ...state,
      routes: state.routes.map((r) => {
        if (r.key === parentRouteKey) {
          return {
            ...r,
            state: {
              routes: [route],
            },
          };
        }

        return r;
      }),
    };
  };

  const getPathFromStateHelper = options?.getPathFromState ?? getPathFromState;

  return getPathFromStateHelper(
    parentRouteKey ? replaceRouteState(root.getState()) : { routes: [route] },
    options?.config
  );
};
```

What does this do?

- Gets the root navigator (same logic as the `useRootState` hook but just returns the navigation object or use the `navigationRef`).
- Gets the parent route's `key`, so it can locate where the current navigation state is in the tree. To get the parent `key`, you can use the `layout` prop on the navigator where you get the parent route and pass it down in a context.
- If the parent route doesn't exist, then we're at the root. So we can pass the state to `getPathFromState` directly.
- If the parent route exists, traverse the state tree and replace the state in the parent route with our own, then pass it to `getPathFromState`.
- For the state object, we pass `{ routes: [route] }` - this just means we're passing a state where our route is the focused route - this way `getPathFromState` will return the path for our route.

It uses `root.getState` in render, which could be problematic, but realistically, it should be fine since if the state structure changed, then our route would be unmounted anyway. Either way, this is a rough implementation but should convey the idea.

The main thing is, we can know state in parent - but not the child. So if a screen wants to know its own `path`, then that's possible since it only needs to know parents and not children.

satya164:
The hook `useStateForPath` is now available in the latest version of React Navigation.

PR #12541, opened 2025-04-09
fix: disable screen readers on stacked transparentModal screens for improved accessibility

theonetheycallneo:
**Motivation**

Web screen readers can see the components on the previous underlying screens.  
I noticed we already do this for iOS and Android via:
```
accessibilityElementsHidden={!focused}
importantForAccessibility={focused ? 'auto' : 'no-hide-descendants'}
```
So, simply adding the 3rd aria-hidden to improve web acessibility.

**Test plan**

1. Create 2 screens, 1 normal and 1 with `presentation: transparentModal`
- first screen should have 1+ text component(s)
- screen 2 can be anything simple
2. Open the transparentModal screen from the previous screen
3. Enable screen reader/voice over
4. Note: The elements on the 2nd screen are no longer visible by the screen reader

Related:
https://github.com/necolas/react-native-web/pull/2773


packages/elements/src/Screen.tsx
```diff
@@ -62,6 +62,7 @@ export function Screen(props: Props) {
 
   return (
     <Background
+      aria-hidden={!focused}
```
trcoffman, line 65: Is this an issue on web as well? or is this only a problem on native?

One thing I found working on the accessibility of one of our components is that simply putting `aria-hidden` on a parent element on web is problematic. If you have certain elements inside of this component like text inputs, the browser will log errors if you put aria-hidden on a parent element without it also being on all the individual children.

You have to put the `inert` attribute on the parent to actually hide the entire tree of children from the screen reader properly.

theonetheycallneo: Nice find, my man!  I've added it in the latest commit

satya164:
Thanks for the PR. This is part of a larger effort to replace all a11y related props with `aria` props, which are now supported in React Native so that we can support all platforms with the samecode. However I haven't been able to finalize it: https://github.com/react-navigation/react-navigation/pull/11848

If you'd like to take over that PR, then feel free. The biggest part left is to test the changes to see if it works, and rebase the PR against latest changes.

PR #12543, opened 2025-04-10
fix(native-stack): align headerBackTitle with iOS behavior 

maciekstosio:
**Motivation**

The iOS has a default behavior how to deal with back title label - it changes its size depending on available space. It didn't work for custom back title, as react-native-screens used custom back button item when custom back title was provided. This changes in https://github.com/software-mansion/react-native-screens/pull/2860, thus we can remove check on react-navigation side.

Corresponding docs update: https://github.com/react-navigation/react-navigation.github.io/pull/1422

**Test plan**

See `Test code and steps to reproduce` in https://github.com/software-mansion/react-native-screens/pull/2860 or test this snippet (very long custom back button title should be changed to `Back`):

<details>
<summary>Source code</summary>

```
import { ParamListBase } from '@react-navigation/native';
import {
  createNativeStackNavigator,
  NativeStackNavigationProp,
} from '@react-navigation/native-stack';


export function FinalScreen({
  navigation,
}: {
  navigation: NativeStackNavigationProp<ParamListBase>;
}) {
  return (
    <View style={{ flex: 1, backgroundColor: 'yellow', justifyContent: 'center', alignItems: 'center' }}>
      <Text>VOID</Text>
      <Button title="Pop to top" onPress={() => navigation.popTo('Home')} />
    </View>
  );
}

export function Home({
  navigation,
}: {
  navigation: NativeStackNavigationProp<ParamListBase>;
}) {
  return (
    <View style={{ flex: 1, backgroundColor: 'yellow' }}>
      <Button
        title="Open screen"
        onPress={() => navigation.navigate('Second')}
      />
    </View>
  );
}

export default () =>{

  return (
      <Stack.Navigator>
        <Stack.Screen name="First" component={Home} />
        <Stack.Screen name="Second" component={FinalScreen} options={{headerBackTitle: 'LongLongLongLongLong'}} />
      </Stack.Navigator>
  );
};

```
</details>

⚠️ Keep in mind that it requires changes from screens to work

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -155,8 +155,9 @@ export type NativeStackNavigationOptions = {
   headerBackVisible?: boolean;
   /**
    * Title string used by the back button on iOS.
-   * Defaults to the previous scene's title, "Back" or arrow icon depending on the available space.
-   * See `headerBackButtonDisplayMode` to read about limitations and customize the behavior.
+   * Defaults to the previous scene's title.
+   * On iOS the text might be shortened to "Back" or arrow icon depending on the available space, following native iOS behaviour.
+   * See `headerBackButtonDisplayMode` to read about limitations and interactions with other props. 
    * Use `headerBackButtonDisplayMode: "minimal"` to hide it.
    *
    * Only supported on iOS and Web.
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -157,7 +157,7 @@ export type NativeStackNavigationOptions = {
    * Title string used by the back button on iOS.
    * Defaults to the previous scene's title.
    * On iOS the text might be shortened to "Back" or arrow icon depending on the available space, following native iOS behaviour.
-   * See `headerBackButtonDisplayMode` to read about limitations and interactions with other props. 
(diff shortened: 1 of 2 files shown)


packages/native-stack/src/types.tsx
```diff
@@ -155,7 +155,8 @@ export type NativeStackNavigationOptions = {
   headerBackVisible?: boolean;
   /**
    * Title string used by the back button on iOS.
-   * Defaults to the previous scene's title.
+   * Defaults to the previous scene's title, "Back" or arrow icon depending on the available space.
+   * See `headerBackButtonDisplayMode` to read about limitations and customize the behavior.
```
kkafar, line 159: I think this is wrong, since react navigation supports this prop also on web and I believe there is no such behaviour as you describe implemented on web. 

What you could do here is either do not modify this description or specify that the "shortening behaviour" applies only to iOS. 

```suggestion
   * Defaults to the previous scene's title.
   * On iOS the text might be shortened to "Back" or arrow icon depending on the available space, following native iOS behaviour.
   * See `headerBackButtonDisplayMode` to read about limitations and interactions with other props. 
```

kkafar (changes requested):
Let's rephrase the comment & we should be good.

kkafar (approved):
We're good thanks. Let's proceed.

PR #12553, opened 2025-04-16
fix: browser history

adamgrzybowski:
**Motivation**
I found two bugs in the current handling of the browser history

**1. The index in createMemoryHistory may be out of sync, and that causes the items array to drop entries in some cases.**  

This causes the navigation state to be lost when going back and forward in the browser. From this moment, it is recreated from the URL instead of saved state objects. 

**2. The `history.get()` function was provided with the relative instead of the absolute position.**

```ts
if (
  nextIndex !== -1 &&
  nextIndex < currentIndex &&
  // We should only go back if the entry exists and it's less than the current index
  history.get(nextIndex - currentIndex)
)
```

If the `nextIndex < currentIndex`, then `nextIndex - currentIndex` is always negative, and this `if` will never be true. It should be `history.get(nextIndex)`

This causes problems with finding the right history entry to go back after a nested stack.

**Test plan**

**Bug 1 - index out of sync**

Besides the videos below, I created e2e tests

❌ before ❌ 

https://github.com/user-attachments/assets/45d639b9-0f9c-45bb-9253-ffd1129ad188

🟢 after 🟢 

https://github.com/user-attachments/assets/c7f7accf-2a8c-4536-904d-fc3468ce839e

**Bug 2 - wrong argument for get function**

❌ before ❌ 

https://github.com/user-attachments/assets/39c72ece-9608-47e0-8c38-ea5463b97992

🟢 after 🟢 

https://github.com/user-attachments/assets/1d157d47-be0c-4436-842b-862d6f6b4cde

### DIFF
--- a/example/e2e/tests/createMemoryHistory.test.ts
+++ b/example/e2e/tests/createMemoryHistory.test.ts
@@ -1,0 +1,28 @@
+import { expect, test } from '@playwright/test';
+
+test('pops to the proper screen after going back and forward in history', async ({
+  page,
+}) => {
+  await page.goto('/native-stack/article/gandalf');
+
+  const buttonToFeed = page.getByRole('button', {
+    name: 'Navigate to feed',
+  });
+
+  await buttonToFeed.click();
+
+  await page.goBack();
+
+  await page.goForward();
+
+  const buttonGoBack = page.getByRole('button', {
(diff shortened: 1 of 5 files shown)


satya164 (approved):
Thank you!

PR #12554, opened 2025-04-17
feat(elements): add forwardRef to HeaderButton and PlatformPressable components

erickreutz:
This PR adds React.forwardRef to the HeaderButton and PlatformPressable components to fix compatibility with Expo Router's <Link asChild /> component. Expo Router's Link component uses @radix-ui/react-slot internally, which requires proper ref forwarding to maintain the component chain. Without forwardRef, using HeaderButton within Link asChild breaks the component chain and causes functionality issues.

### DIFF
--- a/packages/elements/src/Header/HeaderButton.tsx
+++ b/packages/elements/src/Header/HeaderButton.tsx
@@ -1,49 +1,46 @@
-import React from 'react';
+import * as React from 'react';
 import { Platform, StyleSheet } from 'react-native';
 
 import { PlatformPressable } from '../PlatformPressable';
 import type { HeaderButtonProps } from '../types';
 
-export const HeaderButton = React.forwardRef<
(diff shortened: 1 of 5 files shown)


packages/elements/src/Header/HeaderButton.tsx
```diff
@@ -1,39 +1,51 @@
+import React from 'react';
```
satya164, line 1:
```suggestion
import * as React from 'react';
```

satya164 (changes requested):
Thanks for the PR. To avoid changing the git blame information for the component code, prefer the following pattern:

```js
function HeaderButtonInternal(...)

// ...

export const HeaderButton = React.forwardRef(HeaderButtonInternal);
```

satya164 (approved):
Thank you!

PR #12555, opened 2025-04-17
feat: allow setting bottom tab press color and press opacity (Android)

rnigro-rwb:
**Motivation**

Bottom Tabs Navigator has the navigation options of `headerPressColor` and `headerPressOpacity`, but setting these has no effect. 
 Because these options are exposed, I believe that it is a good feature to make them work.   This is an Android only change because the ripple effect is only on Android.

**Test plan**

Create a Bottom Tab Navigator with `headerPressColor` and `headerPressOpacity` set to some non-default value.  This can be done by adding the following to the [BottomTab.tsx](../blob/-/example/src/Screens/BottomTabs.tsx) `Tab.Navigator` `screenOptions` property

```javascript
headerPressColor: 'yellow',
headerPressOpacity: 0.5,
```

When running on an Android device, this will make the press have a ripple color that is not the default grey with an opacity that is not 1.

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -1,4 +1,7 @@
-import type { HeaderOptions } from '@react-navigation/elements';
+import type {
+  HeaderOptions,
+  PlatformPressable,
+} from '@react-navigation/elements';
 import type {
   DefaultNavigatorOptions,
   Descriptor,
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. However, `headerX` options are intended to customize the header. The tab bar has a different design from the header and can be customized to look completely different, so it doesn't make sense to use the colors specified for header to be applied to the tab bar items.

So I'm closing the PR.

You can customize your tab bar items with the `tabBarButton` option to suit your use case.

PR #12558, opened 2025-04-18
fix(useNavigation): queue navigation actions before container is ready

### LINKED ISSUE #12552: [BUG] Navigation is not ready for first useEffect callback in first screen in stack
### Current behavior

```
import { useState, useEffect } from 'react';
import { Text, SafeAreaView, StyleSheet } from 'react-native';
import { useNavigation, NavigationContainer } from '@react-navigation/native';

const AppStackNavigator = createNativeStackNavigator();

const Navigator = () => {
  return (
    <AppStackNavigator.Navigator>
      <AppStackNavigator.Screen name="Home" component={HomeScreen} />
      <AppStackNavigator.Screen name="Paywall" component={PaywallScreen} />
    </AppStackNavigator.Navigator>
  )
}

const HomeScreen = () => {
  const { navigate } = useNavigation();

  useEffect(() => {
    console.log('FIRST MOUNT OF HOME SCREEN AND TRY TO NAVIGATE TO PAYWALL');
    navigate('Paywall');
    console.log("NOTHING HAPPENED AFTER CALLING NAVIGATION")
  }, [navigate]);

  return (
    <Text>Home screen</Text>
  )
}

const PaywallScreen = () => {
  return (
    <Text>Paywall Screen</Text>
  )
}


  const [isLoaded, setIsLoaded] = useState(false);

  useEffect(() => {
    setTimeout(() => {
      setIsLoaded(true);
    }, 1000);
  }, []);

  return (
    <SafeAreaView style={styles.container}>
      <NavigationContainer 
        onUnhandledAction={(action) => {
          console.log('UNHANDLED NAVIGATION ACTION', action);
        }}
        onReady={() => {
          console.log("NAVIGATION IS READY CALLBACK");
        }}>
          {!!isLoaded && (
            <Navigator />
          )}
      </NavigationContainer>
    </SafeAreaView>
  );
}

  container: {
    flex: 1,
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
});

```

Console stack:
```
FIRST MOUNT OF HOME SCREEN AND TRY TO NAVIGATE TO PAYWALL
NOTHING HAPPENED AFTER CALLING NAVIGATION
NAVIGATION IS READY CALLBACK
```

### Expected behavior

Navigate works fine in useEffect callback for first screen in stack

### Reproduction

https://snack.expo.dev/@wanderbon/navigation-bug

### Platform

- [x] Android
- [x] iOS
- [x] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

- [ ] I've removed the packages that I don't use

| @react-navigation/native               | 7.1.6
| @react-navigation/bottom-tabs          | 7.3.10
| @react-navigation/native-stack         | 7.3.10
| react-native-screens                   | 4.10.0
| react-native-safe-area-context         | 5.3.0
| react-native-gesture-handler           | 2.25.0
| react-native-reanimated                | 3.16.7
| react-native-pager-view                | 6.7.0
| react-native                           | 0.77.2
| expo                                   | 52.0.42
| node                                   | 22.13.1
| yarn                            | 1.22.22

```
hermes - true
new architecture - false
```

suojae:
**Fixes #12552**

**Motivation**  
When a screen’s first `useEffect` fires before the `NavigationContainer` has called its `onReady` callback, any call to `navigation.navigate()` is silently dropped. This affects edge‐cases like redirecting from the very first screen.

**What’s changed**  
- Introduce a `wrapWhenReady` helper that checks `root.isReady()` and, if false, queues the original navigation action in a short `setInterval` poll until the container is ready.  
- In `useNavigation`, detect if `NavigationContainerRefContext` provides an `isReady` method, and wrap the five core methods (`navigate`, `dispatch`, `goBack`, `push`, `replace`) via `Array.reduce` into a new `wrappedNav` object.  
- Fall back to the original `nav` object when `isReady` is not available, preserving full backward compatibility.

### DIFF
--- a/packages/core/src/__tests__/useNavigation.test.tsx
+++ b/packages/core/src/__tests__/useNavigation.test.tsx
@@ -1,7 +1,10 @@
-import { beforeEach, expect, test } from '@jest/globals';
-import { render } from '@testing-library/react-native';
+import { beforeEach, expect, jest, test } from '@jest/globals';
+import { act, render } from '@testing-library/react-native';
+import * as React from 'react';
 
 import { BaseNavigationContainer } from '../BaseNavigationContainer';
+import { NavigationContainerRefContext } from '../NavigationContainerRefContext';
+import { NavigationContext } from '../NavigationContext';
 import { Screen } from '../Screen';
 import { useNavigation } from '../useNavigation';
 import { useNavigationBuilder } from '../useNavigationBuilder';
@@ -149,4 +152,47 @@ test('throws if called outside a navigation context', () => {
   };
 
   render(<Test />);
+});
+
+test('queues navigation actions before container is ready', () => {
+  jest.useFakeTimers();
+  const navigateMock = jest.fn();
+  const dispatchMock = jest.fn();
+  const goBackMock = jest.fn();
+  const pushMock = jest.fn();
+  const replaceMock = jest.fn();
+
+  const rootRef = {
+    isReady: () => false,
+    navigate: navigateMock,
+    dispatch: dispatchMock,
+    goBack: goBackMock,
+    push: pushMock,
+    replace: replaceMock,
+  };
+
+  const TestComponent = () => {
+    const navigation = useNavigation<any>();
+    React.useEffect(() => {
+      navigation.navigate('baz');
+    }, [navigation]);
+    return null;
+  };
+
+  render(
+    <NavigationContainerRefContext.Provider value={rootRef as any}>
+      <NavigationContext.Provider value={undefined as any}>
+        <TestComponent />
+      </NavigationContext.Provider>
+    </NavigationContainerRefContext.Provider>
+  );
+
+  expect(navigateMock).not.toHaveBeenCalled();
+
+  act(() => {
+    (rootRef as any).isReady = () => true;
+    jest.advanceTimersByTime(10);
+  });
+
+  expect(navigateMock).toHaveBeenCalledWith('baz');
 });
--- a/packages/core/src/useNavigation.tsx
+++ b/packages/core/src/useNavigation.tsx
@@ -24,6 +24,32 @@ export function useNavigation<
     );
   }
 
-  // FIXME: Figure out a better way to do this
-  return (navigation ?? root) as unknown as T;
+  const nav = (navigation ?? root) as unknown as T;
+
+  if (root && typeof (root as any).isReady === 'function') {
+    const wrapped = { ...(nav as object) } as any;
+    const methodsToWrap = ['navigate', 'dispatch', 'goBack', 'push', 'replace'];
+
+    methodsToWrap.forEach((method) => {
+      if (typeof (nav as any)[method] === 'function') {
+        const original = (nav as any)[method] as (...args: any[]) => any;
+        wrapped[method] = (...args: any[]) => {
+          if ((root as any).isReady()) {
+            original(...args);
+          } else {
+            const interval = setInterval(() => {
+              if ((root as any).isReady()) {
+                clearInterval(interval);
+                original(...args);
+              }
+            }, 10);
+          }
+        };
+      }
+    });
+
+    return wrapped as T;
+  }
+
+  return nav;
 }


satya164 (changes requested):
Thanks for the PR. However this is extremely hacky. Introducing timeout/intervals means it can result in improper order of navigation actions resulting in race conditions.

`useNavigation` is also not the only way to navigate. There shouldn't be any special logic here. If there is any queueing then it needs to be done in a single place (e.g. inside the container at the root) to preserve proper order of any navigation actions dispatched before app was ready, and should synchronously handle all of those actions on ready event instead of an interval.

Or React Navigation should print an error and result in `onUnhandledAction` being called if a navigation happens before it is ready.

suojae:
Sorry for the late reply 🥲

You're right — interval polling is too hacky. I’ll revert the hook changes and move queuing into `BaseNavigationContainer`, flushing synchronously before onReady without timers.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -174,18 +198,26 @@ export const BaseNavigationContainer = React.forwardRef(
...
-    const isReady = useLatestCallback(() => listeners.focus[0] != null);
+    const isReady = useLatestCallback(() => {
+      const ready = listeners.focus[0] != null;
+
+      if (ready && !isReadyRef.current) {
+        isReadyRef.current = true;
+      }
+
+      return ready;
+    });
```
satya164, line 220: This is a get function - it shouldn't be modifying values. `isReadyRef` also shouldn't be relying on some checking `isReady` first.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -122,20 +125,43 @@ export const BaseNavigationContainer = React.forwardRef(
...
+      }
+
+      if (!isReadyRef.current) {
+        pendingActionsRef.current.push(action);
+
+        if (process.env.NODE_ENV !== 'production') {
+          console.warn(
+            'The navigation container is not ready yet. The action has been queued and will be processed when the container is ready.'
+          );
+        }
```
satya164, line 157: If queueing will be a built-in "feature" then I'd rather not print warnings.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -122,20 +125,43 @@ export const BaseNavigationContainer = React.forwardRef(
...
     );
 
+    const dispatch = useLatestCallback((action: NavigationDispatchAction) => {
+      if (listeners.focus[0] == null) {
+        pendingActionsRef.current.push(action);
+        console.error(NOT_INITIALIZED_ERROR);
+        return;
+      }
+
+      if (!isReadyRef.current) {
```
satya164, line 150: Not sure I understand what this is doing. `isReadyRef` is based on `listeners.focus[0]`, but it's printing an error in one case, and a warning in other, and both cases it gets queued either way.

packages/core/src/BaseNavigationContainer.tsx
```diff
@@ -122,20 +125,43 @@ export const BaseNavigationContainer = React.forwardRef(
...
+    const isReadyRef = React.useRef(false);
+
+    const dispatchAction = useLatestCallback(
+      (action: NavigationDispatchAction) => {
         if (listeners.focus[0] == null) {
           console.error(NOT_INITIALIZED_ERROR);
+          return false;
         } else {
           listeners.focus[0]((navigation) => navigation.dispatch(action));
+          return true;
```
satya164, line 138: Who is using these return values?

satya164 (changes requested):
The issue mentions dispatching actions in the screen's `useEffect`, which would typically be using `useNavigation()` or the `navigation` prop, and not the container ref.

So these changes don't quite address the issue mentioned in the PR as this only happens when an action is dispatched via the ref, and queueing happens regardless of if a screen is rendered. 

If queueing is implemented in core, here is what should happen:

- Actions should only be queued when screens have started rendering but the container is not fully ready yet - so that an action from screen's `useEffect` gets queued.

  But if someone dispatches an action without rendering screens, it should instantly print the error.

I'd ask to first add a test demonstrating the issue reported in the PR. This is the most important part - to have a consistent way to reproduce this, so even if you don't manage to address it, someone else can in the future. 

Then, instead of dropping silently, as mentioned in the issue, update it to print the `NOT_INITIALIZED_ERROR` instead - and update the test to check for the error. The error could potentially be logged in `onDispatchAction` function in `BaseNavigationContainer` (but I'm not 100% sure, otherwise check `useOnAction` where it could check for ready state).

Then it'll be good enough to merge for now to address the issue of silently dropping the action.

Then, as a next step in a separate PR, the queueing behavior could be explored, this behavior could be implemented by keeping a list of pending actions in the container, but the code in `useOnAction` should deal with this.

Though I'm not confident that queue can be implemented reliably. In that case having the error instead of silently dropping is enough. Queue can always be implemented in the userland by checking `ref.onReady`

suojae:
Thank you for the thorough and thoughtful review. I wasn't able to find the time to address the remaining feedback properly, so I'm closing this PR for now. I hope someone can pick this up in the future. I'm so sorry

PR #12563, opened 2025-04-23
feat(react-native-tab-view): Expose `onPageSelected` for better tab-switch focus management

ikevin127:
**Motivation**

At [Expensify/App](https://github.com/Expensify/App) we've been struggling for a while with input focus management on iOS: Native after switching between tabs, here's [one of these issues](https://github.com/Expensify/App/issues/59388) for more context.

♻️ Some investigation [details](https://github.com/Expensify/App/issues/59388#issuecomment-2782655033) on what's happening under the hood specifically on iOS: Native:

> The key sequence is:
> 
> 1. User selects the #Room tab.
> 2. Transition animation begins (i.e., `setViewControllers:direction:animated:completion:`).
> 3. Our `useAutoFocusInput` hook assigns focus to the room name input field.
> 4. Upon animation completion, iOS internally calls `removeFromSuperview`, which causes the input to lose focus.

Exposing the `onPageSelected` navigator event method allows us to call focus **after right amount of time** in order to fix [our issue](https://github.com/Expensify/App/issues/59388), targeting the key sequence events described above.

##

ℹ️ We already opened a PR with a `patch` of this exact PRs changes targeting the `react-native-tab-view` package which is part of `react-navigation`, but I've been encouraged to open a PR upstream with the actual patch changes.

**Issue https://github.com/Expensify/App/issues/59388 summary**: On iOS devices, when a user initiates a flow from Tab 1 that navigates to a screen containing a `TextInput` (e.g., a form or chat interface), and then switches to Tab 2 (e.g., a room creation screen), the `TextInput` in Tab 2 unexpectedly loses focus shortly after mount. This results in premature validation errors because of the input onBlur event, even though the user hasn't interacted with the field yet.

##

**Use case**: Exposing the `onPageSelected` navigator event method allows us to pass a function and call focus **after right amount of time** in order to fix [our issue](https://github.com/Expensify/App/issues/59388).

##

**Test plan**

**Added new test coverage for the following:**
- ✅ `passes onPageSelected prop from TabView to Pager` (to existing test file: `index.test.tsx`)
- ✅ `calls onPageSelected when a tab is selected` (to existing test file: `index.test.tsx`)
- ✅ `maintains backward compatibility when onPageSelected is not provided` (to existing test file: `index.test.tsx`)
- ✅ `passes onPageSelected prop to ViewPager` (new test file: `PagerViewAdapter.test.tsx`)

cc @adamgrzybowski for visibility


satya164:
Please provide a summary of the issue and the use case in the PR description. Also, why isn't `onIndexChange` sufficient?

ikevin127:
Well, `onIndexChange` looks to do pretty much the same thing so let me check if using that would work for our issue 👍

ikevin127:
> Also, why isn't onIndexChange sufficient?

`onIndexChange` is only called when [`index !== navigationState.index`](https://github.com/react-navigation/react-navigation/blob/04a82f1a34ebc265c2dac3ebae94170efc3d837a/packages/react-native-tab-view/src/TabView.tsx#L89-L93). In our case, when `jumpToIndex` is called (click the tab), `index` is already equal to `navigationState.index`. If it could be called directly, then exposing `onPageSelected` wouldn’t be necessary.

##

@satya164 I updated the PR description 👍

satya164:
> in Tab 2 unexpectedly loses focus shortly after mount

Is this not a bug in `react-native-pager-view`?

Seems like you have a workaround, but it'd be better if the actual issue could be fixed instead of adding APIs for the workaround.

ikevin127:
> Is this not a bug in `react-native-pager-view`?

No, the issue is caused by our custom auto focus logic (`useAutoFocusInput` hook) which we apply on the `TextInput` component that's part of a tab page, as mentioned in OP:

> 3. Our `useAutoFocusInput` hook assigns focus to the room name input field.
> 4. Upon animation completion, iOS internally calls `removeFromSuperview`, which causes the input to lose focus.

We went with the approach of exposing `onPageSelected` because it offers the right amount of delay, which comes from the Pager component and is more than the delay offered by isFocused from navigation, it's the right amount to help us fix this issue on iOS: Native (quoted above).

We see the `onPageSelected` approach as less of a workaround then using `setTimeout` for example which we always had issues with -> we think that `onPageSelected` should be exposed since is already part of the navigator and can be used to facilitate fixing edge cases like ours without using timeouts.

packages/react-native-tab-view/src/PagerViewAdapter.tsx
```diff
@@ -3,6 +3,7 @@ import { Animated, Keyboard, Platform, StyleSheet } from 'react-native';
 import ViewPager, {
   type PageScrollStateChangedNativeEvent,
 } from 'react-native-pager-view';
+import type { NativeProps } from 'react-native-pager-view/lib/typescript/specs/PagerViewNativeComponent';
```
satya164, line 6: Don't import private files, e.g. `react-native-pager-view/lib/typescript`. You can get the type with `React.ComponentProps<typeof ViewPager>['onPageSelected']`

packages/react-native-tab-view/src/TabView.tsx
```diff
@@ -24,6 +25,7 @@ import type {
 
 export type Props<T extends Route> = Omit<PagerProps, 'layoutDirection'> & {
   onIndexChange: (index: number) => void;
+  onPageSelected?: NativeProps['onPageSelected'];
```
satya164, line 28: Rename `onPageSelected` to `onTabSelect` to match the terminology of the library

packages/react-native-tab-view/src/PanResponderAdapter.tsx
```diff
@@ -24,6 +25,7 @@ type Props<T extends Route> = PagerProps & {
   layout: Layout;
   onIndexChange: (index: number) => void;
   navigationState: NavigationState<T>;
+  onPageSelected?: NativeProps['onPageSelected'];
```
satya164, line 28: This prop should be implemented in `PanResponderAdapter` so that using this API works on platforms other than Android & iOS, not only type

packages/react-native-tab-view/src/__tests__/PagerViewAdapter.test.tsx
```diff
@@ -0,0 +1,85 @@
+import { expect, jest, test } from '@jest/globals';
```
satya164, line 1: This test isn't necessary. The tests for the public API should already use appropriate behavior. So remove this.

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -1,10 +1,60 @@
...
 import { fireEvent, render } from '@testing-library/react-native';
 import React from 'react';
 import { type GestureResponderEvent, View } from 'react-native';
+import type { NativeProps } from 'react-native-pager-view/lib/typescript/specs/PagerViewNativeComponent';
 
 import { SceneMap, TabView } from '../index';
 
+// We need to mock the Pager module before importing TabView
+// to ensure our tests have proper control
+jest.mock('../Pager', () => {
```
satya164, line 11: Don't mock `Pager` with custom implementation - there's already a JS-only implementation (`Pager.tsx`) which can be used for more accurate tests

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -49,3 +102,71 @@ test('switches tabs when the tab header for another route is pressed', () => {
...
   expect(getByTestId('route2')).toBeTruthy();
 });
+
+test('passes onPageSelected prop from TabView to Pager', () => {
+  const pagerMock = require('../Pager');
+  render(<ComponentWithTabView />);
+
+  // Verify onPageSelected was passed to the Pager
+  expect(pagerMock.__mockOnPageSelected).toBeDefined();
+});
```
satya164, line 112: Avoid testing implementation details. They should only cover public API behavior.

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -49,3 +102,71 @@ test('switches tabs when the tab header for another route is pressed', () => {
...
+
+test('calls onPageSelected when a tab is selected', () => {
+  // Reset the mocks before this test
+  require('../Pager').__resetMocks();
```
satya164, line 116: Call `jest.resetAllMocks` in `beforeEach` instead.

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -49,3 +102,71 @@ test('switches tabs when the tab header for another route is pressed', () => {
...
+
+  // Verify onPageSelected was called with the correct event
+  expect(onPageSelectedMock).toHaveBeenCalledTimes(1);
+  expect(
+    (
+      onPageSelectedMock.mock.calls[0][0] as {
+        nativeEvent: { position: number };
+      }
+    ).nativeEvent
+  ).toHaveProperty('position', 1);
```
satya164, line 137: Use `toMatchObject` & `expect.objectContaining` with `toHaveBeenCalledWith`  to check specific properties instead of accessing `.mock` property directly

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -49,3 +102,71 @@ test('switches tabs when the tab header for another route is pressed', () => {
...
+});
+
+test('maintains backward compatibility when onPageSelected is not provided', () => {
```
satya164, line 140: We don't need this test

packages/react-native-tab-view/src/__tests__/index.test.tsx
```diff
@@ -49,3 +102,71 @@ test('switches tabs when the tab header for another route is pressed', () => {
...
+  onPageSelectedMock.mockReset();
+
+  // The initial render should not trigger onPageSelected
+  expect(onPageSelectedMock).not.toHaveBeenCalled();
+
+  // Simulate selecting the second tab
+  fireEvent.press(getByLabelText('Second'), {} as GestureResponderEvent);
+
+  // Verify onPageSelected was called with the correct event
+  expect(onPageSelectedMock).toHaveBeenCalledTimes(1);
```
satya164, line 130: These comments can be removed, they don't say anything that's not clear from reading the code

packages/react-native-tab-view/src/__tests__/index.test.tsx
ikevin127: Added:

```ts
jest.mock('../Pager', () => {
  return jest.requireActual('../Pager');
});
```

But if we just mock pager like this the test will fail at `expect(onTabSelectMock).toHaveBeenCalledTimes(1)` because the `onTabSelect` callback isn't being properly called. Explanation:
1. In the TabView component, the `onTabSelect` prop is passed to the Pager component as `onPageSelected`
2. The Pager component (which resolves to PagerViewAdapter) passes this prop to the underlying ViewPager component
3. When you comment out the mock, Jest can't actually use the real ViewPager component (which is a native module)
4. Without the mock, when the test tries to fire events, the `onPageSelected` callback isn't triggered properly

To fix this I added the mock implementation (see below)  which provides a crucial part: it simulates the `onPageSelected` event when `setPage` or `setPageWithoutAnimation` is called. When a tab is pressed in the test, this mock ensures that the event is properly triggered, which then calls the `onTabSelect` callback.

**Note**: This type of mock was done previously in [`material-top-tabs/index.test.tsx`](https://github.com/react-navigation/react-navigation/blob/8a2451457b7fc2a0b58be3323693af617a16dab0/packages/material-top-tabs/src/__tests__/index.test.tsx#L18-L30).

```ts
jest.mock('react-native-pager-view', () => {
  const React = require('react');
  const { View } = require('react-native');

  return class MockViewPager extends React.Component {
    // eslint-disable-next-line @eslint-react/no-unused-class-component-members
    setPage = (index: number) => {
      if (this.props.onPageSelected) {
        this.props.onPageSelected({
          nativeEvent: { position: index },
        });
      }
    };

    // eslint-disable-next-line @eslint-react/no-unused-class-component-members
    setPageWithoutAnimation = (index: number) => {
      if (this.props.onPageSelected) {
        this.props.onPageSelected({
          nativeEvent: { position: index },
        });
      }
    };

    render() {
      return <View>{this.props.children}</View>;
    }
  };
});
```

packages/react-native-tab-view/src/__tests__/index.test.tsx
ikevin127: 🟢 I **removed** this mock reset as it was not necessary. Replacing it with `jest.resetAllMocks` in `beforeEach` causes all tests in `index.test.tsx` (including existing) to fail with:

```
Trying to detect host component names triggered the following error:

Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `ForwardRef(Switch)`.

There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
Please check if you are using compatible versions of React Native and React Native Testing Library.
```

ikevin127:
@satya164 Applied requested changes and cleaned up previous unnecessary code. Let me know if there's anything else 👍

packages/react-native-tab-view/src/PanResponderAdapter.tsx
```diff
@@ -71,13 +75,25 @@ export function PanResponderAdapter<T extends Route>({
...
   const swipeVelocityThreshold = 0.15;
   const swipeDistanceThreshold = layout.width / 1.75;
 
+  const triggerPageSelected = useLatestCallback((index: number) => {
+    if (onPageSelectedRef.current) {
+      const event = {
+        nativeEvent: {
+          position: index,
+        },
+      } as NativeSyntheticEvent<{ position: number }>;
```
satya164, line 92: Avoid `as`. This type doesn't match `NativeSyntheticEvent`, so it shouldn't say that it's a `NativeSyntheticEvent`.

Either match `NativeSyntheticEvent` or update the types to be a subset that works with both pan responder adapter and pager view adapter.

It maybe just better to do `onTabSelect: ({ index: number }) => void`

packages/react-native-tab-view/src/__tests__/index.test.tsx
satya164: It should use the `PanResponderAdapter`, not `PagerViewAdapter`.

This doesn't do anything - as it mocks `Pager` as `Pager` itself (which is resolved to `Pager.ios.tsx`:

```js
jest.mock('../Pager', () => {
  return jest.requireActual('../Pager');
});
```

To use `Pager.tsx`, you need to specify the extension:

```js
jest.mock('../Pager', () => jest.requireActual('../Pager.tsx'));
```

> To fix this I added the mock implementation

This only tests pager view and not the pan responder implementation.

To test both `PanResponderAdapter` and `PagerViewAdapter`, all the tests could be wrapped in something like:

```js
describe.each([
  { type: 'native' },
  { type: 'js' }
])('$type implementation', ({ type }) => {
  if (type === 'native') {
    jest.mock('react-native-pager-view', () => {
      // ...
    });
  } else {
    jest.mock('../Pager', () => jest.requireActual('../Pager.tsx'));
  }

  afterAll(() => {
    jest.mockRestore();
  });

  // rest of the tests
});
```

satya164 (changes requested):

packages/react-native-tab-view/src/PanResponderAdapter.tsx
ikevin127: Thanks, just pushed a commit for this -> went with `onTabSelect: ({ index: number }) => void`.

packages/react-native-tab-view/src/PanResponderAdapter.tsx
```diff
@@ -71,13 +74,19 @@ export function PanResponderAdapter<T extends Route>({
...
-
+  const onTabSelectRef = React.useRef(onTabSelect);
   const currentIndexRef = React.useRef(index);
   const pendingIndexRef = React.useRef<number>();
 
   const swipeVelocityThreshold = 0.15;
   const swipeDistanceThreshold = layout.width / 1.75;
 
+  const onTabSelectHandler = useLatestCallback((tabIndex = index) => {
+    InteractionManager.runAfterInteractions(() => {
```
ikevin127, line 85: Added `runAfterInteractions` to match the delay of the native `onPageSelected` implementation.

satya164: I've removed it for now since `InteractionManager` is also affected by any other animations or gestures on the page that may not be related to the tab view. If you have any specific scenarios where it's a problem then we can investigate it separately.

satya164 (approved):
Thank you

PR #12600, opened 2025-05-15
Allow Overlay to be focusable in iOS with Full Keyboard Access enabled

ckknight:
**Motivation**

Currently, if you are using Full Keyboard Access in iOS, there is no way to close the drawer once it has been opened, as the Pressable within the Overlay is not focusable.

**Test plan**

in iOS, with Full Keyboard Access enabled at the OS level (System Settings -> Accessibility -> Keyboards and Typing -> Full Keyboard Access -> On)

- Open a drawer
- Try to navigate using only the keyboard to the Overlay to close it
- Press space to close the Overlay


satya164 (approved):
Thanks for the PR

PR #12604, opened 2025-05-17
fix: error thrown when applying transform style to tabBarStyle

Mhp23:

**Motivation**

When trying to apply a `transform` style to `tabBarStyle` in the bottom tabs navigator, the following error is thrown:

```bash
 (NOBRIDGE) ERROR  Warning: Invariant Violation: Transform with key of "translateY" must be number or a percentage. Passed value: {"translateY":0}
```
For example, with this usage:

```tsx
//...
export const Tab = createBottomTabNavigator();
//...
return (
  <Tab.Navigator screenOptions={{tabBarStyle: {
     transform: [{scaleX: -1}],
   }}}>
   {/* ... */}
```

We'll get the error mentioned above. This PR enables applying a custom transform style to the tab bar.

**Test plan**

- Extracting the `transform` style from `tabBarStyle` and merging it with the existing transforms (when not in sidebar mode)
- Merging it with the existing transforms (when not in sidebar mode)
- Applying the remaining styles to the tab bar (in the sidebar mode, applies `tabBarStyle` directly)

Env: RN v0.79.2

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -304,7 +304,7 @@ export function BottomTabBar({
 
   const tabBarBackgroundElement = tabBarBackground?.();
 
-  const {transform = [], ...restTabBarStyle} = (tabBarStyle ?? {}) as ViewStyle;
+  const { transform = [], ...restTabBarStyle } = (tabBarStyle ?? {}) as ViewStyle;
 
   return (
     <Animated.View
--- a/packages/bottom-tabs/src/views/BottomTabBar.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabBar.tsx
@@ -304,7 +304,8 @@ export function BottomTabBar({
 
   const tabBarBackgroundElement = tabBarBackground?.();
 
-  const { transform = [], ...restTabBarStyle } = (tabBarStyle ?? {}) as ViewStyle;
+  const { transform = [], ...restTabBarStyle } = (tabBarStyle ??
+    {}) as ViewStyle;
 
   return (
     <Animated.View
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Passing `transform` should just override the internal style and not result in an error. 

But there doesn't seem to be anything wrong with tab bar's logic. There seems to be a bug in `createAnimatedComponent` that results in the transforms being validated incorrectly.

The proper fix would be to identify what's causing this, and potentially open an issue in React Native repo with a repro.

Mhp23:
Yes, I agree there appears to be an issue with RN Animated’s internal overriding of the transform style. However, since my changes preserve the transform functionality by handling and applying it exclusively, I went ahead and patched it locally and submitted a PR.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -304,6 +304,9 @@ export function BottomTabBar({
 
   const tabBarBackgroundElement = tabBarBackground?.();
 
+  const { transform = [], ...restTabBarStyle } = (tabBarStyle ??
+    {}) as ViewStyle;
```
satya164, line 308: Don't use `as` to ignore types. The type isn't `ViewStyle` but `StyleProp<ViewStyle>` - which can be a nested array and contain booleans and falsy values. Destructuring like this won't work in all scenarios.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -362,6 +368,7 @@ export function BottomTabBar({
                       ],
                     }),
                   },
+                  ...(transform as []),
```
satya164, line 371: We don't merge nested style properties in the library. Any passed properties should override the previous properties.

Also, `[]` is not the type of `transform`. Don't use `as` to specify an incorrect type.

satya164 (changes requested):
I'd rather have this bug fixed in React Native than add an workaround in React Navigation, especially since it requires React Navigation to process incoming styles first before style properties can be read as done in this diff.

PR #12624, opened 2025-06-02
feat: add scrollable bottom tabs

PaulOskarSoe:
**Motivation**

This PR introduces scrollable functionality to react-navigation bottom tabs to address the common issue where apps with many tabs create an overcrowded and unusable bottom navigation experience.

**Testing**
Feature can be tested from BottomTabs example screen
https://github.com/user-attachments/assets/2b3c64be-a5d6-4891-a3ac-a608540b089b

### DIFF
--- a/example/src/Screens/BottomTabs.tsx
+++ b/example/src/Screens/BottomTabs.tsx
@@ -9,6 +9,7 @@ import {
   HeaderBackButton,
   HeaderButton,
   PlatformPressable,
+  Text,
   useHeaderHeight,
 } from '@react-navigation/elements';
 import {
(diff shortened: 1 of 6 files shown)


okwasniewski:
I don't think tabs on mobile are meant to be scrollable. Google specifically outlined not to do that in their Material 3 docs:

Screenshot 2025-06-02 at 15 22 34

PaulOskarSoe:
> I don't think tabs on mobile are meant to be scrollable. Google specifically outlined not to do that in their Material 3 docs:
> Screenshot 2025-06-02 at 15 22 34

Understood. This might be a more specific use case rather than an overall best practices approach

satya164:
Thanks for the PR. However, as @okwasniewski mentioned, this goes against various platform guidelines. So I will close the PR.

PR #12627, opened 2025-06-05
Fix: react-native-tab-view glitching on initial load

adamgrzybowski:

**Motivation**

While working for Expensify, we encountered a bug with glitching material top tabs. After investigation, it turned out that the `react-native-tab-view` causes it. 

It happens on refresh or initial load of the page, when any tab except the first one is selected.

Disclaimer: In our case, we use the `position` value to animate the top bar. 

The same effect can be reproduced in an example app with material top tabs. Hover on the example app, it is not as visible as in the production app due to the performance.

In the example app, it was visible for one or two frames.

https://github.com/user-attachments/assets/a285cce4-333d-4294-854f-a82057820d42

The root cause is:
- `layout.width` determines if we should use a layout that uses the `position` and `translateX` animated values.
- Both `position` and `translateX` are derived from `panX`
- There is a brief moment where `layout.width` is known, so we use `position` and `translateX`, but the `panX` isn't set yet and is still 0.

I changed the code to make sure that switching to the layout that uses `position` and `translateX` is after the `panX` value is updated

After fixing:

https://github.com/user-attachments/assets/f1eeda37-8861-4be7-8d4f-237a7438446a

**Test plan**

See if the glitching occurs

### DIFF
--- a/packages/react-native-tab-view/src/PanResponderAdapter.tsx
+++ b/packages/react-native-tab-view/src/PanResponderAdapter.tsx
@@ -117,11 +117,51 @@ export function PanResponderAdapter<T extends Route>({
     onTabSelectRef.current = onTabSelect;
   });
 
+  const [position, setPosition] =
+    React.useState<Animated.AnimatedDivision<number> | null>(null);
+
+  const [translateX, setTranslateX] =
+    React.useState<Animated.AnimatedMultiplication<number> | null>(null);
+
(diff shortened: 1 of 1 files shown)


adamgrzybowski:
@satya164 resurrecting this PR after a break 😄 I tried your solution with `useLayoutEffect`, and it works like a charm 🪄 Thanks for this suggestion!

PR #12631, opened 2025-06-06
feat: prevent freezing modal screens on iOS

kligarski:
To be merged after https://github.com/react-navigation/react-navigation/pull/12503.

**Motivation**

iOS modals are presented in a seperate hierarchy. When you open a modal screen and push multiple `presentation: 'card'` screens (they are pushed "behind" the modal), the modal screen gets frozen, because it isn't `preloaded`, `focused` nor `belowFocused` (pushed screens are above the modal in `routes` list). You can't interact with the content of the modal, you can only dismiss it.

| before | after |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/b8f22ec4-9072-45da-877d-fe2c9ae9f045" alt="before" /> | <video src="https://github.com/user-attachments/assets/b5914f9c-99e3-4417-a2b7-54008a8ed821" alt="after" />

In this PR, we prevent freezing modal screens on iOS in order to avoid this problem.

The reasoning behind this change is that you shouldn't show more that 1-2 modal screens at the same time - this is also a recommendation in [Apple's Human Interface Guidlines](https://developer.apple.com/design/human-interface-guidelines/modality):
> **Let people dismiss a modal view before presenting another one.** Allowing multiple modal views to be visible at the same time tends to create visual clutter and can make your app seem scattered and disorganized. People need to remember the context they were in before a modal view appears, so presenting multiple views adds to people’s cognitive load, especially when a modal view hides another one by appearing on top of it. Although an alert can appear on top of all other content — including other modal views — you never want to display more than one alert at the same time.

**Test plan**

Run `Test791` in `react-native-screens`' example app.

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -173,8 +173,11 @@ const SceneView = ({
   const insets = useSafeAreaInsets();
   const frame = useSafeAreaFrame();
 
-  // `modal` and `formSheet` presentations do not take whole screen, so should not take the inset.
-  const isModal = presentation === 'modal' || presentation === 'formSheet';
+  // `modal`, `formSheet` and `pageSheet` presentations do not take whole screen, so should not take the inset.
+  const isModal =
+    presentation === 'modal' ||
+    presentation === 'formSheet' ||
+    presentation === 'pageSheet';
 
   // Modals are fullscreen in landscape only on iPhone
   const isIPhone = Platform.OS === 'ios' && !(Platform.isPad || Platform.isTV);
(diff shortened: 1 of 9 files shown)


kligarski:
@kkafar

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -509,8 +512,11 @@ export function NativeStackView({
           // On Fabric, when screen is frozen, animated and reanimated values are not updated
           // due to component being unmounted. To avoid this, we don't freeze the previous screen there
           const shouldFreeze = isFabric()
-            ? !isPreloaded && !isFocused && !isBelowFocused
-            : !isPreloaded && !isFocused;
+            ? !isPreloaded &&
+              !isFocused &&
+              !isBelowFocused &&
+              !(Platform.OS === 'ios' && isModal)
+            : !isPreloaded && !isFocused && !(Platform.OS === 'ios' && isModal);
```
vargajacint: nit, but this check is already hard :D 
```suggestion
const isModalOnIos = (Platform.OS === 'ios' && isModal)

            ? !isPreloaded &&
              !isFocused &&
              !isBelowFocused &&
              !isModalOnIos
            : !isPreloaded && !isFocused && !isModalOnIos;
```

PR #12657, opened 2025-06-27
feat(ios): support UIBarButtonItem in header

johankasperi:
JS and types changes associated with this PR in react-native-screens: https://github.com/software-mansion/react-native-screens/pull/2987

Don't know if further description is needed in this PR? Please let me know if so. Thank you!

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -721,7 +721,7 @@ interface SharedHeaderBarButtonItem {
    *
    * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.property
    */
-  style?: 'Plain' | 'Done' | 'Prominent';
+  style?: 'plain' | 'done' | 'prominent';
   /**
    * The tint color to apply to the button item.
    *
@@ -804,6 +804,24 @@ export interface HeaderBarButtonItemMenuAction {
    * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
    */
   systemImage?: string;
+  /**
+   * State of the menu item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uimenuelement/state
+   */
+  state?: 'on' | 'off' | 'mixed';
+  /**
+   * Style of the menu item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uimenuelement/attributes
+   */
+  attributes?: 'destructive' | 'disabled' | 'hidden' | 'keepsMenuPresented';
+  /**
+   * Discoverability title of the menu item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uiaction/discoverabilitytitle
+   */
+  discoverabilityTitle?: string;
 }
 export interface HeaderBarButtonItemWithMenu extends SharedHeaderBarButtonItem {
   menu: {
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -772,6 +772,8 @@ interface SharedHeaderBarButtonItem {
    * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/identifier
    */
   identifier?: string;
+  accessibilityLabel?: string;
+  accessibilityHint?: string;
 }
 
 export interface HeaderBarButtonItemWithAction
@@ -795,8 +797,22 @@ export interface HeaderBarButtonItemWithAction
   changesSelectionAsPrimaryAction?: boolean;
 }
 
+export interface HeaderBarButtonItemMenuAction {
+  title?: string;
+  onPress: () => void;
+  /**
+   * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
+   */
+  systemImage?: string;
+}
 export interface HeaderBarButtonItemWithMenu extends SharedHeaderBarButtonItem {
-  menu: { title: string; onPress: () => void }[];
+  menu: {
+    title?: string;
+    items: (
+      | HeaderBarButtonItemMenuAction
+      | HeaderBarButtonItemWithMenu['menu']
+    )[];
+  };
 }
 
 export interface HeaderBarButtonItemSpacing {
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -709,6 +709,7 @@ interface SharedHeaderBarButtonItem {
     fontFamily?: string;
     fontSize?: number;
     fontWeight?: string;
+    color?: ColorValue;
   };
   /**
    * Image source for the button.
@@ -795,7 +796,7 @@ export interface HeaderBarButtonItemWithAction
 }
 
 export interface HeaderBarButtonItemWithMenu extends SharedHeaderBarButtonItem {
-  menu: any[];
+  menu: { title: string; onPress: () => void }[];
 }
 
 export interface HeaderBarButtonItemSpacing {
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -703,6 +703,14 @@ interface SharedHeaderBarButtonItem {
    */
   title?: string;
   /**
+   * Style for the button title.
+   */
+  titleStyle?: {
+    fontFamily?: string;
+    fontSize?: number;
+    fontWeight?: string;
+  };
+  /**
    * Image source for the button.
    */
   image?: ImageRequireSource;
@@ -727,12 +735,6 @@ interface SharedHeaderBarButtonItem {
    */
   hidden?: boolean;
   /**
-   * A Boolean value that indicates whether the button is in a selected state.
-   *
-   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/isselected
-   */
-  selected?: boolean;
-  /**
    * A Boolean value that indicates whether the button is in a enabled state.
    */
   enabled?: boolean;
@@ -748,6 +750,27 @@ interface SharedHeaderBarButtonItem {
    * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/possibletitles
    */
   possibleTitles?: string[];
+  /**
+   * A boolean value indicating whether the background this item may share with other items in the bar should be hidden.
+   * Only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/hidessharedbackground
+   */
+  hidesSharedBackground?: boolean;
+  /**
+   * A boolean value indicating whether this bar button item can share a background with other items in a navigation bar or a toolbar.
+   * Only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/sharesbackground
+   */
+  sharesBackground?: boolean;
+  /**
+   * An identifier used to match bar button items across transitions in a navigation bar or toolbar..
+   * Only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/identifier
+   */
+  identifier?: string;
 }
 
 export interface HeaderBarButtonItemWithAction
@@ -756,6 +779,19 @@ export interface HeaderBarButtonItemWithAction
    * Function to call when the button is pressed.
    */
   onPress: () => void;
+  /**
+   * A Boolean value that indicates whether the button is in a selected state.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/isselected
+   */
+  selected?: boolean;
+  /**
+   * A Boolean value that indicates whether the button represents an action or selection.
+   * Only available from iOS 15.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/changesselectionasprimaryaction
+   */
+  changesSelectionAsPrimaryAction?: boolean;
 }
 
 export interface HeaderBarButtonItemWithMenu extends SharedHeaderBarButtonItem {
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -772,6 +772,31 @@ interface SharedHeaderBarButtonItem {
    * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/identifier
    */
   identifier?: string;
+  /**
+   * A badge to be rendered on a bar button item.
+   * Only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitembadge
+   */
+  badge?: {
+    /**
+     * The text to display in the badge.
+     */
+    value: string;
+    /**
+     * The color of the badge.
+     */
+    color?: ColorValue;
+    /**
+     * The background color of the badge.
+     */
+    backgroundColor?: ColorValue;
+    style?: {
+      fontFamily?: string;
+      fontSize?: number;
+      fontWeight?: string;
+    };
+  };
   accessibilityLabel?: string;
   accessibilityHint?: string;
 }
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -12,6 +12,7 @@ import type {
   Theme,
 } from '@react-navigation/native';
 import type {
+  ColorValue,
   ImageRequireSource,
   ImageSourcePropType,
   StyleProp,
@@ -702,13 +703,55 @@ export interface HeaderBarButtonItem {
    */
   title?: string;
   /**
-   * Callback to be called when the button is pressed.
-   */
-  onPress: () => void;
-  /**
    * Image source for the button.
    */
   image?: ImageRequireSource;
+  /**
+   * The style of the item.
+   * "Prominent" only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.property
+   */
+  style?: 'Plain' | 'Done' | 'Prominent';
+  /**
+   * The tint color to apply to the button item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/tintcolor
+   */
+  tintColor?: ColorValue;
+  /**
+   * A boolean that determines the visibility of the item.
+   * Only available from iOS 16.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/ishidden
+   */
+  hidden?: boolean;
+  /**
+   * A Boolean value that indicates whether the button is in a selected state.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/isselected
+   */
+  selected?: boolean;
+  /**
+   * A Boolean value that indicates whether the button is in a enabled state.
+   */
+  enabled?: boolean;
+  /**
+   * The width of the item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/width
+   */
+  width?: number;
+  /**
+   * The set of possible titles to display on the bar button.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/possibletitles
+   */
+  possibleTitles?: string[];
+  /**
+   * Function to be called when the button is pressed.
+   */
+  onPress?: () => void;
 }
 
 export type NativeStackNavigatorProps = DefaultNavigatorOptions<
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -12,6 +12,7 @@ import type {
   Theme,
 } from '@react-navigation/native';
 import type {
+  ImageRequireSource,
   ImageSourcePropType,
   StyleProp,
   TextStyle,
@@ -287,6 +288,18 @@ export type NativeStackNavigationOptions = {
    * Function which returns a React Element to display on the right side of the header.
    */
   headerRight?: (props: NativeStackHeaderRightProps) => React.ReactNode;
+  /**
+   * iOS native UIBarButtomItem to the right side of the header.
+   *
+   * @platform ios
+   */
+  headerRightBarButtonItems?: HeaderBarButtonItem[];
+  /**
+   * iOS native UIBarButtomItem to the left side of the header.
+   *
+   * @platform ios
+   */
+  headerLeftBarButtonItems?: HeaderBarButtonItem[];
   /**
    * String or a function that returns a React Element to be used by the header.
    * Defaults to screen `title` or route name.
@@ -682,6 +695,21 @@ export type NativeStackNavigationOptions = {
    */
   unstable_sheetFooter?: () => React.ReactNode;
 };
+
+export interface HeaderBarButtonItem {
+  /**
+   * Title of the button.
+   */
+  title?: string;
+  /**
+   * Callback to be called when the button is pressed.
+   */
+  onPress: () => void;
+  /**
+   * Image source for the button.
+   */
+  image?: ImageRequireSource;
+}
 
 export type NativeStackNavigatorProps = DefaultNavigatorOptions<
   ParamListBase,
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -697,7 +697,7 @@ export type NativeStackNavigationOptions = {
   unstable_sheetFooter?: () => React.ReactNode;
 };
 
-export interface HeaderBarButtonItem {
+interface SharedHeaderBarButtonItem {
   /**
    * Title of the button.
    */
@@ -748,11 +748,28 @@ export interface HeaderBarButtonItem {
    * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/possibletitles
    */
   possibleTitles?: string[];
-  /**
-   * Function to be called when the button is pressed.
-   */
-  onPress?: () => void;
 }
+
+export interface HeaderBarButtonItemWithAction
+  extends SharedHeaderBarButtonItem {
+  /**
+   * Function to call when the button is pressed.
+   */
+  onPress: () => void;
+}
+
+export interface HeaderBarButtonItemWithMenu extends SharedHeaderBarButtonItem {
+  menu: any[];
+}
+
+export interface HeaderBarButtonItemSpacing {
+  spacing: number;
+}
+
+export type HeaderBarButtonItem =
+  | HeaderBarButtonItemWithAction
+  | HeaderBarButtonItemWithMenu
+  | HeaderBarButtonItemSpacing;
 
 export type NativeStackNavigatorProps = DefaultNavigatorOptions<
   ParamListBase,
--- a/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
+++ b/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
@@ -7,12 +7,36 @@ export const prepareHeaderBarButtonItems = (
   routeKey: string
 ) => {
   return barButtonItems?.map((item, index) => {
+    if ('spacing' in item) {
+      return item;
+    }
+    const image = item.image ? Image.resolveAssetSource(item.image) : undefined;
+    const titleStyle = item.titleStyle
+      ? { ...item.titleStyle, color: processColor(item.titleStyle.color) }
+      : undefined;
+    const tintColor = item.tintColor ? processColor(item.tintColor) : undefined;
+    const processedItem = {
+      ...item,
+      image,
+      titleStyle,
+      tintColor,
+    };
     if ('onPress' in item) {
       return {
-        ...item,
-        ...(item.image ? { image: Image.resolveAssetSource(item.image) } : {}),
-        ...(item.tintColor ? { tintColor: processColor(item.tintColor) } : {}),
+        ...processedItem,
         buttonId: `${item.title ?? item.image}-${index}-${routeKey}`,
+      };
+    }
+    if ('menu' in item) {
+      return {
+        ...processedItem,
+        menu: item.menu.map((menuItem, menuIndex) => ({
+          ...menuItem,
+          // ...(menuItem.image ? { image: Image.resolveAssetSource(menuItem.image) } : {}),
+          // ...(menuItem.tintColor ? { tintColor: processColor(menuItem.tintColor) } : {}),
+          // buttonId: `${menuItem.title ?? menuItem.image}-${menuIndex}-${routeKey}`,
+          menuId: `${menuItem.title}-${menuIndex}-${routeKey}`,
+        })),
       };
     }
     return item;
--- a/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
+++ b/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
@@ -3,14 +3,18 @@ import { Image, processColor } from 'react-native';
 import type { HeaderBarButtonItem } from '../types';
 
 export const prepareHeaderBarButtonItems = (
-  barButtonItems: HeaderBarButtonItem[]
+  barButtonItems: HeaderBarButtonItem[],
+  routeKey: string
 ) => {
-  return barButtonItems?.map((item) => {
-    return {
-      ...item,
-      ...(item.image ? { image: Image.resolveAssetSource(item.image) } : {}),
-      ...(item.tintColor ? { tintColor: processColor(item.tintColor) } : {}),
-      buttonId: Math.random().toString(36).slice(2, 7),
-    };
+  return barButtonItems?.map((item, index) => {
+    if ('onPress' in item) {
+      return {
+        ...item,
+        ...(item.image ? { image: Image.resolveAssetSource(item.image) } : {}),
+        ...(item.tintColor ? { tintColor: processColor(item.tintColor) } : {}),
+        buttonId: `${item.title ?? item.image}-${index}-${routeKey}`,
+      };
+    }
+    return item;
   });
 };
--- a/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
+++ b/packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
@@ -1,6 +1,28 @@
 import { Image, processColor } from 'react-native';
 
-import type { HeaderBarButtonItem } from '../types';
+import type {
+  HeaderBarButtonItem,
+  HeaderBarButtonItemWithMenu,
+} from '../types';
+
+const prepareMenu = (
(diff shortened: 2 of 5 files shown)


packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+      | NativeStackHeaderBarButtonItemMenuAction
+      | NativeStackHeaderBarButtonItemSubmenu
+    )[];
+  };
+}
+
+export interface NativeStackHeaderBarButtonItemSpacing {
+  spacing: number;
+}
+
```
satya164: Please use type aliases instead of interfaces. We typically don't use interfaces in the codebase.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  title?: string;
+  /**
+   * Style for the button title.
+   */
+  titleStyle?: {
+    fontFamily?: string;
+    fontSize?: number;
+    fontWeight?: string;
+    color?: ColorValue;
+  };
```
satya164, line 746: We use `label` terminology for buttons everywhere and not `title`. So this (and the following code using `title`) seems inconsistent. Can we standardize the terminology?

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  /**
+   * Image source for the button
+   */
+  image?: ImageRequireSource;
+  /**
+   * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
+   */
+  sfSymbolName?: string;
```
satya164: What happens if both sf symbol and image are provided?

If one overrides the other then we should have a single property `icon`, which could be an object like in native tabs.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  /**
+   * The style of the item.
+   * "Prominent" only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.property
+   */
+  style?: 'plain' | 'done' | 'prominent';
```
satya164: `style` in react native has a different meaning. We need to use a different name (e.g. `variant`) to avoid confusion.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+    color?: ColorValue;
+    /**
+     * The background color of the badge.
+     */
+    backgroundColor?: ColorValue;
+    style?: {
+      fontFamily?: string;
+      fontSize?: number;
+      fontWeight?: string;
+    };
```
satya164, line 815: `color` and `backgroundColor` are also styles. They should be moved to the `style` property.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  titleStyle?: {
+    fontFamily?: string;
+    fontSize?: number;
+    fontWeight?: string;
+    color?: ColorValue;
+  };
+  /**
+   * Image source for the button
+   */
+  image?: ImageRequireSource;
```
satya164: Is there a reason this needs to be `ImageRequireSource` instead of the same type as `backButtonImageSource`?

satya164: Rename to `imageSource` for consistency with other props.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  /**
+   * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
+   */
+  sfSymbolName?: string;
+  /**
+   * The style of the item.
+   * "Prominent" only available from iOS 26.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.property
```
satya164: We should document what each of these values do; a single line is fine.

Linking to Apple docs is ok, but it should be complementary.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/tintcolor
+   */
+  tintColor?: ColorValue;
+  /**
+   * A boolean that determines the visibility of the item.
+   * Only available from iOS 16.0 and later.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/ishidden
+   */
+  hidden?: boolean;
```
satya164: What's the difference between not providing this item vs providing and setting `hidden`?

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  /**
+   * A Boolean value that indicates whether the button is in a enabled state.
+   */
+  enabled?: boolean;
```
satya164: This is inconsistent with conventional patterns, where you disable a button by specifying `disabled`, instead of enabling based on `enabled`.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+    backgroundColor?: ColorValue;
+    style?: {
+      fontFamily?: string;
+      fontSize?: number;
+      fontWeight?: string;
+    };
+  };
+  accessibilityLabel?: string;
+  accessibilityHint?: string;
+}
```
satya164: Many of the descriptions refer to "item" vs "button", and some say "button item". What's the difference between these? If they are the same, then we should use the same terminology. If "item" is something different (e.g. the container), saying "button's container view" or something like that may be clearer.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uibarbuttonitem/hidessharedbackground
+   */
+  hidesSharedBackground?: boolean;
+};
+
+export type NativeStackHeaderBarButtonItem =
+  | NativeStackHeaderBarButtonItemWithAction
+  | NativeStackHeaderBarButtonItemWithMenu
+  | NativeStackHeaderBarButtonItemSpacing;
```
satya164: For consistency, remove `Bar` terminology - we just call it `Header` everywhere, e.g. `NativeStackHeaderButtonItem`

packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
```diff
@@ -0,0 +1,82 @@
...
+        menuId: `${menuIndex}-${index}-${routeKey}-${side}`,
+      };
+    }),
+  };
+};
+
+export const prepareHeaderBarButtonItems = (
+  barButtonItems: (
+    | NativeStackHeaderBarButtonItem
+    | NativeStackHeaderBarButtonItemWithCustomView<any>
```
satya164: Don't use `any`. If it can be anything, use `unknown` instead.

It'll be better to add a generic to the function and use it for the proper type.

packages/native-stack/src/utils/__tests__/prepareHeaderBarButtonItems.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+function hasCustomView(
+  item: BarButtonItem
+): item is Exclude<
+  BarButtonItem,
+  { spacing: number } | { buttonId: boolean } | { menu: any }
+> & { isSubview: boolean } {
+  return item ? 'isSubview' in item : false;
+}
+
+describe('prepareHeaderBarButtonItems', () => {
```
satya164: we typically don't use describe if the file has a single test suite

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+  discoverabilityTitle?: string;
+}
+
+export interface NativeStackHeaderBarButtonItemSubmenu {
+  title?: string;
+  type: 'submenu';
+  /**
+   * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
+   */
+  sfSymbolName?: string;
```
satya164: Can it only be `sfSymbolName`?

cc @kkafar maybe we should have a consistent type for icons that are used across native stack and bottom tabs?

In this case I'd name the property as `icon` and use an object to make it more future-proof.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uimenuelement/state
+   */
+  state?: 'on' | 'off' | 'mixed';
+  /**
+   * Style of the menu item.
+   *
+   * Read more: https://developer.apple.com/documentation/uikit/uimenuelement/attributes
+   */
+  attributes?: 'destructive' | 'disabled' | 'hidden' | 'keepsMenuPresented';
```
satya164: The property name says `attributes`, but it is union, e.g. accepts a single value. Is this intentional?

It's kinda inconsistent that in most places `hidden` is a direct boolean property, and here it is in a property called `attributes`.

packages/native-stack/src/types.tsx
```diff
@@ -682,6 +708,196 @@ export type NativeStackNavigationOptions = {
...
+}
+
+export interface NativeStackHeaderBarButtonItemMenuAction {
+  title?: string;
+  type: 'action';
+  onPress: () => void;
+  /**
+   * Any SF symbol. Explore them here: https://developer.apple.com/sf-symbols/
+   */
+  sfSymbolName?: string;
```
satya164: Same comment as others regarding icons

packages/native-stack/src/types.tsx
```diff
@@ -281,12 +283,36 @@ export type NativeStackNavigationOptions = {
...
+   * Overrides `headerLeft`.
+   *
+   * @platform ios
+   */
+  headerLeftItems?: (
+    | NativeStackHeaderBarButtonItem
+    | NativeStackHeaderBarButtonItemWithCustomView<NativeStackHeaderLeftProps>
+  )[];
+  /**
+   * Array of iOS native UIBarButtomItem or functions
```
satya164: The phrase "Array of iOS native UIBarButtomItem or functions" doesn't sound right, because you are neither specifying an array of `UIBarButtomItem` instances, nor functions.

Also I assume it should be `Button` and not `Buttom`

Something like "Array of items to display as UIBarButtonItem" maybe more accurate.

packages/native-stack/src/views/useHeaderConfigProps.tsx
```diff
@@ -260,6 +306,74 @@ export function useHeaderConfigProps({
...
+          (item) =>
+            item &&
+            'onPress' in item &&
+            item.buttonId === event.nativeEvent.buttonId
+        );
+        if (pressedItem && 'onPress' in pressedItem && pressedItem.onPress) {
+          pressedItem.onPress();
+        }
+      }
+    : undefined;
```
satya164: Does this kind of code really need to be in React Navigation instead of React Native Screens? It doesn't sound ideal that `onPress` handling needs additional work on React Navigation side.

packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
```diff
@@ -0,0 +1,82 @@
...
+): any => {
+  return {
+    ...menu,
+    items: menu.items.map((menuItem, menuIndex) => {
+      if ('items' in menuItem) {
+        return prepareMenu(menuItem, menuIndex, routeKey, side);
+      }
+      return {
+        ...menuItem,
+        menuId: `${menuIndex}-${index}-${routeKey}-${side}`,
```
satya164: Since this just uses route key and indices for the id, can't this be done in React Native screens instead of React Navigation? We already pass some unique id to each screen that can be used.

packages/native-stack/src/types.tsx
kkafar: That would be good. We haven't settled on consistent API for native-tabs yet though - since we're still exploring various options to implement image loading on native side and what we need from JS depends on that. I'll prioritise it in following days so I have cohesive vision to present.

joacub:
when is going to land this ? react-native-screens is waiting for this too 👯

johankasperi:
> Thanks for the PR.
> 
> I left some comments, most are about consistency with rest of the API and codebase.
> 
> I'm also concerned about `prepareHeaderBarButtonItems`. I don't see why this logic needs to be handled by the consumer.
> 
> If it doesn't have anything to do with navigation, special behavior that react navigation wants to implement or renaming specific options, they belong in React Native Screens.
> 
> Code to assign ids, event handling, properly resolving icons, colors, etc. should be in the core library in React Native Screens. Otheriwse it makes it difficult for both React Navigation and React Native Screens to make changes in the future.

Thanks for the review! I see your point and I agree with this. Will try to move all the logic in `prepareHeaderBarButtonItems` into React Native Screens asap, if not @kkafar disagree with this? When I've done that I will adress the rest of your comments

kkafar:
I'm fine with moving the prop processing logic to screens

packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
johankasperi: Moved all of this to rn screens https://github.com/react-navigation/react-navigation/pull/12657/commits/667519e71984fca67c8510e0e4b20b60083af657

packages/native-stack/src/views/useHeaderConfigProps.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/667519e71984fca67c8510e0e4b20b60083af657

packages/native-stack/src/types.tsx
johankasperi: I'm using the same name for all properties as they have in the iOS implementation. Apple for example calls this property for "style" (https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.property?language=objc). My reasoning for this strategy was that it would be easier to understand what they do if we are not coming up with our own naming convention. But I will happily change this if you think its for the better?

packages/native-stack/src/types.tsx
johankasperi: Same answer as here: https://github.com/react-navigation/react-navigation/pull/12657#discussion_r2380047036 My strategy was to use the same name as Apple. But can change it if you want to?

packages/native-stack/src/types.tsx
johankasperi: I haven't implemented so you can load images from uri. Only local/required images. Maybe we could add the possibility to load images from uri in upcoming versions? @kkafar

packages/native-stack/src/types.tsx
satya164: @johankasperi most react native developers come from js background rather than uikit background, so iOS naming won't always be easier to understand, especially when the name has widespread meaning in the ecosystem already.

packages/native-stack/src/types.tsx
johankasperi: Same answer as https://github.com/react-navigation/react-navigation/pull/12657#discussion_r2380047036 Im using the same name as Apple (https://developer.apple.com/documentation/uikit/uibaritem/isenabled?language=objc)

packages/native-stack/src/utils/prepareHeaderBarButtonItems.ts
johankasperi: `prepareHeaderBarButtonItems` is removed from react-navigation now

packages/native-stack/src/utils/__tests__/prepareHeaderBarButtonItems.test.tsx
johankasperi: `prepareHeaderBarButtonItems` is removed from react-navigation now

packages/native-stack/src/types.tsx
johankasperi: Thats the way Apple has named it and implemented it. `hidden` is not a boolean but a attribute on UIMenuElement. And as I've said before, my strategy has been not to rename and have similar API as they have created.

packages/native-stack/src/types.tsx
johankasperi: Yes the UIMenuElement can only have `sfSymbolName` as of now. It has support for any `image` (https://developer.apple.com/documentation/uikit/uimenuelement/image?language=objc)  as well but I haven't implemented possibility to use `ImageRequireSource` for UIMenuElements. Maybe we can do that in an upcoming PR?

packages/native-stack/src/types.tsx
johankasperi: Same answer as here: https://github.com/react-navigation/react-navigation/pull/12657#discussion_r2380080830

packages/native-stack/src/types.tsx
satya164: React Navigation doesn't try to match Apple-naming 1-1. This is a cross-platform library, so the API needs to make sense in a cross-platform context, not only iOS. There is no guarantee that this API won't be supported on Android or Web in the future.

The API needs to be consistent with other React Navigation APIs and the framework in general. If React Native Screens wants to keep original iOS names, then the properties should just be renamed before passing on React Navigation side.

packages/native-stack/src/types.tsx
johankasperi: The thing is they do different things on different iOS versions. "done" makes the title bold on iOS < 26 but it has the same effect as "prominent" on iOS 26, it gives the liquid glass a background color. I think that documenting this might be difficult because it might become incorrect as soon as Apple makes changes to iOS. I think its better to only reference to the source (Apple). But of course if you disagree I will happily document it 😊

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/92fb593a73eb5ae4cadb1d26009c309feabac949

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/c435978789cd054022866ea9d9b41807f0c99c8c

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/b6a0515a73f8cc7874ce0112549850111f8e81b5

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/5a65519cf06576810eb78ab5981c07443b858f0e

packages/native-stack/src/types.tsx
johankasperi: Good catch! Didn't think of that. There is no difference at all. So maybe we should remove the `hidden` prop from react-native-screens as well? @kkafar

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/28460b003475c3849790af93555383659a0d4e83

johankasperi:
@satya164 I think I've adressed everything in your review now except the image/icon properties (have started a discussion in react-native-screens about that https://github.com/software-mansion/react-native-screens/pull/2987#issuecomment-3335735683) and that all items should use the react-navigation theme by default (will try to fix that asap).

~Regarding your comments about naming of the properties I have used the same names as Apple. I think its clearer to use the same names as they do. But as I've said in my responses above, I will happily change that if you think it's better 😊~ This has been adressed by @satya164 above

packages/native-stack/src/types.tsx
johankasperi: Got it! I will rename it asap

packages/native-stack/src/types.tsx
johankasperi: Yeah I see your point. I will change it asap

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/7693c76702a96431cb084ced6d79032f4b74d7a8

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/646d709276724e3fae826eb3113b97c8f9529fea

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/87b2c226da5fa903d19e2be065cf05c9302c1e82

packages/native-stack/src/types.tsx
johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/d3919cbcc8c769776b64c77cace86ddffcefe281

packages/native-stack/src/types.tsx
johankasperi: Realized that RCTImageLoader support images from uri already. So changed the type here https://github.com/react-navigation/react-navigation/pull/12657/commits/876324db7db6dadb4ea275da44f1a85e89786fe4

johankasperi: https://github.com/react-navigation/react-navigation/pull/12657/commits/876324db7db6dadb4ea275da44f1a85e89786fe4

packages/native-stack/src/types.tsx
johankasperi: Reason for why we can't have a `hidden` as a direct boolean property is because a UIMenuElement can't be both hidden and for example destructive. So all of these attributes have to be controlled by one single property

johankasperi:
> The other thing I forgot - things like colors, fonts, etc., should use React Navigation theme by default, so we need to provide default styles for the items when applicable.

Fixed that in https://github.com/react-navigation/react-navigation/pull/12657/commits/a79b931cfd1c4507db4ff794a0fe8f30579b439b

ajanuar:
@satya164 any ETA for this PR? thank you

satya164:
Thanks a lot for the work @johankasperi 

I'm aiming to get this merged hopefully today, just a few things to clarify that I'll follow up with @kkafar.

ferrannp:
@johankasperi I am testing and found couple of issues, items are reversed, see:

```js
          headerRightItems: () => [
            {
              icon: {
                type: 'sfSymbol',
                name: 'heart',
              },
              onPress: () => {},
              variant: 'prominent',
            },
            {
              icon: {
                type: 'sfSymbol',
                name: 'heart',
              },
              onPress: () => {},
            },
          ],
```

And then, if you have a prominent item, they don't glue together and look at this effect:

https://github.com/user-attachments/assets/93591baf-3c31-422c-b8ad-1d251b4e0969

ferrannp:
image

`done` and `prominent` look the same (seems done is deprecated?). We miss `bordered`.

satya164:
I'm ready to merge, but I'm waiting for the following things:

- https://github.com/software-mansion/react-native-screens/pull/3214
- The issue with reversing items that @ferrannp mentioned - is it a bug? Or do we need to handle it in React Navigation?
- There is `disabled` property, as well as `attributes: 'disabled'`, and apple docs say `attributes` is a struct, whereas it's a union type here.
- I added an explicit `type` property for better TypeScript DX, should backport it to screens?

Once the PR is merged & released, and I get an update for the other issues, I'll merge this PR and make a release.

johankasperi:
> @johankasperi I am testing and found couple of issues, items are reversed, see:
> 
> ```js
>           headerRightItems: () => [
>             {
>               icon: {
>                 type: 'sfSymbol',
>                 name: 'heart',
>               },
>               onPress: () => {},
>               variant: 'prominent',
>             },
>             {
>               icon: {
>                 type: 'sfSymbol',
>                 name: 'heart',
>               },
>               onPress: () => {},
>             },
>           ],
> ```
> 
> And then, if you have a prominent item, they don't glue together and look at this effect:
> 
>  Screen.Recording.2025-10-16.at.18.57.06.mov

Reversing the items on the right side is actually how Apple have implemented it on iOS. Index 0 is the item closest to the edge of the device, so the order is left->right on the left side of the header and right->left on the right side of the header. 

Of course we could reverse the array on the right side if we want it to render left->right as we are used to in RN. But decided not to (at least in the PR to react-native-screens) as that could cause other side effects, for example in the "excessive items"-menu (seen in the video in this comment https://github.com/software-mansion/react-native-screens/pull/2987#issuecomment-3304386706)

johankasperi:
> And then, if you have a prominent item, they don't glue together and look at this effect:

Prominent items cannot share background with regular items if that is what you are referring to?

johankasperi:
> `done` and `prominent` look the same (seems done is deprecated?). We miss `bordered`.

`done` and `prominent` look the same on iOS 26 but on iOS 18 and below `done` makes the label bold and `prominent` isn't supported on iOS 18 at all.

`bordered` was deprecated in iOS 8 https://developer.apple.com/documentation/uikit/uibarbuttonitem/style-swift.enum/bordered 😊

johankasperi:
Thanks for the review! My comments to some of your points below:

> * The issue with reversing items that @ferrannp mentioned - is it a bug? Or do we need to handle it in React Navigation?

See my answer to @ferrannp here: https://github.com/react-navigation/react-navigation/pull/12657#issuecomment-3412564608

> * There is `disabled` property, as well as `attributes: 'disabled'`, and apple docs say `attributes` is a struct, whereas it's a union type here.

The `disabled` property is on UIBarButtonItem (the button in the header). The attributes union is on [UIMenuElement](https://developer.apple.com/documentation/uikit/uimenuelement/attributes) (the dropdown menu that can be opened when pressing a button). Reason for it being a union on the UIMenuElement is because this element can only has one of those attributes listed.

> * I added an explicit `type` property for better TypeScript DX, should backport it to screens?

Nice addition! Yeah maybe we should add it to screens, what do you think @kkafar?

ferrannp:
@johankasperi 

> Prominent items cannot share background with regular items if that is what you are referring to?

If you check the video, I mean when you have a prominent and not-prominent and you press a button, there is a weird effect (in the video something with the color orange getting in the middle of the items).

johankasperi:
> @johankasperi
> 
> > Prominent items cannot share background with regular items if that is what you are referring to?
> 
> If you check the video, I mean when you have a prominent and not-prominent and you press a button, there is a weird effect (in the video something with the color orange getting in the middle of the items).

I'm not seeing anything orange? If you mean that their backgrounds attach and detach when pressing I'm pretty sure this is also a part of iOS/UIKit as I'm not doing anything regarding that

satya164:
> Reason for it being a union on the UIMenuElement is because this element can only has one of those attributes listed.

But it's not a limitation in the native API, correct? Since it's a struct on native side. Making it a union would prevent any future additions to attributes without changing the API.

Theoretically, it also seems that `keepsMenuPresented` and `destructive` could be used together.

> Reversing the items on the right side is actually how Apple have implemented it on iOS

I see. Apple has some weird, non-intuitive APIs.

> as that could cause other side effects, for example in the "excessive items"-menu

Not sure if I understood the problem.

- Without reversing, it'll put the items at the beginning in the overflow menu
- With reversing, it'll put the items at the end in the overflow menu

It seems intuitive that items displayed on the right side would go to the overflow menu located on the right, and seems also appear in reverse order inside the menu. So from this perspective, reversing shouldn't result in any unituitive behavior. Did I misunderstand anything?

If not, then I think it'll be best to reverse it in React Navigation.

Also, is the overflow behavior iOS 26 specific? Seems iOS 18 results in quite a broken state

CleanShot 2025-10-16 at 22 22 16@2x

Also, with `label` being optional, the menu looks broken. I also found an a11y issue - for SF Symbols, Accessibility Inspector tries to infer the label from the icon; however, specifying an image makes it announce it as "disabled, button" - it's not actually disabled, I can tap it.

I'm leaning towards making `label` required. The inference from SF symbol won't always be correct. An explicit label is always better for a11y and prevents broken behavior in menus and images.

kkafar:
I think that adding explicit "type" field to config objects would be nice. We can have that on react-native-screens side. I'll close my eyes and not treat it as breaking change, since react-navigation has not consumed it & no one uses it directly 🙈

kkafar:
@ferrannp The effect you showcase in the video [here](https://github.com/react-navigation/react-navigation/pull/12657#issuecomment-3411795636) is intended & a part of new iOS 26 design. AFAIK this can not be disabled when liquid glass is enabled.

kkafar:
As per item reversing: I want the screens API to stay close to the platform here - if the items are reversed by UIKit, I don't want this changed in `react-native-screens`. We might introduce some logic here to handle this, I don't mind, but I'll be reluctant to tackle this upstream.

johankasperi:
> But it's not a limitation in the native API, correct? Since it's a struct on native side. Making it a union would prevent any future additions to attributes without changing the API.
> 
> Theoretically, it also seems that `keepsMenuPresented` and `destructive` could be used together.

From what I understand it is a limitation in the native API. There is no way of adding several attributes to one UIMenuElement (https://github.com/software-mansion/react-native-screens/blob/3a588b10234657940b627596d0d81a13d267461c/ios/RNSBarButtonItem.mm#L192).

And I don't really understand why a union would make it impossible to add future attributes? Can't you just add a attribute to the union? Could you please elaborate, perhaps with an example? Thank you!

johankasperi:
> It seems intuitive that items displayed on the right side would go to the overflow menu located on the right, and seems also appear in reverse order inside the menu. So from this perspective, reversing shouldn't result in any unituitive behavior. Did I misunderstand anything?
> 
> If not, then I think it'll be best to reverse it in React Navigation.

I think you're right. I guess it was just me being stuck in my thinking that it is more intuitive that index 0 is closest to the edge.

johankasperi:
> Also, is the overflow behavior iOS 26 specific? Seems iOS 18 results in quite a broken state

Yes this is iOS 26 specific

johankasperi:
> Also, with `label` being optional, the menu looks broken. I also found an a11y issue - for SF Symbols, Accessibility Inspector tries to infer the label from the icon; however, specifying an image makes it announce it as "disabled, button" - it's not actually disabled, I can tap it.
> 
> I'm leaning towards making `label` required. The inference from SF symbol won't always be correct. An explicit label is always better for a11y and prevents broken behavior in menus and images.

Regarding the a11y issue you can also use the `accessibilityLabel` and `accessibilityHint` props to make it accessible, so you can argue that this is the responsibility of the developer using the lib 😄 But of course requiring `label` will make it easier to create non-broken interfaces. 

The only thing that might become a problem is frustration from developers wondering why you need to add a label that isn't visible, given that having more than three items in the header might not be very common so they wont see the excessive items menu and understand why a label is required.

satya164:
@johankasperi

> From what I understand it is a limitation in the native API. There is no way of adding several attributes to one UIMenuElement

Here is an example of what I mentioned:

> Theoretically, it also seems that keepsMenuPresented and destructive could be used together.

I made a simple app in Xcode with Swift, and tested it to verify that it's possible. Here I have 3 menu items:

- No attributes - menu item behaves normally
- Destructive - menu item is red
- Destructive + keepsMenuPresented - menu item is red and stays open after pressing

https://github.com/user-attachments/assets/1d7bdb09-8a9b-42b9-bd4c-1cff3941b270


```swift
func handler(action: UIAction) {
  print("\(action.title ?? "") selected")
}

let addAction = UIAction(
  title: "Add",
  image: UIImage(systemName: "plus"),
  handler: handler
)

let copyAction = UIAction(
  title: "Copy",
  image: UIImage(systemName: "clipboard"),
  attributes: .destructive,
  handler: handler
)

let deleteAction = UIAction(
  title: "Delete",
  image: UIImage(systemName: "trash"),
  attributes: [.destructive, .keepsMenuPresented],
  handler: handler
)
    
let menuItems = [
    addAction,
    copyAction,
    deleteAction
]

let myMenu = UIMenu(title: "My menu", image: nil, identifier: nil, options: [], children: menuItems)

let item = UIBarButtonItem(title: "Menu", image: nil, primaryAction: nil, menu: myMenu)

viewController.navigationItem.rightBarButtonItems = [item]
```

So React Native Screens shouldn't limit passing only one attribute.

satya164:
> The only thing that might become a problem is frustration from developers wondering why you need to add a label

yeah, it'll be a bit annoying, but I think it's worth it for proper a11y. we have to document it well to avoid confusion on what it does

ferrannp:
@kkafar this orange connection in this image is iOS intended? Yikes haha. Ok thanks for clarifying!

image

Btw I just tested latest @satya164 commits and reverse order is fixed 👍.

johankasperi:
> So React Native Screens shouldn't limit passing only one attribute.

Now I think I finally realize why we misunderstand each other in this discussion. Weirdly enough UIMenuElementAttributes is a [enum in obj-c ](https://developer.apple.com/documentation/uikit/uimenuelement/attributes?language=objc) and a [struct in swift](https://developer.apple.com/documentation/uikit/uimenuelement/attributes). So it is a limit in the native API, at least in the obj-c API. Thats why I'm unable to add several attributes to a UIAction because React Native Screens is written in obj-c. And you are able to do so in your example because it is in swift. 

Don't know the way forward here? Settle for having attributes as a union because I highly doubt that React Native Screens wants to add Swift?

Thanks for the example btw!

satya164:
> Don't know the way forward here? Settle for having attributes as a union because I highly doubt that React Native Screens wants to add Swift

I'd change it to boolean properties in JS at least, instead of keeping it as a union, so it leaves the possibility to support it in the future without making a breaking change.

johankasperi:
> I'd change it to boolean properties in JS at least, instead of keeping it as a union, so it leaves the possibility to support it in the future without making a breaking change.

But won't a boolean make it confusing because only one of them can be true?

satya164:
> But won't a boolean make it confusing because only one of them can be true?

In the end, it's a limitation in the current implementation of screens, which can change in the future. It'd suck to make a breaking change in the API if it gets resolved in the future, and major version changes also take a long time for libraries like screens and react navigation. I'd rather document the current limitation.

satya164:
I have made a PR in screens repo with a fix https://github.com/software-mansion/react-native-screens/pull/3308

johankasperi:
> I have made a PR in screens repo with a fix [software-mansion/react-native-screens#3308](https://github.com/software-mansion/react-native-screens/pull/3308)

Thank you!

Then we await that PR before updating this?

satya164:
@johankasperi once it is merged and released, I'll merge and release this PR

johankasperi:
I think https://github.com/software-mansion/react-native-screens/pull/3323 needs to be released in react-native-screens before merging this PR

satya164:
@johankasperi gotcha. thanks for the heads up

PR #12658, opened 2025-06-27
refactor(tab-view): Attach event via ref instead of using onPageScroll handler

intergalacticspacehighway:
**Motivation**

We want to allow passing [reanimated event handler](https://github.com/callstack/react-native-pager-view?tab=readme-ov-file#reanimated-onpagescroll-handler) to the underlying pager view. This PR partially enables it by unblocking the usage of `onPageScroll` event callback. 

It would be helpful to add an API to pass reanimated handler in material tabs since it is closely coupled with navigation. It can be used to apply custom animations without relying on `Animated` API (`useAnimatedValue` returns `Animated` value). We can introduce `onPageScroll` prop support in tab view and material tabs since it is not used by the tab view anymore.

**Test plan**

Test all the tab examples, the animation should work as expected.

### DIFF
--- a/packages/react-native-tab-view/src/PagerViewAdapter.tsx
+++ b/packages/react-native-tab-view/src/PagerViewAdapter.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Animated, Keyboard, Platform, StyleSheet } from 'react-native';
+import { Animated, Keyboard, StyleSheet } from 'react-native';
 import ViewPager, {
   type PageScrollStateChangedNativeEvent,
 } from 'react-native-pager-view';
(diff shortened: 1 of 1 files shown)


packages/react-native-tab-view/src/PagerViewAdapter.tsx
```diff
@@ -35,8 +35,6 @@ type Props<T extends Route> = PagerProps & {
   ) => React.ReactElement;
 };
 
-const useNativeDriver = Platform.OS !== 'web';
```
intergalacticspacehighway, line 38: Also tested on web, it is using `PanResponderAdapter` there so i guess this change should be fine. but let me know!

intergalacticspacehighway:
There is a better solution for using reanimated in material tabs - https://github.com/satya164/react-native-animated-observer

Issue #12493, opened 2025-03-19
Material Top Tabs is very unstable when screens contains lists inside


aik117:
### Current behavior

Hi,
I am using Material Top Tabs v 6.6.14, i have a dynamic tab navigator that can contain as much tabs as possible, where the lists item contains separate graphs. I am also hiding tabs label and updating the title of the screen when a tab is swiped. The first problem is when i swipe the tab, the screen title update is very slow. I checked the focus event in the listener where i wanted to update the tab title but the focus event inside listeners is being called after some delay. Also when i quickly swipe between tabs, the tabs start weird behavior and jumping between every tab infinitely. Can someone please help on this, i have spent so much time on this and couldn't find a way to solve.

This is my dynamic tab navigator

        `import React, { useEffect, useState } from 'react';
        import { Dimensions, StyleSheet, Text, View } from 'react-native'
        import { createMaterialTopTabNavigator } from '@react-navigation/material-top-tabs';
        import { getFocusedRouteNameFromRoute, useNavigation } from '@react-navigation/native';
        import WidgetsContainer from './WidgetsContainer';
        import { useSelector } from 'react-redux';
        import TextWrapper from './TextWrapper';
        import LoaderKit from 'react-native-loader-kit';
        import { colorStyles } from '../css/AppStyles';
        
        const Tab = createMaterialTopTabNavigator();

        const UserTabs = (props) => {
      	//console.log("props", props);
      	const navigation = useNavigation();
      	const { setFocusedTabDetails, parentNavigation, removeItem } = props;

	const tabsAndWidgets = useSelector((state) => state.userTabsAndWidgetsReducer.userTabsAndWidgets);

	const getHeaderTitle = (route) => {
		console.log("route", route);
		// If the focused route is not found, we need to assume it's the initial screen
		// This can happen during if there hasn't been any navigation inside the screen
		// In our case, it's "Feed" as that's the first screen inside the navigator
		//const routeName = getFocusedRouteNameFromRoute(route);
		//console.log("routeName", routeName);
		return route.name;
	}

	return (
		<Tab.Navigator 
			initialRouteName={'Home'} 
			screenOptions={{ 
				tabBarStyle: { display: "none" }, 
				lazy: true,
				lazyPlaceholder: () => 
					<View style={styles.lazyPlaceHolderStyle}>
						<LoaderKit
							style={styles.loaderStyle}
							name={'BallSpinFadeLoader'}
							color={colorStyles.headerColor}
						/>
[...]

Issue #12646, opened 2025-06-15
[react-native-tab-view] can not border radius tabbar indicator when turn on auto width


hoangtrunnam:
### Current behavior

I'm encountering an issue when trying to **round all 4 corners** of the tab bar indicator.
My design requires 8px border radius on each corner of the indicator bar.

I've identified where the issue is. It's in this file: https://github.com/react-navigation/react-navigation/blob/main/packages/react-native-tab-view/src/TabBarIndicator.tsx
This code snippet:

`styleList.push(
  { width: width === 'auto' ? 1 : width }, // The error occurs here
  { start: '${(100 / routes.length) * navigationState.index}%' },
  { transform }
);`

on line 160.

The reason is that scaling 1px up to match the tab bar width prevents me from applying border radius. The reason is quite simple: since it's only 1px, any border radius applied to that 1px cannot be visually distinguished by the naked eye whether the 4 corners have been rounded or not, so when scaled up, the rounded corners still cannot be seen. This is not a normal view with width 100 or 200, it's just a view with width of 1 that gets scaled up

I know the reason why they don't use animated width but use scaleX instead is because the library is using React Native's API, so animated width is impossible. You can see here: https://stackoverflow.com/questions/63976219/style-property-width-is-not-supported-by-native-animated-module-need-advice-f

If react-native-tab-view used React Native Reanimated, it could potentially animate width and solve the current corner rounding issue. However, as of June 15, 2025, I haven't seen any pull requests upgrading to use the Reanimated library.

react-native-tab-view: 4.0.10 and the latest: 4.1.2 still get that bug

Even on Android, scaling 1px is currently buggy:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/87139ddc-188e-4e4e-89a3-38cb0e1411e1" />

![Image](https://github.com/user-attachments/assets/7241e1bf-cf90-4c39-b0f9-b745ba85d59d)

![Image](https://github.com/user-attachments/assets/ad55adf5-f192-4794-a91f-165ca17b1df0)

### Expected behavior

I want the tab bar indicator to be able to round all 4 corners when autoWidth={true} is enabled.
I want it to look like the video below:

https://github.com/user-attachments/assets/45412dc7-55ee-42c6-8de2-d8c381176408

### Reproduction

https://github.com/hoangtrunnam/issue-indicator-react-native-tab-view

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

[...]

PR #12664, opened 2025-06-30
fix: Reset offset to 0 on swipe end to prevent indicator jump

AlexCernik:
## Summary

When swiping between tabs, especially in React Native's New Architecture or on some devices, the `offset` value may retain a small negative or positive value (e.g. `-0.02`) even after the swipe ends. This causes a visual glitch where the indicator briefly "jumps" backward before settling on the correct tab.

This PR resets `offset` to `0` when `pageScrollState` becomes `"idle"`, ensuring that the final position is stable and that the indicator aligns correctly with the selected tab.

## Changes

- Added `offset.setValue(0);` to `onPageScrollStateChanged` when state is `'idle'`

## Why

To prevent the tab indicator from jumping slightly backward after swiping, improving visual consistency. #12493 

## Notes

This fix has no functional side effects and only affects the visual animation when the swipe ends.

Tested on:
- React Native 0.79.4
- Fabric enabled
- Android

### DIFF
--- a/packages/react-native-tab-view/src/PagerViewAdapter.tsx
+++ b/packages/react-native-tab-view/src/PagerViewAdapter.tsx
@@ -102,6 +102,7 @@ export function PagerViewAdapter<T extends Route>({
 
     switch (pageScrollState) {
       case 'idle':
+        offset.setValue(0);
         onSwipeEnd?.();
         return;
       case 'dragging': {


satya164 (changes requested):
Thanks for the PR @AlexCernik 

This seems like a bug in `react-native-pager-view`. Can you open a bug report there?

I'd rather not add a workaround and get the actual bug fixed instead.

santhosh-umapathi-wikifolio:
And the change does not fix the problem, still it occurs. also some flickers on the tab indicator on tab press

tom-richtr:
I can confirm that the bug is still present even with the proposed changes.
However I can confirm that refactoring to reanimated could fix the behaviour
See the changes mentioned in
https://github.com/react-navigation/react-navigation/issues/12646
https://github.com/hoangtrunnam/react-navigation/tree/merge_dev

AlexCernik:
To test the fix, update `react-native-tab-view/lib/module/PagerViewAdapter.js` with the proposed solution.

scole-rbi:
> To test the fix, update `react-native-tab-view/lib/module/PagerViewAdapter.js` with the proposed solution.

- `offset.setValue(0)` fix from this PR significantly reduces indicator jump
- **_some residual stutter remains_** but the indicator no longer misaligns after settling.

- also found that removing the start CSS property from TabBarIndicator (which conflicts with the animated translateX during re-renders) further improves the result when combined with this fix

Note: the issue is reproducible on slow swipes too -- it's amplified by heavy screen content
- confirmed with minimal test app (same libraries, lightweight content) that jank does not appear, and that with a fully custom React.memo-blocked tab bar that the issue persists -- proving it originates in the native `onPageScroll` offset pipeline, not tab bar rendering

Tested: RN 0.81.5, Fabric, Android, Expo SDK 54, react-native-tab-view 4.3.0, react-native-pager-view 6.9.1

troZee:
It has been fixed here https://github.com/callstack/react-native-pager-view/releases/tag/v8.0.1

Issue #12685, opened 2025-07-15
Unnecessary rerender of routes, on inner navigator state changes


Ubax:
### Current behavior

In production, when there are state changes in inner navigator, parent navigator is still rerendered. 

The cause for this is this check: https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/useRouteCache.tsx#L26

For example when I have:
```
TabNavigator
|- TabA
  |- Stack
    |- InnerStack
      |- Index
      |- Modal
```

When I navigate from Index to Modal, TabA is unnecessarily rerenderd.

### Expected behavior

The same as in development, caching can be used for routes to prevent rerenders

### Reproduction

https://snack.expo.dev/@jtkacz/moody-green-popcorn

### Platform

- [x] Android
- [x] iOS
- [x] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 7.012 |
| @react-navigation/bottom-tabs          | 7.1.1 |
| @react-navigation/native-stack         | 7.1.13 |
| react-native-screens                   | 4.10.0 |
| react-native-safe-area-context         | 5.3.0 |
| react-native                           | 0.79.1 |
| expo                                   | 53.0.4 |
| npm or yarn                            | npm |

PR #12686, opened 2025-07-15
fix: remove production check in useRouteCache

Ubax:
https://github.com/react-navigation/react-navigation/issues/12685

I think that the cost of unnecessary re-renders of each navigator is higher than the cost of computing map for each route, especially considering complex pages. However, I don't have concrete evidence for that.

**Test plan**

1. Run the app in the production and development


satya164 (approved):
Thanks for the PR

PR #12728, opened 2025-08-22
feat: optional callback added to `useScrollToTop` hook

vargajacint:
**Motivation**

Added an optional `callback` parameter to the `useScrollToTop` hook, allowing us to link code directly to the scroll-to-top event, such as refreshing the list.

**Test plan**

```ts
import { useScrollToTop } from "@react-navigation/native";

const refresh = useCallback(() => ap.refetch(...), [])

useScrollToTop(listRef, refresh);
```

With the example above, `refresh` function gets called after the scroll-to-top is done

### DIFF
--- a/packages/native/src/useScrollToTop.tsx
+++ b/packages/native/src/useScrollToTop.tsx
@@ -50,7 +50,10 @@ function getScrollableNode(ref: React.RefObject<ScrollableWrapper>) {
   }
 }
 
-export function useScrollToTop(ref: React.RefObject<ScrollableWrapper>) {
+export function useScrollToTop(
+  ref: React.RefObject<ScrollableWrapper>,
+  callback?: () => void
+) {
   const navigation = React.useContext(NavigationContext);
   const route = useRoute();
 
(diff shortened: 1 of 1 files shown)


vargajacint:
hey @satya164 
what do you think about it?

satya164:
Thanks for the PR. However, I'm not sure about this. There are a few problems with this approach:

- It's usually possible to scroll to top from other places, for example tapping on the statusbar area, doing different things on tab press only is inconsistent
- For native tabs, eventually this behavior will be handled natively and not via this hook, leading to more inconsistencies
- API-wise, when I look at a callback, I'd expect the callback to be called after the operation happens, but this callback gets called at the same time the scroll starts, which may not be intuitive
- If data refreshes while scroll animation is happening, it may lead to jank
- More of a personal opinion, but as a user, I wouldn't expect data to be refreshed just by scrolling, more so if I have only scrolled a little bit

I think if this functionality is needed, a better API would be to have a callback on the scroll view/list that detects when the user scrolls to top. That'd be a more intuitive API and will avoid inconsistencies.

Technically, this can already be achieved with the current API by passing a custom object for the ref.

PR #12738, opened 2025-09-02
fix(bottom-tabs): fix accessibility hint and label

uffoltzl:
**Motivation**

I have noticed that accessibility labels on tabs are automatically generated for bottom tabs with a tab number for iOS. 
In the codebase, I saw those lines

          const accessibilityLabel =
            options.tabBarAccessibilityLabel !== undefined
              ? options.tabBarAccessibilityLabel
              : typeof label === 'string' && Platform.OS === 'ios'
                ? `${label}, tab, ${index + 1} of ${routes.length}`
                : undefined;

My conviction is that it should be in the value and not the label. 

I have tested some common apps and it seems normal for the screen reader to read by default "tab x of y", just like suggested in react-navigation repo but
- a label is easily overiden by a developper, so we will lose this information using a label
- "tab x of y" is always said be the screen reader and not on command so using accessibility hint does not seem recommended
- [react-native accessibility value documentation](https://reactnative.dev/docs/accessibility#accessibilityvalue) seems to perfectly correspond to what we need

In addition, since "tab" is always said if setting it inside the value, we can remove our fake "button" role.

**Test plan**

Before

image

After

image

I am currently working with @BrittaEvansFenton, @gabrielledevz, @GuillaumeEgret and @tychota on accessibility. They contributed as well.


satya164:
Closing due to no response.

PR #12751, opened 2025-09-11
feat: implement route.history & PUSH_PARAMS

satya164:


packages/routers/src/BaseRouter.tsx
```diff
@@ -42,6 +42,32 @@ export const BaseRouter = {
...
+              ? {
+                  ...r,
+                  params: action.payload.params,
+                  history: [
+                    ...(r.history ?? []),
+                    { type: 'params', params: r.params },
+                  ],
+                }
+              : r
+          ),
```
vargajacint, line 67: I think we don't need the extra iteration, since we know the index. Something like this:

```suggestion
if(state.routes?.[index]) {
 state.routes[index] = {
                  ...r,
                  params: action.payload.params,
                  history: [
                    ...(r.history ?? []),
                    { type: 'params', params: r.params },
                  ],
                }
}

return {
 ...state,
 routes: ...state.routes
}
```

satya164: That would mutate the state which we can't do

PR #12753, opened 2025-09-12
chore: upgrade expo sdk to 54

satya164:


biskyy:
hi. im guessing this is also going to handle the removal of useAnimatedGestureHandler from react reanimated v4 right?

satya164:
@biskyy i don't think we're using that API 🤔

biskyy:
> @biskyy i don't think we're using that API 🤔

are you sure? i've upgraded my project to sdk 54 and it's been giving me these errors:
```
[Reanimated] The `isReanimated3` function is deprecated. Please use the exported variable `reanimatedVersion` instead
[Error: The `useLegacyImplementation` prop is not available with Reanimated 3 as it no longer includes support for Reanimated 1 legacy API. Remove the `useLegacyImplementation` prop from `Drawer.Navigator` to be able to use it.]
```
if i set useLegacyImplementation to false on my Drawer then i get 
```
[TypeError: 0, _reactNativeReanimated.useAnimatedGestureHandler is not a function (it is undefined)]
```
which kinda matches with [react reanimated v4 docs](https://docs.swmansion.com/react-native-reanimated/docs/guides/migration-from-3.x/#removed-useanimatedgesturehandler) so i thought it is related to this pr

maybe im also doing something wrong and this is not related to react navigation at all?

satya164:
@biskyy you're using an old version. `useLegacyImplementation` was removed in React Navigation 7

biskyy:
> @biskyy you're using an old version. `useLegacyImplementation` was removed in React Navigation 7

bruuh.. yeah i didnt realize i was on an older version, sorry. can you mark my comments as off-topic

maniolek:
Hi there, when may we expect this change to be released?

satya164:
@maniolek this upgrades packages in the monorepo. not a user-facing change.

PR #12764, opened 2025-09-18
feat - add header config for hidesSharedBackground for iOS 26

intergalacticspacehighway:
**Motivation**

This option enables us to remove the default glass effect buttons in header right and left items on iOS 26. Allowing us to customise the items as needed.

PR in screens - https://github.com/software-mansion/react-native-screens/pull/3229

**Test plan**

The example is added in screens [here](https://github.com/software-mansion/react-native-screens/pull/3229). Here we only add the types and pass the option to native component.


intergalacticspacehighway:
Closing in favor of https://github.com/react-navigation/react-navigation/pull/12657

PR #12766, opened 2025-09-19
fix: remove aria-selected on elements with role link

### LINKED ISSUE #9418: "aria-selected" should not be set on `role="link"`
**Current Behavior**

the current `DrawerItem` passes `accessibilityState` to `Link`. this is an invalid attribute for `role="link"`

**Expected Behavior**

also exclude `accessibilityState` from `Link` in the default `DrawerItem`

all that really needs to be done is add a line under [this line](https://github.com/react-navigation/react-navigation/blob/0fce8eb2a1e682194d574c65f3366f72e23f2e1c/packages/drawer/src/views/DrawerItem.tsx#L88) with `accessibilityState`, to exclude it from `rest`, and pass it manually to [`PlatformPressable`](https://github.com/react-navigation/react-navigation/blob/0fce8eb2a1e682194d574c65f3366f72e23f2e1c/packages/drawer/src/views/DrawerItem.tsx#L123)

I would open a simple PR, but I'm on a phone and I don't have time to write tests or a computer at the moment.

**How to reproduce**

[Here's a simple snack: https://snack.expo.io/x4ass6Am_](https://snack.expo.io/x4ass6Am_)

**Your Environment**

| iOS or Android                 | Web
| @react-navigation/drawer       | 5.12.4

(other items in the table are irrelevant, as the offending code only exists in the lines of code I linked above)

DilinaCP:
### Motivation

Currently, elements with `role="link"` were being assigned the `aria-selected` attribute, which is incorrect according to accessibility guidelines. `aria-selected` is meant for elements in a selectable group (like tabs or list items) and should not appear on standard links.

This PR removes the `aria-selected` attribute from elements with `role="link"` to improve accessibility compliance and prevent screen readers from announcing misleading information.

Fixes: #9418

### Test Plan

1. Run all existing tests to ensure nothing breaks:

   yarn test

2. Verify snapshots for Link component no longer include `aria-selected` for role="link" elements.  

3. Manually check that web links still navigate correctly and styling remains unaffected.

4. Lint and TypeScript checks:

   yarn lint
   yarn tsc --noEmit

All tests pass, lint is clean, and types are valid.

### DIFF
--- a/example/e2e/tests/Link.test.ts
+++ b/example/e2e/tests/Link.test.ts
@@ -22,6 +22,62 @@ test('loads the article screen', async ({ page }) => {
   );
 });
 
+test('replaces and pushes params', async ({ page }) => {
+  await page.waitForURL('**/link-component/article/gandalf');
+
+  await page.getByRole('button', { name: 'Push params' }).click();
+
+  await page.waitForURL('**/link-component/article/babel-fish');
+
+  await expect(page).toHaveTitle(
+    'Article by Babel fish - React Navigation Example'
+  );
+
+  await page.getByRole('button', { name: 'Push params' }).click();
+
+  await page.waitForURL('**/link-component/article/gandalf');
+
+  await expect(page).toHaveTitle(
+    'Article by Gandalf - React Navigation Example'
+  );
+
+  await page.getByRole('link', { name: 'Go back' }).click();
+
+  await page.waitForURL('**/link-component/article/babel-fish');
+
+  await expect(page).toHaveTitle(
+    'Article by Babel fish - React Navigation Example'
+  );
+
+  await expect(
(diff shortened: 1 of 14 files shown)


DilinaCP:
The E2E failure seems unrelated to my changes. Could a maintainer please re-run the CI to confirm? Thanks!

satya164 (changes requested):
Thanks for the PR.

However, this doesn't seem to fix the linked issue. `DrawerItem`, `TabBarItem`, and `BottomTabBarItem` don't use `Link`, but `PlatformPressable`, and not `Link`.

The proper fix would be to conditionally set `aria-selected` on these based on whether `href` is `undefined`.

```js
aria-selected={href ? undefined : isFocused}
```

PR #12767, opened 2025-09-22
fix: use popTo in popToTop

juliesaia:
**Motivation**

[When `popToTop` was first created](https://github.com/react-navigation/react-navigation/commit/7e050044da27c3e1800dfb2c23a7fb9813037663), `popTo` didn't exist yet, so it was implemented by using `pop` with a count.

This breaks the case where someone starts navigating from a screen deep in a stack, because the top screen hasn't been put in the stack yet. This is most obvious when tapping an already-active bottom tab, which is supposed to navigate to the top screen always.

**Test plan**

New test added in `StackRouter.test.tsx`

Manually, you can create a stack navigator inside a tab navigator, hard navigate to the nested page, then tap the active tab

### DIFF
--- a/packages/native-stack/src/navigators/createNativeStackNavigator.tsx
+++ b/packages/native-stack/src/navigators/createNativeStackNavigator.tsx
@@ -61,7 +61,6 @@ function NativeStackNavigator({
         // This is necessary to know if preventDefault() has been called
         requestAnimationFrame(() => {
           if (
-            state.index > 0 &&
             isFocused &&
             !(e as EventArg<'tabPress', true>).defaultPrevented
           ) {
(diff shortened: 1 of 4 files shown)


satya164:
Thanks for the PR.

However, this changes meaning of `popToTop` - unlike `popTo` where you want to navigate back to a specific screen, `popToTop` is for popping all other screens except the first screen in the current stack. While the initial route can be the first screen in the stack, it won't always be. So going back to the initial route (from `getInitialState`) has a very different meaning - which could be more appropriate if it was `popToInitial`.

This can also break existing code since unlike the previous implementation, this will always be handled by the current stack navigator and can't bubble up to a parent, which can break some assumptions.

The implementation also doesn't apply params, which will break screens that need params. And the API doesn't allow passing params - so it's very rigid.

I'd instead recommend using `popTo(ScreenName)` with explicit screen name if that's the behavior you want. You can also override the action for your use case with `UNSTABLE_router` prop on the navigator.

Because of this, I'm closing the PR.

PR #12790, opened 2025-10-10
feat: useNamedRoute

osdnk:

**Motivation**

**Test plan**


packages/core/src/NavigationRouteContext.tsx
```diff
@@ -1,9 +1,88 @@
...
+  const parent = React.useContext(NavigationRouteParentContext);
+
+  const getRoute = useLatestCallback((name: string) => {
+    if (value?.name === name) {
+      return value;
+    }
+    return parent?.getRoute(name);
+  });
+
+  const routeListeners = React.useRef(new Set<() => void>()).current;
```
satya164, line 46: Use `useLazyValue` to avoid creating the set unncessarily after first render

packages/core/src/NavigationRouteContext.tsx
```diff
@@ -1,9 +1,88 @@
...
+
+  const notifyRouteListeners = useLatestCallback(() => {
+    routeListeners.forEach((listener) => listener());
+  });
+
+  const previousValue = React.useRef(value);
+  if (previousValue.current !== value) {
+    previousValue.current = value;
+    notifyRouteListeners();
+  }
```
satya164, line 71: We can't do this in render. It needs to be in an effect:


```suggestion
  React.useInsertionEffect(() => {
    routeListeners.forEach((listener) => listener());
  }, [value]);
```

Also, storing the previous value and comparing may not be necessary - does `useSyncExternalStore` still cause re-render if the value didn't change?

Though after moving to `useInsertionEffect`, the dependency array will take care of comparing.

Alternatively, this could use `useLayoutEffect`.

Wish there was a way to do this without all these effects.

packages/core/src/NavigationRouteContext.tsx
```diff
@@ -1,9 +1,88 @@
...
+  children: React.ReactNode;
+}) {
+  const parent = React.useContext(NavigationRouteParentContext);
+
+  const getRoute = useLatestCallback((name: string) => {
+    if (value?.name === name) {
+      return value;
+    }
+    return parent?.getRoute(name);
+  });
```
satya164, line 44: I'm wondering if `useLatestCallback` would cause issues. It may get updated too late, which will cause tearing (different components get the value at different times).

Though it's unavoidable.

packages/core/src/useRoute.tsx
```diff
@@ -20,3 +23,36 @@ export function useRoute<T extends RouteProp<ParamListBase>>(): T {
...
+      }
+      return r;
+    },
+    () => {
+      const r = routeWrapper.getRoute(name);
+      if (!r) {
+        throw new Error(`Route '${name}' not found in the navigator`);
+      }
+      return r;
+    }
```
satya164, line 54: since both of them do the same thing, I'd assign it to a function and reuse.

satya164:
I thought of a different approach to implement this that'll avoid all of the problems with tearing.

Basically, the idea is that we create contexts for each route dynamically and store them in a list. When we want to get the route for a specific name, we just look it up in the list and use the corresponding context.

First, the provider may look something like this:

```js
export const NamedRouteContextListContext = React.createContext({})

function RouteContextProvider({ route, children }) {
  const NamedRouteContext = useLazyValue(() => React.createContext(route))

  const parents = useContext(NamedRouteContextListContext)

  const NamedRouteContextList = useMemo(() => {
    ...parents,
    [route.name]: NamedRouteContext,
  }, [NamedRouteContext, parents, route.name])

  return (
    <NamedRouteContextListContext.Provider value={NamedRouteContextList}>
      <NamedRouteContext.Provider value={route}>
        <NavigationRouteContext.Provider value={route}>
          {children}
        </NavigationRouteContext.Provider>
      </NamedRouteContext.Provider>
    </NamedRouteContextListContext.Provider>
  )
}
```

Here, we create a context `NamedRouteContext`, and then keep it in a list `NamedRouteContextListContext`.

Then we have logic inside `useRoute` where we use the appropriate context based on the name:

```js
function useRoute(name) {
  const NamedRouteContextList = useContext(NamedRouteContextListContext);

  if (name !== undefined) {
    const NamedRouteContext = NamedRouteContextList[name];

    if (NamedRouteContext === undefined) {
      throw new Error(
        `Couldn't find a route named '${name}'. Is your component inside the correct screen?`,
      );
    }

    const route = use(NamedRouteContext);

    return route;
  }

  const route = use(NavigationRouteContext);

  if (route === undefined) {
    throw new Error(
      "Couldn't find a route object. Is your component inside a screen in a navigator?",
    );
  }

  return route;
}
```

satya164 (commented):
I thought of a different approach to implement this that'll avoid all of the problems with tearing.

Basically, the idea is that we create contexts for each route dynamically and store them in a list. When we want to get the route for a specific name, we just look it up in the list and use the corresponding context.

First, the provider may look something like this:

```js
export const NamedRouteContextListContext = React.createContext()

function NavigationRouteProvider({ route, children }) {
  const NamedRouteContext = useLazyValue(() => React.createContext(route))

  const parents = use(NamedRouteContextListContext)

  const NamedRouteContextList = useMemo(() => {
    ...parents,
    [route.name]: NamedRouteContext,
  }, [NamedRouteContext, parents, route.name])

  return (
    <NamedRouteContextListContext.Provider value={NamedRouteContextList}>
      <NamedRouteContext.Provider value={route}>
        <NavigationRouteContext.Provider value={route}>
          {children}
        </NavigationRouteContext.Provider>
      </NamedRouteContext.Provider>
    </NamedRouteContextListContext.Provider>
  )
}
```

Here, we create a context `NamedRouteContext`, and then keep it in a list `NamedRouteContextListContext`.

Then we have logic inside `useRoute` where we use the appropriate context based on the name:

```js
function useRoute(name) {
  if (name !== undefined) {
    const NamedRouteContextList = use(NamedRouteContextListContext);

    if (NamedRouteContextList === undefined) {
      throw new Error(
        "Couldn't find a parent screen. Is your component inside a screen in a navigator?"
      );
    }

    const NamedRouteContext = NamedRouteContextList[name];

    if (NamedRouteContext === undefined) {
      throw new Error(
        `Couldn't find a route named '${name}' in any of the parent screens. Is your component inside the correct screen?`,
      );
    }

    const route = use(NamedRouteContext);

    return route;
  }

  const route = use(NavigationRouteContext);

  if (route === undefined) {
    throw new Error(
      "Couldn't find a route object. Is your component inside a screen in a navigator?",
    );
  }

  return route;
}
```

Let's try this instead.

packages/bottom-tabs/src/views/BottomTabBar.tsx
```diff
@@ -440,7 +440,7 @@ export function BottomTabBar({
               key={route.key}
               value={descriptors[route.key].navigation}
             >
-              <NavigationRouteContext.Provider value={route}>
+              <NavigationRouteContextProvider value={route}>
```
satya164, line 443: I added a `NavigationProvider` so we won't need to change all of these places.

Let's move all context for route and navigation there to simplify things.

packages/core/src/__tests__/useRoute.test.tsx
```diff
@@ -37,3 +37,34 @@ test('gets route prop from context', () => {
     </BaseNavigationContainer>
   );
 });
+
+test('gets parent route from context', () => {
```
satya164, line 41: We should add some more tests for `useRoute(name)`:

- returns route for the current route
- returns route for the parent route
- throws error for invalid route
- throws error for child route

PR #12795, opened 2025-10-16
fix: update back button configuration logic to match new logic in `react-native-screens`

kligarski:
**Motivation**

Due to changes in native back button behavior starting from iOS 26, logic handling configuration of back button item needed to change in `react-native-screens`. We managed to handle more cases correctly and simplify the API. For more details, please see [the PR in `react-native-screens`](https://github.com/software-mansion/react-native-screens/pull/3303).

**Backward compatibility**

By using `backButtonUseModernImplementation` flag, we maintain backward compatibility. If outdated version of `react-navigation` or `react-native-screens` will be used, previous implementation will be used. Passing `backButtonDisplayMode` directly in all cases will not break previous implementation. For more details, see `Compatibility with previous versions` section in [PR from `react-native-screens`](https://github.com/software-mansion/react-native-screens/pull/3303).

**Test plan**

Back button should follow configuration table posted in [the PR in `react-native-screens`](https://github.com/software-mansion/react-native-screens/pull/3303). With default configuration on iOS 26, back button should only display the chevron (without any text).

### DIFF
--- a/packages/native-stack/src/views/useHeaderConfigProps.tsx
+++ b/packages/native-stack/src/views/useHeaderConfigProps.tsx
@@ -175,16 +175,6 @@ export function useHeaderConfigProps({
       Platform.OS === 'ios' &&
       headerTransparent !== false);
 
-  const isBackButtonDisplayModeAvailable =
-    // On iOS 14+
-    Platform.OS === 'ios' &&
-    parseInt(Platform.Version, 10) >= 14 &&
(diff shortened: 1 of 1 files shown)


satya164:
This would be a breaking change in React Navigation 7 if it needs a newer version of screens to work. We can do this for React Navigation 8, but we will need to figure out how to keep compatibility in v7.

PR #12796, opened 2025-10-17
feat: move Badge to elements and expose it in public API

Trancever:
**Motivation**

Badge is currently not exposed via public API of the lib which makes it impossible to use it when building custom tab.

**Test plan**

Run example app and check that nothing is broken. This PR only moves the file and exposes component in elements package public API.

I renamed component from Badge to TabBarIconBadge so it's clear what it is for, but if it's not desired, let me know and I will fix it.


satya164 (approved):
Thanks for the PR

PR #12805, opened 2025-10-24
fix: prevent header from disappearing after modal navigation

### LINKED ISSUE #12456: Home screen headerStyle ignored (reset) after new screen in the stack is presented as modal
### Current behavior

- Three screens in the stack Navigator (Home/Second/Modal).
- The first 2 are presented with the default style (Home/Second).
- The third screen has _presentation: 'modal'_ (Modal).
- All three screens have an _headerStyle backgroundColor orange_ assigned as options (Home/Second/Modal).
- The navigation goes from Home to Second, Second to Modal.
- When navigating from the Second screen to the Modal and then back to the Home screen, the header of the Home screen is now white instead of orange.

https://github.com/user-attachments/assets/5d5ae8fe-3ad9-49af-8b53-aed3e55fda0f

### Expected behavior

The header stays consisted with the assigned _headerStyle_ no matter what happens to the navigation stack

### Reproduction

https://github.com/ANTNNJT/ReactNavigationHeaderStyleBug

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| @react-navigation/native               | 7.0.14
| @react-navigation/stack                | 7.1.1
| react-native-screens                   | 4.9.0
| react-native-safe-area-context         | 5.2.0
| react-native-gesture-handler           | 2.24.0
| react-native                           | 0.78.0
| node                                   | 22.9.0
| npm                          | 10.8.3

rthor:
## Problem

Floating headers become invisible after navigating back from a modal screen that uses `ModalPresentationIOS` transition.

### Reproduction Steps

1. Navigate to Screen A (header visible ✓)
2. Navigate to Screen B (header visible ✓)
3. Navigate to Modal Screen C with `presentation: 'modal'`, `headerShown: false`, and `ModalPresentationIOS` transition preset
  (this is how we discovered the issue).
4. Navigate back to Screen B (header visible ✓)
5. Navigate back to Screen A (header **invisible** ✗)

### Root Cause

The `floatingHeader` element in `CardStack.tsx` is conditionally rendered in two different positions in the React tree based on the `isFloatHeaderAbsolute` flag:

```tsx
return (
    {isFloatHeaderAbsolute ? null : floatingHeader}  // Position A
    <MaybeScreenContainer>...</MaybeScreenContainer>
    {isFloatHeaderAbsolute ? floatingHeader : null}  // Position B
  </View>
);
```

Both positions use the same React key (`key="header"`). When `isFloatHeaderAbsolute` changes from `true` to `false` (or vice versa) during navigation, React attempts to reuse the component by moving it between positions in the tree. However, this causes the component to fail to properly commit to the native view hierarchy, resulting in invisible header elements.

## Solution

Use different React keys for different render positions to force React to properly unmount/remount the header when it moves between positions:

```tsx
const floatingHeader = (
  <React.Fragment key={isFloatHeaderAbsolute ? 'header-absolute' : 'header-float'}>
    {floatingHeaderContent}
  </React.Fragment>
);
```

## Testing

> Note: We are running v6.x in our project and that's were we found the bug and developed the fix.

Tested with the reproduction steps above - headers now remain visible throughout the navigation flow.

## Related Issues

Fixes #12456

### DIFF
--- a/packages/stack/src/views/Stack/CardStack.tsx
+++ b/packages/stack/src/views/Stack/CardStack.tsx
@@ -653,7 +653,9 @@ export class CardStack extends React.Component<Props, State> {
     }
 
     const floatingHeader = (
-      <React.Fragment key="header">
+      <React.Fragment
+        key={isFloatHeaderAbsolute ? 'header-absolute' : 'header-float'}
+      >
         {renderHeader({
           mode: 'float',
           scenes,


rthor:
In the case of a same screen, you should maybe use a route key instead. Not really sure what a zindex would do here or what component you want it on? The stacking context isn't really the problem, it's the fact that the previous screen doesn't get rerendered correctly after the modal transition.

satya164:
> In the case of a same screen, you should maybe use a route key instead

Keys are unique to each React tree. Keys from 2 different trees don't conflict with each other. Using `route.key` won't do anything different.

> Not really sure what a zindex would do here or what component you want it on? The stacking context isn't really the problem

You said, "React attempts to reuse the component by **moving it between positions in the tree**. However, this causes the component to fail to properly commit to the native view hierarchy, resulting in invisible header elements".

If this is the issue, then using a `zIndex` to position it on the top instead of re-ordering is a potential fix.

Using a different key isn't the solution. It does the same thing as not using a key. The key is there intentionally, so the element is reused.

rthor:
A good point on intentionality of the key. 

Rendering the item without moving, with a zIndex higher than the content doesn't fix it. The header fades out and doesn't render the "initial" History header on the last back action. 

https://github.com/user-attachments/assets/7ce78dfa-b36f-43c5-ab5b-f08a728de8ec

The key trick works for us as the header never holds a local state, so we can use a patch package for that for now. This of course isn't true for everyone, and definitely would be better find a good solution with you. Got any ideas?

osdnk:
Hi, @rthor 
Thanks for you contribiution. We decided for a slightly different approach, yet eventually resolving the issues.

https://github.com/react-navigation/react-navigation/pull/12833

Indeed, we just reduced the number of held headers to recent 2, not 3.

PR #12810, opened 2025-10-27
fix: Change error handling in useBottomTabBarHeight

philipheinser:
Replace error throw with console warning and return 0. To make components using it more composable.


satya164:
You can use `BottomTabBarHeightContext` directly if you don't want an error to be thrown when not inside bottom tabs as [documented](https://reactnavigation.org/docs/bottom-tab-navigator/#usebottomtabbarheight):

> Alternatively, you can use the `BottomTabBarHeightContext` directly if you are using a class component or need it in a reusable component that can be used outside the bottom tab navigator

The error is intentional so people can catch mistakes early. A warning is worse since you'd have no way to ignore it, and you'd have to use `BottomTabBarHeightContext` either way.

PR #12225, opened 2024-11-04
feat: add support for custom screen transitions with native-stack v7 

maciekstosio:
Continuation of #12204 and #11943. Changes from original version:

- fixes to CST on react-native-screens: https://github.com/software-mansion/react-native-screens/pull/2449
- refactor to CST to move GestureDetector to react-native-screens: https://github.com/software-mansion/react-native-screens/pull/2450
- refactor to CST to use context instead of passing screenRefs: https://github.com/software-mansion/react-native-screens/pull/2476

In general, the idea suggested by @satya164 was that we could move more logic into react-native-screens and avoid leaking it to @react-navigation. 

Original description:

**Motivation**

This PR intents to add implementation of custom screen transitions, recently added in react-native-screens. You can find more about this feature [here](https://github.com/software-mansion/react-native-screens/pull/1913).

**Test plan**

Go to the code of the native stack in example (NativeStack.tsx) and wrap whole navigator with <GestureDetectorProvider> (import from `react-native-screens/gesture-handler`). After that, use `gestureType` prop on the desired screen.

<details>
<summary>Ready to paste code 🍝</summary>

```tsx
/* eslint-disable camelcase */
import * as React from 'react';
import {
  Animated,
  Platform,
  StyleSheet,
  ViewProps,
  ViewStyle,
} from 'react-native';
// @ts-ignore Getting private component
// eslint-disable-next-line import/no-named-as-default, import/default, import/no-named-as-default-member, import/namespace
import AppContainer from 'react-native/Libraries/ReactNative/AppContainer';
import warnOnce from 'warn-once';
import { StackPresentationTypes, ScreensRefsHolder } from '../../types';
import ScreenStack from '../../components/ScreenStack';
import ScreenContentWrapper from '../../components/ScreenContentWrapper';
import { ScreenContext } from '../../components/Screen';
import {
  ParamListBase,
  StackActions,
  StackNavigationState,
  useTheme,
  Route,
  NavigationState,
  PartialState,
} from '@react-navigation/native';
import {
  useSafeAreaFrame,
  useSafeAreaInsets,
} from 'react-native-safe-area-context';
import {
  NativeStackDescriptorMap,
  NativeStackNavigationHelpers,
  NativeStackNavigationOptions,
} from '../types';
import HeaderConfig from './HeaderConfig';
import SafeAreaProviderCompat from '../utils/SafeAreaProviderCompat';
import getDefaultHeaderHeight from '../utils/getDefaultHeaderHeight';
[...]

PR #12811, opened 2025-10-27
chore: Update fullScreenGesture-related docblocks

kmichalikk:

**Motivation**

This PR updates the docblocks for `fullScreenGestureEnabled`, `fullScreenGestureShadowEnabled` to accomodate recent changes in `react-native-screens`.

**Test plan**

--

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -483,8 +483,7 @@ export type NativeStackNavigationOptions = {
    * with `animation` & `animationMatchesGesture` props, but default iOS swipe animation is not achievable.
    *
    * On iOS 26 and up:
-   * `true` by default. Swipe gesture will use the new [contentPopGestureRecognizer](https://developer.apple.com/documentation/uikit/uinavigationcontroller/interactivecontentpopgesturerecognizer)
-   * introduced in iOS 26 if possible. Set the prop to `false` to match iOS 18 default behavior.
+   * `true` by default to match new native behavior. You can still customize it with `animation` & `animationMatchesGesture` props.
    *
    * Doesn't affect the behavior of screens presented modally.
    *
@@ -492,14 +491,13 @@ export type NativeStackNavigationOptions = {
    */
   fullScreenGestureEnabled?: boolean;
   /**
-   * Whether the full screen dismiss gesture has shadow under view during transition. The gesture uses custom transition and thus
-   * doesn't have a shadow by default. When enabled, a custom shadow view is added during the transition which tries to mimic the
-   * default iOS shadow. Defaults to `true`.
+   * iOS 18 and below. Controls whether the full screen dismiss gesture has shadow under view during transition.
+   * The gesture uses custom transition and thus doesn't have a shadow by default. When enabled, a custom shadow view
+   * is added during the transition which tries to mimic the default iOS shadow. Defaults to `true`.
    *
    * This does not affect the behavior of transitions that don't use gestures, enabled by `fullScreenGestureEnabled` prop.
    *
-   * @deprecated since iOS 26, full screen swipe is handled by native recognizer, and this prop is ignored. We still fallback
-   * to the legacy implementation when handling custom animations, but we assume `true` for shadows.
+   * @deprecated since iOS 26.
    *
    * @platform ios
    */


packages/native-stack/src/types.tsx
```diff
@@ -476,9 +476,14 @@ export type NativeStackNavigationOptions = {
    */
   animationMatchesGesture?: boolean;
   /**
-   * Whether the gesture to dismiss should work on the whole screen. Using gesture to dismiss with this option results in the same
-   * transition animation as `simple_push`. This behavior can be changed by setting `animationMatchesGesture` prop. Achieving the
-   * default iOS animation isn't possible due to platform limitations. Defaults to `false`.
+   * Whether the gesture to dismiss should work on the whole screen. The behavior depends on iOS version.
+
```
satya164:
```suggestion
   * Whether the gesture to dismiss should work on the whole screen. The behavior depends on iOS version.
   *
```

packages/native-stack/src/types.tsx
```diff
@@ -476,9 +476,14 @@ export type NativeStackNavigationOptions = {
...
+   * For iOS prior to 26, swiping with this option results in the same transition animation as `simple_push` by default.
+   * It can be changed to other custom animations with `customAnimationOnSwipe` prop, but default iOS swipe animation
+   * is not achievable due to usage of custom recognizer.
```
satya164: `customAnimationOnSwipe` is named as `animationMatchesGesture`.

I think something like this could be nicer (seeing iOS 18 is more glanceable than prior to 26):

```suggestion
   * On iOS 18 and below:
   * If this is enabled, swipe gestures will use `simple_push` transition animation and cannot be customized.
```

For this part:

> It can be changed to other custom animations with `customAnimationOnSwipe` prop

Correct me if I'm wrong. I guess this means that:

> If a custom animation is used, swipe gestures will match the custom animation if `animationMatchesGesture` is set to `true`.

Though I think I'm not sure if the animations specified in `animation` prop can be categorized as "custom animation" or not. So we should clarify.

If I understood correctly, this is what it's referring to: https://github.com/react-navigation/react-navigation/pull/12225 - but it's in draft and isn't merged. So for now, it'd be better to omit this section about custom animation (if this is what it is).

> but default iOS swipe animation is not achievable due to usage of custom recognizer.

I would omit the custom recognizer part since it's an implementation detail and most people won't understand it. We could mention it as a note in docs site instead.

packages/native-stack/src/types.tsx
```diff
@@ -476,9 +476,14 @@ export type NativeStackNavigationOptions = {
...
+   *
+   * For iOS 26 and up, native `interactiveContentPopGestureRecognizer` is used, and this prop controls whether it should
+   * be enabled or not.
```
satya164: What does this mean from the user's point of view? What's the difference between `interactiveContentPopGestureRecognizer` being enabled vs not? We should mention the final result here, i.e. what happens if I set it to `false` on iOS vs `true` (default).

We should also mention it defaults to `true` on iOS 26+ (if I understand correctly).

packages/native-stack/src/types.tsx
```diff
@@ -476,9 +476,14 @@ export type NativeStackNavigationOptions = {
...
    *
    * Doesn't affect the behavior of screens presented modally.
```
satya164, line 488: I think this one also needs a bit of clarification. From testing seems full screen gesture is always enabled on modals, so we should also mention it.

packages/native-stack/src/types.tsx
kmichalikk: [1346b82](https://github.com/react-navigation/react-navigation/pull/12811/commits/1346b82eb4183484d111573982045720f17ed115)

packages/native-stack/src/types.tsx
kmichalikk: [1346b82](https://github.com/react-navigation/react-navigation/pull/12811/commits/1346b82eb4183484d111573982045720f17ed115)

I tried to reword it a bit. Separated iOS 18 / 26, removed the 'custom animation' bit and implementation details, but I left the custom recognizer part, and linked to apple docs - I think this was relatively big change from Apple to allow for content swipe, so whoever uses the prop will probably know what we're trying to say here. As for #12225, I wasn't aware of it tbh - will talk about it to @kkafar; `custom recognizer` refers to our internal implementation of content swipe which used UIPanGestureRecognizer, and in our docs we just want to say that there might be slight differences in behavior because of that. I'm removing this part.

packages/native-stack/src/types.tsx
kmichalikk: [1346b82](https://github.com/react-navigation/react-navigation/pull/12811/commits/1346b82eb4183484d111573982045720f17ed115)

packages/native-stack/src/types.tsx
kmichalikk: I'm not sure what to do with this. Until recently, we had docs for this prop synced, apart from this line, and I don't know why it was only added here.

Anyways, if you add a stack inside a modal, then this prop actually has no effect, and it behaves like native, so false for iOS <= 18, true for iOS >= 26. For now, it's technically correct.

satya164: > if you add a stack inside a modal, then this prop actually has no effect, and it behaves like native, so false for iOS <= 18, true for iOS >= 26. For now, it's technically correct

That's a rare case, and we don't recommend nesting stacks like that. It doesn't make sense to treat that behavior as correct. We should be more specific about what the behavior is, and if the behavior when a stack is nested is different, then we should mention it as well.

We've already had many issues about people not understanding how and when gestures work; I didn't understand either and asked for clarification in the past. Our documentation should be clear about the behavior so there are no assumptions or confusion.

packages/native-stack/src/types.tsx
```diff
@@ -492,6 +498,9 @@ export type NativeStackNavigationOptions = {
    *
    * This does not affect the behavior of transitions that don't use gestures, enabled by `fullScreenGestureEnabled` prop.
    *
+   * @deprecated since iOS 26, full screen swipe is handled by native recognizer, and this prop is ignored. We still fallback
+   * to the legacy implementation when handling custom animations, but we assume `true` for shadows.
```
satya164: For a user, what does "handling custom animations" refer to? Since we don't support custom animations.

We should read these comments from user's point of view who doesn't know anything about screens or iOS native dev. The majority of the users of React Navigation aren't going to be familiar with native iOS dev.

packages/native-stack/src/types.tsx
satya164: I'll emphasize that the majority of React Navigation users aren't iOS devs. Linking to iOS docs, etc. is fine for additional context and those who know, however, it shouldn't be necessary to understand what this comment means.

The JSDoc comments should say what this option does at a glance.

Personally, I'm not an iOS dev, and I do not understand what this option does by reading the JSDoc, so I think it's fair to say it'd be confusing for others too.

> Swipe gesture will use the new `contentPopGestureRecognizer` introduced in iOS 26 if possible.

What does using the new `contentPopGestureRecognizer` entail from the user's point of view? Does using it mean gesture will now match iOS default animation? Do `animation` and `animationMatchesGesture` options have any effect when this is enabled?

packages/native-stack/src/types.tsx
kmichalikk: I 100% agree, but I simply don't know if I'm 50/50 whether I should remove it because it's not needed, or leave it because I didn't change the behavior since it was added here, and it's still technically true?

packages/native-stack/src/types.tsx
kmichalikk: [2e789e2](https://github.com/react-navigation/react-navigation/pull/12811/commits/2e789e2c918b7e04839a939846bf72895c72ac7b)

Okay, I'll just remove it, since it's adding some background, but users may not know what we're talking about anyway.

packages/native-stack/src/types.tsx
satya164: I think we should just update this since this JSDoc is being updated - and you have more context than me to know how exactly this behaves (e.g. in nested stack case). Otherwise, it'll just stay stale.

kmichalikk:
I cleared up the docs a bit. As for shadowEnabled, I'd like to keep the details for iOS 18 as it's a prop that needs some background, and it works for a specific case, and I didn't change that in my PR for screens. For iOS 26 it's not needed anymore*, so it's been deprecated.

*again, more explaining here, if specifying animation, animationMatchesGesture, we still have some non-native implementation that could use the prop, but we choose not to engage it so the 1% case doesn't differ from 99% cases. This is probably nothing regular user cares about, because everything works as one would expect.

If something needs more clarification, let me know

satya164:
@kmichalikk thanks, it's clearer now. the modal part still needs clarification but let's clarify on discord.

satya164 (approved):
Thanks

PR #12813, opened 2025-10-27
feat: Add backgroundComponent property

t0maboro:

**Motivation**

Add a new API to set the custom background when the screen or modal doesn't cover the full display

Corresponding PR in `react-native-screens`: https://github.com/software-mansion/react-native-screens/pull/3348

**Test plan**

Added an example in `react-native-screens`

### DIFF
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -698,9 +698,9 @@ export type NativeStackNavigationOptions = {
   unstable_sheetFooter?: () => React.ReactNode;
   /**
    * Optional custom React element to be rendered as the background of the screen.
-   * 
-   * This component will be rendered behind all foreground screen content. 
-   * You can provide any valid React element, such as an ImageBackground, 
+   *
+   * This component will be rendered behind all foreground screen content.
+   * You can provide any valid React element, such as an ImageBackground,
    * gradient, or a styled View.
    */
   backgroundComponent?: () => React.ReactNode;


satya164:
Can't this already be achieved with the [`layout` prop on the screen](https://reactnavigation.org/docs/screen/#layout) in React Navigation or just by rendering the component as part of the screen?

t0maboro:
> Can't this already be achieved with the [`layout` prop on the screen](https://reactnavigation.org/docs/screen/#layout) in React Navigation or just by rendering the component as part of the screen?

@satya164, please correct me if I'm missing sth important, but `layout` renders the defined component below the `ScreenContentWrapper`, which means it doesn't respect the `contentStyle`. By introducing the `backgroundComponent` prop, our goal is to render the component above the `ScreenContentWrapper`, allowing us to operate independently of the applied `contentStyle` and maintain proper visual layering, especially for integrating it with the `SafeAreaView` component https://github.com/software-mansion/react-native-screens/blob/main/src/components/ScreenStackItem.tsx#L103-L109

Screenshot 2025-10-27 at 20 13 38

https://snack.expo.dev/@tomasz.boron/witty-violet-salsa

I believe that we can achieve the same effect using `layout`, but in some cases (like in the snack above) it may require passing, for example

```js
<>
  <View style={StyleSheet.absoluteFill}>
    <BackgroundComponent />
  </View>
  {children}
</>
```

Having a prop, for which we could just pass `BackgroundComponent`, seems to be more elegant approach

satya164:
> especially for integrating it with the SafeAreaView component

Hmm, why does safe area view need to be handled inside of `StackScreenItem`? ~~It looks like a bad practice to automatically handle safe area this way and provide less control. We should export a component that users can wrap their content with. Users should specify what content should avoid safe area and what shouldn't.~~

~~Using `SafeAreaView` also doesn't match [our current recommendation](https://reactnavigation.org/docs/handling-safe-area/) (we even have a warning), so we have to revisit the approach to make sure we're aligned there and things work properly.~~

Looks like `SafeAreaView` is used to workaround some UIKit bug. In that case I think it's expected that the background should be inside the safe area.

Other than the safe area, it seems like the background is rendered at the same level as the `children`, so it should have a similar behavior.

> seems to be more elegant approach

A new prop may seem more elegant, but it depends - we should only add new APIs for something that's already possible in a slightly more verbose way, only if it's a very common use case, which makes it annoying to write the more verbose API.

So what's the use case for API? Is it common enough that we need an option for this? Without knowing the use case, it's hard for me to say.

t0maboro:
> Looks like SafeAreaView is used to workaround some UIKit bug. In that case I think it's expected that the background should be inside the safe area.

I can imagine some situations when we'd like to have the background extended over system bars.

Screenshot 2025-10-27 at 21 41 17

> So what's the use case for API? Is it common enough that we need an option for this? Without knowing the use case, it's hard for me to say.

The more convenient usage for some (imo) common patterns, like placing at the background:
- linear gradient
- image
- video
- some animated effects, like particles, with the intention to spread them over the whole screen

======

Code for reference where simply passing only `LinearGradient` seems to be more convenient

```js
import { Text, View, StyleSheet } from 'react-native';
import LinearGradient from 'react-native-linear-gradient';
import { SafeAreaView } from 'react-native-screens/experimental'

const CustomBackgroundComponent = ({ children }) => (
  <>
    <LinearGradient
      colors={['#4c669f', '#3b5998', '#c30e1aff']}
      start={{ x: 0, y: 0 }}
      end={{ x: 1, y: 0 }}
      style={StyleSheet.absoluteFill}
      testID="customLayoutComponent"
    />
    {/* simulating SAV from ScreenStackItem */}
    <SafeAreaView edges={{top: true, bottom: true}}>
      {children}
    </SafeAreaView>
  </>
);

function HomeScreen() {
  return (
    <View style={styles.screenContent}>
      <Text style={styles.text}>Home Screen</Text>
    </View>
  );
}

const App = () => {
  return (
    <NavigationContainer>
      <Stack.Navigator
      >
        <Stack.Screen
          name="Home"
          component={HomeScreen}
          options={{
            headerShown: false,
          }}
          layout={CustomBackgroundComponent}
        >
        </Stack.Screen>
      </Stack.Navigator>
    </NavigationContainer>
  );
};

  screenContent: {
    flex: 1,
    justifyContent: 'center',
    alignItems: 'center',
    backgroundColor: '#00000060'
  },
  text: {
    fontSize: 24,
    color: '#fff',
  },
});

export default App;
```

satya164:
> I can imagine some situations when we'd like to have the background extended over system bars.

Yes, however, React Native Screens should never add top inset automatically. In this case, it only seems to be there to fix a bug on iOS, and only if the header is not translucent or hidden. So it shouldn't matter for this case, [as the safe area won't have a top inset if header is hidden](https://github.com/software-mansion/react-native-screens/blob/6f3697f51d83b8a29f2033f83b078327116894e6/src/components/ScreenStackItem.tsx#L240).

This is already one of the use cases the `layout` prop was made for, so users can render something like this once in `screenLayout` for all screens in one place instead of having to duplicate it in each screen. It's slightly more verbose because it's made for a broader use case, e.g., you can apply the padding in your example in the `layout` itself instead of having a separate background and `contentStyle` options, or have more wrappers.

If only using `background`, yeah, it's slightly nicer compared to `layout`, but it's not different enough and not a common enough use case for the verbosity of `layout` prop to be a nuisance.

So for now, I won't accept this PR.

t0maboro:
I completely understand your perspective and I'm okay with that approach. It makes sense that `layout` provides a more flexible and general solution for `react-navigation`. I'll go ahead and close this PR for now. 

For the PR on the `react-native-screens` side, I'll leave that open for internal discussion on our end, as at the moment we're rethinking the proper structure between components in `native-stack` for aligning the approach with `bottom-tabs`.

PR #12826, opened 2025-11-03
feat: accept name argument for useNavigation

satya164:
**Motivation**

This updates `useNavigation` to return more useful types.

- The `useNavigation` hook now accepts an argument - the name of the route
- The argument to `useNavigation` provides auto-completion based on all routes in the project (using `RootParamList`)
- The returned type of `useNavigation` is the navigation object for the provided route
- When no argument is passed, `useNavigation` returns the navigation object for screens in the root navigator

BREAKING CHANGE: `useNavigation` no longer accepts a generic to override the returned type since it's not type-safe. User can do `const navigation = useNavigation() as SomeType` to make it explicit.

https://github.com/user-attachments/assets/bc3397bd-67b3-4341-b514-58fdaf33772a

**Test plan**

- [x] Add unit tests
- [x] Add type tests

### DIFF
--- a/example/__typechecks__/common.check.tsx
+++ b/example/__typechecks__/common.check.tsx
@@ -7,31 +7,40 @@ import type {
 } from '@react-navigation/bottom-tabs';
 import type {
   DrawerNavigationOptions,
+  DrawerNavigationProp,
   DrawerScreenProps,
 } from '@react-navigation/drawer';
 import { Button } from '@react-navigation/elements';
 import {
+  type CompositeNavigationProp,
   type CompositeScreenProps,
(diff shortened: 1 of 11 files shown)


levibuzolic:
This looks great @satya164, will give this a test on our large codebase and see how it goes. 👀

satya164:
@levibuzolic if you're planning to copy the types to your project to test, then this also requires changing how navigation types are defined, so you'd need to patch them https://github.com/react-navigation/react-navigation/commit/8b23fba60b82b25104ce2059dd97a352c1e06dd0

Issue #12401, opened 2025-01-21
[The New Architecture] Gaps between animating screens


Bardiamist:
### Current behavior

I want to migrate big project with high-load screens to the New Architecture but this issue prevents me because navigation looks bad in many places.

In the example I emulated high-load component by `useHighLoad` hook. If you will run the example then you can change `difficulty` parameters based on your environment.

Here is how it looks on the New Architecture

https://github.com/user-attachments/assets/f23a939c-6095-4113-beae-7ec04029faa5

### Expected behavior

Here is how same code works if disable the New Architecture

https://github.com/user-attachments/assets/e9e34952-e683-4b99-849c-aca9df82e5d7

### Reproduction

https://github.com/Bardiamist/diff/tree/react-navigation-new-architecture

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |7.0.14
| @react-navigation/bottom-tabs          |7.2.0
| @react-navigation/stack                |7.1.1
| react-native-screens                   |4.5.0
| react-native-safe-area-context         |5.1.0
| react-native-gesture-handler           |2.22.0
| react-native                           |0.76.6
| node                                   |23.6.0
| npm or yarn                            |npm

Bardiamist:
I provided GitHub repo under my username. Not sure why bot didn't recognize it.

I don't use @react-navigation/drawer. Accidentally checked check box after issue create. Already unchecked.

ronaldkwokhk01:
I have the same issue

petros-g:
same problem here

zispidd:
same issue

Bardiamist:
Also asked here https://github.com/reactwg/react-native-new-architecture/discussions/273

vigor-13:
Same issue

folushooladipo:
[...]

PR #12846, opened 2025-11-14
fix: heavy renders overlap with the animation on transition in stack

osdnk:
It seems that the animation has to wait until the render is finished. `useTransition` (wrapping the navigation action and animation with `startTransition` wrappers does not help). 

fixes: https://github.com/react-navigation/react-navigation/issues/12401 https://github.com/reactwg/react-native-new-architecture/discussions/273

before

https://github.com/user-attachments/assets/a0031343-fa48-4e38-90da-7615f4bc84ef

after

https://github.com/user-attachments/assets/486cc378-5f90-46e3-8ad4-965c28d7f326

### DIFF
--- a/packages/stack/src/__tests__/index.test.tsx
+++ b/packages/stack/src/__tests__/index.test.tsx
@@ -114,6 +114,8 @@ test('fires transition events on navigation', async () => {
 
   fireEvent.press(getByText('Go to B'));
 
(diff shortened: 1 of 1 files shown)


osdnk:
There's a bit more logic, i.e. setting `toValue`. Just wrapping `animate` with `setTimeout` would cause that `animate` from L428 would be invoked. Instead, I suggest adding an argument

PR #12849, opened 2025-11-17
feat(bottom-tabs): use component instead of function

tcK1:
**Motivation**

This solves a potential issue when using the `tabbar` prop on the [bottom tabs navigator dynamically](https://reactnavigation.org/docs/bottom-tab-navigator/?config=dynamic#tabbar), while also using the [React Compiler](https://react.dev/learn/react-compiler).

When creating an inline component directly on the prop, the code works, but it breaks the [react/no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/no-unstable-nested-components.md) rule. When moving the component outside, the rule passes, but the code breaks.

This PR should solve this issue; we've patched the package internally and are running it in production.

**Test plan**

First, [install the Babel plugin](https://react.dev/learn/react-compiler/installation#babel).

With the current implementation, this code will break with the compiler on:

```tsx
function CustomTabBar(props) {
  return <MyTabBar {...props} />
}

function MyTabs() {
  return (
    <Tab.Navigator
      tabBar={CustomTabBar}
    >
      {/* ... */}
    </Tab.Navigator>
  );
}
```

[Here's a reproduction repo](https://github.com/tcK1/navigation-compiler-repro).

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabView.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabView.tsx
@@ -68,13 +68,11 @@ const hasAnimation = (options: BottomTabNavigationOptions) => {
   return Boolean(transitionSpec);
 };
 
-const renderTabBarDefault = (props: BottomTabBarProps) => (
-  <BottomTabBar {...props} />
-);
+const TabBarDefault = (props: BottomTabBarProps) => <BottomTabBar {...props} />;
 
 export function BottomTabView(props: Props) {
   const {
-    tabBar = renderTabBarDefault,
+    tabBar: TabBar = TabBarDefault,
     state,
     navigation,
     descriptors,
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR.

Not sure what "this code will break with the compiler" implies in this case, but the usage is incorrect. The prop accepts a function returning a react element. It shouldn't receive a component.

React Navigation accepts functions and not components. The design is intentional. Accepting a component means it always needs to be defined outside. Accepting functions means you can define your logic either inline or outside, which is more flexible. More importantly, it avoids components from remounting if someone has defined them inline.

This change is breaking; it will break all existing code that passes an inline function, causing the whole tab bar to be remounted on every re-render of the component containing the navigator.

> but it breaks the [react/no-unstable-nested-components](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/no-unstable-nested-components.md) rule

The rule is irrelevant, as it's not a nested component, but a function. The rule doesn't (and cannot) distinguish between them.

I'm closing this since it's both a breaking change and the opposite of the original design of the API.

PR #12852, opened 2025-11-19
chore: disable header height workarounds when new `react-native-screens` implementation is used

kligarski:
**Motivation**

https://github.com/software-mansion/react-native-screens/pull/3402 in `react-native-screens` fixes values reported via `onHeaderHeightChange` event. This makes 2 workarounds in `NativeStackView.native.tsx` redundant. In order to maintain backward compatibility, `react-native-screens` exposes `usesNewAndroidHeaderHeightImplementation` compatibility flag. In this PR, we use this flag to decide whether we need to use the workarounds.

**Test plan**

Run `TestHeaderHeight` in `react-native-screens` example app. Remember to select `react-navigation` implementation.

https://github.com/user-attachments/assets/dfbe9322-5413-4041-bd32-a4107ef4ad1c

### DIFF
--- a/packages/bottom-tabs/src/unstable/NativeScreen/NativeScreen.tsx
+++ b/packages/bottom-tabs/src/unstable/NativeScreen/NativeScreen.tsx
@@ -8,17 +8,12 @@ import * as React from 'react';
 import {
   Animated,
   Platform,
-  StatusBar,
   StyleSheet,
   useAnimatedValue,
   View,
 } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
-import {
-  compatibilityFlags,
-  ScreenStack,
-  ScreenStackItem,
-} from 'react-native-screens';
+import { ScreenStack, ScreenStackItem } from 'react-native-screens';
 
 import type { NativeBottomTabHeaderProps } from '../types';
 import { debounce } from './debounce';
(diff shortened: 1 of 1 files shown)


packages/bottom-tabs/src/unstable/NativeScreen/NativeScreen.tsx
```diff
@@ -116,9 +128,10 @@ export function NativeScreen({ route, navigation, options, children }: Props) {
...
           Platform.OS === 'android' &&
-          (options.headerBackground != null || options.headerTransparent)
+          (options.headerBackground != null || options.headerTransparent) &&
+          !usesNewAndroidHeaderHeightImplementation
         ) {
-          // FIXME: On Android, we get 0 if the header is translucent
+          // On Android, we get 0 if the header is translucent
           // So we set a default height in that case
           setHeaderHeight(ANDROID_DEFAULT_HEADER_HEIGHT + topInset);
           return;
```
satya164: We can remove compat flag from thsi file since this is an unstable API and we can make breaking changes to adapt to latest react-native-screens changes.

kligarski: Fair point, thanks.

https://github.com/react-navigation/react-navigation/pull/12852/commits/255ef4e815c8703da141c49941ae4c2822b6d349

PR #12866, opened 2025-11-24
feat: add `useBottomTabAnimation`

filipengberg:
**Motivation**

This hook makes the animation progress of the Bottom Tab navigator available inside screens so that screen elements can animate in sync with screen changes. 

Its implemented and works the same way as https://github.com/react-navigation/react-navigation/pull/11568 but for the Bottom Tab navigator.

**Test plan**

1. Add the `useBottomTabAnimation` hook to a screen in `/example/src/Screens/BottomTabs.tsx`
2. Add this to the screen:

```ts
const { current } = useBottomTabAnimation();

const style = {
  transform: [
    {
      translateX: current.progress.interpolate({
        inputRange: [-1, 0, 1],
        outputRange: [50, 0, -50],
      }),
    },
  ],
};
```
3. Wrap one of the elements on the screen with `<Animated.View style={style} />`
4. Now if you use the `shift` animation on the bottom tabs the wrapped element will not move when switching screens


filipengberg:
@satya164 Thanks for reviewing! Sounds good, I updated the implementation to use `BottomTabSceneInterpolationProps`.

PR #12871, opened 2025-11-27
feat: add sheetShouldOverflowTopInset prop

t0maboro:
We unified the behavior of form sheets across Android and iOS in `react-native-screens`. On iOS, form sheets respect the top safe area inset and do not extend underneath the status bar. We’ve updated the Android implementation to match this behavior for consistency across platforms.

To maintain backward compatibility, we introduce a new prop `sheetShouldOverflowTopInset` that allows reverting to the previous behavior where the component could extend beneath the status bar.

**Motivation**

This change enhances visual consistency.

Additionally, it allows to handle the case with moving the FormSheet content above the keyboard and avoiding moving it behind the status bar or display cutout

related PRs:
- https://github.com/software-mansion/react-native-screens/pull/3404
- https://github.com/software-mansion/react-native-screens/pull/3248

**Test plan**

Tested with the example added to `react-native screens` in https://github.com/software-mansion/react-native-screens/pull/3404

### DIFF
--- a/packages/native-stack/src/views/NativeStackView.native.tsx
+++ b/packages/native-stack/src/views/NativeStackView.native.tsx
@@ -280,7 +280,7 @@ const SceneView = ({
         sheetCornerRadius={sheetCornerRadius}
         sheetElevation={sheetElevation}
         sheetExpandsWhenScrolledToEdge={sheetExpandsWhenScrolledToEdge}
-        // @ts-expect-error - will be available since react=native-screens@4.19.0
+        // @ts-expect-error - will be available since react-native-screens@4.19.0
         sheetShouldOverflowTopInset={sheetShouldOverflowTopInset}
         statusBarAnimation={statusBarAnimation}
         statusBarHidden={statusBarHidden}


satya164:
Should we support this in 8.x or should I cherry-pick to 7.x only?

Do you plan to always support this behavior, or is it temporary for backwards compatibility? Any use case for extending bottom sheet to underneath status bar?

t0maboro:
> Should we support this in 8.x or should I cherry-pick to 7.x only?

cherry-picking it only to `7.x` is fine; for stack v5 we're planning to rewrite the logic for modals, extracting their logic to a separate component, so with the fresh implementation, this prop won't be needed
 
> Do you plan to always support this behavior, or is it temporary for backwards compatibility? Any use case for extending bottom sheet to underneath status bar?

the main reason is keeping the backward compatibility

t0maboro:
Since `react-native-screens` were upgraded to `4.19` on the `7.x` branch, I'm setting this PR as ready.
Please let me know if there are any further actions on my side.

satya164:
I just picked the commit here https://github.com/react-navigation/react-navigation/commit/ea190e979a93336acf46aa796f2c88bd5bdb3fda

I'll release soon after updating a few more things.

t0maboro:
> I just picked the commit here [ea190e9](https://github.com/react-navigation/react-navigation/commit/ea190e979a93336acf46aa796f2c88bd5bdb3fda)
> 
> I'll release soon after updating a few more things.

Thank you. I can confirm it works fine on my side. I'm closing this PR.

PR #12906, opened 2025-12-26
fix: use Navigation generic type for layout prop

th-ad:
**Motivation**

 When using the `layout` callback on the drawer navigator, I noticed that the `navigation` object being passed to the function was typed using `NavigationHelpers` rather than the `Navigation` generic which is used in other callbacks like `screenLayout`.

This PR updates the type for the navigation object so that it is properly typed.

**Test plan**
Just a type change. I've added a small test to ensure that the layout callback is now using the `Navigation` generic.

### DIFF
--- a/example/__typechecks__/static.check.tsx
+++ b/example/__typechecks__/static.check.tsx
@@ -28,6 +28,12 @@ import {
 import { expectTypeOf } from 'expect-type';
 
 const NativeStack = createNativeStackNavigator({
+  layout: ({ navigation, children }) => {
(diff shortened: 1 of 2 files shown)


satya164:
Thanks for the PR. But this change is not correct. The `layout` for navigator doesn't receive the same `navigation` prop as screens, it uses a different navigation object that's used in navigators - it doesn't have the same functionalities as screen's `navigation` (e.g. you can't add a listener).

So I'm closing the PR.

PR #12926, opened 2026-01-14
feat: add support for material symbols

satya164:
The PR includes the following:

- Script to download material symbol font and generate a list of names - so we can provide TypeScript types
- `MaterialSymbol` component that renders material symbol icon - it uses the glyph map to render the appropriate character for the font (`MaterialSymbolView` in native)
- `MaterialSymbol.getImageSource` that returns an image source with URI - draws the text and creates a bitmap, caches it in fs and uses the file uri (`MaterialSymbolModule` in native)


example/src/material-symbol-names.ts
kkafar, line 1: Why are the the definitions duplicated here? They could be imported from `/native/MaterialSymbolData.tsx` if I understand correctly.

packages/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt
```diff
@@ -0,0 +1,111 @@
...
+        com.facebook.react.bridge.ReadableType.Number -> it.asDouble()
+        com.facebook.react.bridge.ReadableType.Map -> it.asMap()
+        else -> null
+      }
+    }
+
+    val resolvedColor = ColorPropConverter.getColor(colorValue, reactApplicationContext)
+      ?: throw IllegalArgumentException("Could not resolve color")
+
+    val density = reactApplicationContext.resources.displayMetrics.density
```
kkafar, line 46: Not sure whether this gives up-to-date value or not. 

I'm not sure whether this is a value encoded (hardcoded) into application resources or it changes if the user changes zoom / scaling in the phone settings. 

Usually in screens we use `PixelUtil.getDisplayMetricDensity` from `ReactAndroid`, however it had issues on its own. 

To verify it you can just try to test how the icons will behave in case you change screen density in phone options.

packages/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt
```diff
@@ -0,0 +1,111 @@
...
+          } catch (e: Exception) {
+            futures.remove(cacheKey)
+
+            throw e
+          }
+        })
+    }
+
+    try {
+      return future.get()
```
kkafar, line 106: This will block the current thread anyway in case it needs to load the resource. In case it blocks the thread, loading of another resource won't be dispatched anyway. So I'm not really sure this whole setup with executors makes much sense, but I only skim through the code & not fully grasp it.

packages/native/android/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+}
+
+
+apply plugin: "com.android.library"
+apply plugin: "kotlin-android"
+
+apply plugin: "com.facebook.react"
+
+def getExtOrIntegerDefault(name) {
+  return rootProject.ext.has(name) ? rootProject.ext.get(name) : (project.properties["ReactNavigation_" + name]).toInteger()
```
kkafar, line 25: I reckon this will crash the app in case these are not defined in the root project and any of the `ReactNavigation_XXX` variables is missing from the `react-navigation`'s `gradle.properties` file. If this is intended - then ok, but it would be nice to throw an nice error with appropriate error message.

In screens we provide a hardcoded fallback IIRC if none of these is specified.

packages/native/android/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+    sourceCompatibility JavaVersion.VERSION_1_8
+    targetCompatibility JavaVersion.VERSION_1_8
+  }
+}
+
+task copyAssets(type: Copy) {
+  from "../assets/fonts"
+  include "*.ttf"
+  into "${buildDir}/intermediates/react-navigation/fonts"
+}
```
kkafar, line 58: This is nice.

packages/native/android/gradle.properties
```diff
@@ -0,0 +1,5 @@
+ReactNavigation_kotlinVersion=2.0.21
+ReactNavigation_minSdkVersion=24
+ReactNavigation_targetSdkVersion=34
+ReactNavigation_compileSdkVersion=35
+ReactNavigation_ndkVersion=27.1.12297006
```
kkafar, line 5: Since last year August targetSdk 35 is required for any new apps & app updates that want to be admitted to Google Play Store [(link)](https://developer.android.com/google/play/requirements/target-sdk)

Usually you want to target the latest version & keep it in sync with `compileSdkVersion`.

kkafar, line 5: It's worth using NDK above r28 https://developer.android.com/guide/practices/page-sizes#compile-r28
to avoid problems with [support for 16KB sizes on Android](https://developer.android.com/guide/practices/page-sizes#build)

example/src/material-symbol-names.ts
satya164: @kkafar the one in `MaterialSymbolData.tsx` is only type definitions - so we can't use it directly. Could put the actual names there, but there are a lot of names, so it'd increase bundle size unnecessarily.

packages/native/android/build.gradle
satya164: @kkafar it has been like this in create-react-native-library template for a long time and I didn't see any issues, but if it can be improved then I'll checkout the impl in screens and adapt it here as well in CRNL

packages/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt
satya164: i wanted to be future proof but may have gotten carried away
yea you're right, for sync access from JS thread only it doesn't make much sense

packages/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt
satya164: I checked [`react-native-vector-icons`](https://github.com/oblador/react-native-vector-icons/blob/963582925a509b010038b28a80da2f2304d661b6/packages/get-image/android/src/main/java/com/reactnativevectoricons/get_image/VectorIconsModule.kt#L35) initially  and later [`expo-font`](https://github.com/expo/expo/blob/main/packages/expo-font/android/src/main/java/expo/modules/font/FontUtilsModule.kt#L36), which does this (though expo-font impl may have been based on vector icons) - but I'll double-check. Thanks for the tip!

example/src/material-symbol-names.ts
kkafar: +1, you're right.

packages/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt
satya164: Verified that it changes when I change scale in settings

kkafar: That's fine then!

Issue #12922, opened 2026-01-14
Material Top Tabs Navigator is crashing the app after a validation "right operand of 'in' is not an object"


reinaldonetof:
### Current behavior

Hello everyone,

I’m using @react-navigation/material-top-tabs version ^7.4.8 to create top tabs in my app. However, when I try to navigate between tabs, the app crashes and shows the following error:

TypeError: right operand of 'in' is not an object

After investigating, I found that the error occurs in this file and validation logic:

https://github.com/react-navigation/react-navigation/blob/4aae7563d9fb7d823416255c43b586258b34fe0e/packages/core/src/SceneView.tsx#L83C12-L87C59

The issue seems to be that nextRoute.params is coming as a string, containing the name of the tab I want to navigate to, instead of an object.

I was able to work around this by changing the 'in' check to use hasOwnProperty with optional chaining.

I’d like to know if this issue has already been reported or if the problem could be related to my environment or configuration.

If relevant, I can create a PR to help prevent any future issues.

Thanks in advance!

### Expected behavior

Navigate between tabs

### Reproduction

-

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [x] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

package | version
-- | --
@react-navigation/native | ^7.1.24
@react-navigation/bottom-tabs | ^7.8.11
@react-navigation/material-top-tabs | ^7.4.8
@react-navigation/native-stack | ^7.8.5
react-native-tab-view | ^4.2.0
react-native-screens | ^4.18.0
react-native-safe-area-context | ^5.6.2
react-native-gesture-handler | ^2.29.1
react-native-reanimated | 3.19.4
react-native-pager-view | ^6.9.1
react-native | 0.81.5
node (engines) | >=20

reinaldonetof:
After spending another day investigating this issue, I realized that the root cause was actually in our codebase. The project is quite legacy and was passing a **string** as the `params` argument to `navigate`, instead of an object.

The navigation structure is roughly the following:

* Tab Navigator

  * Screen **Home** → `navigate("Profile", "ALL")`

[...]

PR #12937, opened 2026-01-21
refactor: improve parameter checks in SceneView component

reinaldonetof:
**Motivation**

Closes https://github.com/react-navigation/react-navigation/issues/12922

This PR adds a defensive runtime check in SceneView to avoid a crash when route.params is not an object.

As explained in https://github.com/react-navigation/react-navigation/issues/12922 and https://github.com/react-navigation/react-navigation/issues/12922#issuecomment-3749648700
 the crash was caused by a legacy navigation pattern where a string was passed as params to navigate. While this is invalid usage, the current logic assumes params is always an object and uses the in operator, which throws at runtime otherwise.

This change does not alter the public API or expected usage, but prevents hard crashes by making the validation logic more resilient to malformed or legacy params.

**Test plan**
- Run the existing test suite (lint, TypeScript, and tests).
- Verify that normal navigation with object-based params behaves unchanged.
- Confirm that passing invalid legacy params (e.g. a string) no longer crashes the app.

**Screenshot**

Before

https://github.com/user-attachments/assets/5769ff43-ae75-41a9-9b9e-1236933c7fe9

After

https://github.com/user-attachments/assets/ee20dc70-e963-4c23-8bd2-0752c607767e

### DIFF
--- a/packages/core/src/SceneView.tsx
+++ b/packages/core/src/SceneView.tsx
@@ -80,6 +80,7 @@ export function SceneView<
         // This will avoid the navigator trying to handle them again
         if (
           nextRoute.params &&
+          typeof nextRoute.params === 'object' &&
           (('state' in nextRoute.params &&
             typeof nextRoute.params.state === 'object' &&
             nextRoute.params.state !== null) ||


satya164:
What makes params not be an object?

If params is somehow becoming not an object, then there seems to be either a bug in your code or a different bug in React Navigation. Params should always be either `undefined` or an object, and it's how everything is written.

It's not a valid/supported pattern to have non-object params, so defensive coding to handle such a case doesn't make much sense. This assumption will be everywhere, and in any future code as well. It won't make much sense to account for an unsupported pattern in any future code we write.

The proper fix is to find the bug that's causing this and fix it.

satya164:
> the crash was caused by a legacy navigation pattern where a string was passed as params to navigate

Also, I'm not sure it was ever supported. What makes this a legacy navigation pattern?

PR #12942, opened 2026-01-23
feat: add sheetResizeAnimationEnabled prop

t0maboro:
**Motivation**

This PR is a follow-up to https://github.com/software-mansion/react-native-screens/pull/3484, which introduces a prop that allows overriding the default enter/exit animations, enabling users who use the Animated/Reanimated API to disable the default sheet resize animation for `fitToContents` dynamic height update. When the `sheetContentDefaultResizeAnimationEnabled` prop is set, we skip performing our resize + translateY animation and only adjust the screen size to match the current content.

**Test plan**

Tested with react-native-screens example: https://github.com/software-mansion/react-native-screens/blob/3794f4de50cec57220479e73063478b305467fa1/apps/src/tests/Test2560.tsx


t0maboro:
> The code looks good. We should also make a PR to update the documentation.

https://github.com/react-navigation/react-navigation.github.io/pull/1464

PR #12945, opened 2026-01-28
docs: update sheetLargestUndimmedDetentIndex doc for keyboard handling

t0maboro:
**Motivation**

For keyboard handling the DimmingView appears always, even if we set `sheetLargestUndimmedDetentIndex` to `last`, because the keyboard changes sheet's size and it exceeds the `largestUndimmedDetent` threshold

For more context, please look at: https://github.com/software-mansion/react-native-screens/issues/3577#issuecomment-3811404550

**Test plan**

N/A just, doc update


t0maboro:
> ☝🏻 That was not supposed to be approve yet

https://github.com/react-navigation/react-navigation/pull/12945/commits/909fcb36f8801c29d1c86743e59de87b1d272143

PR #12948, opened 2026-01-30
fix(static): add navigation screen prop for parity

sterlingwes:
**Motivation**

We noticed while migrating to the static API that our screens were missing the `navigation` prop. I couldn't find documentation or an issue noting that this was intentional (seems weird to not have parity on this with the dynamic API, would love to understand the rationale if intentional).

**Test plan**

I applied the following patch to our app and verified that `console.log(Object.keys(props.navigation))` in our static navigator screen component logged the keys I'd expect to see on the navigation prop.

(we're currently running `@react-navigation/core: 7.13.6`)

```diff
diff --git a/node_modules/@react-navigation/core/lib/module/StaticNavigation.js b/node_modules/@react-navigation/core/lib/module/StaticNavigation.js
index e20d571..a6f63ad 100644
--- a/node_modules/@react-navigation/core/lib/module/StaticNavigation.js
+++ b/node_modules/@react-navigation/core/lib/module/StaticNavigation.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { isValidElementType } from 'react-is';
 import { useRoute } from "./useRoute.js";
+import { useNavigation } from './useNavigation.js';
 
 /**
  * Flatten a type to remove all type alias names, unions etc.
@@ -32,8 +33,10 @@ const MemoizedScreen = /*#__PURE__*/React.memo(({
   component
 }) => {
   const route = useRoute();
+  const navigation = useNavigation();
   const children = /*#__PURE__*/React.createElement(component, {
-    route
+    route,
+    navigation,
   });
   return children;
 });

```

### DIFF
--- a/example/__typechecks__/static.check.tsx
+++ b/example/__typechecks__/static.check.tsx
@@ -647,6 +647,45 @@ createStackNavigator({
 });
 
 /**
+ * Verify screen component receives both route and navigation props.
+ * This tests that ParamsForScreenComponent correctly infers params
+ * from components using StaticScreenProps (which includes navigation).
+ */
+createStackNavigator({
+  screens: {
+    UserProfile: createStackScreen({
+      screen: ({
+        route,
+        navigation,
+      }: StaticScreenProps<{ userId: string; isAdmin?: boolean }>) => {
+        // Verify route.params has correct type
+        expectTypeOf(route.params).toEqualTypeOf<{
+          userId: string;
+          isAdmin?: boolean;
+        }>();
+
+        // Verify navigation is available with expected methods
+        expectTypeOf(navigation.goBack).toBeFunction();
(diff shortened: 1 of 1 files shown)


satya164:
One of the main points of static config is to infer types from screens. The types can then be used in the navigation object.

Having the navigation object in the screen's signature creates a circular dependency. React Navigation also can't infer the type since props will have to be annotated manually.

`useNavigation` will have even better type inference in 8.x https://reactnavigation.org/blog/2025/12/19/react-navigation-8.0-alpha#better-typescript-types-for-static-configuration

sterlingwes:
Appreciate the quick response @satya164 

Would you accept a PR to add that clarification to the docs?

satya164:
There is already a PR for this but I want to revise it before adding to docs.

PR #12953, opened 2026-02-05
fix(drawer-layout): respect drawerType='back' in animated zIndex

BarakXYZ:
## Summary

The animated `zIndex` in `drawerAnimatedStyle` was overriding the static `zIndex: -1` set for `drawerType='back'`, causing the drawer to appear on top of content instead of behind it.

## Problem

The style array order for the drawer is:
```tsx
style={[
  { zIndex: drawerType === 'back' ? -1 : 0 },  // Static
  drawerAnimatedStyle,  // Animated - comes after, overrides!
  drawerStyle,
]}
```

The original animated zIndex:
```tsx
zIndex: translateX.get() === 0 ? 0 : 1
```

This means for `drawerType='back'`:
- The static style sets `zIndex: -1` ✓
- But the animated style overrides it to `0` or `1` ✗
- When both drawer and content have the same zIndex, DOM render order wins
- The drawer renders after content in the tree, so it appears on top

## Solution

Make the animated zIndex respect `drawerType='back'`:
```tsx
zIndex:
  drawerType === 'back'
    ? translateX.get() === 0 ? -1 : 0
    : translateX.get() === 0 ? 0 : 1
```

This ensures:
- **Open** (translateX=0): drawer zIndex=-1, content zIndex=0 → drawer behind ✓
- **Closed/animating**: drawer zIndex=0, content zIndex=2 → drawer behind ✓

Other drawer types (`front`, `slide`, `permanent`) are unchanged.

## Test Plan

1. Create a drawer with `drawerType="back"`
2. Open and close the drawer via swipe gesture
3. Verify the drawer always stays behind the content
4. Verify other drawer types still work correctly


satya164:
Please provide video comparing before and after since the change is UI related.

> When both drawer and content have the same zIndex

When can this happen?

In the code, the `contentAnimatedStyle` sets `zIndex: 2` for the content, so it should always be above the drawer even though the original `-1` got overwritten to `0` or `1`. The static style is redundant, but there doesn't seem like a bug, and I don't see any visual issue with `drawerType: 'back'`.

BarakXYZ:
Hi @satya164 
You can quickly see the bug if you'll use the 'back' and then scroll a bit the drawer, then try to close the drawer. When you actually interact with the drawer after it opens, it will override the z order and you can visually see the bug. If you can't see it - I'll provide a side by side video tomorrow.

Thanks!!

BarakXYZ:
https://github.com/user-attachments/assets/21fbad2d-5661-43d7-bc1d-2a0ebed58e9c


https://github.com/user-attachments/assets/f321c795-22b1-4635-a71c-0bfde2872d29

@satya164 Here are the videos back to back with and without the fix. Also created a simple repro [here](https://github.com/BarakXYZ/drawer-back-zindex-repro) with all the details and how to trigger the bug (it's a bit more specific than I thought, in my app it will also mess the actual drawer animation. But the styles and decorations like shadows are easier to trigger and see the bug).

Let me know if I can provide anything else.

Thanks!

satya164:
Thanks for the video and repro @BarakXYZ

satya164 (approved):
Thanks for the PR

PR #12968, opened 2026-02-13
fix(bottom-tabs): include tab item appearance in scrollEdgeAppearance for native tabs

Kasendwa:
## Summary

When using `@react-navigation/bottom-tabs` with `implementation: 'native'` on iOS, custom `tabBarLabelStyle` properties (`fontFamily`, `fontSize`, `fontWeight`, `fontStyle`) are not applied on the initial render. The tab bar labels appear with the system default font and only update to the custom font after switching tabs.

## Root Cause

In `BottomTabViewNativeImpl.tsx`, the `scrollEdgeAppearance` prop on each `Tabs.Screen` was hardcoded to an empty object:

```tsx
// FIXME: if this is not provided, ScrollView on lazy tabs glitches on iOS 18
// For now we provide an empty object before adding proper support
scrollEdgeAppearance={{}}
```

On iOS, `UITabBar` uses `scrollEdgeAppearance` when the scroll view content is at the edge of the tab bar — which is the **default initial state**. The `standardAppearance` (which correctly includes font properties via `tabItemAppearance`) only takes effect after a scroll or tab switch forces a layout update.

Since `scrollEdgeAppearance` was `{}`, iOS fell back to the system default font on initial render.

## Fix

Populate `scrollEdgeAppearance` with the same `tabItemAppearance` used in `standardAppearance`:

```diff
-              // FIXME: if this is not provided, ScrollView on lazy tabs glitches on iOS 18
-              // For now we provide an empty object before adding proper support
-              scrollEdgeAppearance={{}}
+              scrollEdgeAppearance={{
+                stacked: {
+                  normal: tabItemAppearance,
+                },
+                inline: {
+                  normal: tabItemAppearance,
+                },
+                compactInline: {
+                  normal: tabItemAppearance,
+                },
+              }}
```

## File Changed

- `packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx`

## Impact

- **iOS only** — `scrollEdgeAppearance` is an iOS-specific `UITabBarAppearance` property
- **No breaking changes** — populates an already-empty appearance object with the same values used in `standardAppearance`
- **No behavioral change on Android** — Android does not use `scrollEdgeAppearance`

## Test Plan

1. Set a custom `tabBarLabelStyle` with a non-system `fontFamily`:
   ```tsx
   screenOptions={{
     tabBarLabelStyle: {
       fontFamily: 'CustomFont',
       fontSize: 14,
       fontWeight: '400',
     },
   }}
   ```
2. Launch the app on iOS with `implementation: 'native'`
3. **Before fix:** Initial tab shows system font labels; custom font only appears after switching tabs
4. **After fix:** Custom font appears immediately on first render
5. Verify no regression with ScrollView behavior on lazy tabs

## Checklist

- [x] Follows conventional commits
- [x] TypeScript passes (`yarn typecheck`)
- [x] ESLint passes (`yarn lint`)
- [x] Minimal, focused change

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -23,10 +23,12 @@ jobs:
       MAESTRO_DRIVER_STARTUP_TIMEOUT: 600000
       JAVA_DISTRIBUTION: 'zulu'
       JAVA_VERSION: '17'
+      ANDROID_HOME: /usr/local/lib/android/sdk
+      ANDROID_USER_HOME: /home/runner/.android
       API_LEVEL: 34
-      EMULATOR_PROFILE: pixel_5
-      EMULATOR_RAM_SIZE: '4096M'
       AVD_NAME: e2e_emulator
+      EMULATOR_PROFILE: pixel_5
+      EMULATOR_OPTIONS: '-no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none -camera-front none -partition-size 4096 -cores 2 -memory 4096'
       ORG_GRADLE_PROJECT_reactNativeArchitectures: x86_64
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


robrechtme:
Thank you so much for this fix! I've been having this issue for a while in my app but I couldn't reproduce it in a minimal reproducible example. Can confirm it's fixed in `@react-navigation/bottom-tabs@7.14.0` 🙏

Kasendwa:
> Thank you so much for this fix! I've been having this issue for a while in my app but I couldn't reproduce it in a minimal reproducible example. Can confirm it's fixed in `@react-navigation/bottom-tabs@7.14.0` 🙏 

You are welcome

PR #12974, opened 2026-02-14
fix(bottom-tabs): pass platform-specific selectedIcon to Tabs.Screen on Android

Kasendwa:
## Summary

`BottomTabViewNativeImpl` passes `selectedIcon` to `Tabs.Screen` using
`selectedIcon?.ios ?? selectedIcon?.shared`, which always prefers the iOS icon format.
On Android, this results in an iOS-formatted icon object (e.g. `{ type: 'templateSource', ... }`)
being passed to `react-native-screens`, which expects `{ type: 'imageSource', ... }` or
`{ type: 'drawableResource', ... }` on Android. The `selectedIcon` is therefore silently ignored.

## Root Cause

The `getPlatformIcon` function for `type: 'image'` returns:

```tsx
{
  ios: { type: 'templateSource', templateSource: icon.source },
  android: undefined,
  shared: { type: 'imageSource', imageSource: icon.source }
}
```

On Android, `selectedIcon?.ios` is `{ type: 'templateSource', ... }` which is truthy,
so the `??` operator never reaches `selectedIcon?.shared` (the correct format for Android).

Meanwhile, the regular `icon` prop passes the full `{ ios, android, shared }` object, and
`react-native-screens` correctly resolves it per platform. The `selectedIcon` didn't get
the same treatment.

## Fix

```diff
-              selectedIcon={selectedIcon?.ios ?? selectedIcon?.shared}
+              selectedIcon={
+                (Platform.OS === 'ios'
+                  ? (selectedIcon?.ios ?? selectedIcon?.shared)
+                  : (selectedIcon?.android ??
+                    selectedIcon?.shared)) as typeof selectedIcon extends undefined
+                  ? undefined
+                  : any
+              }
```

- **iOS:** `selectedIcon.ios` → fallback `shared` (unchanged behavior)
- **Android:** `selectedIcon.android` → fallback `shared` (now gets correct `imageSource` format)

The type assertion is temporary — the current `react-native-screens` types only declare
`selectedIcon` as `PlatformIconIOS`. The companion PR ([react-native-screens#3649](https://github.com/software-mansion/react-native-screens/pull/3649)) will update this type, at which point the assertion can be removed.

## File Changed

| File | Change |
|---|---|
| `packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx` | Platform-aware `selectedIcon` resolution |

## Companion PR: `react-native-screens`

This works in conjunction with [software-mansion/react-native-screens#3649](https://github.com/software-mansion/react-native-screens/pull/3649) which implements `selectedIcon` support on Android.

| Without the RNS PR | Without this PR |
|---|---|
| Correct icon format arrives, but `setSelectedIconImageSource` is a no-op | Native infra handles it, but wrong icon format arrives |

## Testing

1. Create a bottom tab navigator with `tabBarIcon` returning different
   `{ type: 'image', source: ... }` for `focused: true` vs `focused: false`
2. On Android — verify `selectedIcon` resolves to `{ type: 'imageSource', ... }`
3. On iOS — verify no regression, `selectedIcon` still resolves to iOS format

## Checklist

- [x] Follows conventional commits
- [x] TypeScript passes (`yarn typecheck`)
- [x] ESLint passes (`yarn lint`)
- [x] All 574 tests pass (`yarn test`)
- [x] Minimal, focused change
- [x] No behavioral change on iOS

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -23,10 +23,12 @@ jobs:
       MAESTRO_DRIVER_STARTUP_TIMEOUT: 600000
       JAVA_DISTRIBUTION: 'zulu'
       JAVA_VERSION: '17'
+      ANDROID_HOME: /usr/local/lib/android/sdk
+      ANDROID_USER_HOME: /home/runner/.android
       API_LEVEL: 34
-      EMULATOR_PROFILE: pixel_5
-      EMULATOR_RAM_SIZE: '4096M'
       AVD_NAME: e2e_emulator
+      EMULATOR_PROFILE: pixel_5
+      EMULATOR_OPTIONS: '-no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none -camera-front none -partition-size 4096 -cores 2 -memory 4096'
       ORG_GRADLE_PROJECT_reactNativeArchitectures: x86_64
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


Kasendwa:
Hi @satya164, this is exactly what I expected. 

I just wanted to leave this done and if we get the upstream fix, then i'll come back and update it to reflect the changes. 

Thank you!

Kasendwa:
Hi @satya164, let me close this because the companion PR at Screens has been closed and excerpts of it taken into another PR that will provide a lot more value. 

See https://github.com/software-mansion/react-native-screens/pull/3649#issuecomment-3915106062 

It's possible by the time the fix ships, this will be obselete.

PR #12987, opened 2026-02-25
feat(native-stack): Add support for images in nested header menus

Ubax:
**Motivation**

Support for images in nested menus was added to screens in https://github.com/software-mansion/react-native-screens/pull/3645

**This PR upgrades dev version of screens to `4.24.0`**

**Test plan**

Simulator Screenshot - iPhone 17 Pro - 2026-02-25 at 12 58 51

### DIFF
--- a/packages/native-stack/src/views/useHeaderConfigProps.tsx
+++ b/packages/native-stack/src/views/useHeaderConfigProps.tsx
@@ -76,9 +76,7 @@ const processBarButtonItems = (
 
         const { badge, label, labelStyle, icon, ...rest } = item;
 
-        let processedItem:
-          | HeaderBarButtonItemWithAction
-          | HeaderBarButtonItemWithMenu = {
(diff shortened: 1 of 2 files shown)


packages/native-stack/src/views/useHeaderConfigProps.tsx
```diff
@@ -74,27 +76,18 @@ const processBarButtonItems = (
...
-                    imageSource: icon.source,
-                  }
-                : {
-                    type: 'templateSource',
-                    templateSource: icon.source,
-                  }
-              : icon,
-        };
+          icon: transformIcon(icon),
+        } as HeaderBarButtonItemWithAction | HeaderBarButtonItemWithMenu;
```
Ubax: This is needed since `menu.items[].icon` is not compatible here. It is adjusted later in the `if` block

PR #12990, opened 2026-02-25
fix(types): add exactOptionalPropertyTypes support to routers package

YevheniiKotyrlo:
## Motivation

Projects that enable TypeScript's [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) hit type errors in `@react-navigation/routers` when compiling with `skipLibCheck: false`. This flag is increasingly common in strict TypeScript setups and is part of TypeScript's "beyond strict" configuration.

This is PR 1 of 4 in a stacked series to enable `exactOptionalPropertyTypes` across the react-navigation monorepo. This PR fixes all type definitions in the `routers` package by adding explicit `| undefined` to optional properties — the standard pattern for `exactOptionalPropertyTypes` compatibility.

### Stacked PR series

1. **#12990 — routers** (this PR)
2. #12992 — core + native + elements
3. #12994 — navigator packages
4. #12995 — example app + external `@ts-expect-error` + enable flag

### What changed

With `exactOptionalPropertyTypes`, `prop?: T` means "if present, value is exactly `T`" — passing `undefined` explicitly is an error. The fix is `prop?: T | undefined`, which means "optional, and if present, value is `T` or `undefined`".

Applied to all optional properties in:
- **`types.tsx`** — `NavigationAction` (`payload`, `source`, `target`), `Route` (`path`, `history`), `NavigationRoute` (`state`), `NavigationState` (`history`), `PartialRoute` (`key`, `state`), `PartialState` (`stale`), `RouterConfigOptions` (`routeGetIdList` params), `DefaultRouterOptions` (`initialRouteName`)
- **`CommonActions.tsx`** — All action types (`NavigateAction`, `ResetAction`, `SetParamsAction`, `ReplaceParamsAction`, `PushParamsAction`, `PreloadAction`)
- **`StackRouter.tsx`** — `StackActionType` (all 5 variants)
- **`TabRouter.tsx`** — `TabActionType`, `TabRouterOptions`, `TabNavigationState`
- **`DrawerRouter.tsx`** — `DrawerActionType`, `DrawerRouterOptions`
- **`createParamsFromAction.tsx`** / **`createRouteFromAction.tsx`** — local `Options` types

All changes are type-level only — zero runtime impact.

## Test Plan

- `yarn typecheck` — passes with zero errors
- `yarn lint` — passes with zero errors
- `yarn test --maxWorkers=2` — 47 suites, 613 tests, 134 snapshots all passing
[...]

PR #12995, opened 2026-02-26
fix(types): enable exactOptionalPropertyTypes across the monorepo

YevheniiKotyrlo:
## Motivation

PR 4 of 4 in the stacked PR series to enable `exactOptionalPropertyTypes` in the react-navigation monorepo (see #12990 for context and discussion with @satya164).

PRs 1-3 added `| undefined` to all optional property type definitions in react-navigation's own packages (64 files, ~611 properties). This final PR handles the remaining 30 errors — all outside our type definitions — and enables the flag in `tsconfig.json`.

### Error reduction

| Milestone | Errors remaining |
|---|---|
| Baseline (main) | 311 |
| After PR 1 (routers) | ~179 |
| After PR 2 (core + native + elements) | ~78 |
| After PR 3 (navigator packages) | 30 |
| **After this PR (enable flag)** | **0** |

### What changed

**1. External library `@ts-expect-error` (21 call sites)**

These errors occur where react-navigation passes `T | undefined` values to external library components whose types declare `prop?: T` (without `| undefined`). The external libraries need their own `exactOptionalPropertyTypes` fixes — `@ts-expect-error` is the correct bridge until then.

| Library | Files | Errors |
|---|---|---|
| react-native-screens | `NativeStackView.native.tsx`, `useHeaderConfigProps.tsx` | 10 |
| react-native-bottom-tabs | `BottomTabViewNativeImpl.tsx` | 7 |
| NativeComponent codegen | `SFSymbol.ios.tsx`, `MaterialSymbol.android.tsx` | 2 |
| react-native-reanimated | `Drawer.native.tsx` | 1 |
| react-native-gesture-handler | `Card.tsx` | 1 |

**2. Example app type fixes (8 errors)**

- `ComponentsHeaders.tsx` (6) — function parameter contravariance: `{ tintColor?: ColorValue }` → `{ tintColor?: ColorValue | undefined }`
- `ComponentsSFSymbols.tsx` (1) — inline prop type missing `| undefined`
- `playwright.config.ts` (1) — conditional spread for `workers` (Playwright types don't accept `undefined`)

**3. Type assertion test (1 error)**

- `static.check.tsx` — known `expect-type` library bug ([expect-type#128](https://github.com/mmkal/expect-type/issues/128)) where `toExtend` is incompatible with `exactOptionalPropertyTypes`

**4. Enable the flag**

Added `"exactOptionalPropertyTypes": true` to `tsconfig.json`.

## Test plan

- [x] `yarn typecheck` — passes (0 errors)
- [x] `yarn lint` — passes (0 errors)
- [x] `yarn test --maxWorkers=2` — 47 suites, 613 tests, 134 snapshots pass
- [x] `yarn lerna run prepack` — all 12 packages build successfully
- [x] `yarn check-dependency-version-consistency` — passes
- [x] `yarn example expo export --platform all` — Metro bundles web/iOS/Android successfully
- [x] `yarn example web build` — Vite builds successfully
- [x] Post-build `yarn typecheck` — passes
- [x] All lefthook pre-commit hooks pass (types, lint, test, commitlint)

### DIFF
--- a/example/__typechecks__/static.check.tsx
+++ b/example/__typechecks__/static.check.tsx
@@ -1698,6 +1698,7 @@ createStackScreen({
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const ProfileComponent = createComponentForStaticNavigation(Stack, 'Profile');
 
+  // @ts-expect-error — expect-type#128: toExtend incompatible with exactOptionalPropertyTypes
   expectTypeOf<React.ComponentProps<typeof ProfileComponent>>().toExtend<{
     initialRouteName?: 'Profile' | 'Settings';
   }>();
--- a/example/e2e/playwright.config.ts
+++ b/example/e2e/playwright.config.ts
@@ -8,7 +8,7 @@ export default defineConfig({
   fullyParallel: true,
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
-  workers: process.env.CI ? 1 : undefined,
+  ...(process.env.CI ? { workers: 1 } : {}),
   projects: [
     {
       name: 'Chromium',
--- a/example/src/Screens/ComponentsHeaders.tsx
+++ b/example/src/Screens/ComponentsHeaders.tsx
@@ -16,7 +16,11 @@ export function ComponentsHeaders(_: StaticScreenProps<{}>) {
   const { colors } = useTheme();
   const insets = useSafeAreaInsets();
 
-  const headerRight = ({ tintColor }: { tintColor?: ColorValue }) => (
+  const headerRight = ({
+    tintColor,
+  }: {
+    tintColor?: ColorValue | undefined;
+  }) => (
     <>
       <HeaderButton>
         <MaterialCommunityIcons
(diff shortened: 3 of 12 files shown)


satya164:
Thank you for the PR. For now, I can't merge this because it ignores type errors in the libraries we use. While it'd let us enable `exactOptionalPropertyTypes` in the repo, it'll also prevent any new errors from being caught due to incorrect types.

So I'll wait until the libraries are compatible so we don't need to ignore the errors and then enable it in our repo.

YevheniiKotyrlo:
Completely agree — the `@ts-expect-error` suppressions would mask real issues going forward, so waiting for ecosystem compatibility is the right call.

Wanted to say a big thank you for reviewing and merging PRs 1-3. Those alone resolved hundreds of downstream type errors — any project using `@react-navigation/*` with `exactOptionalPropertyTypes` enabled went from 512 errors to zero. That's a huge improvement for the ecosystem, even without this PR.

I'm planning to contribute `exactOptionalPropertyTypes` fixes to the blocking libraries (react-native-screens, react-native-bottom-tabs, react-native-gesture-handler, react-native-reanimated) so the suppressions in this PR can eventually be removed. Happy to close this PR if you prefer and open a fresh one once the external libraries catch up — whatever works best for you.

PR #12996, opened 2026-02-27
test: test suspense with params

osdnk:

**Motivation**

**Test plan**


packages/core/src/__tests__/index.test.tsx
```diff
@@ -3481,3 +3481,222 @@ test('does not throw if while getting current options with empty container', ()
...
+    return undefined;
+  });
+
+  return {
+    read() {
+      if (status === 'pending') throw promise;
+    },
+    resolve: resolveHandle,
+  };
+}
```
satya164: remove this. exclusively use `use(promise)`

packages/core/src/__tests__/index.test.tsx
```diff
@@ -3481,3 +3481,222 @@ test('does not throw if while getting current options with empty container', ()
...
+      <TestNavigator initialRouteName="A">
+        <Screen name="A" component={ScreenA} />
+        <Screen name="B" component={ScreenB} />
+      </TestNavigator>
+    </BaseNavigationContainer>
+  );
+
+  expect(root).toMatchInlineSnapshot(`"[ScreenA]"`);
+
+  // Navigate without startTransition — urgent update
```
satya164: We don't actually have a way to do urgent update in React Native. On React DOM it'd be `flushSync`. So the comment is a bit misleading.

satya164 (changes requested):
Some comments about tests:

- The tests that test "does not show Suspense fallback" and "shows Suspense fallback when navigating" have very different code. we're testing what happens with `startTransition` vs without. That should be the only logic that's different. Better yet, use a single test, navigate without transition first, check output, then navigate without fallback and check output. It will ensure everything else is the exact same.
- Avoid using `any`. In some places it maybe necessary to avoid too much work, but many places `any` is used unnecessarily, like component's return types.
- Instead of assigning `navigate`, `setParams` etc. in a variable, `ref.navigate`, `ref.dispatch(CommonActions.setParams(...))` etc. should work to set params for focused screen. Generally I'm not a fan of assigning in render like this. The proper way is to use a button with `onPress` and trigger it with `userEvents` but ref should be fine for our case

Issue #12999, opened 2026-02-27
Broken animation when preloading one screen and navigating to another


WojtekBoman:
### Current behavior

When a route is preloaded via `navigation.preload()` and the user navigates to a **different** route:

**Broken animation** – The transition looks wrong. The preloaded screen may appear to slide out from under the target screen, or the target screen may not animate in correctly.

**Reproduction steps:**

1. On Screen A, preload Screen C (or preload Screen B).
2. Navigate to Screen B (or Screen C respectively).
3. Observe the broken animation.

### Expected behavior

- Preloading Screen C and then navigating to Screen B should show a normal push animation for Screen B.
- The preloaded screen should not affect the transition.
- Behavior should match navigating without preloading.

### Reproduction

https://github.com/WojtekBoman/PreloadAnimationBug

https://github.com/user-attachments/assets/a810e094-8e3e-4a82-891e-da83a7d03748

### Platform

- [x] Android
- [x] iOS
- [x] Web
- [x] Windows
- [x] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                     | version  |
| --------------------------- | -------- |
| @react-navigation/native    | 7.1.31  |
| @react-navigation/stack     | 7.8.2   |
| react-native-screens        | 4.24.0  |
| react-native-safe-area-context | 5.7.0 |
| react-native-gesture-handler | 2.30.0 |
| react-native-reanimated     | 4.2.1   |
| react-native                | 0.83.2  |
| expo                        | 55.0.3  |
| node                        | 20.19.5 |
| npm                         | 10.8.2  |

PR #13000, opened 2026-02-27
fix(stack): use correct previous focused route when preloaded routes exist

WojtekBoman:
## Motivation

When a route is preloaded via `navigation.preload()` and the user navigates to a **different** route, the stack navigator incorrectly treats the preloaded route as the previously focused route. This leads to broken animations.

https://github.com/react-navigation/react-navigation/issues/12999

**Root cause:** `previousRoutes` is built as `[...routes, ...preloadedRoutes]`, so preloaded routes are appended at the end. The code used `previousRoutes[previousRoutes.length - 1]` to determine the previous focused route. When a preloaded route is last in that array, it gets incorrectly treated as the previous focused route, even though it was never actually focused.

**Example:** On Screen A, preload Screen C, then navigate to Screen B. `previousRoutes` = `[A, C]`, so the last item is C. The code assumes the previous focused route was C, but it was actually A. C then gets treated as a route being "replaced" by B: it's added to `replacingRouteKeys`, inserted into `routes`, and animated out from under B. This causes broken animation and can leave C stuck in state.

**Fix:** Track the previous navigation state in `previousState` and use it to look up the previous focused route instead of using the last item in `previousRoutes`. We use `previousState.routes[previousState.index]` — the route at the focused index in the previous state, which is always the truly focused route.

**Reproduction:** A minimal app that reproduces this bug is available at [PreloadAnimationBug](https://github.com/WojtekBoman/PreloadAnimationBug).

https://github.com/user-attachments/assets/2b5a9434-b9e0-426d-a2ae-143a84618e01

## Test plan

1. **Reproduce the bug (without the fix):**
   - Clone [PreloadAnimationBug](https://github.com/WojtekBoman/PreloadAnimationBug) and run it.
   - On Screen A, preload Screen C (or B).
   - Navigate to Screen B (or C).
   - **Result:** Broken animation.

2. **Verify the fix:**
   - Apply the patch and run the app.
   - On Screen A, preload Screen C.
   - Navigate to Screen B.
   - **Result:** B animates in correctly.

3. **Regression checks:**
   - Navigate A → B (no preload): B animates in.
   - Navigate A → C (C preloaded): C animates in.
   - Pop B → A: pop animation works.
   - Replace A with B: replace animation works.

4. **Manual testing:**
   - Use the [PreloadAnimationBug](https://github.com/WojtekBoman/PreloadAnimationBug) app:
     - Preload C, then navigate to B.
     - Confirm B animates in correctly.

### DIFF
--- a/packages/stack/src/views/Stack/StackView.tsx
+++ b/packages/stack/src/views/Stack/StackView.tsx
@@ -166,10 +166,9 @@ export class StackView extends React.Component<Props, State> {
     // Get previous focused route from previousState (actual focused route, not last in previousRoutes
     // which can be a preloaded route that was never focused)
     const previousFocusedRoute = state.previousState
-      ? (state.previousState.routes[state.previousState.index] as
-          | Route<string>
-          | undefined)
+      ? state.previousState.routes[state.previousState.index]
       : undefined;
+
     const nextFocusedRoute = routes[routes.length - 1];
 
     const isAnimationEnabled = (key: string) => {


packages/stack/src/views/Stack/StackView.tsx
```diff
@@ -317,6 +322,7 @@ export class StackView extends React.Component<Props, State> {
     return {
       routes,
       previousRoutes: [...props.state.routes, ...props.state.preloadedRoutes],
+      previousFocusedRouteKey: props.state.routes?.[props.state.index]?.key,
```
satya164: I'd avoid tracking more state properties derived from the same state if possible. Since more state means logic can become more brittle and there is already a lot.

Maybe instead of `previousRoutes`, it can just store `previousState`. Then the logic can be simpler. It can easily get previous focused route with `previousState.routes[previousState.index]` and get `previousRoutes = [...previousState.routes, ...previousState.preloadedRoutes]`

WojtekBoman: @satya164 I've adjusted it! Could you take a look if it looks better now?

satya164 (approved):
Thank you

PR #13001, opened 2026-02-27
feat: Phase 1 of data loading

osdnk:
Adds `UNSTABLE_getLoaderForRoute` to prefetch data for a route before navigating to it, using the static navigation config.

https://github.com/user-attachments/assets/85ab986d-5cd7-4141-82af-809d9cdbb8c1

- Adds an optional `UNSTABLE_loader` prop to static screen configs 
- `UNSTABLE_getLoaderForRoute` looks up the loader for a given route, traversing nested navigators and composing loaders


packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +939,132 @@ export function createPathConfigForStaticNavigation(
...
+      }
+    }
+  }
+
+  if (loaders.length === 0) {
+    return undefined;
+  }
+
+  return async () => {
+    await Promise.all(loaders.map((l) => l()));
```
satya164: The `.map` seems redundant.

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +939,132 @@ export function createPathConfigForStaticNavigation(
...
+  }
+
+  const loaders: (() => Promise<void>)[] = [];
+
+  if (
+    typeof item === 'object' &&
+    'UNSTABLE_loader' in item &&
+    typeof item.UNSTABLE_loader === 'function'
+  ) {
+    loaders.push(item.UNSTABLE_loader as () => Promise<void>);
```
satya164: Please don't cast types with `as` unless absolutely necessary

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +939,132 @@ export function createPathConfigForStaticNavigation(
 
   return screens;
 }
+
+function findScreenInConfig(
+  config: StaticNavigation<any, any, any>['config'],
```
satya164: It should just need a subset, like `createPathConfigForStaticNavigation` takes. That type could be reused here.

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +939,132 @@ export function createPathConfigForStaticNavigation(
...
+    return focusedRoute.state;
+  }
+
+  const name =
+    nestedTree.config.initialRouteName ??
+    Object.keys(nestedTree.config.screens ?? {})[0];
+
+  if (name) {
+    return { routes: [{ name }], index: 0 };
+  }
```
satya164: Accept `undefined` for the state in `UNSTABLE_getLoaderForState` to traverse children and get the loader. No need to make dummy state objects.

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +939,132 @@ export function createPathConfigForStaticNavigation(
...
+/**
+ * Get a loader function for the focused route from a static navigation config
+ * and a navigation state.
+ *
+ * The function follows the exact focused route path through the state to find
+ * the matching screen in the tree. This avoids ambiguity when multiple screens
+ * share the same name at different nesting levels.
+ *
+ * For example, if the focused route A contains a child navigator whose focused
+ * screen B also has a loader, this returns `() => Promise.all([loaderA(), loaderB()])`.
```
satya164: Claude loves putting instructions/implementation detail in JSDoc but make sure to remove those. This is a public API and the JSDoc should only say what it does, not how.

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +324,18 @@ export type StaticScreenConfig<
    * Optional key for this screen.
    */
   navigationKey?: string;
+
+  /**
+   * Loader function to prefetch data for this screen.
+   * Called before the screen is rendered, e.g. inside `startTransition`.
```
satya164: > Called before the screen is rendered, e.g. inside `startTransition`.

Not sure the comment makes sense

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +324,18 @@ export type StaticScreenConfig<
...
+   * If the screen contains a nested navigator, loaders are composed with `Promise.all`.
```
satya164: > If the screen contains a nested navigator, loaders are composed with `Promise.all`.

Implementation detail, and describes a completely different function - not this property.

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +324,18 @@ export type StaticScreenConfig<
...
+   *
+   * @example
+   * ```js
+   * UNSTABLE_loader: () => prefetchProfileData(),
+   * ```
```
satya164: Don't use prefetch terminology. It's just used for loading data.

satya164 (commented):
There is excessive use of `as` `any` and `!` in these changes. These are unsafe types and should be avoided as much as possible. Please remove them and add them when absolutely necessary (the typing is too complicated, or verbose in case of tests)

Make a separate directory `loader` or something and put the logic there. These files are getting big. There'll be more loader related logic in the future so start with them separated.

packages/core/src/StaticNavigation.tsx
satya164: As per notion doc, loaders should receive `route` object. Being able to fetchi data based on route.params is essential.

satya164 (changes requested):

satya164 (changes requested):
2 more things:

- It should handle params to navigate to nested screens, I added a `getStateForRouteParams` utility which can be used
- Need to figure out how to handle conditional rendering with `if`. If we encounter screens that have `if`, we don't know if it'll be shown or not. I see few ways:
  - Resolve all `if`. This would require making a resolved tree, where the `if` hooks are called in render. But this won't be work when we want to get loader before render, so sounds like a no-go.
  - Don't get loader for screens with `if` and their children. But then how to handle auth?

I'f suggest reading up on how to do SSR with `use` hook for data fetching and if we can leverage it somehow. One way could be to recommend using `use` in `if` hooks, for client, they read context, for server, they read resource. Need to see how this all fits in.

This function is the most critical part, so take your time and consider all edge cases, how it would be used etc.

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +324,16 @@ export type StaticScreenConfig<
...
+
+  /**
+   * Loader function to be called when the screen is focused. This can be used to load data before the screen is rendered.
+   *
+   * @example
+   * ```js
+   * UNSTABLE_loader: (route) => loadProfile(route.params?.id),
+   * ```
+   */
+  UNSTABLE_loader?: (route: Route<string>) => Promise<void>;
```
satya164: The route should use correct param type. also make the argument an object so we can pass more thing in the future (as we have in the notion doc).

```suggestion
  UNSTABLE_loader?: (options: { route: RouteType<Params> }) => Promise<void>;
```

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +937,105 @@ export function createPathConfigForStaticNavigation(
 
   return screens;
 }
+
+function findScreenInConfig(
+  config: TreeForPathConfig['config'],
+  name: string | undefined
```
satya164: Can name be `undefined`? It should always be present if we're getting loader


```suggestion
  name: string
```

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +937,105 @@ export function createPathConfigForStaticNavigation(
...
+    : undefined
+  : undefined {
+  const screens = config.screens;
+
+  if (name === undefined) {
+    if (config.initialRouteName === undefined) {
+      return Object.values(config.screens ?? {})[0];
+    }
+    return screens?.[config.initialRouteName];
+  }
```
satya164: Name shouldn't be undefined, so remove this

```suggestion
```

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +937,105 @@ export function createPathConfigForStaticNavigation(
...
+    }
+  }
+
+  if (loaders.length === 0) {
+    return undefined;
+  }
+
+  return async (route: Route<string>) => {
+    await Promise.all(loaders.map((l) => l(route)));
+  };
```
satya164: this isn't correct. each loader should get the route and params relevant to it

packages/core/src/useNavigationBuilder.tsx
```diff
@@ -279,24 +280,6 @@ const getRouteConfigsFromChildren = <
...
-          name: params.screen,
-          params: params.params,
-          path: params.path,
-        },
-      ],
-    };
-  }
-
-  return undefined;
-};
```
satya164, line 298: Leave this logic here. Use `getStateFromRouteParams`. Rebase main

packages/core/src/StaticNavigation.tsx
```diff
@@ -925,3 +937,105 @@ export function createPathConfigForStaticNavigation(
...
+    }
+  }
+
+  return undefined;
+}
+
+function getNestedTree(item: any): StaticNavigation<any, any, any> | undefined {
+  if (item && typeof item === 'object') {
+    if ('config' in item && item.config?.screens) {
+      return item as StaticNavigation<any, any, any>;
```
satya164: hmm don't the types already say a screen is component or static config? do we need `item: any` and casting here? i'd think checking if item is not a `function` should refine the type to static config.

satya164 (changes requested):
Also lets move loader related stuff to a separate file. `DataLoading.tsx` or somtheing like that to avoid making static config file and tests bigger.

packages/core/src/StaticNavigation.tsx
osdnk: I further expanded to allow PartialRoute. The reason is that the route might not be in the state, so it gets constructed with just a name from the config tree, therefore, missing a key

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +323,18 @@ export type StaticScreenConfig<
...
+  /**
+   * Loader function to be called when the screen is focused. This can be used to load data before the screen is rendered.
+   *
+   * @example
+   * ```js
+   * UNSTABLE_loader: ({ route }) => loadProfile(route.params?.id),
+   * ```
+   */
+  UNSTABLE_loader?: (options: {
+    route: Route<string> | PartialRoute<Route<string>>;
```
satya164, line 336: Let's just make it `{ name, params }`

Also, use proper type for params as used in other properties

packages/core/src/StaticNavigation.tsx
```diff
@@ -322,6 +323,18 @@ export type StaticScreenConfig<
    * Optional key for this screen.
    */
   navigationKey?: string;
+
+  /**
+   * Loader function to be called when the screen is focused. This can be used to load data before the screen is rendered.
+   *
+   * @example
+   * ```js
+   * UNSTABLE_loader: ({ route }) => loadProfile(route.params?.id),
```
satya164, line 332: Optional id for profile loading is probably a confusing example

```suggestion
   * UNSTABLE_loader: ({ route }) => loadProfile(route.params.id),
```

packages/core/src/index.tsx
```diff
@@ -2,6 +2,7 @@ export { BaseNavigationContainer } from './BaseNavigationContainer';
 export { createNavigationContainerRef } from './createNavigationContainerRef';
 export { createNavigatorFactory } from './createNavigatorFactory';
 export { CurrentRenderContext } from './CurrentRenderContext';
+export { UNSTABLE_getLoaderForState } from './DataLoading';
```
satya164, line 5: Let's not export it. We'll see if it's needed based on final API

packages/core/src/DataLoading.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    if (
+      'screen' in item &&
+      item.screen &&
+      typeof item.screen === 'object' &&
+      'config' in item.screen &&
+      item.screen.config?.screens
+    ) {
+      return item.screen;
+    }
```
satya164, line 49: checking for `config?.screens` only will miss configs that only have groups

packages/core/src/DataLoading.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const config = tree.config;
+  const focusedRoute = state?.routes[state.index ?? 0];
+
+  const name =
+    focusedRoute?.name ??
+    config.initialRouteName ??
+    Object.keys(config.screens ?? {})[0] ??
+    Object.values(config.groups ?? {}).flatMap((g) =>
+      Object.keys(g.screens)
+    )[0];
```
satya164, line 84: we should only use focused route from state. reading from screens/initialRouteName has the problem that we discussed - it could be conditional with `if`. reading from state means we use the explicitly provided route.

example/src/Screens/LoaderDemo.tsx
```diff
@@ -0,0 +1,174 @@
...
+import * as React from 'react';
+import { StyleSheet, View } from 'react-native';
+
+const cache = new Map<string, { data: string; promise?: Promise<void> }>();
+
+function fetchData(key: string, delay: number): Promise<void> {
+  const existing = cache.get(key);
+
+  if (existing) {
+    return existing.promise ?? Promise.resolve();
```
satya164, line 17: shouldn't this be

```suggestion
    return existing.promise ?? existing.data;
```

packages/core/src/DataLoading.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  const loaders: (() => Promise<void>)[] = [];
+
+  if (
+    typeof item === 'object' &&
+    'UNSTABLE_loader' in item &&
+    typeof item.UNSTABLE_loader === 'function'
+  ) {
+    const loader = item.UNSTABLE_loader;
+    const route = focusedRoute ?? { name };
```
satya164, line 104: we'll also need to get `initialParams` from the screen config and merge the route params with it

satya164 (changes requested):

PR #13018, opened 2026-03-06
fix(types): make NavigationState/NavigationLeafRoute routes optional

### LINKED ISSUE #9403: Incorrect interfaces for NavigationState and NavigationLeafRoute
**Current Behavior**

As per [documentation](https://reactnavigation.org/docs/3.x/state-persistence/#usage): `persistNavigationState` is an async function that receives single argument - the navigation state object. The function should persist it. So using typescript the function that can be passed to this prop may look like:
```js
async function persistNavigationState(navState: NavigationState) {
  callAnotherFunction(navState.routes);
}
```

I discovered the problem with `NavigationState` interface as it's `routes` property as well as nested `NavigationLeafRoute`'s `routes` property are not optional, but they are missing in the `navState` object.

For reproduction I used an example (https://github.com/react-navigation/react-navigation/tree/4.x/example).
From the screenshot, you can see that the child object (that should be of type `NavigationRoute` misses couple of required properties from it's interface.

Screenshot 2021-03-10 at 13 59 58

**Expected Behavior**

I expect that:
1. `NavigationState` and `NavigationLeafRoute` will have such properties as `routes` optional
2.  `navState` will have default values in the result object like `{ routes: [], key:... }`

**How to reproduce**

- Clone https://github.com/bahkostya/react-navigation-reproduction/ (commit for reproduction [3e3a99330482fed3f2f7b02f9a0951134028a636](https://github.com/bahkostya/react-navigation-reproduction/commit/3e3a99330482fed3f2f7b02f9a0951134028a636))
- Follow the instruction of the example app in README
- Run the app and enable remote JS debugging (Enable Pause On Caught Exceptions for a better debugging experience.)
- Press "Simple Stack" button
- In devtools you should see the same error as on the screenshot.

**Your Environment**

| Android                 |9.0 (Pie) - API 28
| react-navigation               |4.x
| react-navigation-stack         |2.10.4
| react-native                   |0.63.30
| expo                           |39.0.3
| yarn                    |1.18.0

SergioChan:
## Summary
- make `NavigationState.routes` optional in the v4 TypeScript definitions
- make `NavigationLeafRoute.routes` optional to match runtime persisted state payloads
- keep scope minimal to the issue-reported typing mismatch

## Testing
- Not run locally: this repository's full TypeScript/lint/test workflow requires installing the monorepo dependencies, which is heavy for this constrained cron environment.
- Validation performed: inspected and diff-verified that only the two type fields reported in issue #9403 were changed.

## Related
Fixes #9403

### DIFF
--- a/packages/react-navigation/typescript/react-navigation.d.ts
+++ b/packages/react-navigation/typescript/react-navigation.d.ts
@@ -75,7 +75,7 @@ export interface NavigationState {
    * Index refers to the active child route in the routes array.
    */
   index: number;
-  routes: NavigationRoute[];
+  routes?: NavigationRoute[];
   isTransitioning: boolean;
   key: string;
   params?: NavigationParams | undefined;
@@ -117,7 +117,7 @@ export interface NavigationLeafRoute<Params = NavigationParams> {
   /**
    * Array containing the navigator's routes
    */
-  routes: NavigationRoute[];
+  routes?: NavigationRoute[];
   /**
    * Flag that indicates the transition state of the route
    */


satya164:
Thanks for the PR. However, React Navigation 4 is frozen and won't have new releases. So we don't accept PRs for it.

Issue #13020, opened 2026-03-07
[native-stack][Android] Using LinearGradient in headerBackground with transparent headerStyle breaks header/content layout


DragomirAndrei19:
### Current behavior

I observed this issue while trying to use LinearGradient + headerBackground property. Tested it on a physical too, Samsung Galaxy A54 (Android 16/IOne UI 8). Also using CLI, not expo.

**_IMPORTANT: Weirdly enough, if you rotate the screen in landscape and then portrait again, it displays correctly (temporarily, at least)._**. For the record, `react-native-edge-to-edge` is used (v1.8.1).

I prepared a single file app repo. Feel free to delete any code related to the bootsplash library.

```
/* eslint-disable react-native/no-inline-styles */
import React from 'react';
import {
  Platform,
  Pressable,
  ScrollView,
  StyleSheet,
  Text,
  View,
} from 'react-native';
import { GestureHandlerRootView } from 'react-native-gesture-handler';
import { SafeAreaProvider, SafeAreaView, useSafeAreaInsets } from 'react-native-safe-area-context';
import { NavigationContainer } from '@react-navigation/native';
import { createNativeStackNavigator } from '@react-navigation/native-stack';
import LinearGradient from 'react-native-linear-gradient';
import { SystemBars } from 'react-native-edge-to-edge';
import BootSplash from 'react-native-bootsplash';

type ReproStackParamList = {
  Home: undefined;
};

const Stack = createNativeStackNavigator<ReproStackParamList>();

const headerBackgroundGradient = () => (
  <LinearGradient
    colors={['#0E3A59', '#2B6A8D']}
    start={{ x: 0.5, y: 0 }}
    end={{ x: 0.5, y: 1 }}
    style={styles.gradient}
  />
);

function HomeLiteScreen() {
  const insets = useSafeAreaInsets();

  return (
    <View style={{paddingTop: insets.top, paddingBottom: insets.bottom, paddingLeft: insets.left, paddingRight: insets.right}}>
    <ScrollView contentContainerStyle={styles.content}>
      <Text style={styles.title}>Home Lite Screen</Text>
      <Text style={styles.subtitle}>
        Same navigation setup, only enough content to reproduce
        the header issue.
      </Text>

      <View style={styles.card}>
        <Text style={styles.cardLabel}>Total Alerts (7 days)</Text>
        <Text style={styles.cardValue}>253</Text>
      </View>

      <View style={styles.card}>
[...]

Issue #12944, opened 2026-01-28
[Android] native-stack headerBackground broken


erikchan002:
### Current behavior

On Android, when `headerBackground` is set in `screenOptions` (`native-stack` `7.9.0`-`7.11.0`; `react-native-screens` `~4.16.0`):
```typescript
<Stack.Navigator
  screenOptions={{
    headerBackground: () => (
      <View style={{ flex: 1, backgroundColor: 'pink' }} />
    ),
  }}>
  <Stack.Screen
    name="Title"
    component={() => (
      <View style={{ flex: 1, backgroundColor: 'lightblue' }}>
        <Text>Content</Text>
      </View>
    )}
  />
</Stack.Navigator>
```
<img width="340" height="700" alt="Image" src="https://github.com/user-attachments/assets/f714d1d8-7a8a-4ada-9e71-29ba6254db79" />

Adding `headerTransparent: false` does not help and has the same result.
Forcing a fixed size in `style` of the `headerBackground` component also has the same result.
This issue does not occur on iOS. I have not tested on other platforms.

`native-stack` `7.11.0`: https://snack.expo.dev/@erikchan002/react-navigation-native-stack-7-11-0-headerbackground
`native-stack` `7.9.0`: https://snack.expo.dev/@erikchan002/react-navigation-native-stack-7-9-0-headerbackground

Reverting `native-stack` to `7.8.6` can workaround this issue. 
Forcing `react-native-screens` to `4.19.0` with a development build can also workaround this issue. 

### Expected behavior

`native-stack` `7.8.6`:
<img width="349" height="700" alt="Image" src="https://github.com/user-attachments/assets/0f72760d-d2db-4d48-b2ff-2b69fbca84f5" />

https://snack.expo.dev/@erikchan002/react-navigation-native-stack-7-8-6-headerbackground

This is just an example. I am aware that I can use `headerStyle` for plain `backgroundColor`. I encountered this issue with a `LinearGradient` `headerBackground`.

[...]

PR #13021, opened 2026-03-07
fix(native-stack): robust Android header-height normalization for out-of-order events

DragomirAndrei19:
**Motivation**
A fix attempt for https://github.com/react-navigation/react-navigation/issues/13020 and possibly https://github.com/react-navigation/react-navigation/issues/12944.

This is my first PR for **react-navigation**, so please take everything with a grain of salt, I might be wrong in my statements/suppositions (and also ding a lot of back and forth), but this does seem to fix (at least apparently) the issue from my tests so far.

This PR attempts to fix an Android header sizing issue in `@react-navigation/native-stack` that I reproduced when using `headerBackground` (linear gradient) together with `headerStyle.backgroundColor: 'transparent'`.

I also prepared a patch file.

[@react-navigation+native-stack+7.14.4.patch](https://github.com/user-attachments/files/25815518/%40react-navigation%2Bnative-stack%2B7.14.4.patch)

**Full code I tested this on is below.**

From my understanding, this is mainly visible with `headerBackground` because that path renders a separate JS background view whose height directly follows headerHeight; when Android emits an out-of-order short height (like 56.17), that layer visibly clips, while solid-color headers are mostly painted natively and tend to hide the same transient mismatch.

On my physical test device (**Samsung Galaxy A54, Android 16 / OneUI 8.0** - and curiosly enough, on a Pixel 9a emulator from Android Studio this wasn't happening), I observed out-of-order native header height events on initial render. In Logcat, I consistently saw this sequence (I also tested with orientation changes):

```txt
rawHeaderHeight: ~84.62
then rawHeaderHeight: ~56.17
```

`native-stack` already has an Android correction branch, but it depended on this threshold:

`headerHeight <= ANDROID_DEFAULT_HEADER_HEIGHT`

On Android this constant is effectively `56`, so a value like `56.17` can bypass correction (`56.17 > 56`) and be applied as-is.

When `headerBackground` is enabled, this becomes visually obvious as clipped/misaligned header background/layout. Rotating to landscape and back often "fixes" it temporarily by triggering a new layout/event sequence.

So, based on my tests, the concrete problem fixed by this PR is threshold brittleness in the Android correction path.

I also reviewed this discussion/commit context:
https://github.com/react-navigation/react-navigation/commit/0467f9ad829e4494fdb9d585c758c9535ef9b099

I used it as context, but this PR is based on direct reproduction + logs, not on assuming that commit alone is the cause.

**What changed**
Instead of hard-threshold correction, this PR uses adaptive normalization on Android:

```txt
rawHeaderHeight = headerHeight
correctedHeaderHeight = headerHeight + insets.top
expectedHeaderHeight = defaultHeaderHeight
choose candidate closer to expectedHeaderHeight
```

This removes dependency on strict `<= 56` checks and handles rounded values like `56.17` across devices/densities.

**Test plan**
Manual validation (physical device):

1. Configure native-stack header with:
   - `headerBackground` as a gradient component
   - `headerStyle: { backgroundColor: 'transparent' }`
2. Launch app on a screen using this header.
3. Verify header is not clipped/misaligned on first render.
4. Rotate to landscape.
5. Rotate back to portrait.
6. Verify header remains stable and gradient remains correctly rendered.
7. I think this should also be tested in cases such as nesting the stack inside a tab layout (native-tabs).

This is the code the PR was tested on 

```
/* eslint-disable react-native/no-inline-styles */
import React from 'react';
import {
  Platform,
  Pressable,
  ScrollView,
  StyleSheet,
  Text,
  View,
} from 'react-native';
import { GestureHandlerRootView } from 'react-native-gesture-handler';
import { SafeAreaProvider, SafeAreaView, useSafeAreaInsets } from 'react-native-safe-area-context';
import LinearGradient from 'react-native-linear-gradient';
import { SystemBars } from 'react-native-edge-to-edge';
import BootSplash from 'react-native-bootsplash';

type ReproStackParamList = {
  Home: undefined;
};

const Stack = createNativeStackNavigator<ReproStackParamList>();

const headerBackgroundGradient = () => (
  <LinearGradient
    colors={['#0E3A59', '#2B6A8D']}
    start={{ x: 0.5, y: 0 }}
    end={{ x: 0.5, y: 1 }}
    style={styles.gradient}
  />
);

function HomeLiteScreen() {
  const insets = useSafeAreaInsets();

  return (
    <View style={{paddingTop: insets.top, paddingBottom: insets.bottom, paddingLeft: insets.left, paddingRight: insets.right}}>
    <ScrollView contentContainerStyle={styles.content}>
      <Text style={styles.title}>Home Lite Screen</Text>
      <Text style={styles.subtitle}>
        Same navigation setup, only enough content to reproduce
        the header issue.
      </Text>

      <View style={styles.card}>
        <Text style={styles.cardLabel}>Total Alerts (7 days)</Text>
        <Text style={styles.cardValue}>253</Text>
      </View>

      <View style={styles.card}>
        <Text style={styles.cardLabel}>Critical Alerts</Text>
        <Text style={styles.cardValue}>19</Text>
      </View>

      <View style={styles.buttonRow}>
        <Pressable style={styles.button}>
          <Text style={styles.buttonText}>Action 1</Text>
        </Pressable>
        <Pressable style={styles.button}>
          <Text style={styles.buttonText}>Action 2</Text>
        </Pressable>
      </View>

      <View style={styles.spacer} />
    </ScrollView>
    </View>
  );
}

function App() {
  React.useEffect(() => {
    BootSplash.hide({ fade: true }).catch(error => {
      console.error('BootSplash hide failed:', error);
    });
  }, []);

  return (
    <GestureHandlerRootView style={{ flex: 1 }}>
      <SafeAreaProvider>
        <SystemBars style={systemBarStyle} />

        <NavigationContainer>
          <Stack.Navigator
            screenOptions={{
              headerTitleAlign: 'center',
              statusBarTranslucent: true,
              headerBackground: headerBackgroundGradient,
              headerStyle: {
                backgroundColor:
                  Platform.OS === 'android' ? 'green' : 'transparent',
              },
              headerTintColor: '#FFFFFF',
              headerTitleStyle: {
                fontSize: 22,
                fontWeight: '700',
              },
              contentStyle: {
                backgroundColor: '#F3F5F7',
              },
            }}
          >
            <Stack.Screen
              name="Home"
              component={HomeLiteScreen}
              options={{ title: 'Home' }}
            />
          </Stack.Navigator>
        </NavigationContainer>
      </SafeAreaProvider>
    </GestureHandlerRootView>
  );
}

const systemBarStyle = {
  statusBar: 'light',
  navigationBar: 'dark',
} as const;

  gradient: {
    flex: 1,
  },
  content: {
    padding: 16,
    gap: 12,
  },
  title: {
    fontSize: 24,
    fontWeight: '700',
    color: '#102131',
  },
  subtitle: {
    fontSize: 15,
    color: '#334E64',
  },
  card: {
    backgroundColor: '#FFFFFF',
    borderRadius: 12,
    padding: 14,
    borderWidth: StyleSheet.hairlineWidth,
    borderColor: '#D0D9E1',
  },
  cardLabel: {
    fontSize: 14,
    color: '#4A657A',
    marginBottom: 6,
  },
  cardValue: {
    fontSize: 28,
    fontWeight: '700',
    color: '#102131',
  },
  buttonRow: {
    flexDirection: 'row',
    gap: 12,
  },
  button: {
    backgroundColor: '#0E3A59',
    borderRadius: 10,
    paddingVertical: 12,
    paddingHorizontal: 16,
  },
  buttonText: {
    color: '#FFFFFF',
    fontWeight: '600',
  },
  spacer: {
    height: 500,
  },
});

export default App;
```

Device used:

1. Samsung Galaxy A54 ( Android 16 / OneUI 8.0)

Regression check observed in logs:

1. Out-of-order heights (`~84.62` then `~56.17`) were present before.
2. With this change, the incorrect short event is normalized instead of being accepted as final height.

**About react-native-screens**
This still looks related to native event ordering/measurement timing from the Android header pipeline in `react-native-screens.`
My patch is intentionally in native-stack as a robust consumer-side normalization, not a claim that react-native-screens is definitively at fault in all cases.

Long story short, after this fix, the single-file application provided inside https://github.com/react-navigation/react-navigation/issues/13020

Should no longer show

| `backgroundColor="transparent" ` (incorrect layout)|` backgroundColor="solid_color" ` (correct layout)|
|----------|----------|
| Image| Image |

but instead (although now I'm noticing a bit of extra padding inside the home screen content - left image...unless we remove `paddingTop: insets.top` from our app home screen code OR better remove ` headerBackground: headerBackgroundGradient,` - which we shouldn't need anyway if the background is only a solid color
)

| `backgroundColor="transparent" ` (now correct layout)|` backgroundColor="solid_color" ` (correct layout)|
|----------|----------|
| Image| Image |

With the current patch, if we remove `headerBackground: headerBackgroundGradient` too it correctly renders this

Without the current patch, if we leave headerBackgroundGradient and only change "green" to "transparent" here `( Platform.OS === 'android' ? 'green' : 'transparent')` this happens (notice the insets.top padding in our app screen code is only applied after orientation change):

https://github.com/user-attachments/assets/dd6a2316-0861-4c9f-aad0-ae695e45cca7

`


satya164 (changes requested):
Thanks for the PR. As I understand, this is the main issue:

> On Android this constant is effectively 56, so a value like 56.17 can bypass correction (56.17 > 56) and be applied as-is.

If this is the case, I think changing `headerHeight <= ANDROID_DEFAULT_HEADER_HEIGHT` to `Math.round(headerHeight) <= ANDROID_DEFAULT_HEADER_HEIGHT` should fix the issue. This is a hacky workaround, and it's best not to make the logic more complex.

The hack only works because if we know it's equal to the header height, then it's very likely that it doesn't include top inset. if it's something different, then we don't really know why so it'd be unsafe to do this. We don't want to pick the candidate closer to default header height, as future changes such as large title can easily break this.

DragomirAndrei19:
> Thanks for the PR. As I understand, this is the main issue:
> 
> > On Android this constant is effectively 56, so a value like 56.17 can bypass correction (56.17 > 56) and be applied as-is.
> 
> If this is the case, I think changing `headerHeight <= ANDROID_DEFAULT_HEADER_HEIGHT` to `Math.round(headerHeight) <= ANDROID_DEFAULT_HEADER_HEIGHT` should fix the issue. This is a hacky workaround, and it's best not to make the logic more complex.
> 
> The hack only works because if we know it's equal to the header height, then it's very likely that it doesn't include top inset. if it's something different, then we don't really know why so it'd be unsafe to do this. We don't want to pick the candidate closer to default header height, as future changes such as large title can easily break this.

Hello, thanks for the reply. I suspected afterwards that the more complex normalization approach could break cases such as large titles.

I removed that logic and kept the existing correction, but changed the condition to use `Math.round(headerHeight)`, exactly as you suggested, to avoid brittle cases like 56.17. Hopefully no device decides to get creative and report something like 57.12.

This seems to address the main issue on my side. Not sure if it resolves the `headerBackground` reports in all cases, but it might help some.

And hopefully React Native will soon give some attention to the out-of-order events problem that made this hack necessary in the first place. I bumped the thread after the react-native-screens maintainers, so fingers crossed.

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -281,6 +281,11 @@ const SceneView = ({
                   // and then the one without status bar height
                   // This is hack to include status bar height if it's not already included
                   // It only works because header height doesn't change dynamically on Android
+                  //
+                  // Keep the existing correction logic, but round the measured height
+                  // before comparing against the default Android header height.
+                  // This avoids brittle cases such as values like 56.17 bypassing the
+                  // correction even though they are effectively the default 56dp header.
```
satya164: The comment doesn't make sense here - "existing logic" is the logic that's here already
```suggestion
```

packages/native-stack/src/views/NativeStackView.native.tsx
```diff
@@ -271,7 +271,7 @@ const SceneView = ({
                 if (
                   Platform.OS === 'android' &&
                   headerHeight !== 0 &&
-                  headerHeight <= ANDROID_DEFAULT_HEADER_HEIGHT
+                  Math.round(headerHeight) <= ANDROID_DEFAULT_HEADER_HEIGHT
```
satya164, line 275:
```suggestion
                  // On some devices, height maybe slightly off (e.g. 56.17 instead of 56)
                  Math.round(headerHeight) <= ANDROID_DEFAULT_HEADER_HEIGHT
```

satya164 (changes requested):

satya164 (approved):
Thanks

Issue #13016, opened 2026-03-06
`autoFocus` property of `TextInput` component does not work


WojtekBoman:
### Current behavior

When the user navigates to a screen that contains a text input with `autoFocus`, the input briefly receives focus (cursor may appear for a moment) and then immediately loses it. The user must tap the input manually to start typing.

### Expected behavior

When the user navigates to a screen that contains a text input with `autoFocus`, the input should automatically receive focus and stay focused so the user can start typing immediately without tapping the input first.

#### RN TextInput (bug: focus then onBlur)

1. Start the app: `npm run web`.
2. Open DevTools console (F12 or Cmd+Option+I).
3. From the home screen, tap **/rn-text-input**.
4. Observe the console logs.

**Expected:** `[RN TextInput] onFocus - gained focus` only (input stays focused).

**Actual:** `[RN TextInput] onFocus - gained focus` followed by `[RN TextInput] onBlur - lost focus` (input loses focus immediately after gaining it).

#### Live Markdown (works correctly)

1. Start the app and open DevTools console.
2. From the home screen, tap **/live-markdown**.
3. Observe the console logs.

**Expected:** `[MarkdownTextInput] onFocus - gained focus` only (input stays focused).

**Actual:** Matches expected. Input gains focus and remains focused.

https://github.com/WojtekBoman/FocusInputRepro

https://github.com/user-attachments/assets/3a9b2801-38ae-4c2a-8129-49cff99a8e82

### Platform

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| Package | Version |
| -------------------------------------- | ------- |
| @react-navigation/native | 7.1.33 |
| @react-navigation/stack | 7.7.2 |
| @react-navigation/native-stack | 7.14.2 |
| react-native-screens | 4.24.0 |
| react-native-safe-area-context | 5.7.0 |
| react-native-gesture-handler | 2.30.0 |
| react-native-reanimated | 4.2.1 |
| react-native | 0.83.2 |
| expo | 55.0.5 |
| node | 20.20.0 |
[...]

PR #13022, opened 2026-03-09
fix(stack): preserve autoFocus on TextInput when navigating forward

WojtekBoman:
## Motivation

Fixes [react-navigation/react-navigation#13016](https://github.com/react-navigation/react-navigation/issues/13016).

When the user navigates to a screen that contains a text input with `autoFocus`, the input briefly receives focus (cursor may appear for a moment) and then immediately loses it. The user must tap the input manually to start typing.

**Expected:** The input should automatically receive focus and stay focused so the user can start typing immediately without tapping the input first.

## Test plan

1. **RN TextInput (bug: focus then onBlur):**
   - Start the app: `npm run web`
   - Open DevTools console (F12 or Cmd+Option+I)
   - From the home screen, tap **/rn-text-input**
   - Observe the console logs

   **Expected:** `[RN TextInput] onFocus - gained focus` only (input stays focused).

   **Actual (before fix):** `[RN TextInput] onFocus - gained focus` followed by `[RN TextInput] onBlur - lost focus` (input loses focus immediately after gaining it).

2. **Lint and types:**
   ```bash
   npm run lint
   npx tsc --noEmit
   ```

3. **Tests:**
   ```bash
   npm test
   ```

### DIFF
--- a/packages/stack/src/utils/useKeyboardManager.tsx
+++ b/packages/stack/src/utils/useKeyboardManager.tsx
@@ -1,7 +1,13 @@
 import * as React from 'react';
 import { type HostInstance, Keyboard, TextInput } from 'react-native';
 
-export function useKeyboardManager({ enabled }: { enabled: boolean }) {
+export function useKeyboardManager({
+  enabled,
+  focused,
+}: {
+  enabled: boolean;
+  focused: boolean;
+}) {
   // Numeric id of the previously focused text input
   // When a gesture didn't change the tab, we can restore the focused input with this
   const previouslyFocusedTextInputRef = React.useRef<HostInstance>(undefined);
@@ -109,12 +115,14 @@ export function useKeyboardManager({ enabled }: { enabled: boolean }) {
   // This handles the "navigate forward" case so we don't dismiss the new screen's
   // auto-focused input from handleTransition.
   React.useLayoutEffect(() => {
-    if (enabledRef.current && !enabled) {
+    if (enabledRef.current && !focused) {
       Keyboard.dismiss();
     }
+  }, [focused]);
 
+  React.useLayoutEffect(() => {
     enabledRef.current = enabled;
-  }, [enabled]);
+  });
 
   React.useEffect(() => {
     return () => clearKeyboardTimeout();
--- a/packages/stack/src/views/Stack/CardContainer.tsx
+++ b/packages/stack/src/views/Stack/CardContainer.tsx
@@ -105,7 +105,7 @@ function CardContainerInner({
   const enabled = focused && options.keyboardHandlingEnabled !== false;
 
   const { onPageChangeStart, onPageChangeCancel, onPageChangeConfirm } =
-    useKeyboardManager({ enabled });
+    useKeyboardManager({ enabled, focused });
 
   const handleOpen = () => {
     const { route } = scene.descriptor;
--- a/packages/stack/src/utils/__tests__/useKeyboardManager.test.tsx
+++ b/packages/stack/src/utils/__tests__/useKeyboardManager.test.tsx
@@ -10,7 +10,7 @@ describe('useKeyboardManager', () => {
   describe('onPageChangeConfirm', () => {
     test('calls onPageChangeCancel when closing is false', () => {
       const { result } = renderHook(() =>
-        useKeyboardManager({ enabled: true })
+        useKeyboardManager({ enabled: true, focused: true })
       );
 
       const blurMock = jest.fn();
@@ -36,7 +36,7 @@ describe('useKeyboardManager', () => {
       const dismissSpy = jest.spyOn(Keyboard, 'dismiss');
 
       const { result } = renderHook(() =>
-        useKeyboardManager({ enabled: true })
+        useKeyboardManager({ enabled: true, focused: true })
       );
 
       act(() =>
@@ -54,7 +54,7 @@ describe('useKeyboardManager', () => {
 
     test('blurs previously focused input when closing with gesture and active', () => {
       const { result } = renderHook(() =>
-        useKeyboardManager({ enabled: true })
+        useKeyboardManager({ enabled: true, focused: true })
       );
 
       const blurMock = jest.fn();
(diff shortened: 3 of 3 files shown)


packages/stack/src/views/Stack/CardContainer.tsx
```diff
@@ -157,7 +157,10 @@ function CardContainerInner({
 
     const { route } = scene.descriptor;
 
-    if (!gesture) {
+    if (!gesture && closing) {
+      // Only dismiss when going back (closing). When opening a new screen,
+      // the previous screen's useLayoutEffect in useKeyboardManager handles dismiss on focus loss.
+      // Calling onPageChangeConfirm(true) when opening would dismiss the new screen's auto-focused input.
       onPageChangeConfirm?.(true);
```
satya164: It'd be better to have this `useKeyboardManager.tsx`. referring to something in a different file is easier to go out of date.

Probably best to change `onPageChangeConfirm` to accept an object `{ gesture: boolean, active: boolean, closing: boolean }` and move all of the logic calling into `onPageChangeConfirm` function itself. So all of this logic stays in one file.

packages/stack/src/utils/useKeyboardManager.tsx
```diff
@@ -91,6 +94,17 @@ export function useKeyboardManager(isEnabled: () => boolean) {
     }
   }, [clearKeyboardTimeout, isEnabled]);
 
+  // Dismiss keyboard when screen loses focus (e.g. when pushing a new screen).
+  // This handles the "navigate forward" case so we don't dismiss the new screen's
+  // auto-focused input from handleTransition.
```
satya164, line 116: This effect checks if `isEnabled` is changed, and not whether screen focus changed. While `isEnabled` includes focus state, it's confusing to use `isEnabled` to detect a change in focus. Same about the comment on the ref.

packages/stack/src/utils/useKeyboardManager.tsx
```diff
@@ -7,6 +7,9 @@ export function useKeyboardManager(isEnabled: () => boolean) {
   const previouslyFocusedTextInputRef = React.useRef<HostInstance>(undefined);
   const startTimestampRef = React.useRef<number>(0);
   const keyboardTimeoutRef = React.useRef<NodeJS.Timeout>(undefined);
+  // Previous value of isEnabled() (screen focused + keyboard handling enabled).
+  // Used to detect when this screen loses focus and dismiss its keyboard when navigating forward.
+  const prevKeyboardHandlingEnabledRef = React.useRef(isEnabled());
```
satya164: `prev` won't be correct as it always represents the current screen's value. It's only previous in the `useLayoutEffect` right before it's updated. Rename to `enabledRef` instead.

packages/stack/src/utils/useKeyboardManager.tsx
```diff
@@ -91,6 +94,17 @@ export function useKeyboardManager(isEnabled: () => boolean) {
...
+  // Dismiss keyboard when screen loses focus (e.g. when pushing a new screen).
+  // This handles the "navigate forward" case so we don't dismiss the new screen's
+  // auto-focused input from handleTransition.
+  React.useLayoutEffect(() => {
+    const enabled = isEnabled();
+    if (prevKeyboardHandlingEnabledRef.current && !enabled) {
+      Keyboard.dismiss();
+    }
+    prevKeyboardHandlingEnabledRef.current = enabled;
+  }, [isEnabled]);
```
satya164: `isEnabled` uses `navigation.isFocused()`, which isn't intended to be used during render, as it's not reactive. This logic only works because the descriptor changes on focus change, recreating the callback. If a later optimization changed this, then this will stop working as the function won't change, keeping dep array the same and preventing effect from running.

To do this correctly, instead,
- change `useKeyboardManager` to accept an object instead of a callback
- pass `{ enabled: boolean, focused: boolean }` (`focused` is available in scope from props in `CardContainer`)
- store `enabled` in a ref, and use that ref in places where `isEnabled()` is used (in `onPageChangeX` callbacks)
- then use `enabled` in the dependency array of the `useLayoutEffect` and rest is as usual

Also, nit: use newlines between code blocks.

WojtekBoman:
@satya164  I've addressed your comments and added the tests, so it should be fine now 🤞  The only thing that's failing is 1 check (e2e ios) :( but that's probably unrelated to my changes

packages/stack/src/utils/useKeyboardManager.tsx
satya164: This one doesn't seem to be addressed. While logically it is correct since `enabled` includes focus, it is confusing to use it to detect focus change. We also don't want this effect to run if `enabled` is changed without focus changing.

Pass `focused` to `useKeyboardManager` and use `focused` for the dependency array and `!enabled` check instead.

satya164:
@WojtekBoman yea, it's a maestro issue.

seems one review is left. i'll merge after it is addressed.

packages/stack/src/utils/useKeyboardManager.tsx
```diff
@@ -89,7 +69,58 @@ export function useKeyboardManager(isEnabled: () => boolean) {
...
+  // Dismiss keyboard when screen loses focus (e.g. when pushing a new screen).
+  // This handles the "navigate forward" case so we don't dismiss the new screen's
+  // auto-focused input from handleTransition.
+  React.useLayoutEffect(() => {
+    if (!enabled) {
+      Keyboard.dismiss();
+    }
+
+    enabledRef.current = enabled;
+  }, [enabled, focused]);
```
satya164: Even though `focused` is in dep array now, it also includes `enabled` and the logic checks `enabled`. This hook shouldn't be running if `keyboardHandlingEnabled` changed but focus state of screen didn't change.


```suggestion
  React.useLayoutEffect(() => {
    if (enabledRef.current && !focused) {
      Keyboard.dismiss();
    }
  }, [focused]);

  React.useLayoutEffect(() => {
    enabledRef.current = enabled;
  });
```

WojtekBoman: Ok, this is what I wanted to ask if I should use it here, I'll adjust it in a moment

WojtekBoman: Should be ok now, thanks for your help! 🙇

WojtekBoman:
@satya164 It's ready 🫡

PR #13029, opened 2026-03-11
fix(tab-view): reset Android offset when pager becomes idle

christophby:
# react-native-tab-view

## What does this fix?
- Reset the tab-view pager `offset` to `0` when `pageScrollState` becomes `idle` on Android.
- Prevent offset drift that can leave the indicator visually between tabs after swipe completion.
- Keep scope limited to `PagerViewAdapter.native.tsx`; iOS and web behavior remain unchanged.

**Before** (glitches start at 0:14)

https://github.com/user-attachments/assets/a7cdbf2f-11b9-4684-96d2-48db28239fac

**After**

https://github.com/user-attachments/assets/20ad9d6d-b345-4a8b-9d93-6529118dc6c4

## Reproduction
* [Repo](https://github.com/christophby/rn-true-sheet-repro)
* run `npm i`
* `npm run android`
* Navigate to "Profile" Tab. -> You should see the issue
* Run `npx patch-package`
* Reload app
* Issue should be gone

### DIFF
--- a/packages/react-native-tab-view/src/PagerViewAdapter.native.tsx
+++ b/packages/react-native-tab-view/src/PagerViewAdapter.native.tsx
@@ -90,6 +90,10 @@ export function PagerViewAdapter({
 
     switch (pageScrollState) {
       case 'idle':
+        if (Platform.OS === 'android') {
+          // Work around pager offset drift that can leave indicator in-between tabs.
+          offset.setValue(0);
+        }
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
thanks for the pr. this looks like a bug in react-native-pager-view. can you open an issue there? I'd prefer to avoid workarounds here.

PR #13035, opened 2026-03-15
fix(types): handle missing canPreventDefault in EventListenerCallback generic

YevheniiKotyrlo:
## Problem

When downstream consumers wrap react-navigation navigators and tsc emits `.d.ts`, `EventListenerCallback`'s 3rd generic parameter resolves to `unknown` for events that don't declare `canPreventDefault`. This `unknown` violates the `extends boolean | undefined` constraint, producing TS2344 for any consumer with `skipLibCheck: false`.

Repro: https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault-direct ([CI](https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault-direct/actions/runs/23257409850/job/67616198044))

## Fix

Two changes to `EventListenerCallback` in `packages/core/src/types.tsx`:

1. **Remove the constraint** on `EventCanPreventDefault` — allows `unknown` (from expanded generics) without TS2344
2. **Flip the conditional** from `undefined extends EventCanPreventDefault ? false : EventCanPreventDefault` to `EventCanPreventDefault extends true ? true : false` — always produces `boolean`, satisfying `EventArg`'s constraint

The behavior is identical: only events with `canPreventDefault: true` (like `beforeRemove`) get `preventDefault()`. All others resolve to `false`.

1 file, +2/−4 lines. Zero runtime changes.

### DIFF
--- a/packages/core/src/types.tsx
+++ b/packages/core/src/types.tsx
@@ -220,13 +220,11 @@ export type EventArg<
 export type EventListenerCallback<
   EventMap extends EventMapBase,
   EventName extends keyof EventMap,
-  EventCanPreventDefault extends
-    | boolean
-    | undefined = EventMap[EventName]['canPreventDefault'],
+  EventCanPreventDefault = EventMap[EventName]['canPreventDefault'],
 > = (
   e: EventArg<
     EventName,
-    undefined extends EventCanPreventDefault ? false : EventCanPreventDefault,
+    EventCanPreventDefault extends true ? true : false,
     EventMap[EventName]['data']
   >
 ) => void;


satya164:
Can you provide a repro showing the issue?

YevheniiKotyrlo:
Repro: https://github.com/YevheniiKotyrlo/repro-react-navigation-canpreventdefault ([CI run](https://github.com/YevheniiKotyrlo/repro-react-navigation-canpreventdefault/actions/runs/23198335034))

```bash
npm install --legacy-peer-deps
npm run check          # 23 TS2344 errors
npm run check:fixed    # 0 errors
```

`repro.ts` uses the exact generic arguments from expo-router's emitted [StackClient.d.ts](https://unpkg.com/expo-router@55.0.3/build/layouts/StackClient.d.ts). `repro-fixed.ts` is identical but with `undefined` instead of `unknown` — that's the entire fix.

### Root cause

Events like `focus` in `EventMapCore` are `{ data: undefined }` with no `canPreventDefault`. When tsc expands `EventMap[EventName]['canPreventDefault']` on the concrete type during `.d.ts` emission, the missing property resolves to `unknown` (not through the `EventMapBase` constraint). This violates `EventCanPreventDefault extends boolean | undefined` → TS2344 for any downstream consumer with `skipLibCheck: false`.

### Runtime parity

I verified that all 23 events are emitted at runtime without `canPreventDefault` — the type annotation matches the actual behavior:

| Package | Non-preventable events | Runtime emission |
|---------|----------------------|------------------|
| core | `focus`, `blur`, `state` | [`useFocusEvents.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/useFocusEvents.tsx), [`useNavigationBuilder.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/useNavigationBuilder.tsx#L880) |
| native-stack | `transitionStart`, `transitionEnd`, `gestureCancel`, `sheetDetentChange` | [`NativeStackView.native.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/native-stack/src/views/NativeStackView.native.tsx#L499) |
| bottom-tabs | `tabLongPress`, `transitionStart`, `transitionEnd` | [`BottomTabBar.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/views/BottomTabBar.tsx#L422), [`BottomTabViewCustom.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/views/BottomTabViewCustom.tsx#L137) |
| stack | `transitionStart`, `transitionEnd`, `gestureStart`, `gestureEnd`, `gestureCancel` | [`StackView.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Stack/StackView.tsx#L445) |
| drawer | `transitionStart`, `transitionEnd`, `gestureStart`, `gestureEnd`, `gestureCancel` | [`DrawerView.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/src/views/DrawerView.tsx#L127) |
| material-top-tabs | `tabLongPress`, `swipeStart`, `swipeEnd` | [`MaterialTopTabBar.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/material-top-tabs/src/views/MaterialTopTabBar.tsx#L220), [`MaterialTopTabView.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/material-top-tabs/src/views/MaterialTopTabView.tsx#L96) |

None of these `emit()` calls pass `canPreventDefault`. The 4 events that do — `beforeRemove`, `tabPress`, `drawerItemPress` — already have `canPreventDefault: true` in their type definitions and are unaffected by this PR.

The guard is in [`useEventEmitter.tsx`](https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/useEventEmitter.tsx#L95) — `preventDefault()` is only attached when `canPreventDefault` is truthy.

### Scope

I expanded the PR from 3 to 6 packages to cover all affected events. 6 files, +44/−23, zero runtime changes. All quality gates pass.

satya164:
Please share a demo using react navigation from npm with a practical example, not a replication of the types. Otherwise it makes hard to understand where the issue is and whether the fix makes sense.

YevheniiKotyrlo:
Here's a practical repro — a minimal Expo project using expo-router from npm: https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault

```bash
npm install --legacy-peer-deps
npm run check    # 64 TS2344 errors
```

The app is just a root layout with `<Stack />` and a home screen — standard expo-router setup. The only tsconfig change is `skipLibCheck: false`.

CI output: [tsc --noEmit (skipLibCheck: false)](https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault/actions/runs/23239242377/job/67551062853#step:5:1) — 64 TS2344 errors from expo-router's bundled `StackClient.d.ts` (28) and `TabsClient.d.ts` (36), where tsc baked `unknown` into `EventListenerCallback`'s 3rd parameter for events that don't declare `canPreventDefault`.

satya164:
Please provide a repro that uses react navigation API directly with the minimal code required to reproduce the issue. I don't know the code of expo-router and how it's using it/

YevheniiKotyrlo:
Here's a repro using `@react-navigation/native-stack` directly: https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault-direct ([CI](https://github.com/YevheniiKotyrlo/repro-navigation-canpreventdefault-direct/actions/runs/23257409850/job/67616198044))

```bash
npm install --legacy-peer-deps
npm run repro
```

`wrap-navigator.ts` calls `createNativeStackNavigator()` and wraps the navigator in `forwardRef` with `Omit<ComponentProps<T>, 'children'>` — same pattern expo-router's `withLayoutContext` uses. When tsc emits `.d.ts` for this wrapper, `EventListenerCallback`'s default 3rd parameter resolves to `unknown` for events that don't declare `canPreventDefault`. You can see 15 occurrences of `unknown` in the emitted `dist/wrap-navigator.d.ts`.

To be transparent: the `unknown` doesn't affect developer-visible types in IDE or at call sites. The conditional in `EventListenerCallback` — `undefined extends EventCanPreventDefault ? false : EventCanPreventDefault` — evaluates to `false` regardless of whether `EventCanPreventDefault` is `unknown` or `undefined`, so `EventArg`'s shape is the same either way. The issue is strictly that `unknown` violates the `extends boolean | undefined` constraint in emitted `.d.ts`, producing TS2344 for consumers with `skipLibCheck: false`.

That said, the emitted `unknown` is semantically wrong — `beforeRemove` correctly declares `canPreventDefault: true` and emits `true`, so the other events should emit `undefined` (not preventable), not `unknown` (unspecified). Adding `canPreventDefault?: undefined` makes all events explicitly declare their preventability, which is what `EventMapBase`'s contract (`canPreventDefault?: boolean`) intends.

satya164:
Thanks for the repro.

From the repro, it seems like an issue with the way Expo Router wraps React Navigation Navigators and not a bug in React Navigation itself. We use generics for type inference in many places, so it'd be ideal to wrap it in a way that doesn't change the actual types.

The type changes don't represent actual code as this property will never be present, and adds unnecessary noise and verbosity, so I'm not willing to make such a change.

YevheniiKotyrlo:
Took a different approach based on your feedback. Instead of annotating every event, the fix is now a 2-line change to `EventListenerCallback`'s generic — removes the constraint on the 3rd parameter and flips the conditional so it always resolves to `boolean`.

No event types are touched. The behavior is identical: only `canPreventDefault: true` events get `preventDefault()`.

satya164:
Since Expo has forked React Navigation codebase (https://github.com/expo/expo/pull/44176), I'm closing this as such changes won't be propagated to Expo Router.

Issue #12583, opened 2025-05-06
Accessibility bug when using a Stack with animation on web


mathias-berg:
### Current behavior

When using a stack I would expect only the top most screen in the stack to be visible in the accessibility tree.
And it is, as long as you don't add an animation in `screenOptions` of the stack or `options` for the screen.

For this very simple set up with two screens (Home Screen and Details Screen) it looks like this when on the Details Screen:
<img width="851" alt="Image" src="https://github.com/user-attachments/assets/2bde810f-3e9b-4a1d-8f3d-3090510cf77d" />

But as soon as I add animation, the previous screen will be listed in the accessibility tree.
<img width="915" alt="Image" src="https://github.com/user-attachments/assets/30fc1152-d2cc-43d7-a3e5-908077c0c3c3" />

As you see in the image above the `main` element of the previous screen (Home Screen) is exposed and accessible from a screen reader and if you use the keyboard to tab to a focusable element. 

There is also an warning in the console related to this:
<img width="1262" alt="Image" src="https://github.com/user-attachments/assets/2c15fb41-a07a-4e68-ba10-75043306f5e3" />

### Expected behavior

The expected behaviour is that it should work the same way regardless if the animation is set or not.
The previous screen should not be accessible to a screen reader or you should not be able to keyboard tab to the previous screen.

A side note is that is also the same if I change the `headerMode` to `float`, then the previous header is also accessible from the current screen which it should not be.

This behaviour is not really reproducible on Snack. The previous screen is visible in the accessibility tree for a short moment and then is goes away. Why I don't know, but I will add a small repo that you can test this on and also the Snack link.
https://snack.expo.dev/@fyranollfyra/accessibility-bug

This behaviour started on version `7.2.1`, `7.2.0` works fine. Not sure if the console warning was there on `7.2.1`, but it is at the latest version tested at least.

### Reproduction

https://github.com/mathias-berg/react-navigation-a11y-bug

### Platform

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
[...]

PR #13038, opened 2026-03-18
fix: improve aria-hidden handling during stack animations

ihabshea:
**Motivation**

Fixes https://github.com/react-navigation/react-navigation/issues/12583

When stack animations are enabled on web, non-focused screens remain reachable via keyboard Tab navigation. The existing `aria-hidden={!focused}` hides screens from screen readers but doesn't prevent keyboard focus on focusable elements (buttons, links, etc.).

Without animations, `visibility: hidden` is applied to non-focused screens, which blocks both screen reader access and keyboard focus. But with animations enabled, `visibility` stays `visible` (needed for the animation to render), so keyboard focus leaks through to the previous screen.

This adds the `inert` attribute to non-focused screens, which blocks both screen reader access and keyboard focus — matching the behavior that `visibility: hidden` already provides when animations are disabled.

**Test plan**

- Added unit tests for `CardA11yWrapper` verifying `aria-hidden` behavior on focused and unfocused screens
- All existing tests pass (`yarn test packages/stack/` — 36 tests, 5 suites)
- Lint, typecheck, and commitlint all pass
- The `inert` attribute is a web DOM feature that can be verified by inspecting elements in browser DevTools during stack transitions

### DIFF
--- a/packages/stack/src/views/Stack/CardA11yWrapper.tsx
+++ b/packages/stack/src/views/Stack/CardA11yWrapper.tsx
@@ -27,6 +27,11 @@ export const CardA11yWrapper = React.forwardRef(
     return (
       <View
         aria-hidden={!focused}
+        // On web, aria-hidden alone doesn't prevent keyboard Tab navigation
+        // to focusable elements. The inert attribute blocks both screen reader
+        // access and keyboard focus, which is needed when animations keep the
+        // screen visible in the DOM.
+        inert={!focused || undefined}
         style={[
           StyleSheet.absoluteFill,
           {
--- a/packages/stack/src/views/Stack/__tests__/CardA11yWrapper.test.tsx
+++ b/packages/stack/src/views/Stack/__tests__/CardA11yWrapper.test.tsx
@@ -1,0 +1,74 @@
+import 'react-native-gesture-handler/jestSetup';
+
+import { expect, jest, test } from '@jest/globals';
+import { Text } from '@react-navigation/elements';
+import { NavigationContainer } from '@react-navigation/native';
+import { act, fireEvent, render } from '@testing-library/react-native';
+import { Button, View } from 'react-native';
+
+import { createStackNavigator, type StackScreenProps } from '../../../index';
+
+type StackParamList = {
+  A: undefined;
+  B: undefined;
+};
+
+jest.useFakeTimers();
+
+function findA11yWrapper(element: any) {
+  let node = element.parent;
+  while (node && node.props['aria-hidden'] === undefined) {
+    node = node.parent;
+  }
+  return node;
+}
+
+test('focused screen is not aria-hidden', () => {
+  const Stack = createStackNavigator<StackParamList>();
+
+  const { getByText } = render(
+    <NavigationContainer>
+      <Stack.Navigator>
+        <Stack.Screen name="A">{() => <Text>Screen A</Text>}</Stack.Screen>
+        <Stack.Screen name="B">{() => <Text>Screen B</Text>}</Stack.Screen>
+      </Stack.Navigator>
+    </NavigationContainer>
+  );
+
+  const wrapper = findA11yWrapper(getByText('Screen A'));
+
+  expect(wrapper).not.toBeNull();
+  expect(wrapper!.props['aria-hidden']).toBe(false);
+});
+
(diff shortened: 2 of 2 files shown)


satya164:
Thanks for the PR. Can you provide more details on why is it a problem? What do you mean by "abruptly lose context"? Animations are just visuals and shouldn't have any more context than what was already on the screen when it was focused.

I need a practical example to understand the problem.

ihabshea:
This PR addresses #12583 — when stack animations are enabled on web, previous screens remain visible in the accessibility tree because `aria-hidden` is determined solely by focus state. The fix uses the `closing` state to properly hide screens that are animating out.

satya164:
> when stack animations are enabled on web, previous screens remain visible in the accessibility tree because aria-hidden is determined solely by focus state. The fix uses the closing state to properly hide screens that are animating out

that doesn't make sense. screens that are closing aren't focused anymore - so checking for focus should already cover this. screens that are closing also get removed from DOM after the animation ends so they won't stay in a11y tree. the linked issue also demonstrates the opposite, where a screen is pushed, whereas `closing` is `true` when a screen is popped.

also, the code sets `isA11yHidden = !focused && (!animated || closing)`, which means a screen is hidden if it's not focused **and** either not animated or closing. so it will never set `aria-hidden` for unfocused screens, which are animated, so they will stay in a11y tree.

ihabshea:
@satya164 you're right — the `closing` approach was wrong. I dug deeper into the root cause.

The issue is that `aria-hidden` doesn't prevent keyboard Tab navigation. When animations are disabled, `visibility: hidden` is applied to non-focused screens, which blocks both screen reader access *and* keyboard focus. But when animations are enabled, `visibility` stays `visible` (so the animation can render), and `aria-hidden` alone only hides from screen readers — focusable elements like buttons and links on the previous screen are still reachable via Tab.

I've reworked the fix to add the `inert` attribute to non-focused screens instead. `inert` blocks both screen reader access and keyboard focus, matching what `visibility: hidden` already provides when animations are off. The change is now just a single line addition to `CardA11yWrapper` with no changes needed to `CardContainer`.

satya164:
`inert` is already used on main branch.

https://github.com/react-navigation/react-navigation/blob/a4078f1cb406fea7da4607ad70ac078484eeba9c/packages/stack/src/views/Stack/CardStack.tsx#L708

ihabshea:
You're right, I missed that `ActivityView` already handles `inert` via the `Container` component. My change in `CardA11yWrapper` is redundant.

Looking at the actual flow, I think the issue is in how `activityMode` is determined in `CardStack.tsx`. During a push animation, the previous screen gets `activityMode: 'normal'` because `isFocusing` is true for it (line 678-682 — it's true when the next screen is in `openingRouteKeys`). So the previous screen stays fully interactive and accessible throughout the transition, and only becomes `'inert'` after the animation completes.

Would the right approach be to separate the interactivity concern from `isFocusing`? The previous screen needs to render normally for the animation to look right, but it shouldn't be accessible to keyboard/screen readers during the transition. Or is there a better way to handle this that I'm not seeing?

satya164:
What's the exact issue that you're trying to fix?

ihabshea:
The issue is #12583. On web, when a stack screen with animations is pushed, the previous screen's focusable elements (buttons, links) remain reachable via keyboard Tab. The reporter has a repro repo here: https://github.com/mathias-berg/react-navigation-a11y-bug

The issue started in v7.2.1 — on v7.2.0 the previous screen is properly hidden from both screen readers and keyboard navigation after a push.

satya164:
> On web, when a stack screen with animations is pushed, the previous screen's focusable elements (buttons, links) remain reachable via keyboard Tab
> During a push animation, the previous screen gets activityMode: 'normal' because isFocusing is true for it (line 678-682 — it's true when the next screen is in openingRouteKeys). So the previous screen stays fully interactive and accessible throughout the transition, and only becomes 'inert' after the animation completes.

In a practical case, `isFocusing` also means `focused` will be `true` as the screen which is animating in is the same as focused screen. So far the changes have been unrelated.

So if you're trying to fix the bug, please test it properly and post a video showing the previous and current behavior.

satya164:
Closing due to inactivity and as the change doesn't actually do what it says it does.

Issue #13033, opened 2026-03-13
Cannot add new property 'CONSUMED_PARAMS'


thomasdelobel:
### Current behavior

When using a nested StackNavigator and navigating from a screen to another in it, I get an error : `[TypeError: Cannot add new property 'CONSUMED_PARAMS']` coming from there : https://github.com/react-navigation/react-navigation/blob/5f5a5335e479a5730308b79b4f1bd51385be54af/packages/core/src/useNavigationBuilder.tsx#L733
It appears that sometimes, the `route.params` object is not exetensible, probably sealed or freezed. I can't figure exactly why. 

My navigation structure :
```
StackNavigator (App) :
|__ Signup : SignupNavigator _(receives route.params { context: "whatever" })_
                      |__ SignupUsername
                      |__ SignupPicture
                      |__ ...
|__ ...
````

When navigating from SignupUsername to SignupPicture, via the (SignupNavigator's) navigation `navigate` function, with params, like this : `navigation.navigate("Signup", { context: SignupContext.WHATEVER }, { pop: true })`, it appears that the `route.params` object in `useNavigationBuilder` becomes non-extensible.

To patch it quickly, I will add a `Object.isExtensible(route.params)` check on my own. But I wonder if it is a stable solution or if I'm doing something wrong in my use of React Navigation.

### Expected behavior

Get no error

### Reproduction

https://github.com/react-navigation/react-navigation/issues

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               |
| @react-navigation/material-top-tabs    |
| @react-navigation/stack                |
| @react-navigation/native-stack         |
| react-native-tab-view                  |
| react-native-screens                   |
| react-native-safe-area-context         |
| react-native-gesture-handler           |
| react-native-reanimated                |
| react-native-pager-view                |
| react-native                           |
| expo                                   |
| node                                   |
| npm or yarn                            |

thomasdelobel:
[...]

PR #13042, opened 2026-03-19
fix(core): guard CONSUMED_PARAMS against non-extensible params

gtokman:
## Summary

- Adds an `Object.isExtensible()` check before the `Object.defineProperty(route.params, CONSUMED_PARAMS, ...)` call in `useNavigationBuilder`
- When using React Compiler (which freezes objects) or other tools that make `route.params` non-extensible, the current code throws `TypeError: Cannot add new property 'CONSUMED_PARAMS'`
- This is a minimal, safe fix — if params are frozen, the property simply won't be set, and the next navigation will re-consume params as expected

## Related

- https://github.com/react-navigation/react-navigation/issues/13033

Code 2026-03-19 11 14 57

### DIFF
--- a/packages/core/src/ConsumedParamsContext.tsx
+++ b/packages/core/src/ConsumedParamsContext.tsx
@@ -5,8 +5,6 @@ type ConsumedParamsContextValue = {
   setRef: (ref: WeakRef<object>) => void;
 };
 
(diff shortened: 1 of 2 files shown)


satya164 (approved):
Thank you

Issue #13043, opened 2026-03-23
Headers on Android lose accessibility ability when headerTransparent: true


helblinglilly:
### Current behavior

When the header has the option `headerTransparent: true`, then the entire header becomes inaccessible to screen readers on Android.

TalkBack cannot navigate into the header anymore.

This does not exist on iOS - VoiceOver will read out and select the headers fine.

Screenshots picture use of `scrcpy` which highlights the select a11y element, but can be reproduced with native TalkBack options as well.

https://github.com/user-attachments/assets/5a6eeb64-300d-4e57-9ae9-c0b0c35f953d

### Expected behavior

The behaviour should be the same as iOS, or when `headerTransparent: false` whereby the header has the first tab index, can can be navigated to by swiping on the screen.

https://github.com/user-attachments/assets/3caedb62-4053-4244-8e68-e3dc9c71121d

### Reproduction

https://snack.expo.dev/x9MuliDzjthfDDLt1Hx1Q

### Platform

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [x] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 7.1.34|
| @react-navigation/stack                | 7.8.6|
| react-native-safe-area-context         |~5.6.0|
| react-native-gesture-handler           |~2.28.0|
| expo                                   |54.0.0|

helblinglilly:
I can't access those latest versions via the Snack yet, but can confirm that the issue can still be reproduced on those versions

lcsy1234:
I want to try fixing this bug, and recently I'm planning to submit a pull request.

matt-dalton:
Did this ever get resolved? We still seem to be getting something like this happening on our E2E tests

PR #13075, opened 2026-04-16
fix: preserve accessibility on android when header transparent is true

lcsy1234:
## Summary
The bug happens because when headerTransparent: true, the focused header is forced into absolute positioning, and on Android TalkBack this layer setup can make the back button and title hard to focus.
The fix removes this forced absolute positioning only for the focused header, so it stays in normal layout and remains accessible.
The change is limited to one condition in packages/stack/src/views/Header/HeaderContainer.tsx, with no API, gesture, routing, or render-structure changes.
So the impact is very small, while absolute positioning for non-focused headers and transition behavior is still preserved, keeping regression risk low.

## Related issue
https://github.com/react-navigation/react-navigation/issues/13043

### DIFF
--- a/packages/elements/src/Screen.tsx
+++ b/packages/elements/src/Screen.tsx
@@ -88,7 +88,12 @@ export function Screen(props: Props) {
 
               setHeaderHeight(height);
             }}
-            style={[styles.header, headerTransparent ? styles.absolute : null]}
+            style={[
+              styles.header,
+              headerTransparent
+                ? [styles.absolute, { height: headerHeight }]
+                : null,
+            ]}
           >
             {header}
           </View>
--- a/packages/elements/src/Screen.tsx
+++ b/packages/elements/src/Screen.tsx
@@ -91,7 +91,11 @@ export function Screen(props: Props) {
             style={[
               styles.header,
               headerTransparent
-                ? [styles.absolute, { height: headerHeight }]
+                ? [
+                    styles.absolute,
+                    // Specify an explicit min height for Android screen readers
+                    { minHeight: headerHeight },
+                  ]
                 : null,
             ]}
           >
--- a/packages/elements/src/Screen.tsx
+++ b/packages/elements/src/Screen.tsx
@@ -82,12 +82,6 @@ export function Screen(props: Props) {
       {headerShown ? (
         <NavigationProvider navigation={navigation} route={route}>
           <View
-            ref={headerRef}
-            onLayout={(e) => {
-              const { height } = e.nativeEvent.layout;
-
-              setHeaderHeight(height);
-            }}
             style={[
               styles.header,
               headerTransparent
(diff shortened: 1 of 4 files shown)


satya164 (changes requested):
Thanks for the PR, but the main purpose of `headerTransparent` is to make header transparent and absolute, so content can be shown underneath. Making it not absolutely positioned breaks its functionality.

lcsy1234:
> Thanks for the PR, but the main purpose of `headerTransparent` is to make header transparent and absolute, so content can be shown underneath. Making it not absolutely positioned breaks its functionality.

Thanks for your advice，I made some revisions again.

1. **Reason:** The header in headerTransparent mode is still positioned absolutely and overlaid on the content, but some wrapper layers do not have an explicit height that is stable and can participate in accessibility calculations. Android TalkBack may not be able to focus on the titles and buttons.

2. **Fixing solution:** While retaining the absolute positioning, add the onLayout-measured headerHeight to the transparent header container.
3. **Why three changes:** stack (CardContainer), elements (Screen), and native-stack (custom header) are three separate header mounting paths that do not share. Each needs to be supplemented to cover all areas.
4. **There is a similar approach in the repository:** CardStack.tsx explicitly sets the height on the absolute positioning container of the floating header (the comment states it is for Maestro and other accessible tools to read the title); HeaderContainer.tsx avoids accessibility tools not finding the control by "avoiding using absolute positioning for the focused floating header". This time, we continue the "ensuring visible area / accessible traversal" approach on headerTransparent and other mounting paths.

lcsy1234:
@satya164 I don't know why the iOS E2E test failed. It seems to have timed out. Do you know what I need to do to handle this? I can't find the retry button on my end. Thanks.

satya164:
@lcsy1234 the tests are flaky because of maestro. nothing can be done about it.

satya164 (changes requested):
Thanks for the updates. Let's do `minHeight` instead of `height`, and add a comment above the `minHeight: headerHeight` line with something like: "Specify an explicit height for screen readers on Android".

satya164:
FYI I'm currently OOO, so it'll take time to test and merge these changes.

lcsy1234:
> Thanks for the updates. Let's do `minHeight` instead of `height`, and add a comment above the `minHeight: headerHeight` line with something like: "Specify an explicit height for screen readers on Android".

Thank you for reviewing. I have fixed it according to your instructions.

lcsy1234:
> FYI I'm currently OOO, so it'll take time to test and merge these changes.

okok

satya164 (approved):
Thank you

Bardiamist:
This change breaks layout for modal screens with `headerTransparent: true` in `@react-navigation/stack` 7.8.11.

In `CardContainer.tsx` the header style now becomes `[styles.header, { minHeight: headerHeight }]` when `headerTransparent` is true. Previously `styles.header` was just `{ zIndex: 1 }`, so the transparent header didn't take up space in the flex layout and the scene rendered underneath it — which is the whole point of `headerTransparent`.

With the new `minHeight`, the header now occupies its full height in the layout and pushes the scene content down by exactly that amount. So the screen content ends up below the transparent header instead of behind it.

Reproduces both for regular screens and modal presentations. Reverting to 7.8.10 restores the correct behavior. Looks like the same regression reported in #13096.

@lcsy1234 @satya164

PR #13084, opened 2026-04-25
feat: phase 2 init

osdnk:
https://github.com/user-attachments/assets/4cdcfe6a-a22a-49f5-ac5e-d95a4078d80d


packages/core/src/useNavigationBuilder.tsx
```diff
@@ -876,6 +878,32 @@ export function useNavigationBuilder<
...
+      state
+    );
+    if (!composedLoader) return;
+    staticContext.abortControllerRef.current?.abort();
+    const controller = new AbortController();
+    staticContext.abortControllerRef.current = controller;
+    composedLoader(controller.signal).catch(() => {});
+    // Only the initial state
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [staticContext]);
```
satya164: What's this needed for? Initial mount?

For mount, the screen itself should load the data. We should only call loaders on navigation actions.

This means when using React Query, the `useSuspenseQuery` will trigger the data fetching.

packages/core/src/useOnAction.tsx
```diff
@@ -102,8 +106,21 @@ export function useOnAction<State extends NavigationState>({
...
+              : undefined;
+
+            if (loader && staticContext) {
+              staticContext.abortControllerRef.current?.abort();
+              const controller = new AbortController();
+              staticContext.abortControllerRef.current = controller;
+              loader(controller.signal).catch(() => {});
+              React.startTransition(() => {
+                setState(result);
+              });
```
satya164: I'm unsure if we should always abort other data fetches if a new one begins. For Web, it makes sense as a single screen is rendered at one point. For native, not sure.

If I push one screen, and then the next before the previous one finishes, if we abort the previous screen's loader, then when you go back, it'll be in error state, which is not ideal behavior.

Maybe aborting makes sense if data isn't fetched yet and the screen gets unmounted. Or maybe we shouldn't bother with aborting at all.

In any case, if we keep abort behavior, we should clear the abort controller in `finally` for loader, so we don't unnecessarily call abort signal listeners.

satya164: The `startTransition` may mess with native state updates. For now lets move it to callsite that does `navigation.navigate`. And create a ticket to review how to integrate well. I'm thinking we can integrate it in `useLinkProps` by default.

packages/core/src/StaticNavigation.tsx
```diff
@@ -680,10 +682,33 @@ export function createComponentForStaticNavigation<
...
+    const abortControllerRef = React.useRef<AbortController | null>(null);
+    const isOutermost = parent == null;
+
+    React.useEffect(() => {
+      if (!isOutermost) return;
+      return () => {
+        abortControllerRef.current?.abort();
+        abortControllerRef.current = null;
+      };
+    }, [isOutermost]);
```
satya164: Same as the other abort related comment. Maybe we don't need it. I'd put it for investigation as another step.

example/src/Screens/LoaderDemo.tsx
```diff
@@ -2,134 +2,142 @@ import { Button, Text } from '@react-navigation/elements';
...
+    </View>
   );
 }
 
 const LoaderStack = createNativeStackNavigator({
+  layout: ({ children }) => (
+    <ErrorBoundary
+      onReset={() => {
+        config.shouldFail = false;
+        queryClient.clear();
```
satya164: Not sure it makes sense to reset the whole query cache when error occurs.

For `ErrorBoundary`, I'd also wrap the screen (in `screenLayout`) instead of the navigator, similar to suspense. When error happens, for retry, the user should just call `refetchQuery` instead of resetting.

example/src/Screens/LoaderDemo.tsx
```diff
@@ -2,134 +2,142 @@ import { Button, Text } from '@react-navigation/elements';
...
+      UNSTABLE_loader: ({ params, signal }) => {
+        const delay = getDelay(params);
+        signal.addEventListener('abort', () => {
+          queryClient.cancelQueries({ queryKey: ['detail-data', delay] });
+        });
+        return queryClient
+          .ensureQueryData(detailQuery(delay))
+          .then(() => undefined);
       },
     },
```
satya164: Use `createNativeStackScreen` for the screen and use a pattern for `linking` so `params` has the proper type, instead of a `getDelay` helper.

example/src/Screens/LoaderDemo.tsx
```diff
@@ -2,134 +2,142 @@ import { Button, Text } from '@react-navigation/elements';
...
+
   return (
     <View style={styles.content}>
       <Text style={styles.heading}>Loader Demo</Text>
       <Text style={styles.description}>
-        Data can be prefetched for a route before navigating to it. This avoids
-        showing Suspense fallbacks if not desired.
+        UNSTABLE_loader runs before the screen mounts. The screen reads via
+        useSuspenseQuery, so the same fetch is shared between the loader and the
+        screen through TanStack Query's cache.
```
satya164: Codex loves putting description of example inside the example. Lets keep it more realistic.

example/src/Screens/LoaderDemo.tsx
```diff
@@ -2,134 +2,142 @@ import { Button, Text } from '@react-navigation/elements';
...
+    }),
+});
 
-  return cache.get(key)?.data ?? '';
-}
+const getDelay = (params: unknown): number =>
+  (params as { delay?: number } | undefined)?.delay ?? 1000;
 
 function HomeScreen() {
+  const navigation = useNavigation<typeof LoaderStack>();
```
satya164: Use the name `LoaderHome` for the home screen and then:

```suggestion
  const navigation = useNavigation('LoaderHome');
```

example/src/Screens/LoaderDemo.tsx
```diff
@@ -2,134 +2,142 @@ import { Button, Text } from '@react-navigation/elements';
...
+      <Text style={styles.description}>
+        Loading state was {wasLoadingVisible ? 'visible' : 'not visible'}
+      </Text>
     </View>
   );
 }
 
-function DetailScreen() {
+function LoadingFallback() {
+  debug.wasLoadingVisible = true;
```
satya164: This is a bit weird. We shouldn't be mutating things that are rendered during render, even if for the example. The closer we keep the example to real-world usage, the closer our testing is to identify gaps in DX as well as implementation issues.

satya164 (commented):
I'd also put a nested navigator (probably native tabs) in the example, so we test with more realistic situation.

PR #13085, opened 2026-04-28
fix(bottom-tabs): disable touch events on unfocused tab screens

fabriziocucci:
## Summary

Disable touch events on unfocused tab screens in the custom (JS) bottom tab view.

## Problem

When using bottom tabs, unfocused tab screens are kept alive for performance (via `ActivityView`). However, these inactive screens can still intercept touch events, causing "ghost taps" — where a user taps on content in the focused tab but the tap is received by an element in an unfocused tab behind it.

## Precedent

The same fix was applied to expo-router's native tabs implementation in [expo/expo#44778](https://github.com/expo/expo/pull/44778).

## Fix

Add `pointerEvents={isFocused ? 'box-none' : 'none'}` to the `ActivityView` wrapping each tab screen in `BottomTabViewCustom.tsx`.

- `'box-none'` (focused): the container doesn't block touches, children receive them normally
- `'none'` (unfocused): all touch interaction is disabled on the inactive tab's view tree

This is a 1-line change affecting only `BottomTabViewCustom.tsx`. The native implementation (`BottomTabViewNativeImpl.tsx`) is not affected as touch handling is managed natively there.

## Test Plan

- [x] Taps on focused tab content work as expected
- [x] Taps no longer pass through to unfocused tab content behind the focused tab
- [x] Tab switching continues to work correctly

### DIFF
--- a/packages/bottom-tabs/src/views/BottomTabViewCustom.tsx
+++ b/packages/bottom-tabs/src/views/BottomTabViewCustom.tsx
@@ -323,6 +323,7 @@ export function BottomTabViewCustom({
               mode={isFocused ? 'normal' : isActive ? 'inert' : 'paused'}
               visible={isFocused || isAnimatingRoute}
               style={{ ...StyleSheet.absoluteFill, zIndex: isFocused ? 0 : -1 }}
+              pointerEvents={isFocused ? 'box-none' : 'none'}
             >
               {content}
             </ActivityView>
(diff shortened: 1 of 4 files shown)


cortinico:
see also https://github.com/react-navigation/react-navigation/pull/13086 I can drop mine

fabriziocucci:
Closing this PR — the fix is already present on `main` (v8) via the `inert` mechanism introduced in #12998 (Feb 2026). The `Container` component already sets `pointerEvents: 'none'` when `inert={true}`, which covers all unfocused tab screens.

Will open a new PR targeting the v7 branch instead, where this fix is still needed.

cortinico:
> Will open a new PR targeting the v7 branch instead, where this fix is still needed.

@fabriziocucci yes that's what my PR is doing #13086

PR #13086, opened 2026-04-28
fix(bottom-tabs): disable pointer events on inactive tab screens

cortinico:
**Motivation**

Inactive tab screens are rendered as full-screen overlays with `StyleSheet.absoluteFill`. Without explicit `pointerEvents` handling, these invisible screens intercept touch events meant for the focused tab, blocking user interaction.

Set `pointerEvents` to `'none'` on unfocused screens so touches pass through, and `'box-none'` on the focused screen so the container itself doesn't block events but its children remain interactive.

cc @jorge-cab

**Test plan**

Essentially the same patch as:
- https://github.com/expo/expo/pull/44778


cortinico:
cc @satya164 if you could look into this one 🙏

satya164 (approved):
thank you

PR #13093, opened 2026-05-04
feat: treat preloaded screens similar to regular screens

satya164:
- receive events from parent navigator
- allow param updates
- allow navigation in nested navigators
- include nested navigator state in getRootState
- don't treat screen as being removed for `preventRemove`


satya164:
Superseded by #13094

Issue #12755, opened 2025-09-15
Using `animation` in BottomTabs, sometimes renders blank screen


Ubax:
### Current behavior

Originally reported in https://github.com/expo/expo/issues/39514

https://github.com/user-attachments/assets/221da870-08a4-4ca3-943e-919b84475700

I suspect the issue is related to [code here](https://github.com/react-navigation/react-navigation/blob/main/packages/bottom-tabs/src/views/BottomTabView.tsx#L304) (setting the activity state using `Animated`) and integration with `react-native-screens`. I can only reproduce this when `detachInactiveScreens` is `true`.

Note that the issue is only happening when tabs are changed rapidly.

### Expected behavior

The animation shouldn't cause screen to be blank

### Reproduction

https://snack.expo.dev/@jtkacz/moody-yellow-graham-crackers

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | 7.1.17 |
| @react-navigation/bottom-tabs          | 7.4.7 |
| react-native-screens                   | 4.16.0 |
| react-native                           | 0.81.4
| expo                                   | 54

yonitou:
+1, having the exact same issue. I'm pretty sure it happened when upgrading from expo 53 to 54 which implicitly means, when upgrading `react-native-reanimated` from v3 to v4 and `react-native` from v0.79 to v0.81

eugene-zfort:
Faced this issue during the upgrade to RN 0.80. Still presented on 0.81. Setting `detachInactiveScreens` to `false` resolves the issue

DvzZDev:
I have the same issue after upgrading to Expo 54.
Adding detachInactiveScreens={false} apparently fix it.

HAizelf:
+1
Setting detachInactiveScreens to false resolves the issue

duylun:
> +1 Setting detachInactiveScreens to false resolves the issue

[...]

PR #13097, opened 2026-05-06
fix: Disable native driver for iOS and web on BottomTabViewCustom.tsx

steve228uk:
**Motivation**

This PR fixes the the longstanding issue reported in #12755

**What Changed**

The issue stems from `useNativeDriver` being set to true in `packages/bottom-tabs/src/views/BottomTabViewCustom.tsx` for all platforms other than web.

When a detached tab with `activityState=0` becomes focussed, React tries to set `activityState=2` but there's a race between the native animation thread (holding `0`) and the JS update (`2`). The native thread wins -> screen never re-attaches -> blank rendered.

With `useNativeDriver` set to false, the animation runs on the JS thread which means there's no cross-thread conflict and the JS side correctly transitions the prop. `react-native-screens` re-attaches the screen, and the animation completes as expected.

**Test plan**

Check the issue reported in #12755 and quickly navigate between tabs with `detachInactiveScreens={true}` and the `shift` animation set. You should notice that on occasion the native thread wins the race and a blank screen will be rendered.

Then, test it against this PR and you should see the issue is resolved.

**Patch**

```
diff --git a/node_modules/@react-navigation/bottom-tabs/lib/module/views/BottomTabView.js b/node_modules/@react-navigation/bottom-tabs/lib/module/views/BottomTabView.js
index 39d48b8..3e965ef 100644
--- a/node_modules/@react-navigation/bottom-tabs/lib/module/views/BottomTabView.js
+++ b/node_modules/@react-navigation/bottom-tabs/lib/module/views/BottomTabView.js
@@ -29,7 +29,7 @@ const NAMED_TRANSITIONS_PRESETS = {
     }
   }
 };
-const useNativeDriver = Platform.OS !== 'web';
+const useNativeDriver = false;
 const hasAnimation = options => {
   const {
     animation,
```


packages/bottom-tabs/src/views/BottomTabViewCustom.tsx
```diff
@@ -56,7 +55,7 @@ const NAMED_TRANSITIONS_PRESETS = {
   },
 } as const;
 
-const useNativeDriver = Platform.OS !== 'web';
+const useNativeDriver = false;
```
kaloudis: since the issue only affects iOS, I believe you can set this to `Platform.OS === 'android'`

kaloudis (changes requested):

packages/bottom-tabs/src/views/BottomTabViewCustom.tsx
steve228uk: Ah, thought this was affecting Android too. Changed :)

kaloudis (approved):

PR #13099, opened 2026-05-08
fix: 🐛 prevent scroll jitter when CardContent switches from page to card styles

angelalagao:
**Motivation**

This PR fixes a mobile web scrolling issue caused by `CardContent` switching from page styles to card styles during transient layout changes.

On mobile web (notably iOS Safari), rapid layout updates during scroll can temporarily report smaller layout heights. In more complex stack screens (for example, long scrollable content with sticky/fixed bottom actions), this can cause:

- visible layout jitter/reflow
- sticky footer shake
- occasional scroll position reset

In `CardContent`, page style keeps the document scrollable (`minHeight: 100%`), while card style applies `overflow: hidden`. Unexpected switches to card style during scroll can cause the jitter/jump behaviour above.

This change stabilizes that behaviour by keeping `CardContent` in page style on touch devices when the width remains unchanged, while preserving strict exact-match behaviour for non-touch environments. 

**Test plan**

1. Run the example app on web (iOS safari) and view `Stack -> Card + Modal` 
2. There should be no regressions

Unfortunately, I wasn’t able to reproduce this in the example app. The issue appears to require a more complex page composition than the example setup. But I've validated the Stack Card example and found no regressions.
I've added screenshots below.

**Screenshots**

**Before**

https://github.com/user-attachments/assets/cf679b39-dc63-4e38-91ae-b2946cdf540c

**After**

https://github.com/user-attachments/assets/670e4848-bf86-4d56-b126-c1459f9753d1

### DIFF
--- a/packages/stack/src/views/Stack/CardContent.tsx
+++ b/packages/stack/src/views/Stack/CardContent.tsx
@@ -78,7 +78,7 @@ export function CardContent({ enabled, layout, style, ...rest }: Props) {
 
       if (isTouchDevice) {
         // Keep page mode stable on touch devices while width remains unchanged to
-        // avoid transient re-layout flips to `overflow:hidden card mode.
+        // avoid transient re-layout flips to `overflow:hidden` card mode.
         return isSameWidth;
       }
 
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
Thanks for the PR. With this change, looks like the height will keep increasing based on screen sizes and never decrease back for smaller screens.

The main problem sounds like is why the height is changing during scroll? Shouldn't this be based on the content size itself, which should change when scrolling?

angelalagao:
@satya164 Thanks for the feedback 🙏 I've updated the solution to be for the content size

PR #13101, opened 2026-05-10
feat: update native tabs to use the new API in react-native-screens

satya164:
**User-facing changes**

- The minimum version required for `react-native-screens` is now `4.25.0`
- The `scrollEdgeEffects` option has been removed. Users need to use `ScrollViewMarker` from `react-native-screens` instead.

**Internal**

This updates the state management for native tabs to use the new provenance model from screens for more reliable state changes.


packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -241,68 +393,23 @@ export function BottomTabViewNative({
         ? tabBarElement
         : null}
       <Tabs.Host
+        navStateRequest={{
+          selectedScreenKey: focusedRouteKey,
+          baseProvenance: nativeState.confirmed.provenance,
+        }}
+        rejectStaleNavStateUpdates
```
kkafar, line 407: First, let me note that you opt into stale navigation updates rejection mechanism. 

This is completely fine & ok, I just want to make sure it is intentional, because right now the rejected updates are just ignored & the information on failed navigation is not propagated to the user. I'll discuss this topic more in next note. 

Alternative here is to disable the mechanism. Every update will go through then.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -99,21 +160,112 @@ export function BottomTabViewNative({
...
+      dispatch({
+        type: 'CONFIRM',
+        confirmed,
+      });
+
+      return;
+    }
+
+    navigate(route, confirmed);
+  };
```
kkafar, line 274: Okay, so here you consume the information that the navigation attempt has been rejected and simply trigger state update to the most current one. 

This is fine, however we could consider here exposing the information to the user, so that the programmer e.g. can retry the navigation if that's needed.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -99,21 +160,112 @@ export function BottomTabViewNative({
...
+    if (!route || focusedRouteKey === route.key) {
+      dispatch({
+        type: 'CONFIRM',
+        confirmed,
+      });
+
+      return;
+    }
+
+    navigate(route, confirmed);
```
kkafar, line 273: I haven't run the code, just reading through it & I don't understand the call to `navigate` here. 

From what I can see in the code above, `navigate` dispatches `SET_PENDING` to the reducer and `Navigate` to the router. It looks like you want to repeat the operation until it succeeds, no matter other conditions. If that's the desired effect, maybe you should consider disabling `rejectStaleNavStateUpdates`?

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -99,21 +160,112 @@ export function BottomTabViewNative({
...
+    const route = state.routes.find((route) => route.key === selectedScreenKey);
+
+    if (!route) {
+      dispatch({
+        type: 'CONFIRM',
+        confirmed,
+      });
+
+      return;
+    }
```
kkafar, line 214: If you fail to find a route for a returned key, that's rather a hard error, right? It's an attempt to navigate to a route that does not exist. Why dispatching `CONFIRM` then? 

Could you elaborate?

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -99,21 +160,112 @@ export function BottomTabViewNative({
...
+
+    navigate(route, confirmed);
+  };
```
kkafar, line 240: The current code ignores the possibility of `actionOrigin === 'implicit'`, which can be triggered on iPads due to more navigation controller & app resize interaction. IIRC such update will be lost from reducer state right now.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -99,21 +160,112 @@ export function BottomTabViewNative({
...
+        from: previousRouteKeyRef.current,
+        to: route.key,
+      },
+    });
+
+    navigation.dispatch({
+      ...CommonActions.navigate(route.name, route.params),
+      target: state.key,
+    });
+  };
```
kkafar, line 193: I think that `SET_PENDING` should be dispatched only on js-side-request. There is no need to update `navStateRequest` on `Tabs.Host` on every action. With some unfortunate timings it might lead to unexpected navigation actions.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
kkafar: I read through the code again & it seems you don't. `pending` state is not translated directly to the request here, it's only used for some decisions regarding content rendering later. I don't understand it's semantics clearly.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
kkafar: From second read through it seems that this is never called on js navigation right? You trigger router update directly and then this components gets rendered with a new state & new `focusedRouteKey` which later determines the `navStateUpdateRequest` for `Tabs.Host`.

kkafar (commented):
This mostly looks fine, beside the code lacking memoization & referential stability for the callbacks etc. (unless the react-compiler is in play). 

I want to bring your attention to a couple of places in the code I either don't quite understand or have some concerns for 👇🏻

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: @kkafar it's only for controlling when we apply `Activity mode="hidden"`, the naming could be clearer.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: @kkafar i think it's an edge case, for example there was a re-render that changed tabs before we received an event. tho we should probably print an error.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: @kkafar we early return for `actionOrigin === 'programmatic-js'` but always call `navigate` for other cases, so `'implicit'` should also be handled right?

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: > you trigger router update directly and then this components gets rendered with a new state & new focusedRouteKey which later determines the navStateUpdateRequest for Tabs.Host.

yup. the `pending` state is only used for `Activity`, not for navigation state. I'll rename it so it makes more sense.

satya164: renamed it to `lastTransition` as it just tracks the last navigation transition that happened

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: @kkafar removed `CONFIRMED` and added `console.error`

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: do you think it'll be a common enough case where users may want to handle it?

kkafar: We obviously have neither feedback nor usage data here. I can only guess. At screens level I've decided to expose it, because we provide lower level API & I didn't want to tie anybody's hands unnecessarily. On react-navigation level - I guess both approached might be fine. We can always settle with approach that if it will be needed, we'll hear appropriate feedback.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
kkafar: Yeah, it mislead me first time I read. It's much better now. Thanks.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
satya164: @kkafar gotcha. so i'll wait before exposing it

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
kkafar: Yes. My comment was based on misunderstood `SET_PENDING` semantics. Now I see that you'll update `confirmed` state via call to `navigate` and dispatch of `TRACK_TRANSITION`. Seems fine.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
kkafar: Reads better now. Thanks.

kkafar (approved):
The provenance-related part of the code looks ✅ 

Appearance related code looks also fine, but I've only skimmed. Nothing raising my suspicions there.

PR #13103, opened 2026-05-11
feat(iOS): implement unstable_toolbarItems for stack navigator

johankasperi:
**Motivation**

Adds an api to the native stack navigator that adds support for the suggested toolbar items API introduced in https://github.com/software-mansion/react-native-screens/pull/4013

**Test plan**

Run the example app in react-native-screens and open the "Toolbar" example to test it


johankasperi:
Will open a PR in this repo instead after https://github.com/software-mansion/react-native-screens/pull/4013 has landed

PR #13105, opened 2026-05-13
fix(core): throw on conflicting router path patterns

### LINKED ISSUE #7653: Throw error when providing conflicting paths when configuring a router

vivekjm:
## Summary
- add a conflict check in `createPathParser` to fail fast when 2 routes resolve to the same exact non-wildcard matcher
- keep existing `null`/empty-string wildcard behavior unchanged
- add regression tests for both route-config and `paths`-override conflict cases

Closes #7653

## Test Plan
- `yarn test packages/core/src/routers/__tests__/PathHandling.test.js`
- pre-commit hook suite (auto-ran on commit): `eslint`, `tsc --noEmit --composite false`, full `jest`

## Notes
- this change only rejects ambiguous exact path patterns (including param-name aliases like `user/:id` vs `user/:userId`) to avoid unpredictable route selection.

### DIFF
--- a/packages/core/src/routers/__tests__/PathHandling.test.js
+++ b/packages/core/src/routers/__tests__/PathHandling.test.js
@@ -554,6 +554,48 @@ const performRouterTest = (createTestRouter) => {
     expect(action.type).toEqual(NavigationActions.NAVIGATE);
     expect(action.routeName).toEqual('baz');
   });
+
+  it('throws when route configs resolve to conflicting path patterns', () => {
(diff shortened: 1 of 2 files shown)


vivekjm:
Updated this PR branch to use a commit authored by @vivekjm and aligned project-level git identity/auth in my local automation workspace.

satya164:
thanks for the PR. but react navigation 4 is a deprecated version and not supported. so i'll close the pr.

PR #13110, opened 2026-05-14
fix(tab-view): forward pageMargin to adapters

### LINKED ISSUE #13065: pageMargin prop not forwarded to Pager in TabView
https://github.com/user-attachments/assets/322387a0-af48-4c23-8b28-421f0ee227bd

### Current behavior

pageMargin is accepted in TabView's type definition via PagerProps (which extends PagerViewProps), but it's never actually passed through to the underlying pager component. This means setting pageMargin on TabView has no effect on any platform.

Two issues exist:

TabView.tsx destructures specific props but doesn't include pageMargin, so it's never forwarded to <Pager>.
PanResponderAdapter.tsx (the JS-based pager used on web) doesn't implement pageMargin at all — it ignores it even if received.
Note: PagerViewAdapter.tsx (native pager, used on iOS/Android) already spreads ...rest onto the native PagerView, so it would work correctly if it received the prop.

### Expected behavior

Setting pageMargin={16} on TabView should create a 16px gap between pages on all platforms.
LIKE THE iOS GAP BETWEEN TABS ON INSTAGRAM

### Reproduction

no need

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [x] react-native-tab-view 4.3.0

### Environment

REPRODUCTION:

<TabView
  pageMargin={16}
  // ... other props
/>
// No gap appears between pages

Versions:
react-native-tab-view: 4.3.0
react-native: 0.79

vivekjm:
## Summary
- thread `pageMargin` through `TabView` into adapter props
- pass `pageMargin` to native pager adapter so `react-native-pager-view` receives it
- add JS fallback support in `PanResponderAdapter` by applying inter-page spacing and matching translate/page width math
- add a regression test to verify `TabView` forwards `pageMargin` to the adapter

Closes #13065

## Testing
- `yarn test packages/react-native-tab-view/src/__tests__/index.test.tsx`
- pre-commit hook (auto-ran): `test`, `types`, `lint`

### DIFF
--- a/packages/react-native-tab-view/src/PagerViewAdapter.native.tsx
+++ b/packages/react-native-tab-view/src/PagerViewAdapter.native.tsx
@@ -21,7 +21,9 @@ export type PagerViewAdapterProps = AdapterProps &
     | 'onPageSelected'
     | 'onPageScrollStateChanged'
     | 'children'
-  >;
+  > & {
+    pageMargin?: PagerViewProps['pageMargin'] | undefined;
+  };
 
 const useNativeDriver = Platform.OS !== 'web';
 
@@ -36,6 +38,7 @@ export function PagerViewAdapter({
   children,
   style,
   animationEnabled,
+  pageMargin,
   ...rest
 }: PagerViewAdapterProps) {
   const { index } = navigationState;
@@ -160,6 +163,7 @@ export function PagerViewAdapter({
         }}
         onPageScrollStateChanged={onPageScrollStateChanged}
         scrollEnabled={swipeEnabled}
+        pageMargin={pageMargin}
       >
         {children}
       </AnimatedViewPager>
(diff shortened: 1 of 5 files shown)


satya164:
thanks for the pr.

`pageMargin` is not valid on all supported adapters, and it's not possible to have proper type-safety when adding adapter-specific props to `TabView`. it also doesn't make sense to add only `pagerMargin` to `TabView` props when non of the other adapter specific props are added.

the recommended approach is to use `renderAdapter` to pass the adapter-specific props. you can use `DefaultAdapter` to keep default behavior https://reactnavigation.org/docs/8.x/tab-view#renderadapter

PR #13114, opened 2026-05-20
feat: add prop for disabling default top inset application

t0maboro:
**Motivation**

In `react-native-screens`, we refined the approach for applying internal padding to headers due to issues with padding being applied multiple times to nested stacks. Starting from 4.25, we're exposing a new prop to prevent that behavior.  When the padding is intended to offset the header from the status bar for the outermost stack, it doesn't apply when the outermost stack isn't associated with the outermost header (e.g. stack nested in tabs).

For additional context, please see:
- https://github.com/software-mansion/react-native-screens/pull/3793
- https://github.com/software-mansion/react-native-screens/pull/3835

**Test plan**

Tested on the `react-native-screens` side


satya164 (approved):
Thank you

PR #13121, opened 2026-05-26
fix(core): reset lastNotifiedStateRef on cleanup for StrictMode compatibility

### LINKED ISSUE #13025: Deep links dropped in StrictMode when navigator mount is deferred (since core@7.14.1)
### Current behavior

Deep links are silently ignored when using React's `StrictMode` with a deferred navigator mount (e.g. returning `null` while loading auth/session state before rendering the navigator). The user lands on the initial route instead of the deep link target.

### Expected behavior

Deep links should resolve correctly regardless of StrictMode and deferred mounting.

### Reproduction

https://github.com/NGMarmaduke/deep-link-repro

```bash
git clone https://github.com/NGMarmaduke/deep-link-repro
cd deep-link-repro
yarn install
yarn web
```

Navigate to http://localhost:8081/profile — lands on Home instead of Profile.

### Platform

- [x] Android
- [x] iOS
- [x] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| package                            | version |                                                                                                                                                                                                    
| ---------------------------------- | ------- |                                                                                                                                                                                                      
| @react-navigation/core             | 7.14.1  |                                                                                                                                                                                                      
| @react-navigation/native           | 7.1.33  |                                                                                                                                                                                                      
| @react-navigation/native-stack     | 7.9.0   |                                                                                                                                                                                                      
| react-native-screens               | 4.16.0  |                                                                                                                                                                                                    
| react-native-safe-area-context     | 5.6.2   |                                                                                                                                                                                                      
| react-native                       | 0.81.5  |                                                                                                                                                                                                      
| expo                               | 54.0.33 |
| node                               | 22.12.0 |                                                                                                                                                                                                      
| yarn                               | 4.12.0  |

## Root cause

The `lastNotifiedStateRef` guard added in #12985 (`useNavigationBuilder.tsx` line 749) prevents `setState` from being called on StrictMode's second effect invocation:

```js
if (!getIsInitial() && lastNotifiedStateRef.current !== state) {
  setState(state);
  lastNotifiedStateRef.current = state;
}
```

In StrictMode, React double-invokes effects: mount → cleanup → remount. The `state` object is the same reference across both invocations. On the first mount, `setState(state)` propagates the deep link state to the container and saves the reference. The cleanup then calls `setCurrentState(undefined)`, wiping the container's state. On the second mount, the guard sees `lastNotifiedStateRef.current === state` (same reference) and skips `setState`. The container store stays `undefined`, and on re-render the navigator creates a fresh default state — losing the deep link.

This only affects deferred navigator mounts because:
- `getIsInitial()` must be `false` for the effect to call `setState` at all
- StrictMode must double-invoke the effect, which only happens when the navigator appears as a new component due to a state change within the StrictMode boundary

## Workaround

Pin `@react-navigation/core` to 7.14.0:

```json
{
  "resolutions": {
    "@react-navigation/core": "7.14.0"
  }
}
```

Jasleen-Kaur96:
## Summary

Fixes deep links being silently dropped when using React's StrictMode with a deferred navigator mount (e.g. behind an auth/loading gate).

## Root cause

The `lastNotifiedStateRef` guard in `useNavigationBuilder.tsx` (introduced in #12985) prevents `setState` from re-propagating state on StrictMode's second effect invocation. The cleanup wipes the container state via `setCurrentState(undefined)`, but `lastNotifiedStateRef` still holds the same reference — so the guard skips `setState` on remount, leaving the container with `undefined` state and losing the deep link.

## Fix

Reset `lastNotifiedStateRef.current = null` in the effect cleanup. This allows StrictMode's second mount to correctly re-propagate state.

This does not affect the `<Activity mode="hidden">` loop prevention since the effect has `[]` deps and only runs once per mount cycle.

## Test plan

- Verified with reproduction repo: https://github.com/NGMarmaduke/deep-link-repro
- Before fix: navigating to `/profile` lands on Home
- After fix: navigating to `/profile` correctly lands on Profile

Fixes #13025

### DIFF
--- a/packages/core/src/__tests__/index.test.tsx
+++ b/packages/core/src/__tests__/index.test.tsx
@@ -616,6 +616,51 @@ test('cleans up state when the navigator unmounts', () => {
   expect(onStateChange).toHaveBeenLastCalledWith(undefined);
 });
 
+test('preserves initial state when navigator mount is delayed in StrictMode', async () => {
+  const TestNavigator = (props: any): any => {
+    const { state, descriptors, NavigationContent } = useNavigationBuilder(
+      MockRouter,
+      props
+    );
+
+    return (
+      <NavigationContent>
+        {descriptors[state.routes[state.index].key].render()}
+      </NavigationContent>
+    );
+  };
+
+  const TestScreen = ({ route }: any): any => `[${route.name}]`;
+
+  const navigation = createNavigationContainerRef<ParamListBase>();
+
+  const Test = ({ condition }: { condition: boolean }) => (
+    <React.StrictMode>
+      <BaseNavigationContainer
+        ref={navigation}
+        initialState={{
+          index: 1,
+          routes: [{ name: 'foo' }, { name: 'bar' }],
+        }}
+      >
+        {condition ? (
+          <TestNavigator>
+            <Screen name="foo" component={TestScreen} />
+            <Screen name="bar" component={TestScreen} />
+          </TestNavigator>
+        ) : null}
+      </BaseNavigationContainer>
+    </React.StrictMode>
+  );
+
+  const root = await renderAsync(<Test condition={false} />);
+
+  await root.rerenderAsync(<Test condition />);
+
+  expect(root).toMatchInlineSnapshot(`"[bar]"`);
+  expect(navigation.getCurrentRoute()?.name).toBe('bar');
+});
+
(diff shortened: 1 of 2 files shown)


satya164 (approved):
thanks for the PR

Issue #12512, opened 2025-03-26
Bottom tab navigator popToTopOnBlur actually pops to top on refocus instead of on blur


laurens-lamberts:
### Current behavior

I use a bottom tab navigator.
One of my tabs contains a stack of it's own.
Whenever the user leaves that tab, and returns to it, I'd like the stack to be back at it's root (reset).

However, when I;
1.  apply popToTopOnBlur to that certain screen in my bottom tab navigator
2. Open a screen inside the stack within the tab
3. Go to another tab
4. Return to the tab where I opened the screen

...I find that the stack inside the tab is actually reset whenever I come back to it. I see that happening in an animation.

Therefore, I'd say that the current implementation does not actually pop to top on blur, but on **refocus**.

### Expected behavior

Is there currently a way to have the stack reset whenever I leave the tab, or otherwise disable the animation for this popToTop action? Currently the user would return to the tab after a while and suddenly sees an animation happening, which I think is undesirable.

All the best,
Laurens

### Reproduction

Will be added soon

### Platform

- [ ] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [x] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package                                | version |
| -------------------------------------- | ------- |
| @react-navigation/native               | ^7.0.15
| @react-navigation/bottom-tabs          | ^7.2.1
| @react-navigation/material-top-tabs    | ^7.1.1
| @react-navigation/native-stack         | ^7.2.1
| react-native-tab-view                  | ^4.0.5
| react-native-screens                   | ~4.4.0
| react-native-safe-area-context         | 4.12.0
| react-native-gesture-handler           | ~2.20.2
| react-native-reanimated                | ~3.16.7
| react-native-pager-view                | 6.5.1
| react-native                           | 0.76.7
| expo                                   | ^52.0.37
| node                                   | 20.18.0
| npm or yarn                            | npm

laurens-lamberts:
Snack added and tested with latest versions (same behavior)

laurens-lamberts:
> Couldn't find version numbers for the following packages in the issue:
> 
> * `react-native-drawer-layout`
> 
[...]

PR #13123, opened 2026-05-26
fix(bottom-tabs): dispatch popToTopOnBlur even when tab transition is interrupted

brojor:
**Motivation**

Related to #12512.

`popToTopOnBlur: true` on a Bottom Tab sometimes leaves the nested stack at its previous deep route after switching tabs — the pop is silently dropped, not deferred. `BottomTabViewCustom` dispatches `popToTopAction`
from `Animated.parallel(...).start(cb)` gated on `finished`. If the effect re-runs before `cb` fires (any parent re-render that churns `state.routes` / `descriptors`), a second `Animated.parallel` preempts the first, the first callback gets `finished: false`, and the captured action is dropped.

The fix removes the `finished` gate. The sibling `transitionEnd` emit inside the same callback already runs unconditionally, and `BottomTabViewNativeImpl` already dispatches `popToTopAction` synchronously with no animation gate — so this brings Custom in line with Native. Semantically the option is "pop on blur", not "pop on
animation complete".

Not using `Closes` for #12512 because that issue describes a slightly different visual symptom (pop animating on refocus rather than the dispatch being dropped). This PR may or may not fully resolve it — worth leaving the issue open for the maintainer to triage after merge.

**Test plan**

Regression test added: `Animated.parallel` is mocked so `start()`'s callback fires with `finished: false`, deterministically simulating the race. Without the fix the test fails (Tab A's stack stays `['Root', 'Detail']`); with the fix it passes (`['Root']`).

Interactive Snack:
https://snack.expo.dev/@brojor/poptotoponblur-race-condition-repro

The Snack uses `screenListeners.state` to force a parent re-render on every navigation state change — that's what real-world context churn does in larger apps and is required to trigger the race in a minimal example.

Steps in the Snack:

1. Tap "Open Detail" — nested stack is `[HomeRoot, Detail]`.
2. Tap "Other" tab → tap back to "Home".

Without the fix: Detail still shown.  With the fix: HomeRoot every time.

`yarn lint`, `yarn typecheck`, `yarn test` (49 suites, 816 tests) all pass.


satya164 (approved):
thanks

PR #13131, opened 2026-06-02
feat: add support for shared URLs in linking config

satya164:
This adds support for shared URLs with a `shared` property.

For static config, this is automatic - if the same screen (compared by reference) appears in multiple navigators with the same path pattern, it'll be marked as shared automatically.

So the screen keeps one URL, even when it lives in different navigators.

This is useful when the same screen is used in multiple navigators, e.g., tabs with stacks containing a profile screen in each stack.

https://github.com/user-attachments/assets/64c3b280-e6db-4966-ab6f-437441b91d91


Bram-dc:
This looks great!

Issue #13144, opened 2026-06-09
useNavigation('Screen') resolves to root navigator (depth-limit fallback) in @react-navigation/core@8.0.0-alpha.20 — regression from alpha.19


Bram-dc:
### Current Behavior

In `@react-navigation/core@8.0.0-alpha.20`, `useNavigation('SomeScreen')` no longer resolves to the navigator that actually contains `SomeScreen` once the navigation tree is moderately large. It falls back to the **root** navigator, so navigating to a sibling screen fails to type-check even though the call is valid at runtime.

Given a stack `FeatureStack` that contains both `FeatureList` and `FeatureDetails`:

```ts
const navigation = useNavigation('FeatureList')
navigation.push('FeatureDetails', { type: 'example' })
```

`tsc` reports:

```
error TS2345: Argument of type '["FeatureDetails", { type: string; }]' is not assignable to
parameter of type '[screen: "MainTabs", params: NavigatorScreenParams<...>]'.
  Type '["FeatureDetails", { type: string; }]' is not assignable to type '[screen: "Form", params: { formId: number; }]'.
      Type '"FeatureDetails"' is not assignable to type '"Form"'.
```

i.e. `navigation` was typed as the **root** stack's prop (`MainTabs | Form | …`) rather than `FeatureStack`'s.

The resolution is also **unstable**: the validity of an identical `navigation.navigate(...)` call can flip depending on how many *unrelated* modules a file imports (i.e. the total amount of type-instantiation work in the program). Adding an unrelated `import` to a passing file can make the same navigation call start failing. This is the fingerprint of `NavigationListForNested` exceeding TypeScript's instantiation-depth budget and silently degrading to the root fallback.

### Expected Behavior

`useNavigation('FeatureList')` resolves to `FeatureStack`'s navigation prop, so `push('FeatureDetails', { type })` type-checks. This was the case in `@react-navigation/core@8.0.0-alpha.19` — the regression was introduced in `alpha.20`.

### Reproduction

Minimal, self-contained repro: **https://github.com/Bram-dc/rn-usenavigation-instability-repro**

```bash
git clone https://github.com/Bram-dc/rn-usenavigation-instability-repro
cd rn-usenavigation-instability-repro
npm install
npm run check
```

- On the pinned `alpha.20` set → **1 error** (shown above).
- Swap to the `alpha.19` set (table in the README) and reinstall → **0 errors**, identical source.

[...]

PR #13145, opened 2026-06-09
fix(core): prevent useNavigation(name) from degrading on large trees

Bram-dc:
**Motivation**

`useNavigation('SomeScreen')` stops resolving to the navigator that actually contains
`SomeScreen` once the navigation tree is moderately large — it falls back to the **root**
navigator (so navigating to a sibling screen fails to type-check), or collapses to `any`. The
resolution is also unstable: the validity of an identical `navigation.navigate(...)` call can
flip depending on how many *unrelated* modules a file imports, which is the fingerprint of the
type exceeding TypeScript's instantiation-depth budget.

This was correct in `@react-navigation/core@8.0.0-alpha.19` and regressed in `alpha.20`, when the
`useNavigation(name)` overload was rewritten to:

```ts
export function useNavigation<
  const Navigator = RootNavigator,
  const RouteName extends string = string,
>(
  name: RouteName & keyof NavigationListForNested<Navigator>
): NavigationListForNested<Navigator>[RouteName & keyof NavigationListForNested<Navigator>];
```

Two problems: `NavigationListForNested<Navigator>` — already a deep, recursive type — is
instantiated **three times** (constraint, parameter, return), and `RouteName` is unconstrained,
so the name is re-derived via the `RouteName & keyof …` intersection in both the parameter and
the index. On a real-world tree that tips over the depth budget and the type silently degrades.

**The change**

Compute `NavigationListForNested<Navigator>` **once** (bind it to a `List` type parameter) and
constrain `RouteName` to `keyof List`:

```ts
export function useNavigation<
  const Navigator = RootNavigator,
  const List = NavigationListForNested<Navigator>,
  const RouteName extends keyof List = keyof List,
>(name: RouteName): List[RouteName];
```

This is the `alpha.19` behaviour with the heavy type evaluated a single time. No change to the
type machinery (`NavigationListForNested` etc.) is needed.

Note: constraining `RouteName extends keyof List` means a value typed as a plain `string` is no
longer accepted as `name` (you must pass a known route-name literal) — same as `alpha.19`. Keeping
the loose `RouteName extends string` form, even with `List` computed once, does **not** fix the
regression (it still degrades to `any`), so the constraint is load-bearing.

Fixes the issue tracked here: https://github.com/react-navigation/react-navigation/issues/13144

**Test plan**

- `yarn typecheck` — passes (existing `useNavigation` `@ts-expect-error` type tests stay green).
- `yarn jest packages/core/src/__tests__/useNavigation.test.tsx` — 6/6 pass (runtime is unchanged;
  this is a types-only change).
- Standalone reproduction (on the same `alpha.20` dependency set), before vs. after the fix:
  https://github.com/Bram-dc/rn-usenavigation-instability-repro
  - Before: `useNavigation('FeatureList').push('FeatureDetails', { type })` errors because the prop
    is resolved as the root navigator's.
  - After: it resolves to the correct `FeatureStack` composite prop (not the root navigator and not
    `any`); valid sibling navigations type-check and unknown screens / wrong params are still
    rejected.

### DIFF
--- a/packages/core/src/useNavigation.tsx
+++ b/packages/core/src/useNavigation.tsx
@@ -29,7 +29,6 @@ export function useNavigation<
 >(): NavigationListForNavigator<Navigator>[keyof NavigationListForNavigator<Navigator>];
 export function useNavigation<
   const Navigator = RootNavigator,
-  const List = NavigationListForNested<Navigator>,
   const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
 >(name: RouteName): NavigationListForNested<Navigator>[RouteName];
 export function useNavigation(name?: string): unknown {
--- a/packages/core/src/useNavigation.tsx
+++ b/packages/core/src/useNavigation.tsx
@@ -30,7 +30,7 @@ export function useNavigation<
 export function useNavigation<
   const Navigator = RootNavigator,
   const List = NavigationListForNested<Navigator>,
-  const RouteName extends keyof List = keyof List,
+  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
 >(name: RouteName): List[RouteName];
 export function useNavigation(name?: string): unknown {
   const root = React.use(NavigationContainerRefContext);
--- a/packages/core/src/useNavigation.tsx
+++ b/packages/core/src/useNavigation.tsx
@@ -29,7 +29,8 @@ export function useNavigation<
 >(): NavigationListForNavigator<Navigator>[keyof NavigationListForNavigator<Navigator>];
 export function useNavigation<
   const Navigator = RootNavigator,
-  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
+  const RouteName extends keyof NavigationListForNested<Navigator> =
+    keyof NavigationListForNested<Navigator>,
 >(name: RouteName): NavigationListForNested<Navigator>[RouteName];
 export function useNavigation(name?: string): unknown {
   const root = React.use(NavigationContainerRefContext);
--- a/packages/core/src/useNavigation.tsx
+++ b/packages/core/src/useNavigation.tsx
@@ -29,11 +29,9 @@ export function useNavigation<
 >(): NavigationListForNavigator<Navigator>[keyof NavigationListForNavigator<Navigator>];
 export function useNavigation<
   const Navigator = RootNavigator,
-  const RouteName extends string = string,
->(
-  name: RouteName & keyof NavigationListForNested<Navigator>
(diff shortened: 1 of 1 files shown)


packages/core/src/useNavigation.tsx
```diff
@@ -29,11 +29,9 @@ export function useNavigation<
 >(): NavigationListForNavigator<Navigator>[keyof NavigationListForNavigator<Navigator>];
 export function useNavigation<
   const Navigator = RootNavigator,
-  const RouteName extends string = string,
->(
-  name: RouteName & keyof NavigationListForNested<Navigator>
-): NavigationListForNested<Navigator>[RouteName &
-  keyof NavigationListForNested<Navigator>];
+  const List = NavigationListForNested<Navigator>,
+  const RouteName extends keyof List = keyof List,
```
satya164: to avoid regressing public API, it should be:
```suggestion
  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
```
assigning the type before doesn't seem to have a performance impact in benchmarks

packages/core/src/useNavigation.tsx
```diff
@@ -29,11 +29,9 @@ export function useNavigation<
...
 export function useNavigation<
   const Navigator = RootNavigator,
-  const RouteName extends string = string,
->(
-  name: RouteName & keyof NavigationListForNested<Navigator>
-): NavigationListForNested<Navigator>[RouteName &
-  keyof NavigationListForNested<Navigator>];
+  const List = NavigationListForNested<Navigator>,
+  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
+>(name: RouteName): List[RouteName];
```
satya164:
```suggestion
>(name: RouteName): NavigationListForNested<Navigator>[RouteName];
```

packages/core/src/useNavigation.tsx
```diff
@@ -29,11 +29,9 @@ export function useNavigation<
 >(): NavigationListForNavigator<Navigator>[keyof NavigationListForNavigator<Navigator>];
 export function useNavigation<
   const Navigator = RootNavigator,
-  const RouteName extends string = string,
->(
-  name: RouteName & keyof NavigationListForNested<Navigator>
-): NavigationListForNested<Navigator>[RouteName &
-  keyof NavigationListForNested<Navigator>];
+  const List = NavigationListForNested<Navigator>,
+  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
```
satya164:
```suggestion
  const RouteName extends keyof NavigationListForNested<Navigator> = keyof NavigationListForNested<Navigator>,
```

satya164 (approved):
thanks for the pr

PR #13153, opened 2026-06-12
fix(core): propagate beforeRemove to nested navigators on reset when route keys are unchanged

### LINKED ISSUE #9031: beforeReplace event not fired on browser back button
**Current Behavior**

Using a Stack Navigator in a web context, the browser's back button is not firing a beforeReplace event when the navigation will keep the user inside the app. Using `navigation.goBack()` does trigger the beforeReplace event properly, and using the back button when it would take you to another site also properly triggers the beforeReplace event.

Here's a snack that shows the behavior by using `alert()`s to show that blur gets fired but not beforeReplace
https://snack.expo.io/0XeAfnORv

**Expected Behavior**

The browser back button should trigger the beforeReplace event, just like `navigation.goBack()` does.

**How to reproduce**

Minimal repro I could come up with on snack: https://snack.expo.io/0XeAfnORv

**Your Environment**

│ │   ├─┬ workbox-navigation-preload@3.6.3
├─┬ @react-navigation/bottom-tabs@5.10.6
├─┬ @react-navigation/native@
│ ├─┬ @react-navigation/core@5.14.0
│ │ ├─┬ @react-navigation/routers@5.6.0
├─┬ @react-navigation/stack@

| @react-navigation/native       | 5.8.6
| @react-navigation/stack        | 5.12.3
| react-native-gesture-handler   | 1.6.1
| react-native-safe-area-context | 3.0.7
| react-native-screens           | 2.9.0
| react-native                   | 0.62.2
| expo                           | 38.0.8
| node                           | 12.18.2
| npm or yarn                    | yarn 1.22.10

TaduJR:
**Motivation**

`usePreventRemove` / `beforeRemove` never fire when screens are removed by a `RESET` that keeps a route's key but replaces its **nested** state `shouldPreventRemove` only inspects the keys of the navigator's own routes, so such a reset silently destroys nested screens without consulting their listeners.

On the web this is the common case: browser back goes through `useLinking`'s `resetRoot(record.state)` a root-targeted `RESET` with unchanged route keys where only a nested stack is popped so any `usePreventRemove` screen inside a nested navigator loses unsaved data on browser back with no event fired.

**The change** (`packages/core`):

- A route kept by the action whose nested `state` changed propagates the check down through its keyed `beforeRemove` listener, recursing into nested navigators regardless of which action caused the change. Outright removals behave exactly as before.
- `ChildBeforeRemoveListener` gains an optional `nextState`: the kept route's next nested state, or `undefined` when the route itself is removed.
- Both call sites diff against `getRoutesToCompare(nextState, routeNames)`. A partial next state is **not** rehydrated `getRehydratedState` generates fresh keys, a side effect a read-only check must not have. The diff instead mirrors rehydration's effect on route keys: unknown names are dropped (filtered on commit = genuinely removed), and keyless routes need no handling (fresh keys on commit, so they never match a current route). This also fixes the navigator's own level, where a stale payload consulted kept screens as removed.
- Confirm-and-proceed is unchanged (`VISITED_ROUTE_KEYS`, focused-first). Index-only changes and kept routes with unchanged nested state emit nothing new events only fire where screens were previously destroyed silently.

**Test plan**

6 new tests in `usePreventRemove.test.tsx` alongside the existing 9 each hardcodes the initial state, dispatches a hardcoded reset state, and asserts only the resulting `getRootState()`: nested prevent + confirm-proceed, kept-screen allow, multiple nested preventers, grand child screen, index-only change, and a stale reset state.

Fixes #9031


satya164 (changes requested):
Thanks for the PR.

When dealing with state changes, the type of action that triggered the state change is not relevant. The only thing that should matter is what changed. While special-casing `RESET` helps with this specific case, a proper fix should be generic and work with any action that changes the same state.

So it should only check if nested state changed, not the type of the action.

In addition, `shouldPreventRemove` can't use `nextState.routes` directly. The `route.state` can be a partial state, which means it needs to be rehydrated to get the full state first.

We also don't need an example for this. Unit tests should be enough.

Unit tests should also be simpler. It can just hardcode initial state and dispatch hardcoded state instead of simulating navigations and then calculating changes. It should also not test implementation details like what internal methods got called. It should only test the public behavior.

TaduJR:
@satya164 

Thanks for the review! I really appreciate your quick feedback. 

Addressed your concerns, one question a route popped but retained via `retainedRouteKeys` still counts as removed (matching the navigator's own level today), even though its instance is preserved. Should retention suppress the event? Can handle it here.

satya164:
> one question a route popped but retained via retainedRouteKeys still counts as removed

are you sure it counts as removed? the detection of removed routes is based on `routes` array. when a route is retained, it's still in `routes` array, so there doesn't seem like a reason it should count as removed.

TaduJR:
> are you sure it counts as removed? the detection of removed routes is based on routes array. when a route is retained, it's still in routes array, so there doesn't seem like a reason it should count as removed.

Hmm Yea a retained route stays in `routes` (just moves past `index`) and `retainRoutes` runs inside `getStateForAction`, so `shouldPreventRemove` always sees it by key and never counts it as removed. 

Nothing to handle, getting used to the codebase. 

Thanks a lot!

TaduJR:
@satya164 Would you be open to a 7.x backport too? 

We needed it at Expensify since v8 is alpha.

Thanks!

satya164:
hey @TaduJR, these should already be backported to 7.x. let me know if I'm misremembering.

TaduJR:
Thanks for the fast reply @satya164! 

I still double checked and this specific one isn't on 7.x yet. [`shouldPreventRemove` on 7.x](https://github.com/react-navigation/react-navigation/blob/680d8891630ccf512a2b0c44758392ee50318bd0/packages/core/src/useOnPreventRemove.tsx#L33-L47) still only loops over removed routes, without the kept-route / nested-state branch from [`60a484b5`](https://github.com/react-navigation/react-navigation/commit/60a484b558d1de3f8435c6509a00218df85c4d5a), so a root reset that keeps the route keys still fires no `beforeRemove`. Same on the published `7.21.11`.

Looks like a few other reset/`beforeRemove` fixes landed on 7.x recently but not this one. I can open a backport PR against `7.x` if you'd like.

satya164:
@TaduJR picked and released

TaduJR:
Thanks a lot @satya164

PR #13166, opened 2026-06-25
feat: add support for extending SFSymbol name types

satya164:
this makes it possible to specify additional SFSymbol names via module augmentation:

```ts
declare module '@react-navigation/native' {
  interface SFSymbolNames {
    'my-custom-symbol': true
  }
}
```


stachbial (approved):
Works great for me!

PR #13167, opened 2026-06-25
feat: support custom asset-catalog symbols in SFSymbol

stachbial:
**What**

Custom, app-provided SF Symbols now render through the existing `sfSymbol` `Icon` type — no new type. The native `<SFSymbol>` view falls back `UIImage(systemName:) ?? UIImage(named:)` and applies the same `SymbolConfiguration`, so rendering modes (monochrome / hierarchical / palette / multicolor) and symbol-effect animations work on a custom asset-catalog symbol just like a built-in one.

Since `PlatformIcon` landed on `main`, every `Icon` consumer (`Button`, `HeaderIcon`, `DrawerItem`, `DrawerToggleButton`, `MaterialTopTabBar`, native bottom tabs) already routes through `<SFSymbol>`, so this is a native-only change.

**Scope change (per review)**

Split out of the earlier `image`/`sfSymbol` bundle:

- Non-square image sizing moved to a separate PR as `aspectRatio` (replacing the earlier `width` / `height` proposal).
- This PR is now just the SFSymbol custom-symbol support, which is mergeable independently of react-native-screens.

**Depends on**

- #13166 — custom SF Symbol *names* (`SFSymbolNames` augmentation makes `{ type: 'sfSymbol', name }` accept them).

**Test plan**

Built and ran on iOS. A custom asset-catalog symbol renders via `{ type: 'sfSymbol', name }` in `Button`, the header back button, drawer items, and material top tabs, and on the native bottom tab bar — including palette / hierarchical modes and animations on the custom symbol. Built-in SF Symbols unaffected.

### DIFF
--- a/packages/native/ios/ReactNavigationSFSymbolView.swift
+++ b/packages/native/ios/ReactNavigationSFSymbolView.swift
@@ -532,15 +532,19 @@ import UIKit
       let variableValue = min(max(props.variableValue, 0), 1)
 
       if #available(iOS 16.0, *) {
-        return UIImage(
+        if let systemImage = UIImage(
           systemName: name,
           variableValue: Double(variableValue),
           configuration: configuration
-        )
+        ) {
+          return systemImage
+        }
       }
     }
 
+    // Fall back to a custom symbol from the app's asset catalog.
     return UIImage(systemName: name, withConfiguration: configuration)
+      ?? UIImage(named: name)?.applyingSymbolConfiguration(configuration)
   }
 
   @available(iOS 26.0, *)


packages/bottom-tabs/src/views/TabBarIcon.tsx
```diff
@@ -163,6 +163,27 @@ function renderIcon({
...
+            source={{ uri: iconValue.name }}
+            resizeMode="contain"
+            style={{
+              width: size,
+              height: size,
+              tintColor:
+                iconValue.tintingMode === 'original' ? undefined : color,
+            }}
+          />
+        );
```
satya164: I wouldn't add separate types since `sfSymbol` and `image` already handle this. My goal is `type: 'image'` and `type: 'sfSymbol'` should be consistent with `Image` and `SFSymbol` component APIs, i.e., support the same values. So for new types, we should only add if the existing shape isn't sufficient.

satya164:
Maybe we should extend screens `type: 'sfSymbol'` to accept a custom `SFSymbol` instead of `xcasset` 🤔 If it's a custom "SFSymbol" but then `type: 'sfSymbol'` doesn't handle it, it's kinda confusing.

We'll also want to pass more configuration for the SFSymbol in the future like all customization options we have here https://reactnavigation.org/docs/8.x/icons#sf-symbols, and afaik they aren't going to work with any xcasset catalog image even though both are used with `UIImage(named: "myIcon")`. Having them in `type: 'sfSymbol'` instead of `type: 'xcasset'` would mean we can provide proper TS types. 

We'll also need to add custom SF Symbol support in the `SFSymbol` component in React Navigation as currently it uses only `systemName` https://github.com/react-navigation/react-navigation/blob/323cca48fb875959d99062fb92b6bebb8b7fca65/packages/native/ios/ReactNavigationSFSymbolView.swift#L536

satya164 (approved):
Thanks for the PR

Issue #4209, opened 2018-05-12
creating simple stacknavigator using createStackNavigator


dheerajsinghanktech:
Using react-navigation less 2.0 everythings running perfectly. But as I upgraded the react-navigation "react-navigation": "^2.0.1" and then I am getting bellow error.

Attempted to assign to readonly property.
popProvider
    ReactNativeRenderer-dev.js:8624:10
unwindWork
    ReactNativeRenderer-dev.js:10368:20
completeUnitOfWork
    ReactNativeRenderer-dev.js:12288:31
renderRoot
    ReactNativeRenderer-dev.js:12450:44
performWorkOnRoot
    ReactNativeRenderer-dev.js:13034:34
performWork
    ReactNativeRenderer-dev.js:12951:26
performSyncWork
    ReactNativeRenderer-dev.js:12913:16
requestWork
    ReactNativeRenderer-dev.js:12799:6
scheduleWorkImpl
    ReactNativeRenderer-dev.js:12666:24
scheduleWork
    ReactNativeRenderer-dev.js:12622:28
scheduleRootUpdate
    ReactNativeRenderer-dev.js:13324:17
_updateContainerAtExpirationTime
    ReactNativeRenderer-dev.js:13360:6
updateContainer
    ReactNativeRenderer-dev.js:13385:8
render
    ReactNativeRenderer-dev.js:14119:35
renderApplication
    renderApplication.js:58:21
run
    AppRegistry.js:104:10
runApplication
    AppRegistry.js:196:26
__callFunction
    MessageQueue.js:353:47
<unknown>
    MessageQueue.js:118:26
__guardSafe
    MessageQueue.js:316:6
callFunctionReturnFlushedQueue
    MessageQueue.js:117:17

dcvz:
I have a similar issue on the same line but my issue reads as `undefined is not an object (evaluating 'context._currentValue = currentValue')

dcvz:
Solved my particular issue! Looks like [this commit] https://github.com/facebook/react/commit/ba245f6f9b0bf31c2ebff5c087c21bcae111e6c3 changed the way context works in React `16.3.1+`.

Downgrading back to v16.3.0 fixed my issue.

rizwanahmed19:
I'm having the same problem :(

zacherkkila:
Same issue here, downgrading to 2.0.0 worked for me as well

react-native:0.54.4

alex-mironov:
are there any plans to fix this?

[...]

PR #13168, opened 2026-06-25
feat(bottom-tabs): per-tab icon size & active-indicator sizing on the native tab bar

stachbial:
**What**

Bottom-tab icon sizing options, on the `native` bar:

- `tabBarIconSize` (dp) — per-tab icon size. On the `custom` (JS) bar it sizes the icon on **all platforms**; on the `native` bar it maps to react-native-screens' `drawableIconSize` — the bar boxes to the largest size across tabs and insets smaller ones (Android only). Unset uses the default.
- `tabBarActiveIndicatorWidth` / `tabBarActiveIndicatorHeight` (dp) — bar-wide active-indicator size, taken from the focused tab. Native bar only; unset auto-scales to wrap the icon box.

Android drawable tinting on the native bar is driven by `image`'s `tinted` (no `tintingMode` in the public API) — `tinted: false` keeps the drawable's own colors.

Also fixes `tabBarIconStyle`'s type (`TextStyle` → `ViewStyle`) — it styles the icon container View.

**Changes since review**

- Rebased onto `PlatformIcon`.
- `tabBarIconSize` is no longer restricted to Android on the `custom` implementation — it is plain RN layout and works on iOS too. It stays Android-only on the `native` implementation, where it maps to `drawableIconSize`.
- Non-square image sizing moved out to `aspectRatio` (#13194); this PR is stacked on it.

**Depends on**

- #13194 (icon `aspectRatio`) and software-mansion/react-native-screens#4209 (native Android support). Stays a draft until #4209 publishes — the cross-repo `tinted` on `PlatformIconAndroid` isn't in the released screens types yet.

**Test plan**

Native bottom tab navigator on Android, with `react-native-screens` from #4209:

- `tabBarIconSize` on one screen scales that tab's icon; others stay default.
- `tabBarActiveIndicatorWidth` / `tabBarActiveIndicatorHeight` in `screenOptions` reshape the indicator.
- `tinted: false` on a multicolor drawable keeps its colors as a tab icon.
- On the custom (JS) bar, `tabBarIconSize` scales the icon on iOS and Android.

### DIFF
--- a/packages/bottom-tabs/src/types.tsx
+++ b/packages/bottom-tabs/src/types.tsx
@@ -174,7 +174,7 @@ type BottomTabCustomOptions = {
    *
    * Only supported with `custom` implementation.
    */
-  tabBarIconStyle?: StyleProp<TextStyle>;
+  tabBarIconStyle?: StyleProp<ViewStyle>;
 
   /**
    * Whether the tab bar gets hidden when the keyboard is shown.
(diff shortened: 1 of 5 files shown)


packages/bottom-tabs/src/types.tsx
```diff
@@ -402,6 +402,42 @@ export type BottomTabNavigationOptions = {
    */
   tabBarLabelVisibilityMode?: TabBarItemLabelVisibilityMode;
 
+  /**
+   * Size (in dp) of this tab's icon. Can be set globally via `screenOptions`
+   * and/or overridden per screen; the largest value across tabs sets the bar's
+   * icon box, and smaller icons are inset to their own size. Unset tabs use the
+   * default 24dp.
+   *
+   * Only supported with `native` implementation.
```
satya164: I guess this one can easily be supported for custom implementation as well.

packages/bottom-tabs/src/views/BottomTabViewNativeImpl.tsx
```diff
@@ -747,6 +753,7 @@ function getPlatformIcon(icon: Icon): PlatformIcon {
         android: {
           type: 'drawableResource',
           name: icon.name,
+          tintingMode: icon.tintingMode,
```
satya164: for `image` we have a `tinted: boolean`, so lets use the same property for react navigation's public API consistency

PR #13170, opened 2026-06-30
feat(native-stack): add options to disable inset application for header

t0maboro:
*Note: This API will be released in react-native-screens@4.26*

**Motivation**

Previously, only the top inset could be opted out of (managed internally via `headerTopInsetEnabled`). Recently, in `react-native-screens` we added support for disabling left, right, and bottom paddings for the header directly on the native side. I am exposing these props for the native-stack to control opting out of padding applied by Screens, when someone is handling it manually.

**Test plan**

Tested on react-native-screens side; associated commit: https://github.com/software-mansion/react-native-screens/pull/4220


satya164 (approved):
Thanks

PR #13175, opened 2026-07-03
fix: avoid scheduling updates in insertion effect in useNavigationState

### LINKED ISSUE #13174: Error/Regression - useInsertionEffect must not schedule updates at every screen navigation after upgrading to @react-navigation/core@7.21.4
### Current behavior

Apologies for not providing a minimal reproduction repository. I am short on time, but I was able to compare installed package versions and library source before and after the issue.

After upgrading React Navigation packages, the app logs this error when navigating from the Home screen to other screens:

```text
ERROR  useInsertionEffect must not schedule updates.
ERROR  useInsertionEffect must not schedule updates.
```

## App Structure

The app uses nested navigation:

```text
NavigationContainer
└─ Root routing
   └─ Drawer Navigator
      ├─ Home screen
      ├─ Other drawer screens
      └─ Screens/routes that use native-stack navigation
```

The error appears during navigation from Home to other screens in this nested navigator setup.

## Environment

```text
react@19.2.3
react-native@0.86.0
```

## Versions Where The Error Happens

```text
@react-navigation/native@7.3.6
@react-navigation/drawer@7.12.5
@react-navigation/native-stack@7.17.8
@react-navigation/core@7.21.4
@react-navigation/elements@2.9.28
```

## Versions Where The Error Does Not Happen

The error no longer happens after forcing:

```json
"overrides": {
  "@react-navigation/core": "7.21.2",
  "@react-navigation/elements": "2.9.26"
}
```

Verified installed tree:

```text
@react-navigation/core@7.21.2
@react-navigation/elements@2.9.26
```

## Confirmed Code Difference

In the working install, `@react-navigation/core@7.21.2` has this in `useScheduleUpdate.tsx`:

```ts
scheduleUpdate(callback);
useClientLayoutEffect(flushUpdates);
```

In the failing install, `@react-navigation/core@7.21.4` has this in `useScheduleUpdate.tsx`:

```ts
React.useInsertionEffect(() => {
  scheduleUpdate(callback);
});

useClientLayoutEffect(flushUpdates);
```

Also, `@react-navigation/core@7.21.4` contains multiple `React.useInsertionEffect` usages in navigation internals. The app itself does not call `useInsertionEffect` directly.

## Confirmed Result

With `@react-navigation/core@7.21.4`, the error happens.

With `@react-navigation/core@7.21.2` forced via npm overrides, the error no longer happens after reinstalling dependencies and rebuilding.

### Expected behavior

No error

### Reproduction

Reproduction repository: not available (investigation included below).

### Platform

- [x] Android
- [x] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/drawer
- [ ] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| package                             | version |
| ----------------------------------- | ------- |
| @react-navigation/native            | 7.3.6 |
| @react-navigation/bottom-tabs       | N/A |
| @react-navigation/drawer            | 7.12.5 |
| @react-navigation/material-top-tabs | N/A |
| @react-navigation/stack             | N/A |
| @react-navigation/native-stack      | 7.17.8 |
| react-native-drawer-layout          | N/A |
| react-native-tab-view               | N/A |
| react-native-screens                | 4.25.2 |
| react-native-safe-area-context      | 5.8.0 |
| react-native-gesture-handler        | 2.32.0 |
| react-native-reanimated             | 4.5.1 |
| react-native-pager-view             | N/A |
| react-native                        | 0.86.0 |
| expo                                | No |
| node                                | 24.15.0 |
| npm or pnpm                         | npm (using `overrides`) |

Michaelmala:
**Motivation**

Fixes #13174.

After upgrading to `@react-navigation/core@7.21.4`, the app logs
`useInsertionEffect must not schedule updates.` on **every** navigation, for any
component that reads `useNavigationState` and stays mounted across navigations
(e.g. a custom tab bar).

The cause is in `NavigationStateListenerProvider`, which notifies its
subscribers from a `useInsertionEffect` (introduced in e0672284). Those
subscribers are backed by `useSyncExternalStore`, so being notified schedules a
state update — and React forbids scheduling updates inside insertion effects.

The insertion effect was chosen for a reason (insertion effects run before
layout effects and bottom-up, so nested navigators queue their updates before
any flush). So a plain revert to `useLayoutEffect` would reintroduce the
nested-navigator ordering issue that e0672284 was solving.

Instead, this ports the concurrent-safe approach already used on `main` (8.x, in
b390bec3 "make useNavigationState work with concurrent features") to the `7.x`
line:

- `useNavigationState` no longer relies on `useSyncExternalStore`. It reads the
  selected value **during render** via `getState` (always up-to-date, no
  tearing, independent of effect timing) and subscribes with a manual
  `checkForUpdates` that schedules a re-render only when the selected value
  actually changes.
- `NavigationStateListenerProvider` now receives the live `getState` from the
  navigator and notifies subscribers from a `useLayoutEffect`, where scheduling
  updates is allowed. Because consumers read fresh state during render, moving
  the notification to the layout phase does not cause stale reads, and the
  ordering guarantees are preserved.

This keeps `useInsertionEffect` everywhere it is genuinely needed and only
changes the one place where it was scheduling updates.

**Test plan**

- `yarn lint` — passes
- `yarn typecheck` — passes
- `yarn test packages/core` — passes (356 tests)
- Added a regression test in `useNavigationState.test.tsx`
  ("doesn't schedule updates in an insertion effect when navigating") that
  reproduces #13174: a persistently-mounted `useNavigationState` subscriber
  navigating between screens. On the previous code it fails with
  `useInsertionEffect must not schedule updates.`; with this change it passes.
- Also verified the consuming navigators are unaffected:
  `yarn test packages/native-stack packages/bottom-tabs packages/drawer packages/stack packages/elements` — passes (61 tests).


satya164 (approved):

satya164:
Thanks

PR #13183, opened 2026-07-14
feat(native-stack): Support XCAsset header icons

maxlapides:
## Summary

- add the XCAsset icon variant to native-stack's iOS header-item icon type
- add static type coverage for an XCAsset-backed header button

## Context

`react-native-screens` already supports iOS platform icons in the form `{ type: 'xcasset', name: string }`, and native-stack already passes non-image platform icon objects through at runtime. This change exposes that supported shape in native-stack's public TypeScript API.

### DIFF
--- a/example/__typechecks__/static.check.tsx
+++ b/example/__typechecks__/static.check.tsx
@@ -30,6 +30,7 @@ import {
 import {
   createNativeStackNavigator,
   createNativeStackScreen,
+  type NativeStackNavigationOptions,
 } from '@react-navigation/native-stack';
 import {
   createStackNavigator,
@@ -60,6 +61,19 @@ const NativeStack = createNativeStackNavigator({
     },
   },
 });
+
+const nativeStackOptions = {
+  unstable_headerRightItems: () => [
+    {
+      type: 'button',
+      label: 'Archive',
+      icon: { type: 'xcasset', name: 'archive' },
+      onPress: () => {},
+    },
+  ],
+} satisfies NativeStackNavigationOptions;
+
+expectTypeOf(nativeStackOptions).toExtend<NativeStackNavigationOptions>();
 
 createStaticNavigation(NativeStack);
 
--- a/packages/native-stack/src/types.tsx
+++ b/packages/native-stack/src/types.tsx
@@ -824,7 +824,18 @@ export type NativeStackNavigationOptions = {
   inactiveBehavior?: 'pause' | 'unmount' | 'none' | undefined;
 };
 
-type IconIOS = Extract<Icon, { type: 'image' | 'sfSymbol' }>;
+type IconIOS =
+  | Extract<Icon, { type: 'image' | 'sfSymbol' }>
+  | {
+      /**
+       * - `xcasset` - Use an image from the iOS asset catalog.
+       */
+      type: 'xcasset';
+      /**
+       * Name of the image in the iOS asset catalog.
+       */
+      name: string;
+    };
 
 type SharedHeaderItem = {
   /**


satya164:
Thanks for the PR. We're not planning to expose `type: 'xcasset'` at the moment as it's largely redundant:

- To use images from the asset catalog, `type: 'image'` already handles it (similar to React Native's `Image` component), e.g. `{ type: 'image', source: { uri: 'asset_name' } }`
- For custom symbols, [we're looking to update `type: 'sfSymbol'` to support them](https://github.com/software-mansion/react-native-screens/pull/4209)

Let me know if there's anything else `type: 'xcasset'` provides that's not acheivable with the above.

maxlapides:
@satya164 Actually, there is a specific reason why we want to use `xcasset` instead of a PNG 🙂

For native-stack header items, React Navigation maps `type: 'image'` to an image/template source, which react-native-screens resolves through `RCTImageLoader`. react-native-screens notes that synchronous loading through this path isn’t guaranteed in release builds. By contrast, `type: 'xcasset'` is resolved directly with `UIImage imageNamed:`.

We hit this distinction in a release build: UIKit laid out the native header item before the image-source callback completed, causing a flash of missing button icons:

621674862-fe31ad92-fb51-481f-b164-fb1abf911bae

Using `xcasset` resolved the image synchronously and fixed the layout race. So the two forms can address the same catalog resource, but aren’t equivalent in loading semantics.

The linked custom `sfSymbol` work could also solve this if the `imageNamed:` fallback is applied to `RNSBarButtonItem`; its current diff appears scoped to native tabs.

satya164:
if it's necessary, then we can internally use `xcasset` when an asset name is specified for `image`. though if images are not appearing in header items at all then that seems like a bug. can you report it in screens repo as well?

maxlapides:
@satya164 The images do appear in the header buttons after a brief flash where they are missing, while they are loaded in asynchronously. Here's a screen recording of the effect that I'm trying to avoid:

https://github.com/user-attachments/assets/f96e3f65-ee9d-4fde-acbd-b1f6e9aad4a4

I'm not sure if this will be classified as a bug in the screens repo, since the asynchronous loading seems to be intentional, but I opened an issue there anyway to see what they think: https://github.com/software-mansion/react-native-screens/issues/4347

satya164:
@maxlapides thanks for the details. i'd still consider it a bug if the image loader behaves like this but let me consult with screens team and get back to the PR.

maxlapides:
Note that this issue only affects production builds on physical iOS devices. The issue cannot be reproduced on an iOS simulator.

satya164:
@maxlapides looking at the code in react native, it seems like it's still synchronous

https://github.com/react/react-native/blob/f91191160497e431798bbb87b0c7c220ffd857bc/packages/react-native/Libraries/Image/RCTBundleAssetImageLoader.mm#L52-L57
https://github.com/react/react-native/blob/ff4f96a23b85b30d8de9331dfac2895f385f67b0/packages/react-native/React/Base/RCTUtils.mm#L902-L908

i'm not super familiar with this code, so maybe i'm missing something. but it would seem that the bug is not because of asynchronous image loading, but something else.

have you been using patched `type: 'xcasset'` in prod without this issue? and to clarify, with `type: 'image'`, are you using bundler assets or image names from the asset catalog? so you notice a difference between `{ type: 'image', source: { uri: 'name' }}` vs `{ type: 'xcasset', name: 'name' }`?

maxlapides:
Thanks for digging into this and challenging the original diagnosis.

Yes, we've been using patched type: 'xcasset' in prod without this issue. But, since opening this PR, our downstream app upgraded Expo and migrated these header icons to the existing supported form:

```ts
{
  type: 'image',
  source: { uri: 'asset_name' },
}
```

This uses the same named assets from the iOS asset catalog without requiring the `xcasset` type or our local TypeScript patch.

We’re validating this path in our production app on a physical device, but the proposed type addition is no longer needed by our codebase, so I’m closing this PR.

However, I still think we should be able to use regular PNG assets (without adding them to the iOS asset catalog), but we can continue that discussion here: https://github.com/software-mansion/react-native-screens/issues/4347

Thanks again for the guidance!

Issue #12137, opened 2024-09-07
White flicker on drawer@7 when having nested navigators on new architecture


hsjoberg:
### Current behavior

Hi, when there are nested navigators \w @react-navigation/drawer@7, flickering occurs on Android when pulling the drawer.
This only happens on new architecture.

https://github.com/user-attachments/assets/5c56946a-2207-4187-8408-4af9d3c00243

There's also a warning from react-reanimated:
```
[react-native-gesture-handler] None of the callbacks in the gesture are worklets. If you wish to run them on the JS thread use '.runOnJS(true)' modifier on the gesture to make this explicit. Otherwise, mark the callbacks as 'worklet' to run them on the UI thread.
```

But I do not believe this is related.

It seems like the screens gets dismounted or something, because I have the core background color on my real other app as dark ( `<item name="android:windowBackground">@color/dark</item>`), and there the flickering is black instead.

The code is just this. Just nest navigators. I also uploaded a repro to github, link below.

```Typescript
import React from 'react';
import {createStackNavigator} from '@react-navigation/stack';
import {View} from 'react-native';
import {NavigationContainer} from '@react-navigation/native';
import {createDrawerNavigator} from '@react-navigation/drawer';

function Screen() {
  return (
    <View style={{ flex: 1, backgroundColor: '#000' }} />
  );
}

const Stack = createStackNavigator();
function MyStack() {
  return (
    <Stack.Navigator>
      <Stack.Screen name="Screen" component={Screen} />
    </Stack.Navigator>
  );
}

const Drawer = createDrawerNavigator();
function MyDrawer() {
  return (
    <Drawer.Navigator
      screenOptions={{
        swipeEdgeWidth: 400,
      }}>
      <Drawer.Screen name="Test" component={MyStack} />
    </Drawer.Navigator>
  );
}

export default function App() {
  return (
    <NavigationContainer>
      <MyDrawer />
    </NavigationContainer>
  );
}
```

### Expected behavior

There should be no flickering.

### Reproduction

https://github.com/hsjoberg/reactnavigationflashbug

### Platform

- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [X] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [ ] @react-navigation/native-stack
- [ ] react-native-tab-view

### Environment

[...]

PR #13187, opened 2026-07-16
fix: avoid Android Fabric drawer flash from hitSlop remounts

rd273001:
## Summary

On Android (New Architecture), opening/closing a drawer that wraps nested navigators can flash a frame of the underlying scene mid-animation.

**Cause:** root pan `hitSlop` was keyed off `isOpen`, so it updated when `open` flipped at animation start and remounted `GestureDetector` mid-spring.

**Fix (minimal):** drive `hitSlop` from **settled** open state (update after the spring finishes / when already at target). No public API changes. Content a11y via `useAnimatedProps` remains as on `main`.

Related: #12137

## Example repro

Examples → **Drawer → Nested Tabs + Stack**

Nesting: **Drawer → Bottom Tabs → Native Stack**

Red `sceneStyle` makes the remount flash obvious on light screen content.

## Tests

Unit tests in `react-native-drawer-layout` cover both `settledOpen` sync paths (already-at-target + spring finished).

## Test plan

Android, New Architecture enabled:

1. Run the example app
2. Open **Drawer → Nested Tabs + Stack**
3. Open/close the drawer frequently (~10–20×) via hamburger, edge swipe, and overlay tap
4. **On `main` (without this fix):** brief RED flash on the screen (~2 in 4–5 toggles on mid-range devices; rarer on newer/faster devices). Bottom tabs should not flash.
5. **On this PR:** same steps — no RED flash
6. Confirm edge-swipe open and overlay-tap close still work

Verified locally on mid-range Android 10 and a faster Android 16 device.

### DIFF
--- a/packages/react-native-drawer-layout/src/types.tsx
+++ b/packages/react-native-drawer-layout/src/types.tsx
@@ -1,6 +1,9 @@
 import * as React from 'react';
 import type { StyleProp, View, ViewStyle } from 'react-native';
-import type { PanGestureConfig } from 'react-native-gesture-handler';
+import type {
+  PanGesture,
+  PanGestureConfig,
+} from 'react-native-gesture-handler';
 import type { SharedValue } from 'react-native-reanimated';
 
 export type DrawerProps = {
@@ -157,4 +160,6 @@ export type OverlayProps = React.ComponentProps<typeof View> & {
   progress: SharedValue<number>;
   onPress: () => void;
   accessibilityLabel?: string | undefined;
+  /** Optional pan for swipe-to-close from the dimmed overlay area. */
+  panGesture?: PanGesture | undefined;
 };
--- a/packages/react-native-drawer-layout/src/views/Drawer.native.tsx
+++ b/packages/react-native-drawer-layout/src/views/Drawer.native.tsx
@@ -130,15 +130,25 @@ export function Drawer({
     onGestureCancel?.();
   });
 
+  // Root pan hitSlop stays edge-only forever (never keyed off isOpen). Changing
+  // hitSlop mid-spring remounts GestureDetector and flashes on Android Fabric
+  // (#12137). Full-width root hitSlop while closed steals content horizontal pans.
+  //
+  // Swipe-to-close from the dimmed content / drawer body uses separate pans
+  // (below) so root hitSlop can stay edge-only forever.
+  // Widen the open-edge via swipeEdgeWidth if the default feels too small.
   const hitSlop = React.useMemo(
     () =>
       isRight
-        ? // Extend hitSlop to the side of the screen when drawer is closed
-          // This lets the user drag the drawer from the side of the screen
-          { right: 0, width: isOpen ? undefined : swipeEdgeWidth }
-        : { left: 0, width: isOpen ? undefined : swipeEdgeWidth },
-    [isRight, isOpen, swipeEdgeWidth]
+        ? // Edge-only hit area so the user can drag the drawer open from the side
+          { right: 0, width: swipeEdgeWidth }
+        : { left: 0, width: swipeEdgeWidth },
+    [isRight, swipeEdgeWidth]
   );
+
+  // Defer aria-hidden until the spring finishes. Flipping it with isOpen mid-spring
+  // Fabric-recommits the content tree and can flash even when hitSlop is stable.
+  const [settledOpen, setSettledOpen] = React.useState(isOpen);
 
   const [initialWidth] = React.useState(() => Dimensions.get('window').width);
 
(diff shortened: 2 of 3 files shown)


rd273001:
`codecov/patch` is at ~70.37% (target ~72.24%). The uncovered lines are native gesture / spring paths in `Drawer.native.tsx` and `Overlay.native.tsx`.

`react-native-drawer-layout` currently has no unit tests for these gesture flows, and covering them meaningfully needs a React Native + gesture-handler setup rather than a light Jest mock. Happy to add tests if maintainers prefer a specific approach; otherwise this may need a Codecov exception for this native path.

satya164:
Thanks for the PR. Can you provide before/after videos and a test plan with repro?

rd273001:
Thanks @satya164 — happy to add those.

### Before / after
- **Before:** https://drive.google.com/file/d/101s5i_Sq7Afrigrr8nFnOKyQlWR_mHO6/view?usp=sharing — Android, New Arch (`newArchEnabled=true`), RN 0.81, nested Drawer → Stack → Tabs. Opening/closing the drawer intermittently flashes a light frame over the scene (often over the header; easier to see with a non-white header background). Captured in our production app (DFC) with that nested setup — I haven’t re-verified on the bare template example.
  - Clearest flashes in this recording: ~**1s**, ~**24–25s**, and ~**28–29s**.
  - Some flashes that were obvious live didn’t make it into the recording — they’re very brief, so device screen-capture frame rate can miss them.
- **After:** https://drive.google.com/file/d/1Kpp9iNkrxzM8YjR75UJiGhwA9aAixrBf/view?usp=sharing — same device / same gestures on this branch; flash completely gone. Clip is intentionally >1 min to show it stays clean & smooth across many open/close attempts (hamburger + edge drag open; overlay tap + drag close).

### Repro
Related to #12137.

Root cause we hit: root pan `hitSlop` was `width: isOpen ? undefined : swipeEdgeWidth`, so open/close remounted `GestureDetector` mid-spring on Android Fabric (and flipping `aria-hidden` with `isOpen` during the spring could recommit content).

Minimal setup (what we used in DFC; should also apply to a minimal drawer demo):
- Android, `newArchEnabled=true`
- `@react-navigation/drawer` + `react-native-drawer-layout` wrapping a nested Stack (or Stack → Tabs), ideally with a colored header
- Open/close repeatedly (hamburger + edge swipe + overlay tap) — flash is intermittent but usually shows within ~10–15 toggles

### Test plan
1. Android device/emulator, New Architecture on.
2. Nested navigators: Drawer → Stack → Tabs (or Drawer → Stack with a colored header).
3. Open drawer via menu button ~15× — no white/light flash over content/header.
4. Close via overlay tap ~15× — same.
5. Edge-swipe open / swipe-close still works.
6. Horizontal pans inside content (e.g. FlatList) still work when the drawer is closed (edge-only hitSlop).
7. iOS smoke: open/close still looks normal (this was an Android Fabric issue).

rd273001:
@satya164 I have updated the PR description with the follow-up fix and clearer repro notes.

**What changed since the first version**
- `settledOpen` alone still left a rare flash, so content `aria-hidden` / Android `importantForAccessibility` now track drawer progress via `useAnimatedProps` (plus overlay `accessibilityViewIsModal`).
- Clarified that the flash is the underlying/previous scene background (not always white) — e.g. green/blue from the example home when nested Drawer → Tabs is open.
- Added before/after videos from the example app. Nested Tabs + Stack wiring and stronger background colors were local-only for reproduction/video clarity and are **not** in this PR — commits only contain the `react-native-drawer-layout` fix.

satya164 (changes requested):
Hey @rd273001 . The PR diff is now much bigger with a lot of refactors and unnecessary changes to the public API and behavior. Please keep the changes to a minimum and change only what's necessary for the fix without changing public API.

satya164:
Also, please provide a reproducible example app where I can actually see the issue. Video isn't sufficient without a test plan that I can actually verify.

rd273001:
@satya164 Updated as per your review feedback:

**Minimal fix only** — `hitSlop` now follows settled open state (no public API / gesture refactors).

**Runnable repro** — Examples → **Drawer → Nested Tabs + Stack** (Drawer → Bottom Tabs → Native Stack). Red `sceneStyle` makes the flash obvious.

### Test plan
1. Android + New Architecture
2. Open **Drawer → Nested Tabs + Stack**
3. Open/close ~10–20× (hamburger / edge swipe / overlay tap)
4. **On `main`:** brief RED flash on the screen (~2 in 4–5 on mid-range; rarer on newer/faster devices). Bottom tabs should not flash.
5. **On this PR:** same steps — no RED flash

Happy to adjust further if needed.

rd273001:
@satya164 Quick follow-up on the latest commits:

- Added unit tests for the `settledOpen` / `hitSlop` sync paths (`codecov/patch` is green).
- Branch is merged with latest `main`.
- Repro remains: Examples → **Drawer → Nested Tabs + Stack** (single fixed-path example; compare vs `main` for the flash).

Please ignore older comments that mentioned missing unit tests or said the Nested Tabs example wasn’t in the PR — those are outdated.

rd273001:
@satya164  Friendly ping — I've addressed the review feedback (minimal fix, runnable repro, tests, CI green). Whenever you have a moment, could you take another look? Happy to adjust anything else. Thanks!

PR #13188, opened 2026-07-17
fix: preserve static param list inference with TypeScript 7

WookieFPV:
**Motivation**

TypeScript v7 (and typescript 6 with --stableTypeOrdering) broke static navigation type inference for me: 
`TS2315: Type 'StaticParamList' is not generic.`
`StaticParamList` started being treated as non-generic, which caused inferred route names/params to collapse.

This small type-only change fixes the issue by tightening the internal static screen map constraint from `Record<string, any>` to `Record<string, unknown>`. I’m not fully sure why this specific shape trips TS7’s circular type alias handling, but this keeps the existing inference working.

**Test plan**

Added a type-level regression test for nested static navigation inference.

**Related**

- [https://github.com/react-navigation/react-navigation/discussions/12891](https://github.com/react-navigation/react-navigation/discussions/12891)
- https://github.com/microsoft/typescript-go/issues/2698

### DIFF
--- a/example/__typechecks__/static.check.tsx
+++ b/example/__typechecks__/static.check.tsx
@@ -53,6 +53,32 @@ expectTypeOf<StaticParamList<typeof NativeStack>>().toMatchTypeOf<{
   Foo: undefined;
 }>();
 
+const NestedStaticStack = createNativeStackNavigator({
+  screens: {
+    Profile: (_: StaticScreenProps<{ id: string }>) => null,
+  },
+});
+
+const StaticRootStack = createNativeStackNavigator({
+  screens: {
+    Home: () => null,
+    Nested: NestedStaticStack,
+  },
+});
+
+createStaticNavigation(StaticRootStack);
+
+type StaticRootParamList = StaticParamList<typeof StaticRootStack>;
+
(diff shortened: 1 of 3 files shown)


satya164 (changes requested):
Thanks for the PR.

Lets remove the changes from `StaticNavigation.test.tsx` - that file is for runtime tests.

We have type tests in `__typechecks__` with `StaticParamList` already https://github.com/react-navigation/react-navigation/blob/b0135879fbe4aa4a657798845483b7e78c307778/example/__typechecks__/static.check.tsx#L1136

But these don't produce any errors with TypeScript 7 (e.g. in this WIP PR https://github.com/react-navigation/react-navigation/pull/13163). So there maybe some other factors.

Edit: Maybe this is a React Navigation 7 specific issue, haven't tested those types with TS7 yet, and since TS7 doesn't have API for ESLint plugin etc. yet, can't upgrade yet.

WookieFPV:
Thank you for the guidance!

These tests fit much better. With the new test, `tsc` fails without the type change in `packages/core/src/types.tsx`, so it now catches the regression directly.

satya164 (approved):
Thanks

PR #13189, opened 2026-07-18
fix: prevent hidden stack screens from receiving keyboard focus on web

### LINKED ISSUE #12583: Accessibility bug when using a Stack with animation on web

end-y:
**Motivation**

Fixes #12583

On the web, when transition animation is enabled, the previous screen stays in the DOM and remains keyboard-focusable — `aria-hidden` and `pointerEvents: 'none'` don't block Tab focus, so focus moves to invisible elements and the "Blocked aria-hidden" console warning appears. This PR sets the [`inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) attribute on the unfocused card's DOM element (via ref, since React Native Web's `View` doesn't support the prop) — the same approach `main` uses in `ActivityView` (#12964). Animations and native behavior are unchanged.

**Test plan**

Verified with the repro app from the issue: Tab no longer reaches hidden screens after animated transitions, and screens become focusable again after going back. Added unit tests for `CardA11yWrapper`; lint, typescript and all `packages/stack` tests pass.


satya164:
thanks for the PR. my main concern is whether people would consider it breaking.

the second thing is if we add it, then:

- it should use a similar approach as 8.x - a reusable view that uses a div on web and applies `inert` declaratively instead of in an effect, which would ensure it also works for SSR
- the view should be used in all navigators, not only stack, so it's handled consistently

end-y:
Thanks for taking a look, makes sense. I gave it a try on stack first.

It's a shared Inert component in elements now, and the attribute is set declaratively instead of in an effect, so SSR is covered. It ended up close to the one you already have in react-native-drawer-layout on main, where the div wraps the content inside the existing View rather than replacing it. I didn't go through ActivityView/Container because that one uses React's Activity, which isn't available on 7.x since it still supports React 18.

One detail: inert is passed as a string rather than a boolean, because React 18 doesn't know the attribute and skips boolean values for it. The value doesn't matter since it's a boolean attribute, the browser only cares that it's there, and I checked that in Chromium, WebKit and Firefox. Browsers without inert support just ignore it, so nothing changes for them.

Does this look like the right shape to you? If it does, I'll do the same for drawer and bottom tabs.

satya164:
thanks for the update @end-y 

the current approach isn't ideal, because `<Inert>` adds a wrapper on web and not on native so there will be inconsistencies. we also need to handle `aria-hidden` separately for native, then `inert` separately for web.

the simpler approach would be to copy [`Container`](https://github.com/react-navigation/react-navigation/blob/main/packages/elements/src/Container.tsx) from `main` and export it in elements named as `UNSTABLE_Container` (root export as we don't have subpath exports in 7.x). `Container` itself doesn't need React 19 features, so it'll be fine.

then we can use it in place of `View` in multiple navigators, header etc. and pass `inert` which would set the a11y props on all platforms.

PR #13198, opened 2026-07-30
fix(native): provide linking context to the fallback element

### LINKED ISSUE #13051: UnhandledLinkingContext throws while linking is not ready (!isLinkingReady branch omits providers)

KallinikosMil:
**Motivation**

Closes #13051

While `NavigationContainer` is restoring state asynchronously (deep link or persisted state), it renders the `fallback` element through an early return. That early-return branch wraps `fallback` in `LocaleDirContext` and `ThemeProvider`, but — unlike the ready branch — it does **not** mount `LinkingContext.Provider`.

`LinkingContext`'s default value throws (`Couldn't find a LinkingContext context.`) as soon as its `options` are read. So anything rendered in `fallback` that touches linking — a `Link`, `useLinkBuilder`, `useLinkProps`, `useRoutePath`, or portaled UI — crashes during the restore window instead of rendering.

**Fix**

Wrap the `fallback` branch in `LinkingContext.Provider` with the same `linkingConfig` value used by the ready branch, so the linking context is consistently available whenever `NavigationContainer` is mounted.

**Test plan**

- Added a test in `NavigationContainer.web.test.tsx`: with a never-resolving `persistor` (container stays on `fallback`), a `fallback` that reads `LinkingContext` now renders instead of throwing `Couldn't find a LinkingContext context.`. It fails on `main` and passes with this change.
- `yarn test`, `yarn lint`, `yarn typecheck` pass locally (pre-commit hook runs the full suite).

### DIFF
--- a/packages/native/src/NavigationContainer.tsx
+++ b/packages/native/src/NavigationContainer.tsx
@@ -246,9 +246,14 @@ export function NavigationContainer<ParamList extends {} = RootParamList>({
     (isLinkStateResolved && isPersistedStateResolved);
 
   if (!isStateReady) {
+    // Keep the same linking context available as in the ready branch, so that
+    // anything rendered in `fallback` (e.g. a `Link`) doesn't hit the default
+    // `LinkingContext` value, which throws.
     return (
       <LocaleDirContext.Provider value={direction}>
-        <ThemeProvider value={theme}>{fallback}</ThemeProvider>
+        <LinkingContext.Provider value={linkingConfig}>
+          <ThemeProvider value={theme}>{fallback}</ThemeProvider>
+        </LinkingContext.Provider>
       </LocaleDirContext.Provider>
     );
   }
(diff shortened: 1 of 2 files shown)


satya164 (changes requested):
thanks for the PR.

it doesn't make sense to wrap fallback in `LinkingContext`. a fallback is for placeholder content and it shouldn't have navigation related logic such as links or components using `useNavigation`, as react navigation can't handle any navigation when the fallback is rendered.

can you please provide a valid use case that needs this or breaks unexpectedly?

KallinikosMil:
Fair point — thanks for the quick review.

I ran into this from an edge case where a `<Link>` ended up inside the fallback tree, but you're right that a fallback is placeholder content and navigation isn't meaningful while it's rendered. I don't have a use case strong enough to justify the change, so I'll close this.

Thanks for taking the time to look at it.

PR #13202, opened 2026-08-04
fix: don't offset the card underneath during the fade transition

### LINKED ISSUE #13128: Android: Header briefly flashes white when going back after overriding NavigationContainer theme colors.card
### Current behavior

Description

When overriding the React Navigation theme and changing colors.card, Android shows a brief white flash in the top-left area of the header during back navigation transitions.

This only happens for a single frame while returning from a pushed screen.

https://github.com/user-attachments/assets/900ccabf-d3fc-4171-8baa-208511dfafdb

Image

### Expected behavior

Back navigation transition should remain visually stable without any white flash or transparent frame.

### Reproduction

https://github.com/chunghe/react-navigation-flicker

### Platform

- [x] Android (pixel 9, android simulator)
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| @react-navigation/native               |7.2.5
| @react-navigation/bottom-tabs          |7.16.2
| @react-navigation/stack                |7.9.3
| react-native                           |0.83.6
| expo                                   |55.0.26
| node                                   |24.14.0
| npm or yarn                            |yarn

TaduJR:
**Motivation**

Closes #13128.

`forFadeFromRightAndroid` (the default on Android 14+) slides the card underneath 96dp while the card above is still fading in. The card's background is inside the transform, so it slides too. Nothing paints behind the cards, so the strip it leaves is empty, and the card above is still see-through, so you end up looking at whatever is behind the navigator. Happens on push and back. You only notice it if `colors.card` doesn't match the window, which is why the issue mentions overriding the theme.

The offset itself is AOSP's, `activity_open_exit` moves the covered activity the same 96dp. But AOSP pairs it with `<extend>`, which outsets the window on the exposed side so that region still has the window's own pixels. We don't have anything like that.

Painting a colour behind the cards doesn't help either. `headerMode` is `screen` on Android, so the strip covers the header and the body and no single colour is right for both. I tried a layer inside the card as an approximation, it fixed iOS but Android still seamed, and it hurt the animation.

I couldn't find a way to keep the offset, so this drops it. The fade and the incoming slide are unchanged and nothing else is touched. Downside is the screen underneath doesn't slide anymore.

Second commit is unrelated, `Card` wasn't clearing the timeout that kicks off its mount animation, so unmounting mid-transition left it firing on a card that's gone.

**Test plan**

Reproduction: https://github.com/TaduJR/react-navigation-stack-header-flash-reproduction

A theme where `colors.card` differs from `colors.background`, two screens, default transition:

1. Push a screen, then go back.
2. Watch the right edge, header down to the bottom.

Before: a strip of the window colour. 

https://github.com/user-attachments/assets/c167c17a-988b-4889-bec3-ead14dc2ceb8

https://github.com/user-attachments/assets/bf376750-61a8-402d-aec9-e1bdcffbf2e9

After: nothing and smooth transition animation.

https://github.com/user-attachments/assets/fee06ba8-98de-4331-9769-c8305e1b2fc9

https://github.com/user-attachments/assets/d30435b1-a439-450a-900a-08c892b3e029


satya164 (changes requested):
thanks for the PR

however changing the animation isn't a proper fix, and is a breaking change. the description even says "Standard Android-style fade from right for Android 14" - but after the change it's not.

also i don't mind LLM assisted PRs, but please try to write the PR description in your own words, and keep it brief. there PR description is very verbose, including a lot of unnecessary details and i still don't think i understood the actual reason for the issue.

TaduJR:
Thanks for the review @satya164. Sorry for the verbosity on the description. I personally worked on the code changes but made a full trust on LLM for the description. I wanted it be descriptive and informative by feeding my research to LLM to produce a better version of it than my hand written one, but it seems not a good idea 😅. Updated the description. 

> however changing the animation isn't a proper fix, and is a breaking change.

I did try to keep the offset. AOSP does the same slide but pairs it with `<extend>` on the side it exposes, and there's no equivalent for that here. Painting a colour behind the cards doesn't work either since `headerMode` is `screen` on Android, so the strip spans two colours. Closest I got was a layer inside the card, which fixed iOS but Android still seamed and it hurt the animation.

Dropping the offset was the only thing that worked on both, but yeah, it changes the animation. Is there a direction you'd rather I go?

satya164:
thanks for updating @TaduJR. i always find human written descriptions much more readable. if something is unclear i can always use LLM or ask for clarification.

regarding the issue, i think there might be a misunderstanding (unless I'm not reading correctly). the original issue is about this strip on the left:

CleanShot 2026-08-07 at 11 00 23@2x

i believe this is not caused by animation offset, as the offset is translating unfocused screen to the right. but in this case it seems the unfocused screen became invisible before the animation finished exposing the the window color on the left side of the focused screen. unsure which version you tested with, if it was react navigation 7, it could be related to react-native-screens.

the animation exposing the window background is a limitation, we could kind of approximate it by using fake 1px width line that adds sections based on card and header background color, but the original issue doesn't seem to be about that.

TaduJR:
I will try to get back on this in the coming week. My schedule have been very tight. Sorry for the delay 🙏.

TaduJR:
@satya164 

I am very sorry about the delay. Also Thank you so much for the review

You're right it's the detach, not the offset.

From my investigation perspective after a push finishes, the screen underneath is detached: `activityState` interpolates `[0, 1 - EPSILON, 1] > [1, 1, 0]`, so it's 0 at progress exactly 1. Going back lifts it off 0 immediately, but re-attaching doesn't land in the same frame that the card above starts translating, so you get a frame of window color on the left.

That's been there since 7.0.0 the interpolation is unchanged through 7.10.24, though the code around it moved out of `render` into `getDerivedStateFromProps` around 7.6. `detachPreviousScreen: false` avoids it. Gone on main, where `CardStack` renders `ActivityView` no flash on device against `8.0.0-alpha.53`.

On 7.x one line does it: `activeScreensLimit = Math.max(activeScreensLimit, 2)`, the same last-two bound main uses. Over 10 backs on 7.10.24 the left edge goes from 10 leaking frames to 0, right edge unaffected (51 vs 50). The rework itself can't be backported `ActivityView` is built on React's `<Activity>`, and main's stack peers `react >= 19.2.0` against 7.10.24's `>= 18.2.0`.

So this PR is a different bug. We can close it and send the unrelated `Card.tsx` unmount fix on its own. Want the backport as a PR?

PR #13211, opened 2026-08-09
fix(stack): defer pop on gesture-driven close until the animation ends

### LINKED ISSUE #13210: Stack swipe-back release hitches on the new architecture: pop is dispatched 16ms into the close animation
On gesture-driven close, `@react-navigation/stack` dispatches the pop **16ms after** the close animation starts ([`Card.tsx`, `GestureState.END`](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/views/Stack/Card.tsx)):

```js
if (shouldClose) {
  // We call onClose with a delay to make sure that the animation has already started
  pendingGestureCallbackRef.current = setTimeout(() => {
    onClose();
    ...
  }, 16);
}
```

That state update (navigator re-render + screen focus changes + view mount/unmount) lands right as the close animation's first frames are being rendered. On the **new architecture**, mounting runs on the main thread — the same thread driving the native-driver animation — so the animation visibly freezes for a beat right after you lift your finger, then continues. Very noticeable jank on a swipe-back release.

Button-initiated pops don't hitch, because the ordering is reversed: the pop (and all its React work) happens **before** the animation starts — the screen stays rendered via `closingRouteKeys` and animates out cleanly afterwards.

### Proposed fix

Remove the 16ms `setTimeout` entirely and let the pop come from `animate()`'s `onFinish → onClose` when the close animation **ends**. `StackView.handleCloseRoute` already supports exactly this path — its comment even documents it:

```js
if (state.routes.some((r) => r.key === route.key)) {
  // If a route exists in state, trigger a pop
  // This will happen in when the route was closed from the card component
  // e.g. When the close animation triggered from a gesture ends
  navigation.dispatch({ ...StackActions.pop(), ... });
}
```

This mirrors the button ordering: motion first, React work after — never both competing for the main thread.

**Trade-off:** navigation state (URL, focus events, listeners) updates ~one transition-duration later than today for gesture pops. That's the same ordering a button pop produces, just mirrored, and it matches what the user sees on screen.

### Verified on device

iPhone, RN 0.81.5, `newArchEnabled: true`, expo-router 6 / Expo SDK 54, `@react-navigation/stack` 7.10.17. Applied via `patch-package` (delete the `setTimeout` block, keep `onGestureEnd?.()`): the release hitch disappears completely; swipe-back release is now indistinguishable from the button-back animation. The same code exists unchanged in 7.10.22 and 8.0.0-alpha.50.

Happy to send a PR — either removing the timeout outright or gating it behind an option if you prefer to keep the current default.

sromero1905:
Fixes #13210

## Problem

On a gesture-driven close, the pop is dispatched via a 16ms `setTimeout` right after the close animation starts. That state update (navigator re-render, focus changes, view unmounts) lands mid-animation. On the new architecture, mounting work runs on the UI thread — the same thread driving the native-driver animation — so the animation visibly freezes for a beat right after the finger lifts, then continues.

`goBack`-triggered pops never hitch because their ordering is reversed: the state update happens *before* any motion (the screen stays rendered via `closingRouteKeys` and animates out afterwards).

## Fix

Remove the early dispatch and let the pop come from `animate`'s `onFinish` when the close animation ends — a path `StackView.handleCloseRoute` already supports and documents:

```js
// If a route exists in state, trigger a pop
// This will happen in when the route was closed from the card component
// e.g. When the close animation triggered from a gesture ends
```

Motion first, React work after — the button ordering, mirrored. The `pendingGestureCallbackRef`/`pendingOnCloseCallbackRef` timers existed only to schedule the early pop and its re-open check, so they're removed; the re-open safety remains covered by `maybeAnimate` in `onFinish`.

## Trade-off

Navigation state (URL, focus events, listeners) updates when the close animation ends rather than ~16ms after release. That matches what's on screen, and is the same ordering `goBack` produces.

## Test plan

Verified on device (iPhone, RN 0.81.5, `newArchEnabled: true`, Expo SDK 54 + expo-router 6) with `@react-navigation/stack` 7.10.17 patched via `patch-package` with this change: the release hitch disappears entirely — swipe-back release becomes indistinguishable from the `goBack` animation. Repro repo linked in the issue.

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -156,10 +156,6 @@ function Card({
   const lastToValueRef = React.useRef<number | undefined>(undefined);
 
   const animationHandleRef = React.useRef<number | undefined>(undefined);
-  const pendingGestureCallbackRef =
-    React.useRef<ReturnType<typeof setTimeout>>(undefined);
-  const pendingOnCloseCallbackRef =
-    React.useRef<ReturnType<typeof setTimeout>>(undefined);
 
   const [isClosing] = React.useState(() => new Animated.Value(FALSE));
 
@@ -200,8 +196,6 @@ function Card({
       const animation =
         spec.animation === 'spring' ? Animated.spring : Animated.timing;
 
-      clearTimeout(pendingGestureCallbackRef.current);
-
       if (animationHandleRef.current !== undefined) {
         cancelAnimationFrame(animationHandleRef.current);
       }
@@ -232,8 +226,6 @@ function Card({
           useNativeDriver,
           isInteraction: false,
         }).start(({ finished }) => {
-          clearTimeout(pendingGestureCallbackRef.current);
-
           if (finished) {
             onFinish();
           }
@@ -274,18 +266,12 @@ function Card({
       if (animationHandleRef.current) {
         cancelAnimationFrame(animationHandleRef.current);
       }
-
-      clearTimeout(pendingGestureCallbackRef.current);
-      clearTimeout(pendingOnCloseCallbackRef.current);
     };
   }, []);
 
   const timeoutRef = React.useRef<ReturnType<typeof setTimeout>>(undefined);
 
   const maybeAnimate = useLatestCallback(() => {
-    clearTimeout(pendingGestureCallbackRef.current);
-    clearTimeout(pendingOnCloseCallbackRef.current);
-
     if (!didInitiallyAnimate.current) {
       // Animate the card in on initial mount
       // Wrap in setTimeout to ensure animation starts after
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR. However, the code you removed has a comment saying why it's necessary.

The delayed check animates the screen back if it wasn't removed, e.g., in `beforeRemove` event. So this change will break this functionality and wait till animation ends, which may feel unnatural.

Please also provide a repro to test the issue to see if there's a different approach.

Generally, we don't recommend doing expensive work during animation. Especially if a screen is closing, it's best to leave everything as is as the screen will be removed anyway.

You can use `transitionStart` and `transitionEnd` events to see whether an animation is currently happening.

PR #13214, opened 2026-08-12
fix: preserve opening animation for stacked screens

### LINKED ISSUE #13213: Second stacked modal skips its opening animation on Android
### Current behavior

On Android with `@react-navigation/stack`, opening a second modal above an existing modal does not visibly animate. The second modal appears immediately at its final position, even though:

- closing the second modal animates normally;
- `transitionStart` and `transitionEnd` report a normal opening transition of approximately 260 ms; and
- the same configuration animates correctly for the first modal.

The minimal flow is:

1. Open `modal-1` from `Home`.
2. Open `modal-2` above `modal-1` using the same route name and distinct `getId` values.
3. Observe that `modal-2` appears without its bottom-sheet opening animation.

The route options are:

```tsx
{
  animation: 'default',
  cardStyleInterpolator: CardStyleInterpolators.forBottomSheetAndroid,
  gestureEnabled: false,
  presentation: 'modal',
}
```

This regressed in `@react-navigation/stack@7.6.7`:

- `7.6.6`: opening animation works;
- `7.6.7`: opening animation is skipped; and
- `7.10.22`: still affected.

The `7.6.6...7.6.7` comparison contains [commit d98552bced](https://github.com/react-navigation/react-navigation/commit/d98552bced693a8ed04a5b9b4952838bc8660a65), which removed the strict screen-depth guard while fixing blank screens. Without that guard, a deeply covered route can receive an animated activity state during the second modal push instead of remaining numerically inactive, changing the native screen composition during the transition.

This logic originated in [PR #11315](https://github.com/react-navigation/react-navigation/pull/11315), which prevents an already inactive screen from briefly reattaching during `A → B → C → popTo(B)`.

Setting `detachPreviousScreen: false` makes the opening animation visible, but it permanently retains the previous screen and is not a safe workaround.

### Expected behavior

Each modal should animate in and out with the configured interpolator, including when another modal is already open, without reattaching deeply covered screens or reintroducing blank frames during pop, replacement, or rapid navigation.

### Reproduction

https://github.com/hgray-instawork/react-navigation-js-stack-modal-repro

### Platform

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [x] @react-navigation/stack

### Environment

| package                        | version |
| @react-navigation/native       | 7.3.16  |
| @react-navigation/stack        | 7.10.22 |
| react-native-screens           | 4.11.1  |
| react-native-safe-area-context | 5.4.0   |
| react-native-gesture-handler   | 2.24.0  |
| react-native                   | 0.79.6  |
| expo                           | 53.0.27 |
| node                           | 24.14.1 |
| npm                            | 11.11.0 |

hgray-instawork:
### Motivation
Opening another screen over an already-covered JS stack can skip its visible opening animation on Android, even though transition events run for the expected duration.

This regressed in `@react-navigation/stack@7.6.7` when [d98552bced](https://github.com/react-navigation/react-navigation/commit/d98552bced693a8ed04a5b9b4952838bc8660a65) removed the strict depth guard while fixing blank screens.

The guard came from [#11315](https://github.com/react-navigation/react-navigation/pull/11315), which prevents inactive routes from briefly reattaching during transitions such as `A → B → C → popTo(B)`. Restoring it unconditionally fixes the skipped animation but reintroduces blank frames in overlapping transitions.

### Approach
This change handles the activity state according to the transition lifecycle rather than checking for modal presentation:
- Apply the depth guard only when exactly one route is opening, it is the focused route, and no route is closing.
- Keep deeply covered screens numerically inactive during that opening transition.
- Preserve the existing previous-activity-state guard in other transition states.
- After a route is removed, recreate the activity interpolation immediately below the active window so that screen can appear during a later close.
There are no checks for `presentation`, modal interpolators, or modal-specific options.
This allows the incoming screen to animate without reactivating deeply covered screens, while preserving the interpolation needed to reveal the correct underlying screen during subsequent closing transitions.

| Before | After |
| -- | -- |
| before | after |

Fixes #13213
### Validation
Revalidated with:
- `@react-navigation/stack@7.10.23`
- `@react-navigation/native@7.3.17`
- React Native 0.79.6
- Android 16 emulator

Stock `7.10.23` still reproduces the skipped second opening animation. The updated implementation preserves the animation without reintroducing blank frames.

Validated scenarios:
- `Home → modal-1 → modal-2`
- Closing `modal-2`, followed by `modal-1`
- Equivalent non-modal vertical card transitions
- `A → B → C → popTo(B)`
- Push-style replacement ordering
- Reset with shifted route indices
- Multi-pop route ordering
- Overlapping opening transitions
- Simultaneous opening and closing routes
- Rapid heavy-render navigation sequences

A local-only `CardStack` activity-state harness covered the index-sensitive scenarios above. All seven cases passed; the temporary harness is not included in this PR.
Repository checks also passed:
- 752 tests
- TypeScript
- Declaration generation
- ESLint

### Blank-frame comparison:

Before historical fix (`7.6.6`):

Stock 7.6.6 blank-frame contact sheet

After historical fix (`7.6.7`):

Stock 7.6.7 contact sheet

Modal regression and patched behavior:

- [Stock `7.10.22` recording](https://github.com/user-attachments/assets/4cb55efc-e202-45be-96e0-0700203dad3c)
- [Patched `7.10.22` recording](https://github.com/user-attachments/assets/d352625f-8c0b-490b-8753-a1c625974cc8)

### Test plan
Android:
- Verify both opening transitions in `Home → modal-1 → modal-2`.
- Close `modal-2`, then `modal-1`, and verify the correct underlying route remains visible.
- Verify the same behavior with a non-modal vertical transition.
- Verify `A → B → C → popTo(B)` without a blank frame.
- Exercise replacement, reset, multi-pop, and overlapping transition sequences.
- Run the heavy-render rapid-navigation sequence and inspect the recorded frames.

Shared-platform regression check:
- Verify single and stacked JS-stack transitions animate correctly on iOS.

### DIFF
--- a/packages/stack/src/views/Stack/CardStack.tsx
+++ b/packages/stack/src/views/Stack/CardStack.tsx
@@ -506,6 +506,10 @@ export class CardStack extends React.Component<Props, State> {
         }
       }
 
+      const isTopScreenModal =
+        scenes[props.routes.length - 1]?.descriptor.options.presentation ===
+        'modal';
+
       activeStates = props.routes.map((_, index, self) => {
         // The activity state represents state of the screen:
         // 0 - inactive, the screen is detached
@@ -519,9 +523,19 @@ export class CardStack extends React.Component<Props, State> {
 
         const lastActiveState = state.activeStates[index];
         const activeAfterTransition = index >= self.length - activeScreensLimit;
+        // Recreate the activity interpolation below a remaining modal.
+        const shouldReactivateForModal =
+          props.routes.length < state.routes.length &&
+          index === self.length - activeScreensLimit - 1 &&
+          isTopScreenModal;
 
-        if (lastActiveState === STATE_INACTIVE && !activeAfterTransition) {
-          // screen was inactive before and it will still be inactive after the transition
+        if (
+          (isTopScreenModal && index < self.length - activeScreensLimit - 1) ||
+          (lastActiveState === STATE_INACTIVE &&
+            !activeAfterTransition &&
+            !shouldReactivateForModal)
+        ) {
+          // screen is deeply covered by a modal, or was inactive and will remain inactive
           activityState = STATE_INACTIVE;
         } else {
           const sceneForActivity = scenes[self.length - 1];
--- a/packages/stack/src/views/Stack/CardStack.tsx
+++ b/packages/stack/src/views/Stack/CardStack.tsx
@@ -506,9 +506,11 @@ export class CardStack extends React.Component<Props, State> {
         }
       }
 
-      const isTopScreenModal =
-        scenes[props.routes.length - 1]?.descriptor.options.presentation ===
-        'modal';
+      const focusedRouteKey = props.state.routes[props.state.index]?.key;
+      const isSingleOpeningTransition =
+        props.openingRouteKeys.length === 1 &&
+        props.openingRouteKeys[0] === focusedRouteKey &&
+        props.closingRouteKeys.length === 0;
 
       activeStates = props.routes.map((_, index, self) => {
         // The activity state represents state of the screen:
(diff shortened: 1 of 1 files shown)


satya164 (changes requested):
Thanks for the PR. But it's not correct to have special logic only for modals. For JS stack, it's just a vertical animation. So the bug could happen with any animation.

The logic needs to be generic instead of special-casing modals to handle it properly.

hgray-instawork:
Thanks for the review @satya164. I've updated the logic to use transition state rather than presentation type. This still fixes the stacked modal case and also works with other JS stack animations; I verified it with a card presentation using the same vertical interpolator.

hgray-instawork:
@satya164 is there anything else you need from me on this?

PR #13217, opened 2026-08-14
fix: prevent state reset when history.go() takes longer than timeout on web

### LINKED ISSUE #11145: Navigation state is reset after calling navigation.navigate on Firefox
### Current behavior

While on Firefox, when I call `navigation.navigate`, the navigation state is reset. This does not happen on Chrome or Safari.
After looking at `useLinking.tsx`, specifically at this block code:

```
// If navigation didn't happen within 100ms, assume that it won't happen
// This may not be accurate, but hopefully it won't take so much time
// In Chrome, navigation seems to happen instantly in next microtask
// But on Firefox, it seems to take much longer, around 50ms from our testing
// We're using a hacky timeout since there doesn't seem to be way to know for sure
const timer = setTimeout(() => {
  const index = pending.findIndex((it) => it.ref === done);

          if (index > -1) {
    pending[index].cb();
    pending.splice(index, 1);
  }
}, 100);
```

I found out that our navigation takes longer than 100ms to complete on Firefox. Once I increase that timeout, the navigation works fine. Would it be possible to make it configurable?

### Expected behavior

Navigation is not reset after calling `navigation.navigate`

### Reproduction

https://snack.expo.dev/

### Platform

- [ ] Android
- [ ] iOS
- [X] Web
- [ ] Windows
- [X] MacOS

### Packages

- [X] @react-navigation/drawer
- [X] @react-navigation/stack
- [ ] react-native-tab-view
- [ ] flipper-plugin-react-navigation

### Environment

| @react-navigation/native               | 5.9.8 |
| @react-navigation/bottom-tabs          | 5.11.15 |
| @react-navigation/drawer               | 5.12.19 |
| @react-navigation/stack                | 5.12.3 |
| react-native-safe-area-context         | 4.4.1 |
| react-native-screens                   |  3.18.2 |
| react-native-gesture-handler           | 2.8.0 |
| react-native-reanimated                | 2.12.0 |
| react-native-tab-view                  | 3.1.1 |
| react-native-pager-view                | 6.0.1 |
| react-native                           | 0.70.5
| expo                                   | 47.0.8 |
| node                                   | 14.20.1 |
| yarn                            | 1.22.19 |

collectioneur:
**Motivation**

Fixes #11145

On web, `createMemoryHistory().go(n)` calls `window.history.go(n)` and waits for the `popstate` event to know when the traversal has finished. Since the browser provides no API to detect a traversal that will never happen, a 100ms timeout was used as a fallback: if `popstate` didn't arrive in time, the navigation was assumed to have failed.

The problem is that the 100ms assumption doesn't hold on Firefox. A real-world example of the impact: [Expensify/App#94571](https://github.com/Expensify/App/issues/94571) (a newly created chat immediately closes on Firefox because the late `popstate` resets navigation to the previous screen), which was fixed by patching this exact timeout in [Expensify/App#95980](https://github.com/Expensify/App/pull/95980) ([the patch](https://github.com/Expensify/App/blob/main/patches/react-navigation/%40react-navigation%2Bnative%2B7.1.33%2B003%2Bincrease-history-go-popstate-fallback-timeout.patch)). When the timeout fired before the traversal finished:

1. The pending `go()` was rejected and removed from the `pending` queue.
2. The late `popstate` then arrived with an empty `pending` queue, so `listen()` misclassified it as a user-initiated back/forward navigation and reset the app to a stale route - visible as lost screens/state after calling `navigation.navigate()` or going back on Firefox.

This PR increases the timeout to 1000ms. Besides covering the measured traversal times, keeping the `go()` promise pending longer has a second benefit: navigation state updates are serialized, so while the promise is pending, subsequent `pushState`/`replaceState` calls are queued instead of racing against the in-flight traversal. Without this, a user navigating in the app during the stuck traversal could end up with the browser history modified underneath the traversal, which lands on an unpredictable entry (deltas are resolved at execution time).

**Trade-off**

The timeout is only ever waited out in full when `popstate` never arrives - i.e. when the in-memory history is out of sync with the real browser history (edge cases around reloads/session restore). In that already broken case, the URL bar sync now lags up to 1s instead of 0.1s, on all browsers. The app itself is unaffected: screens render immediately and user interactions are never blocked. On Chrome the traversal completes in the next microtask, so the timeout value is effectively never reached in normal operation.

A timeout of any length remains a heuristic: a traversal delayed beyond 1000ms would still be misclassified. This is not fully fixable with the History API (no way to detect or cancel a pending traversal). The section below explains the proper long-term fix.

**In future: Navigation API**

The [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API) removes the need for this heuristic entirely:

- [`navigation.traverseTo(key)`](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/traverseTo) targets a specific entry and returns `{ committed, finished }` promises, so both the "late popstate" and the "traversal never happens" cases are detected deterministically.
- The `navigate` event distinguishes traversal sources, removing the `pending`-based classification in `listen()`.

The API is baseline newly available since January 2026 (Chrome/Edge 102+, Firefox 147+, Safari 26.2+). Since older browsers will need the `popstate` fallback for years, adopting it would be a progressive enhancement and is left as a follow-up.

**Test plan**

- Unit test: `packages/native/src/__tests__/createMemoryHistory.web.test.tsx` - new test `resolves go when popstate arrives before the timeout` simulates a traversal that takes 600ms, it fails with the previous 100ms timeout and passes with this change.

### DIFF
--- a/packages/native/src/__tests__/createMemoryHistory.web.test.tsx
+++ b/packages/native/src/__tests__/createMemoryHistory.web.test.tsx
@@ -250,7 +250,7 @@ test('can go back in browser history after a previous attempt failed', async () 
 
   const timedOutNavigation = history.go(-1);
 
-  jest.advanceTimersByTime(100);
+  jest.advanceTimersByTime(1000);
 
   await expect(timedOutNavigation).rejects.toThrow(
     'History was changed during navigation.'
@@ -271,7 +271,7 @@ test('can go back in browser history after a previous attempt failed', async () 
   unlisten();
 });
 
-test("doesn't treat late popstate from own traversal as user navigation", async () => {
+test('resolves go when popstate arrives before the timeout', async () => {
   jest.useFakeTimers();
 
   const state: NavigationState = {
@@ -290,36 +290,28 @@ test("doesn't treat late popstate from own traversal as user navigation", async 
   const listener = jest.fn();
   const unlisten = history.listen(listener);
 
-  // Simulate a slow traversal (e.g. Firefox when the main thread is busy)
-  // where `popstate` doesn't fire before the timeout in `go`
+  // Simulate a slow traversal, e.g. on Firefox `history.go(n)` can take
+  // several hundred milliseconds when the main thread is busy
+  const originalGo = window.history.go.bind(window.history);
   const windowGoSpy = jest
     .spyOn(window.history, 'go')
-    .mockImplementation(() => {});
+    .mockImplementation((delta) => {
+      setTimeout(() => originalGo(delta), 600);
+    });
 
-  const timedOutNavigation = history.go(-1);
+  const navigation = history.go(-1);
 
-  jest.advanceTimersByTime(100);
-
-  await expect(timedOutNavigation).rejects.toThrow(
-    'History was changed during navigation.'
-  );
-
-  windowGoSpy.mockRestore();
-
-  // The browser performs the traversal after the timeout and fires a late `popstate`
-  window.history.go(-1);
+  jest.advanceTimersByTime(600);
   jest.runAllTimers();
 
-  // The late `popstate` is from our own traversal, so it shouldn't call the listener
+  // The traversal arrived before the timeout, so it should resolve
+  // and not be treated as a user-initiated navigation
+  await expect(navigation).resolves.toBeUndefined();
+
   expect(listener).not.toHaveBeenCalled();
   expect(history.index).toBe(0);
 
-  // A user-initiated navigation afterwards should still call the listener
-  window.history.go(1);
-  jest.runAllTimers();
-
-  expect(listener).toHaveBeenCalledTimes(1);
-  expect(history.index).toBe(1);
+  windowGoSpy.mockRestore();
 
   unlisten();
 });
--- a/packages/native/src/__tests__/createMemoryHistory.web.test.tsx
+++ b/packages/native/src/__tests__/createMemoryHistory.web.test.tsx
@@ -270,3 +270,56 @@ test('can go back in browser history after a previous attempt failed', async () 
 
   unlisten();
 });
+
+test("doesn't treat late popstate from own traversal as user navigation", async () => {
+  jest.useFakeTimers();
+
+  const state: NavigationState = {
+    key: 'stack-123',
+    index: 0,
+    routeNames: ['One'],
+    routes: [{ name: 'One', key: 'One-23' }],
+    type: 'stack',
+    stale: false,
+  };
+  const history = createMemoryHistory();
+
+  history.replace({ path: '/route-one', state });
+  history.push({ path: '/route-two', state });
+
+  const listener = jest.fn();
+  const unlisten = history.listen(listener);
+
+  // Simulate a slow traversal (e.g. Firefox when the main thread is busy)
+  // where `popstate` doesn't fire before the timeout in `go`
+  const windowGoSpy = jest
+    .spyOn(window.history, 'go')
+    .mockImplementation(() => {});
+
+  const timedOutNavigation = history.go(-1);
+
+  jest.advanceTimersByTime(100);
+
+  await expect(timedOutNavigation).rejects.toThrow(
+    'History was changed during navigation.'
+  );
+
+  windowGoSpy.mockRestore();
+
+  // The browser performs the traversal after the timeout and fires a late `popstate`
+  window.history.go(-1);
+  jest.runAllTimers();
+
+  // The late `popstate` is from our own traversal, so it shouldn't call the listener
+  expect(listener).not.toHaveBeenCalled();
+  expect(history.index).toBe(0);
+
+  // A user-initiated navigation afterwards should still call the listener
+  window.history.go(1);
+  jest.runAllTimers();
+
+  expect(listener).toHaveBeenCalledTimes(1);
+  expect(history.index).toBe(1);
+
+  unlisten();
+});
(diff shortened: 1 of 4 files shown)


collectioneur:
@satya164 I didn't add a repro to the example app - it could only artificially simulate
a busy main thread anyway (real-world case linked in the description).
Happy to add one and/or an e2e test if you think it's useful 🙂

packages/native/src/__tests__/createMemoryHistory.web.test.tsx
```diff
@@ -270,3 +270,44 @@ test('can go back in browser history after a previous attempt failed', async ()
...
+  const windowGoSpy = jest
+    .spyOn(window.history, 'go')
+    .mockImplementation((delta) => {
+      setTimeout(() => originalGo(delta), 600);
+    });
+
+  const navigation = history.go(-1);
+
+  jest.advanceTimersByTime(600);
+  jest.runAllTimers();
```
satya164: if you're running all timers either way, `advanceTimersByTime` is redundant. remove all timers and use `advanceTimersByTime` to explictly advance by time

packages/native/src/__tests__/createMemoryHistory.web.test.tsx
```diff
@@ -270,3 +270,44 @@ test('can go back in browser history after a previous attempt failed', async ()
...
+
+  await expect(navigation).resolves.toBeUndefined();
+
+  expect(listener).not.toHaveBeenCalled();
+  expect(history.index).toBe(0);
+
+  windowGoSpy.mockRestore();
```
satya164: lets add `restoreAllMocks` in `afterEach` at the top-level instead.

packages/native/src/createMemoryHistory.tsx
```diff
@@ -189,10 +189,13 @@ export function createMemoryHistory() {
...
 
-        // If navigation didn't happen within 100ms, assume that it won't happen
+        // If navigation didn't happen within 1000ms, assume that it won't happen
         // This may not be accurate, but hopefully it won't take so much time
         // In Chrome, navigation seems to happen instantly in next microtask
-        // But on Firefox, it seems to take much longer, around 50ms from our testing
+        // But on Firefox, the traversal can take much longer when the main thread
+        // is busy, e.g. it was measured to take up to ~900ms while rendering a new screen
+        // While the promise is pending, subsequent `pushState`/`replaceState` are queued,
+        // which avoids racing them against the in-flight traversal
```
satya164: this happens in a different place so this comment can get out of sync. lets remove these lines.

packages/native/src/__tests__/createMemoryHistory.web.test.tsx
```diff
@@ -270,3 +270,44 @@ test('can go back in browser history after a previous attempt failed', async ()
 
   unlisten();
 });
+
+test('resolves go when popstate arrives before the timeout', async () => {
```
satya164: I'm not sure this test is useful. it tests an implementation detail rather than expected behavior

satya164 (commented):
Thanks for the PR. We should add some tests to ensure the behavior:

I came up with these tests with help of Codex. Lets add them to `useLinking.web.test.tsx`:

1. `keeps the latest navigation when programmatic back is delayed`  
   Delay `history.go(-1)` by 600 ms, call `navigation.goBack()`, then navigate to another screen. Assert the latest screen, URL, and back history remain correct.

2. `queues replace while programmatic back is delayed`  
   Start the same delayed back, then dispatch `StackActions.replace` or update route params. Assert the replacement affects the correct entry without adding history.

3. `applies multiple updates queued during delayed history traversal`  
   Queue multiple navigations while the traversal is pending. Assert the final route, URL, and browser history match the update order.

4. `syncs a queued navigation after history traversal times out`  
   Make `window.history.go` emit no `popstate`, navigate elsewhere while pending, advance 1000 ms, and assert the visible route updates immediately and the URL eventually catches up.

5. `preserves the latest navigation when popstate arrives after the timeout`  
   Delay traversal beyond 1000 ms, allow a queued navigation to sync, then deliver the late event. Assert it does not reset the app. This likely exposes the remaining bug.

6. `rolls back prevented browser back when forward traversal is delayed`  
   Use `usePreventRemove`, delay the positive `history.go` used for rollback, and assert the route and URL stay on the protected screen.

7. `syncs a delayed multi-entry programmatic pop`  
   Pop multiple stack entries with a delayed traversal. Assert the destination route, URL, and forward history are correct.

8. `handles browser navigation during a delayed programmatic traversal`  
   Trigger browser back or forward while the app’s `go()` is pending. Assert neither action is lost or applied twice.

Also update the existing `preserves history entries when traversal is slower than the fallback timeout` test. Its 150 ms delay is now below the 1000 ms timeout, so it no longer tests its stated scenario.

Re: `Navigation` API, I wanted to use it but it's too new. Also using it means we'd no longer be testing our example with the history API which is the common case, so I'd wait sometime before using it.

collectioneur:
All comments are resolved ✅  

A quick note on the `advanceTimersByTime(100)` calls in the tests:
Since jsdom runs `history.go()` asynchronously, the traversal and popstate dispatch are queued via `setTimeout(..., 0)` and don't fit into the `advanceTimersByTime(600)` window. The extra advance lets jsdom deliver the popstate (it just needs to stay under the 1000ms fallback timeout). The same applies to `advanceTimersByTime(1)` after `history.back()`.

As for the missing test (preserving the latest navigation when popstate arrives after the timeout):
I skipped adding it because it currently fails. This PR mitigates the issue in practice by raising the timeout, but the core bug remains: if popstate arrives late, the History API can't tell it apart from a user-initiated navigation. A proper fix will probably have to wait for the Navigation API, which is realistically a few years away given browser support. 
LMK if you'd like me to add it as `test.failing`

satya164 (approved):
thanks

PR #13218, opened 2026-08-15
fix(stack): silence onAnimatedValueUpdate warning by registering gesture listener

### LINKED ISSUE #11564: Sending `onAnimatedValueUpdate` with no listeners registered `warning` when swiping back using ios gesture. JS Stack Navigator
### Current behavior

This is only happening on JS Stack Navigator 

I keep getting warning that says "Sending `onAnimatedValueUpdate` with no listeners registered" every time i'm swiping back to the previous screen using ios gesture. 

I'm not sure why since i'm following the installation and setup in the docs correctly. 

## Screenshot

image

## Video
https://github.com/react-navigation/react-navigation/assets/46083126/b69c46af-49d4-48e5-a605-c5aa9d79685e

### Expected behavior

There should not be warning for gesture back behavior. This is sometimes really annoying when debugging or developing my apps.

### Reproduction

https://github.com/bryanltobing/rnav-test

### Platform

- [ ] Android
- [X] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [X] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| @react-navigation/native               | 6.1.7
| @react-navigation/stack                | 6.3.17
| expo                                   | 49.0.8
| node                                   | 20.3.1
| npm or yarn                            | yarn 1.22.19

ErfanBagheri404:
Fixes #11564

When the swipe-back gesture updates the `Animated.Value` with `useNativeDriver` on iOS, React Native warns about `onAnimatedValueUpdate` being sent with no listeners registered. This happens on every swipe-back gesture in the JS Stack navigator.

The fix attaches a no-op listener to the gesture `Animated.Value` on mount and removes it on unmount, so the native side knows JS is observing the value and stops emitting the warning.

### DIFF
--- a/packages/stack/src/views/Stack/Card.tsx
+++ b/packages/stack/src/views/Stack/Card.tsx
@@ -362,6 +362,20 @@ function Card({
     maybeAnimate,
   ]);
 
+  React.useEffect(() => {
+    // Register a no-op listener on the gesture value.
+    // When the gesture is updated from the native side (e.g. on iOS swipe-back)
+    // with `useNativeDriver`, React Native warns about `onAnimatedValueUpdate`
+    // being sent with no listeners registered. Attaching a listener tells the
+    // native side that JS is observing the value, silencing the warning.
+    // cf. https://github.com/react-navigation/react-navigation/issues/11564
+    const id = gesture.addListener(() => {});
+
+    return () => {
+      gesture.removeListener(id);
+    };
+  }, [gesture]);
+
   const interpolationProps = React.useMemo(
     () => ({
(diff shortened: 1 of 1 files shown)


satya164:
Thanks for the PR. But I don't want to add an empty listener as we don't want to receive the updates on JS every frame and do extra work.

Listening to animated value only natively is a valid use case. This seems to be a React Native issue where it unnecessarily sends the value and prints a warning.

ErfanBagheri404:
i agree on your take, if you want i can close the PR. 
thanks for the review!

PR #13221, opened 2026-08-18
fix(native): accept SectionList refs in useScrollToTop under RN 0.87 strict types

### LINKED ISSUE #13220: RN 0.87, TypeScript – `useScrollToTop` no longer accepts `SectionList` refs
### Current behavior

Passing a `SectionList` ref (or `SectionListInstance`) to `useScrollToTop` produces a TypeScript error under React Native 0.87. The same code works fine for `ScrollView` and `FlatList`.

### Minimal repro

```tsx
import { useRef } from 'react';
import { ScrollViewInstance, FlatListInstance, SectionList, SectionListInstance } from 'react-native';
import { useScrollToTop } from '@react-navigation/native';

const App = () => {
  const scrollViewRef = useRef<ScrollViewInstance>(null);
  useScrollToTop(scrollViewRef); // All good

  const flatListRef = useRef<FlatListInstance>(null);
  useScrollToTop(flatListRef); // All good

  const ref1 = useRef<SectionListInstance>(null);
  useScrollToTop(ref1); // Error TS2345

  const ref2 = useRef<SectionList>(null);
  useScrollToTop(ref2); // Also Error TS2345

  return <SectionList ref={ref1} sections={[]} />;
};
```

Error produced:

```
App.tsx(21,18): error TS2345: Argument of type 'RefObject<SectionListInstance | null>' is not assignable to parameter of type 'RefObject<ScrollableWrapper>'.
  Type 'SectionListInstance | null' is not assignable to type 'ScrollableWrapper'.
    Type 'SectionListInstance' is not assignable to type 'ScrollableWrapper'.
      Type 'SectionList<any, DefaultSectionT>' is not assignable to type '{ getScrollResponder(): ReactNode | (((props: ...) => ReactNode) & Readonly<...>); }'.
        The types returned by 'getScrollResponder()' are incompatible between these types.
          Type 'ScrollViewImperativeMethods | null | undefined' is not assignable to type 'ReactNode | (((props: ...) => ReactNode) & Readonly<...>)'.
            Type 'ScrollViewImperativeMethods' is not assignable to type 'ReactNode | (((props: ...) => ReactNode) & Readonly<...>)'.
```

### Root cause

RN 0.87 enables the [strict TypeScript API](https://reactnative.dev/blog/2025/08/12/react-native-0.87#strict-typescript-api) by default.
* With legacy types, `ScrollView` was a class, and the type `ScrollView` referred to the **instance** (which structurally includes `ScrollViewImperativeMethods`).
* With strict types, `ScrollView` is now the **component** type (`typeof ScrollViewWrapper & ScrollViewComponentStatics`). The instance type is a separate export, `ScrollViewInstance`.

`useScrollToTop`'s `ScrollableWrapper` type relies on the old meaning of `ScrollView`:
https://github.com/react-navigation/react-navigation/blob/cec5e05aa7f329f624a34858f8d89c27daa72a87/packages/native/src/useScrollToTop.tsx#L19-L22

Under the strict API, `SectionList.getScrollResponder()` returns `ScrollResponderType` (= `ScrollViewImperativeMethods`), which is no longer assignable to `React.ReactNode | ScrollView` because `ScrollView` now means the *component*, not the instance.

Why only `SectionList` breaks: `ScrollableWrapper` is a union, and `ScrollView`/`FlatList` match the `ScrollableView` branch structurally (`ScrollView` has `scrollTo`, `FlatList` has `scrollToOffset`). `SectionList` only exposes `scrollToLocation`, so it matches none of the `ScrollableView` members and falls back to the `getScrollResponder` branch, which is the one broken by the type change.

### Expected behavior

`useScrollToTop` should accept a `SectionList` / `SectionListInstance` ref without a type error, consistent with `ScrollView` and `FlatList`.

### Reproduction

https://github.com/SimpleCreations/SEE_CURRENT_BEHAVIOR

### Platform

- [ ] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| package                  | version  |
| ------------------------ | -------- |
| @react-navigation/native | 7.3.16   |
| react-native             | 0.87.0   |
| react                    | 19.2.3   |
| @types/react             | 19.2.18  |
| typescript               | 6.0.3    |
| node                     | 24.19.0  |

ErfanBagheri404:
### Problem

RN 0.87 enables the strict TypeScript API by default. `ScrollView` is now the component type (`typeof ScrollViewWrapper`), not the instance type. When a `SectionList` ref is passed to `useScrollToTop`, it falls through to the `getScrollResponder()` branch in `ScrollableWrapper`, which fails to type-check because `SectionList.getScrollResponder()` returns `ScrollResponderType` that no longer matches `React.ReactNode | ScrollView`.

### Fix

Add `scrollToLocation` to the `ScrollableView` union so `SectionList` matches directly on the `ScrollableView` branch, avoiding the broken `getScrollResponder` path entirely. This also adds the runtime `scrollToLocation` handler in the scroll-to-top logic.

### Reproduction

```tsx
const ref = useRef<SectionListInstance>(null);
useScrollToTop(ref); // TS2345 before this fix
```

Fixes #13220


packages/native/src/__tests__/useScrollToTop.test.tsx
```diff
@@ -0,0 +1,165 @@
+import { expect, jest, test } from '@jest/globals';
```
satya164: please remove this test. it doesn't test anything useful, but exact code that was just added.

packages/native/src/useScrollToTop.tsx
```diff
@@ -103,6 +110,12 @@ export function useScrollToTop(ref: React.RefObject<ScrollableWrapper>) {
...
               } else if ('scrollResponderScrollTo' in scrollable) {
                 scrollable.scrollResponderScrollTo({ y: 0, animated: true });
+              } else if ('scrollToLocation' in scrollable) {
+                scrollable.scrollToLocation({
+                  sectionIndex: 0,
+                  itemIndex: 0,
+                  animated: true,
+                });
               }
             }
```
satya164: move `scrollToLocation` to before `scrollResponderScrollTo`. `scrollResponderScrollTo` should be the last fallback.

satya164 (changes requested):
thanks for the pr. i have a couple of changes

satya164:
from testing, it seems the only change we need is the return type of `getScrollResponder`. `scrollToLocation` may also throw for empty lists so it's not safe to call unconditionally. so I have updated the PR.

PR #13222, opened 2026-08-22
fix(stack): register gesture listener to silence onAnimatedValueUpdate warning

### LINKED ISSUE #11564: Sending `onAnimatedValueUpdate` with no listeners registered `warning` when swiping back using ios gesture. JS Stack Navigator

ErfanBagheri404:
Fixes #11564.

When the swipe-back gesture is updated from the native side with `useNativeDriver`, React Native warns about `onAnimatedValueUpdate` being sent with no listeners registered. Attaching a no-op listener tells the native side that JS is observing the value, silencing the warning.

The linking change from earlier commits has been reverted — see discussion above.

### DIFF
--- a/packages/native/src/useLinking.native.tsx
+++ b/packages/native/src/useLinking.native.tsx
@@ -1,4 +1,5 @@
 import {
+  CommonActions,
   getActionFromState as getActionFromStateDefault,
   getStateFromPath as getStateFromPathDefault,
   type NavigationContainerRef,
@@ -209,7 +210,10 @@ export function useLinking<ParamList extends ParamListBase>(
             );
           }
         } else {
-          navigation.resetRoot(state);
+          // Use CommonActions.reset via dispatch (same as useBuildAction / linkTo)
+          // instead of resetRoot, so the action goes through the normal pipeline
+          // and stays consistent with programmatic linkTo() calls.
+          navigation.dispatch(CommonActions.reset(state));
         }
       }
     };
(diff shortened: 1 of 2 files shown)


satya164:
Can you provide a repro showing what it fixes?

While `resetRoot` and `CommonActions.reset` aren't the exact same, both perform a reset. It doesn't make sense that it would "leave open modals untouched" when the whole state is being reset.

ErfanBagheri404:
You're right 
I dug into `resetRoot` and it already dispatches `CommonActions.reset` internally via the focus listener, so they're equivalent for modals. The premise was wrong.

The linking commits have been reverted anyway. The remaining change here is just the Card gesture listener fix for #11564. Want me to split that into a fresh PR, or is it fine here?

PR #13224, opened 2026-08-26
fix: provide linking context while state loads

### LINKED ISSUE #13051: UnhandledLinkingContext throws while linking is not ready (!isLinkingReady branch omits providers)

huytdps13400:
**Motivation**

`NavigationContainer` renders its fallback while the initial linking state is unresolved, but that early branch currently omits `LinkingContext.Provider`. Any fallback content or portaled consumer that reads the public linking context receives its throwing default instead of the active linking configuration.

The v8 codebase no longer contains the `UnhandledLinkingContext` mentioned in #13051, but `LinkingContext` retains the same failure mode. This change gives the pending branch the same `LocaleDirContext → LinkingContext` outer tree as the ready branch while preserving the existing theme fallback and keeping `BaseNavigationContainer` unmounted until state resolution completes.

Fixes #13051.

**Test plan**

- Added a regression whose `getInitialURL()` remains pending while the fallback reads `LinkingContext`; it failed before the change with `Couldn't find a LinkingContext context.`
- The regression also verifies that the actual pending config is provided (`enabled: true`, prefix `test://`), not merely a non-throwing placeholder.
- Focused `NavigationContainer` suite: 8/8 passed.
- Full Jest: 81 suites, 1,297 tests, and 169 snapshots passed in both projects.
- Root `pnpm lint` passed.
- Root `pnpm typecheck` passed.
- Package builds passed in dependency order: `@react-navigation/routers`, `@react-navigation/core`, then `@react-navigation/native`.
- Commitlint passed.

The parallel Lefthook test job reproduced an unrelated existing test-environment race twice (`window.dispatchEvent is not a function` in core tests, plus one elements timeout on the first run). The same full Jest command passed twice when run standalone; Lefthook's lint, root/standalone/declaration type checks all passed. The commit was therefore created with Lefthook disabled for that one command and validated with commitlint immediately afterward.

No visual output or public API changes.


satya164:
https://github.com/react-navigation/react-navigation/pull/13143#issuecomment-4649318305

https://github.com/react-navigation/react-navigation/pull/13198#pullrequestreview-4820492178

PR #13229, opened 2026-09-03
fix: upgrade query-string to v9 to fix DoS in decode-uri-component

hughevans:
**Motivation**

`@react-navigation/core@7.x` depends on `query-string@^7.1.3`, which pulls in `decode-uri-component@^0.2.2`. That version is affected by [GHSA-vcc3-ghjq-m6fr](https://github.com/advisories/GHSA-vcc3-ghjq-m6fr) / CVE-2026-45822: when the decoder hits malformed percent-encoded input it repeatedly splits and retries, so the time it takes grows exponentially with the input length. Fixed in `decode-uri-component@0.5.0`.

This is reachable from deep links. `getStateFromPath` hands the query part of an incoming URL straight to `queryString.parse`, so an attacker-supplied link can freeze the JS thread:

| `%C3` tokens | query size | `query-string@7.1.3` | `query-string@9.5.1` |
| --- | --- | --- | --- |
| 100 | 304 B | 59 ms | 0 ms |
| 400 | 1.2 kB | 1,373 ms | 0 ms |
| 800 | 2.4 kB | 5,505 ms | 0 ms |
| 1,200 | 3.6 kB | **13,945 ms** | 0 ms |
| 200,000 | 600 kB | – | 39 ms |

`main` already uses `query-string@9`, so this just brings `7.x` in line. There's no way to fix it within v7: `query-string@8` still depends on `decode-uri-component@^0.2.2`, and only v9.5.0 moved to `^0.5.0`.

**What changed**

- `query-string: ^7.1.3` -> `^9.5.0` in `packages/core`. Pinning to `^9.5.0` rather than `^9.4.0` matters here – `9.4.x` still resolves `decode-uri-component@^0.4.1`, which the advisory also covers.
- The two call sites switch from `import * as queryString` to a default import. v9's `index.js` exports only a default, so `import * as` would leave `queryString.parse` undefined at runtime.
- `query-string`, `decode-uri-component`, `filter-obj` and `split-on-first` added to `transformIgnorePatterns` so Jest can transform them – v9 is ESM-only. This matches what `main`'s `jest.config.js` already does.

On the ESM-only point: `@react-navigation/core` already ships ESM only (`lib/module/package.json` is `{"type":"module"}`, and `exports` has no `require` condition), so this doesn't change what consumers need from their bundler. Consumers who run Jest against the real package will need the same `transformIgnorePatterns` entries, which they already need for `@react-navigation/*` itself.

**Test plan**

`parse(query)` and `stringify(params, { sort: false })` are the only two APIs used here, so I diffed both between 7.1.3 and 9.5.1 across 66 cases – plus signs, `%20`, unicode, emoji, malformed sequences (`%`, `%ZZ`, `%C3`), reserved characters, arrays, repeated keys, `null`/empty/absent values, key ordering under `sort: false`, and stringify -> parse round trips. **Zero differences.** Notably `+` still decodes to a space: `decode-uri-component@0.5.0` dropped that, but `query-string@9` does the substitution itself.

Also added a regression test in `getStateFromPath.test.tsx` covering the decoding semantics that a major bump could plausibly change (`hello+world`, `caf%C3%A9`, `%F0%9F%98%80`, and a malformed `%C3` left as a literal). It passes on both `query-string@7` and `@9`, so it locks the behaviour rather than encoding the upgrade.

In this repo, on Node 25.6.0:

- `yarn test` – 759 tests, 54 suites, all passing
- `yarn lint` – clean
- `yarn lerna run prepack && yarn typecheck` – clean
- `yarn check-dependency-version-consistency .` – clean

And against the built ESM output, to confirm the default import resolves at runtime:

```
getStateFromPath('user/42?name=hello+world&city=caf%C3%A9&broken=%C3', config)
-> { name: 'Profile', params: { id: '42', name: 'hello world', city: 'café', broken: '%C3' } }

getStateFromPath(`user/1?q=${'%C3'.repeat(1200)}`, config)   // ~14s before
-> 1 ms
```


EmilJunker:
Please merge this already. The latest 7.x version of `@react-navigation/core` still depends on `query-string@^7.1.3`: https://github.com/react-navigation/react-navigation/blob/7.x/packages/core/package.json#L47

React Navigation is a very common library. The fact that it has had this transitive vulnerability for over two weeks and it still hasn't been fixed is honestly unbelievable.

satya164 (changes requested):
thanks for the PR, but this is a breaking change

> On the ESM-only point: @react-navigation/core already ships ESM only (lib/module/package.json is {"type":"module"}, and exports has no require condition), so this doesn't change what consumers need from their bundler. Consumers who run Jest against the real package will need the same transformIgnorePatterns entries, which they already need for @react-navigation/* itself.

`@react-navigation/core` being ESM is irrelevant. the dependency is changing from CommonJS to ESM which means this may need changes in bundler and test setup like how changes to `transformIgnorePatterns` was needed here.

EmilJunker:
If I understand it correctly, that means we have to wait for the release of v8 (where the dependency is already updated and this problem doesn't exist). 

What is the release plan for v8? I see that it's currently still in alpha.

satya164:
we don't need all features of `query-string`, so I have replaced it with a custom implementation in [4f0702c](https://github.com/react-navigation/react-navigation/commit/4f0702ce4af69d59693524ae03d6d4f50d893eb3)

PR #13233, opened 2026-09-09
docs: clarify default animationTypeForReplace is push

maxRN:
**Motivation**

Fixes documentation mismatch. The default is actually "push", not "pop".

**Test plan**

Not applicable.


satya164 (approved):
thanks for the pr

PR #13235, opened 2026-09-10
fix: use reactApplicationContext.currentActivity in MaterialSymbolModule

eduardoborges:
**Motivation**

On Android, `MaterialSymbolModule.kt` doesn't compile in our app (RN 0.87.1, Kotlin 2.2.0, AGP 9, Gradle 9.4.1):

```
e: file:///.../@react-navigation/native/android/src/main/java/org/reactnavigation/MaterialSymbolModule.kt:49:19 Function invocation 'getCurrentActivity()' expected.
```

`getCurrentActivity()` in `ReactContextBaseJavaModule` is a Kotlin function, not a Java getter, so Kotlin won't let you read it as a `currentActivity` property. It has also been deprecated since RN 0.80, and its `ReplaceWith` points to `reactApplicationContext.currentActivity`, which is what this change uses.

The new code only relies on `ReactContext.getCurrentActivity()`, and `reactApplicationContext` was already used on the same line. I checked the RN sources:

| RN | `ReactContext` | `ReactContextBaseJavaModule` |
|---|---|---|
| 0.76, 0.79 | Java, public `getCurrentActivity()` | Java |
| 0.81, 0.83, 0.85, 0.87 | Java, public `getCurrentActivity()` | Kotlin |
| 0.88.0-rc.0, main | Java, public `getCurrentActivity()` | Kotlin |

Kotlin exposes a Java getter as a property, so `reactApplicationContext.currentActivity` resolves on all of these. The change doesn't narrow the supported RN range, and the module stops depending on the deprecated function.

I haven't figured out why the example app (Expo 56, RN 0.85.3) builds fine with the current code. It may be the Kotlin version, but I haven't confirmed that.

**Test plan**

- In our app (RN 0.87.1, Kotlin 2.2.0), ran `./gradlew :react-navigation_native:compileDebugKotlin`. It fails with the error above without this change and passes with it.
- Our Android build has used an equivalent local patch (calling `getCurrentActivity()` directly) since `8.0.0-alpha.41`.


satya164 (approved):
Thanks for the PR

PR #11366, opened 2023-05-10
fix: workaround for mChrome empty space when navigating after address bar collapses

BeeMargarida:
**Motivation**

Related to https://github.com/react-navigation/react-navigation/issues/11274 and https://github.com/Expensify/App/issues/15749.

In mobile Chrome, the address bar collapses, which causes some problems: in an SPA and using a `flex: 1; height: 100%` style for the document, if the navigation is made with the address bar collapsed, it will show an empty space at the bottom of the screen (only visible if there are differences in background color). 
A possible fix is updating the body height using viewport height units based on the `window.innerHeight`, which should have the correct height taking into consideration the collapsed address bar. 

Resources:
- https://developer.chrome.com/blog/url-bar-resizing/
- https://css-tricks.com/the-trick-to-viewport-units-on-mobile/ 

**Test plan**

Run the example app in mobile chrome 

<details>
<summary>AuthFlow example code modified to test this exact scenario</summary>

```tsx
import { ParamListBase, useTheme } from '@react-navigation/native';
import {
  createStackNavigator,
  StackScreenProps,
} from '@react-navigation/stack';
import * as React from 'react';
import {
  ActivityIndicator,
  ScrollView,
  StyleSheet,
  TextInput,
  View,
} from 'react-native';
import { Button, Title } from 'react-native-paper';
import { SafeAreaProvider } from 'react-native-safe-area-context';

type AuthStackParams = {
  Home: undefined;
  SignIn: undefined;
  Chat: undefined;
};

const AUTH_CONTEXT_ERROR =
  'Authentication context not found. Have your wrapped your components with AuthContext.Consumer?';

const AuthContext = React.createContext<{
  isSignedIn: boolean;
  signIn: () => void;
  signOut: () => void;
}>({
  isSignedIn: false,
  signIn: () => {
    throw new Error(AUTH_CONTEXT_ERROR);
  },
  signOut: () => {
    throw new Error(AUTH_CONTEXT_ERROR);
  },
});

const SplashScreen = () => {
  const { colors } = useTheme();

  return (
    <View style={styles.content}>
      <ActivityIndicator color={colors.primary} />
    </View>
  );
};

const SignInScreen = ({
  navigation,
}: StackScreenProps<AuthStackParams, 'SignIn'>) => {
  const { signIn } = React.useContext(AuthContext);
  const { colors } = useTheme();

  return (
    <ScrollView>
      <View style={styles.content}>
        <TextInput
          placeholder="Username"
          style={[
            styles.input,
[...]

PR #13237, opened 2026-09-11
fix: let native-stack collapse the mobile address bar on scroll

rjmreis:
**Motivation**

On mobile browsers the address bar collapses as the page scrolls, but only when `document.body` is the element that actually scrolls - not a nested fixed-height container. `packages/stack` already had a mechanism for this (`CardContent`), but `packages/native-stack` had no equivalent, so native-stack screens on web never collapsed the address bar.

This PR adds that behavior to native-stack, brings the two packages to parity, and removes a long-standing Chrome workaround that turned out to be incompatible with the mechanism.

---

**What changed**

**1. native-stack can now let `document.body` scroll**

New `useFillsWindow` hook (`packages/native-stack/src/utils/useFillsWindow.ts`) that compares the navigator's frame against `document.body`. When they match, the focused screen switches from a fixed-height container to one that lets its content overflow into the document.

It is applied only to the screen that can safely be in the document's natural flow:
- only the **focused** screen (native-stack keeps several screens mounted for back-history and transitions, and only one can be in normal flow at a time)
- only **full-page** presentations (`presentation` unset or `'card'`) - modals, sheets and transparent presentations are untouched
- screens opt out with `contentStyle={{ flex: 1 }}`

**2. `packages/stack` gains the same opt-out**

`contentStyle={{ flex: 1 }}` now opts a screen out of the page-overflow behavior, matching native-stack.

**3. Removed the mobile-Chrome viewport workaround from both packages**

Both packages carried a workaround for [#11366](https://github.com/react-navigation/react-navigation/issues/11366): after navigating while the address bar was already collapsed, Chrome could leave empty space at the bottom of the page equal to the address bar's height. It fixed this by injecting a `<style>` tag forcing `body { height: calc(var(--vh) * 100) }` and recomputing `--vh` on every `resize`.

This is removed rather than repaired, because forcing an explicit height on `body` is fundamentally in tension with the mechanism this PR relies on - which needs `body` to be free to grow taller than the viewport. Every attempt to keep both produced a different regression:

- On `main` the workaround was gated on `navigator.maxTouchPoints > 0`, i.e. any touch-capable device, so it was silently active on iOS Safari too. There it fought Safari's own address-bar animation and caused a visible flicker. (It had never been exercised on Safari before, because the example app's `#root` override - see below - stopped body from scrolling at all.)
- Re-gating it to Android Chrome only fixed Safari, but on Android the forced height then got stuck at a stale value: a plain `resize` doesn't reliably fire when only the address bar auto-hides.
- Also listening on `visualViewport`'s `resize` fixed the stale value, but then the workaround's height rewrites and the fill re-measurement fought each other throughout the ~300ms toolbar animation.

Removing it is a deliberate trade-off. The original bug is cosmetic (temporary empty space, resolved on the next scroll or resize), this code uses percentage-based `minHeight: '100%'` rather than raw `vh` units so the original root cause may not apply here at all, and Chrome 108+ has `dvh`/`svh`. If #11366 resurfaces it should be fixed in a way that doesn't pin `body` to an explicit height.

**4. Example app `#root` override removed**

`example/src/index.tsx` unconditionally set `height: 100svh; overflow: auto` on `#root`, which captured the scroll before it could ever reach `document.body`. This is almost certainly why the existing stack behavior never worked on Safari.

**5. Supporting change in `packages/elements`**

`Screen` gains an optional `fill` prop that swaps its inner content wrapper from `flex: 1` to `minHeight: '100%'`. Default behavior is unchanged.

**Note on the duplicated hook:** `useFillsWindow.ts` and `CardContent.tsx` deliberately keep their own copies of the same detection logic rather than sharing an abstraction - the two packages apply the result to differently shaped container trees. They are intended to stay mirrored; please change both together.

---

**Root cause of the native-stack flicker**

This PR was originally opened as a draft because native-stack flickered continuously on real devices while stack did not, despite both using identical detection logic. That is now root-caused and fixed.

The cause was a feedback loop through the frame measurement. native-stack's fill originally also promoted the `ActivityView` container from `StyleSheet.absoluteFill` into normal flow (`minHeight: '100%'`). `Container`'s defaults (`flexShrink: 0`, `flexBasis: 'auto'`) make an in-flow box content-sized, so the screen's content height propagated up into the navigator's own frame - which is the value `useFillsWindow` compares against. Enabling fill therefore invalidated its own precondition:

```
t=10598 f=1 b=775 l=775  i=775
t=10616 f=0 b=775 l=2192 i=775   <- layout jumps to the content height
t=10631 f=1 b=775 l=775  i=775
```

`document.body` stays at 775 throughout; `layout` alternates between the viewport height and the content height roughly every frame - 556 toggles in one short session, against 0 for stack on the same page.

`packages/stack` was never affected because `CardContent` sits below `Card`'s `flex: 1` (`flex-basis: 0%`) box, which absorbs the growth so it never reaches the measured frame.

The fix keeps native-stack's `ActivityView` container `StyleSheet.absoluteFill` unconditionally. Out-of-flow content cannot change an ancestor's height, so the frame stays pinned to the viewport while the content still overflows normally into the document. The two packages now have the same shape: one swap point, with a size-absorbing container above it.

---

**Test plan**

- `pnpm lint`, `pnpm typecheck` and `pnpm test` pass (83 suites / 1312 tests / 169 snapshots).
- New unit tests for `CardContent` and `useFillsWindow` covering the match/no-match cases, sub-pixel tolerance (`document.body.clientHeight` is an integer while the layout is sub-pixel precise), and the `contentStyle={{ flex: 1 }}` opt-out.
- Manually verified on a real iPhone, the iOS Simulator and an Android emulator against the example app's `stack-basic` and `native-stack` article demos: the address bar collapses on scroll, no flicker on either navigator, and modal presentations are unaffected. Both platforms were re-verified after the workaround removal, including on a fresh private-mode tab.
- Verified in the DOM that with the fix the navigator's frame stays at the viewport height while the document still overflows, so the collapse is still driven by `document.body`.


satya164 (approved):
Thanks

PR #13247, opened 2026-09-13
feat: add tab bar badge support to MaterialTopTabs

### LINKED ISSUE #8744: How to show tabBarBadge in material top navigator
**Current Behavior**

- I am using the material top tab navigator
```
const TobTab = createMaterialTopNavigator();
const TopTabNav = ({ route }) => {
  const {
     notifications
    } = route.params;
  return (
    <NetworkContext.Provider value={route.params}>
      <TopTab.Navigator>
        <TopTab.Screen
          name="Home"
          component={Home}
        />
        <TopTab.Screen
          name="Settings"
          component={Settings}
          options={{
            tabBarBadge: notifications.length,
          }}
        />
      </TopTab.Navigator>
    </NetworkContext.Provider>
  );
};
```
- In the material bottom tab navigator, there is a property called tabBarBadge. So that it is very easy to show the badge/notification count.
- But in material top navigator when I apply the same thing... it is not showing the badge count as expected

**Expected Behavior**

- tabBarBadge property should be added to the material top navigator so that it will be very easy to show the badge counts
- Or what is the way to show badge count in material top navigator ?

**How to reproduce**

[Example project code ](https://snack.expo.io/@gayathri_perumal/tab-navigation-%7C-react-navigation)

**Your Environment**

| software                            | version |
| ----------------------------------- | ------- |
| @react-navigation/native            |^5.7.3
| @react-navigation/material-top-tabs |^5.2.16
| react-native-tab-view               |^2.15.1

raajnadar:
**Motivation**

Fixes #8744, open since 2020.

In material-top-tabs, `tabBarBadge` only accepts a function that returns an element. To show a simple unread count, you must write and style your own badge.

Bottom tabs already accept a string or a number, with `tabBarBadgeStyle` for the colours. This makes Material top tabs do the same.

```js
options={{ tabBarBadge: 3 }}
```

The function form still works, so this is not a breaking change. For a string or a number, we render the shared `Badge` from `@react-navigation/elements` and apply `tabBarBadgeStyle`. The function form ignores `tabBarBadgeStyle`, because you render your own element there.

**Test plan**

Added unit tests for the string, number and function forms.

Screenshot_1789910717

Screenshot_1789911011

Run the example app and open the Material Top Tabs screen. Chat shows the number badge `3`. Albums shows the text badge `new` with a custom background colour.

Lint, TypeScript and tests pass.


satya164 (changes requested):
Thanks for the PR.

A proper built-in badge needs to be next to the label/icon. Rendering the badge component by itself isn't enough and looks bad.

Material docs have examples of how it should look like: https://m3.material.io/components/tabs/specs

raajnadar:
@satya164 updated as per your review, could you please review again?

PR #13248, opened 2026-09-14
feat: emit `bottomAccessoryPlacementChange` event from native bottom tabs

vargajacint:
**Motivation**

`bottomAccessory` is rendered for both placements at once and UIKit shows whichever matches the tab bar state, so an app cannot tell which one is visible. With `tabBarMinimizeBehavior: 'onScrollDown'` the accessory goes inline and the strip it reserved above the tab bar disappears, which leaves a floating button hanging. This adds the missing signal, using the `placement` vocabulary the navigator already uses.

**Changes**

Adds a `bottomAccessoryPlacementChange` event to `BottomTabNavigationEventMap`, carrying `{ placement: 'regular' | 'inline' }` as data. The native bottom tabs view emits it on the focused route when react-native-screens reports an accessory environment change.

Requires react-native-screens with `TabsHostPropsIOS.onBottomAccessoryEnvironmentChange`: https://github.com/software-mansion/react-native-screens/pull/4656

**Test plan**

`pnpm lint`, `pnpm typecheck`, `pnpm typecheck:declarations`, both standalone typechecks and `pnpm test` (82 suites, 1354 tests) all pass, with the screens change linked locally.

```tsx
React.useEffect(
  () =>
    navigation.addListener('bottomAccessoryPlacementChange', (e) => {
      setPlacement(e.data.placement);
    }),
  [navigation]
);
```


satya164 (changes requested):
thanks for the PR. since this is an event, it should use the events API and be emitted as an event with `navigation.emit`

vargajacint:
Switched to the events API as suggested: the option is gone and this now emits a `bottomAccessoryPlacementChange` event with `{ placement: 'regular' | 'inline' }` as data, on the focused route, alongside the existing `transitionStart`/`transitionEnd` emits.

PR #13255, opened 2026-09-17
fix(stack): clip inactive cards inside Card so their shadow stays visible

### LINKED ISSUE #13254: [Web] Stacked card drop shadow is clipped — inactive CardA11yWrapper overflow: hidden has no opt-out
### Current behavior

On web, I style each card as a floating panel with cardStyle: rounded corners, a border and a boxShadow. While a card is on top this looks right. As soon as I push another screen over it, the shadow on the card underneath is cut off. Only its 1px border stays. When I go back, the shadow shows again.

The cause is that the stack wraps every screen that is not on top in a View with overflow: 'hidden'. In 7.11.1 this is in CardA11yWrapper (overflow: active ? undefined : 'hidden'), on main the same rule sits on ActivityView in CardStack.tsx. It was added in 522fa47a so a taller inactive page does not make the document scroll.

There is no option to turn this off. Returning containerStyle: {overflow: 'visible'} from cardStyleInterpolator does not help, that style lands inside the wrapper that clips. cardOverlayEnabled, cardShadowEnabled, detachPreviousScreen and presentation: 'transparentModal' do not change it either. My page has overflow: hidden on body and the stack sits in a fixed-size container, so the scroll problem the clip was meant to fix cannot happen here. The clip only removes visuals.

### Expected behavior

A card that is no longer on top should keep looking the same as when it was on top, shadow included. Or an option to skip the clip, in the style of cardOverlayEnabled or cardShadowEnabled, for layouts where the page cannot scroll anyway.

### Reproduction

https://snack.expo.dev/@bartekholinice/stacked-card-drop-shadow

The packages in the Snack are a bit older than what I run in the app, but they show the exact same problem. .
If the preview does not pick up the code, click “Devices” in the bottom right corner and hit Reload or Update changes.

### Platform

- [ ] Android
- [ ] iOS
- [x] Web
- [ ] Windows
- [x] MacOS

### Packages

- [x] @react-navigation/stack
- [ ] react-native-tab-view

### Environment

| @react-navigation/native               | 7.1.33  |
| @react-navigation/bottom-tabs          | 7.15.5  |
| @react-navigation/material-top-tabs    | 7.4.19  |
| @react-navigation/stack                | 7.8.5   |
| @react-navigation/native-stack         | 7.14.5  |
| react-native-tab-view                  | 4.3.0   |
| react-native-screens                   | 4.25.0  |
| react-native-safe-area-context         | 5.6.2   |
| react-native-gesture-handler           | 2.32.0  |
| react-native-reanimated                | 4.5.5   |
| react-native-pager-view                | 9.0.4   |
| react-native                           | 0.86.0  |
| expo                                   | 57.0.8  |
| node                                   | v26.5.0 |
| npm or pnpm                            | npm 11.17.0 |

BartekObudzinski:
**Motivation**

Fixes #13254.

`CardStack` wraps every card that is not on top in an `ActivityView` with `overflow: 'hidden'`. The wrapper fills the whole stack, so it clips anything the card paints outside its own box, including its `boxShadow` on the sides where the card touches the stack edge.

You hit this when `cardStyle` makes the card smaller than the stack and gives it a shadow, for example a floating panel docked to the right. Push a second card on top and the one below loses its shadow on those sides and shows a hard edge instead.

Per the discussion in the issue, this does not add an option. It moves the clip onto the `Animated.View` inside `Card` that receives `cardStyle`. That view paints the card, and `overflow: 'hidden'` does not clip an element's own shadow. The shadow of the card underneath stays visible, and the content is still clipped to the card, so a buried larger page still cannot grow the scroll area on web. `CardContainer` passes `active` down to `Card` for this.

Side effect: the built-in `styles.shadow` edge view is a child of that view, so inactive cards now clip it. The active card covers that edge anyway.

**Test plan**

Verified in Expensify/App on web, where each stacked right-hand panel is a floating card with its own border and shadow set through `cardStyle`.

1. Open a wide floating card, then push a narrower one over it.
2. Before: the wide card behind lost its shadow on the sides touching the stack edge.
3. After: the wide card behind keeps its shadow and its content is still clipped to the card.

`pnpm typecheck`, `pnpm lint` and `pnpm exec jest packages/stack` pass. Two tests in `packages/core` (`useNavigationState`, `useOnAction`) fail on current `main` without this change as well.


packages/stack/src/views/Stack/Card.tsx
```diff
@@ -600,7 +602,16 @@ function Card({
         <GestureDetector gesture={panGesture}>
           <Animated.View
             needsOffscreenAlphaCompositing={hasOpacityStyle(cardStyle)}
-            style={[styles.card, cardStyle]}
+            style={[
+              styles.card,
+              {
+                // Avoid unfocused larger pages increasing scroll area
+                // e.g. when a smaller screen is pushed over a larger one on web
+                // Clipping here instead of the wrapper keeps the card's own shadow visible
```
satya164: don't mention historical details in code comments

```suggestion                
```

satya164 (changes requested):

packages/stack/src/views/Stack/Card.tsx
BartekObudzinski: Done. Kept only the original two lines from CardStack, dropped the rest.

satya164 (approved):
thanks

Issue #13258, opened 2026-09-18
[drawer-layout] `onOpen`/`onClose` are the only callbacks without a stable identity and are called unconditionally from a layout effect — inline handlers cause an infinite render loop


yinminqian:
**Package:** `react-native-drawer-layout`
**Versions:** 4.2.10 (`latest`) and 5.0.0-alpha.18 (`next`) — both affected
**Platform:** reproduced on iOS (New Architecture, RN 0.86.2); the code path itself is platform-independent

## Summary

In `src/views/Drawer.native.tsx`, every callback prop is given a stable identity through `useLatestCallback` — `onGestureBegin`, `onGestureFinish`, `onGestureAbort`, `onAnimationStart`, `onAnimationEnd`, `updateLayout`, `hideKeyboard` — **except `onOpen` and `onClose`**. Those two are used raw, and they sit in the dependency array of `toggleDrawer`:

```js
const toggleDrawer = React.useCallback((open, velocity) => {
  'worklet'
  // ...
  if (open) {
    scheduleOnRN(onOpen)    // 4.2.10: runOnJS(onOpen)()
  } else {
    scheduleOnRN(onClose)   // 4.2.10: runOnJS(onClose)()
  }
}, [/* ... */, onOpen, onClose])

React.useLayoutEffect(() => {
  toggleDrawer(open)
}, [animatingTo, open, toggleDrawer])
```

Two things combine:

1. `onOpen` / `onClose` are invoked **unconditionally** every time `toggleDrawer` runs — including when the layout effect re-runs with an unchanged `open` value, i.e. when nothing actually opened or closed.
2. Because they are raw props inside the `useCallback` deps, **any caller that passes an inline arrow function changes `toggleDrawer`'s identity on every render**, which re-runs the layout effect.

So for a very ordinary call site:

```jsx
<Drawer
  open={this.state.open}
  onOpen={() => this.setState({ open: true })}
  onClose={() => this.setState({ open: false })}
  renderDrawerContent={...}
>
```

the loop is:

```
render → new onClose identity → toggleDrawer identity changes
       → useLayoutEffect re-runs → toggleDrawer(open) → onClose()
       → setState (a class component re-renders even when the value is unchanged)
       → render → ...
```

This is self-sustaining. We measured **~45 renders/second, 6,707 renders in 2.5 minutes**, with the drawer closed and nobody touching it.

## Impact (what this actually looks like in an app)

[...]

PR #13259, opened 2026-09-18
fix: give drawer onOpen and onClose a stable identity

ahmdshrif:
**Motivation**

`Drawer.native.tsx` runs every callback prop through `useLatestCallback` — except `onOpen`/`onClose`, used raw in the `toggleDrawer` worklet and in its deps. With inline handlers `toggleDrawer` changes identity each render, so the `panGestureConfig` memo recomputes, the public `configureGestureHandler` prop is re-invoked and the gesture reconfigured. This wraps them like the other seven.

Scope: the loop in #13258 is real on 4.2.10, but on `main` the `translationX === toValue` / `animatingTo` early returns already prevent it — the repro gives 1 render, 0 `onClose` calls. Firing them only on real transitions is a behaviour change, left out.

**Test plan**

First tests in this package, covering behaviour preservation: `onClose` is not called while the drawer stays closed; `onOpen`/`onClose` fire once each on toggle.

I dropped a third test counting `configureGestureHandler` calls: `useMemo` cache survival is not a React guarantee — it passed locally, failed once on CI.


ahmdshrif:
Closing this — @satya164 pushed the same fix directly to `main` in f599852bde0331932c346890807a9cca3222b19c, which already closes #13258. Thanks!

PR #13261, opened 2026-09-20
fix(bottom-tabs): ignore the empty keyboard frame reported on interface rotation

### LINKED ISSUE #13262: Bottom tab bar disappears permanently after rotating on iOS 26+ with tabBarHideOnKeyboard
### Current behavior

On iOS 26+, rotating the device permanently hides the tab bar of any navigator using `tabBarHideOnKeyboard`. It does not come back while the app stays in that orientation, so tab navigation is unreachable until the user rotates back or restarts the app.

iOS 26+ posts a keyboard *show* notification on every interface rotation, with an empty keyboard frame, and never posts a matching hide. Subscribing to all four keyboard events and rotating twice on an iPhone running iOS 27.1:

```
keyboardWillShow  height=0  screenY=871
keyboardDidShow   height=0  screenY=871
keyboardWillHide  — never fires
keyboardDidHide   — never fires
```

`useIsKeyboardShown` treats any show notification as a real keyboard, so `isKeyboardShown` latches `true` and nothing clears it. `BottomTabBar` then evaluates

```js
const shouldShowTabBar = !(tabBarHideOnKeyboard && isKeyboardShown);
```

as `false` and animates the bar out with `translateY: layout.height`, `position: 'absolute'` and `pointerEvents: 'none'` — permanently, because the hide that would restore it never arrives.

`Keyboard.isVisible()` also returns `true` for these notifications, so it cannot be used to distinguish them from a real keyboard. The empty frame is the only signal I found that can.

This is not device-specific. The notifications come from the platform, so any rotating iOS 26+ device with a `tabBarHideOnKeyboard` navigator should be affected. The same unguarded handler is present on `7.x` and on `main`, so v7 and the v8 alphas are both affected.

### Expected behavior

Rotating the device should not hide the tab bar. `tabBarHideOnKeyboard` should react only to an actual keyboard.

### Reproduction

https://github.com/olivier-bouillet/rn-bottom-tabs-rotation-repro

```sh
npm install
npx expo run:ios
```

Run on an iOS 26+ device or simulator, then rotate. The tab bar disappears and does not return. Each screen prints the keyboard notifications it receives, so the empty-frame show events are visible on screen.

### Platform

- [x] iOS

### Packages

- [x] @react-navigation/bottom-tabs

### Environment

| package | version |
| --- | --- |
| @react-navigation/bottom-tabs | 7.19.2 |
| @react-navigation/native | 7.4.1 |
| react-native | 0.86.3 |
| expo | 57.0.24 |
| node | 23.11.1 |
| Xcode | 27.0 |

A fix is proposed in #13261.

olivier-bouillet:
Closes #13262

### Current behavior

On iOS 26+, rotating permanently hides the tab bar of any navigator using `tabBarHideOnKeyboard`.

Rotation posts a keyboard *show* notification with an empty frame (`height: 0`) and never a matching hide, so `useIsKeyboardShown` latches `true` and the bar is animated away for good. `Keyboard.isVisible()` is `true` for these too.

### Expected behavior

Rotating should not hide the tab bar.

### Reproduction

https://github.com/olivier-bouillet/rn-bottom-tabs-rotation-repro

`npm install && npx expo run:ios`, then rotate.

### The change

Ignore show notifications whose `endCoordinates.height` is `0`. An event with no frame keeps the previous behaviour. Trace in the comment below.

### Platform

- [x] iOS

### Packages

- [x] @react-navigation/bottom-tabs

### Environment

| package | version |
| --- | --- |
| @react-navigation/bottom-tabs | 7.19.2 |
| react-native | 0.86.3 |
| expo | 57.0.24 |


olivier-bouillet:
Full trace and analysis, kept out of the description to stay under the length limit.

## What the platform sends

Subscribing to all four keyboard events in a `createBottomTabNavigator` screen and rotating twice, on an iPhone running iOS 27.1:

```
keyboardWillShow  height=0  screenY=871
keyboardDidShow   height=0  screenY=871
keyboardWillHide  — never fires
keyboardDidHide   — never fires
```

`screenY` equals the window height, so the frame is an empty strip parked at the bottom edge. No hide notification ever arrives.

## Why the tab bar never comes back

`useIsKeyboardShown` treats any show notification as a real keyboard:

```js
const handleKeyboardShow = () => setIsKeyboardShown(true);
```

so the flag latches `true`. `BottomTabBar` then evaluates

```js
const shouldShowTabBar = !(tabBarHideOnKeyboard && isKeyboardShown);
```

as `false` and animates the bar out with `translateY: layout.height`, `position: 'absolute'` and `pointerEvents: 'none'`. Because the hide that would clear the flag never comes, the bar stays off screen until the app is restarted, and tab navigation is unreachable.

## Why not `Keyboard.isVisible()`

It returns `true` for these notifications as well, so it cannot distinguish them from a real keyboard. The empty frame is the only signal I found that can.

## Scope

Not specific to a device or a foldable — the notifications come from the platform. Any rotating iOS 26+ device with a `tabBarHideOnKeyboard` navigator should show it. The same unguarded handler is on `7.x` and on `main`, so v7 and the v8 alphas are both affected. Happy to open a `7.x` backport.

## Tests

Three cases, run against the package suite locally (4 suites, 15 tests green):

- a keyboard with a height still reports shown, and a hide still clears it
- an empty frame does not report shown
- an event carrying **no** frame still reports shown

That last one matters: `tab bar cannot be tapped when hidden` in `index.test.tsx` calls `listener({})`, and an earlier revision of this patch crashed on it. With the fix reverted locally, only the rotation case fails — 1 of 15.

## A related note for iOS 26+

Unrelated to this bug, but it blocks the reproduction: an app built against the iOS 26+ SDK that does not adopt the UIScene lifecycle is refused at launch by UIKit. The repro sets `ios.enableSceneSupport` through `expo-build-properties` for that reason.

